PR #19, opened 2024-09-12
Setup the build environment, basic scaffolding, and block validation (minus KES)

AndrewWestberg:
KES validation can come in a future PR.


amaru-node/src/main.rs
```diff
@@ -0,0 +1,37 @@
...
+        Ok(level) => match level.to_lowercase().as_str() {
+            "error" => tracing::Level::ERROR,
+            "warn" => tracing::Level::WARN,
+            "info" => tracing::Level::INFO,
+            "debug" => tracing::Level::DEBUG,
+            "trace" => tracing::Level::TRACE,
+            _ => tracing::Level::INFO,
+        },
+        Err(_) => tracing::Level::INFO,
+    };
```
scarmuega, line 29: we could probably use the following instead of manual mapping:

https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html#

amaru-node/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "amaru-node"
```
scarmuega, line 2: I agree with @rvcas here, calling the node just `amaru` seems reasonable (unless we're thinking on treating this crate as a library, which I don't think it's the case).

AndrewWestberg:
Refactoring to move many things into ouroboros repository. closing.

PR #26, opened 2024-10-19
Consensus ledger integration

KtorZ:
- :round_pushpin: **feat: Add block header validation**
  
- :round_pushpin: **fix: reinstate phase-2 validation**
    This relies on a temporary branch of Aiken which makes use of the unreleased pallas crate.

- :round_pushpin: **fix: create client connection outside of gasket pipeline**
    This gets rid of the underlying hack with tokio, and simply pass the
  client as a reference.

  More generally, we probably want to embrace some kind of
  resource-allocation bracket pattern where necessary resources such as
  connection handles or database connections are acquired only once at
  the beginning of the program.

- :round_pushpin: **chore: rename 'validate' to 'header_validation' for clarity.**
  
- :round_pushpin: **chore: move CSV parsing code in its own separate module.**
    Because it's demo code, doesn't mean we have to be absolutely reckless about code, hard-wired values/behaviours and organization. There's no such thing as a temporary hack, so anything we write shall be written from the perspective that it might stay forever. From that viewpoint, even if we're going to throw away modules, tests, and functions; it's better to keep things tidy.

- :round_pushpin: **chore: decouple consensus/ledger from demo scenario**
    Ideally, the consensus never needs to commit to a concrete
  representation of the ledger. So I had to create the ledger mock
  outside of the consensus fully. The mock data is pulled on startup and
  used to create a mock that satisfies the given scenario.

  As a side-effect, this allows to make the mock data dynamic and
  provided at runtime; so that we can more easily run and re-run the
  demo across various scenarios.

  NOTE: The data/stake_pools/172.csv that is currently versioned is NOT
  the data from epoch 172. My PreProd is actually still syncing (54%)
  and I only generated the list from whatever ongoing epoch it is at.


amaru/src/consensus/header_validation.rs
```diff
@@ -0,0 +1,179 @@
...
+    epoch
+}
+
+#[derive(Stage)]
+#[stage(name = "validate", unit = "PullEvent", worker = "Worker")]
+pub struct Stage {
+    peer_session: Arc<Mutex<PeerClient>>,
+
+    ledger: Arc<Mutex<dyn LedgerState>>,
+    epoch_to_nonce: HashMap<Epoch, Hash<32>>,
```
KtorZ, line 43: @AndrewWestberg I feel like this should be part of the `LedgerState` trait to begin with. I would expect the ledger to be keeping track of the rolling nonce, and providing a way for the consensus to access the current nonce when needed. 

What do you think?

amaru/src/consensus/header_validation.rs
```diff
@@ -0,0 +1,179 @@
...
+
+                match header {
+                    MultiEraHeader::BabbageCompatible(_) => {
+                        let minted_header = header.as_babbage().unwrap();
+                        let epoch = epoch_for_preprod_slot(minted_header.header_body.slot);
+
+                        // TODO: This is awkward, and should probably belong to the LedgerState
+                        // abstraction? The ledger shall keep track of the rolling nonce and
+                        // provide some endpoint for the consensus to access it.
+                        let epoch_nonce = stage
```
abailly, line 114: Interestingly, in the Haskell node this state is maintained on the consensus side, In `PraosState` alongside operational certificates counters map.

amaru/src/consensus/header_validation.rs
```diff
@@ -0,0 +1,179 @@
...
+    let byron_epoch_length = 10 * byron_protocol_consts_k;
+    let byron_slots = byron_epoch_length * shelley_transition_epoch;
+    let shelley_slots = slot - byron_slots;
+    let epoch = (shelley_slots / shelley_epoch_length) + shelley_transition_epoch;
+
+    epoch
+}
+
+#[derive(Stage)]
+#[stage(name = "validate", unit = "PullEvent", worker = "Worker")]
```
abailly, line 38: I am not familiar with this `Stage` thing, is this from gasket?

amaru/src/consensus/header_validation.rs
```diff
@@ -0,0 +1,179 @@
...
+    let shelley_epoch_length = 432000; // 5 days in seconds
+    let shelley_transition_epoch: u64 = 4;
+    let byron_protocol_consts_k: u64 = 2160;
+    let byron_epoch_length = 10 * byron_protocol_consts_k;
+    let byron_slots = byron_epoch_length * shelley_transition_epoch;
+    let shelley_slots = slot - byron_slots;
+    let epoch = (shelley_slots / shelley_epoch_length) + shelley_transition_epoch;
+
+    epoch
+}
```
KtorZ, line 35: @Quantumplation here's something that can be its own stream of work: slotting arithmetic. In the long run, we will need code that can perform a variety of conversion from slots, epochs, posix times, etc.. from (evolving) protocol parameters and multi-era sizes. In particular, it should also handle notions of forecast into the future given a certain stability window's size. 

I believe this to be something sufficiently well-scoped (I can spend more time drafting that) and isolated that it can even live in its own crate. Is something Sundae would be interested in assisting with?

amaru/src/consensus/header_validation.rs
```diff
@@ -0,0 +1,179 @@
...
+                        // TODO: This is awkward, and should probably belong to the LedgerState
+                        // abstraction? The ledger shall keep track of the rolling nonce and
+                        // provide some endpoint for the consensus to access it.
+                        let epoch_nonce = stage
+                            .epoch_to_nonce
+                            .get(&epoch)
+                            .ok_or(miette!("epoch nonce not found"))
+                            .or_panic()?;
+
+                        let active_slots_coeff: FixedDecimal =
```
abailly, line 120: Don't we want to move those values to a `ProtocolParameter` data structure?

amaru/src/consensus/header_validation.rs
```diff
@@ -0,0 +1,179 @@
...
+                let block = {
+                    info!("fetching block...");
+                    let mut peer_session = stage.peer_session.lock().unwrap();
+                    let client = (*peer_session).blockfetch();
+                    let block = client.fetch_single(point.clone()).await.or_restart()?;
+                    info!("fetched block done.");
+                    block
+                };
+
+                // info!(?block, "fetched block");
```
abailly, line 149: delete line? or update line 145 to add info about the block?

amaru/src/consensus/header_validation.rs
```diff
@@ -0,0 +1,179 @@
...
+                    }
+                    MultiEraHeader::Byron(_) => {
+                        trace!("byron header, skipping validation");
+                    }
+                }
+
+                let block = {
+                    info!("fetching block...");
+                    let mut peer_session = stage.peer_session.lock().unwrap();
+                    let client = (*peer_session).blockfetch();
```
abailly, line 143: In "reality", we don't always fetch blocks, we only do so if we can extend our best chain with the retrieved header IIRC. Should there not be a separate "actor" to fetch blocks?

abailly (commented):
Some probably minor comments, feel free to ignore as I am not sure they are pursuant to the goal of this PR

amaru/src/consensus/header_validation.rs
KtorZ: I know, and always found it to be a weird design choice? Do you see this to be preferable?

amaru/src/consensus/header_validation.rs
KtorZ: Indeed.

amaru/src/consensus/header_validation.rs
KtorZ: We do; and I was already tempted to do it. I still might before the demo recording.

amaru/src/consensus/header_validation.rs
KtorZ: This is what @AndrewWestberg is hinting at in the module comments I believe. At this point, the validation worker is pulling the block directly; but we do want to move that into a separate stage of the pipeline.

amaru/src/consensus/header_validation.rs
abailly: I remember discussions between consensus and ledger teams about the fact some code lives in cardano-ledger but is really about consensus, e.g things like nonce manipulation or protocol-level crypto (VRF, KES).
The nonce is a protocol thing IMO and its evolution is driven by the protocol so I think it makes sense to keep state on the protocol side. We could change the ledger without changing the protocol, and still need to compute nonce.

amaru/src/consensus/header_validation.rs
abailly: Unrelated comment, but is `Hash<32>` a good type? Don't we want it to be more `Hash<Blake2b_256>` or `Hash<Blake2b<32>>`?

amaru/src/consensus/header_validation.rs
Quantumplation: Yes, absolutely!

AndrewWestberg: This work is already done inside of cncli. It probably needs to be cleaned up, but the arithmetic is all there.

AndrewWestberg: https://github.com/cardano-community/cncli/blob/develop/src/nodeclient/leaderlog/mod.rs#L200

https://bitbucket.org/muamw10/jormanager/src/caa5b601c70b4836219000c3990bd1979e87f11e/src/main/kotlin/com/swiftmako/jormanager/controllers/utils/BlockUtils.kt#lines-58

amaru/src/consensus/header_validation.rs
AndrewWestberg: @KtorZ I'm not sure if it belongs in Ledger, or in Chain.  In most of my projects I separate out the concept of the Chain (only deals with block headers), and Ledger (deals with the whole block, txns, etc...).

The rolling nonce and epoch nonce is something that's calculated purely from the headers. In CNCLI, I don't even bother with block fetch to do leaderlogs.

I like the idea of separating these concerns also because blockfetch() can be made more performant if we let it sync larger chunks of blocks than just one at a time.

For my tools that do both chainsync and blockfetch like JorManager, I separate these concerns. Chainsync runs ahead of blockfetch when catching up to the tip.  Blockfetch can then grab stuff in bigger chunks and it also takes it longer to process transactions.

amaru/src/consensus/header_validation.rs
AndrewWestberg: I would add a TODO to the code and move on. Don't be tempted by PR scope creep. Before you know it, you're trying to put every piece we're building into a single PR.

amaru/src/consensus/header_validation.rs
AndrewWestberg: I left an empty Stage for moving this to a fetch module. However, I didn't think I could get all that done before I needed to leave next week for Dubai and get it properly handed off to @KtorZ .

amaru/src/consensus/header_validation.rs
AndrewWestberg: These can probably all be switched to `trace!()` or just removed. I only really put them in there when I was debugging the multi-threaded dead-locks to figure out where the code was getting hung up.

amaru/src/sync/mod.rs
```diff
@@ -38,23 +50,33 @@ fn define_gasket_policy() -> gasket::runtime::Policy {
     }
 }
 
-pub fn bootstrap(config: Config) -> miette::Result<Vec<Tether>> {
-    let mut pull = pull::Stage::new(
-        config.upstream_peer.clone(),
-        config.network_magic,
-        config.intersection.clone(),
-    );
+pub fn bootstrap(config: Config, client: &Arc<Mutex<PeerClient>>) -> miette::Result<Vec<Tether>> {
```
AndrewWestberg, line 53: The client probably doesn't need to be Arc-Mutex as @scarmuega was saying. It can instead be destructured into individual protocols. The chainsync can be owned by the pull Stage, and the blockfetch can be owned by the validate Stage, or the fetch Stage if we separate that out.

AndrewWestberg (approved):
Ready to merge so others can continue iterating on this. I'm comfortable with where it's at.

amaru/src/consensus/header_validation.rs
KtorZ: @AndrewWestberg, I was hinting at something quite more sophisticated here. Since the problem we have isn't really to perform slot arithmetic from known parameters (that's indeed straightforward), but to provide an abstraction that (1) keeps track of the ever evolving parameters, (2) ensures that we follow heuristic similar to the Haskell implementation (e.g. no time conversation in the future outside of the stability window).

amaru/src/consensus/header_validation.rs
```diff
@@ -0,0 +1,179 @@
...
+use pallas_traverse::MultiEraHeader;
+use std::{
+    collections::HashMap,
+    sync::{Arc, Mutex},
+};
+use tracing::{info, trace};
+
+pub type UpstreamPort = gasket::messaging::InputPort<PullEvent>;
+pub type DownstreamPort = gasket::messaging::OutputPort<ValidateHeaderEvent>;
+
```
KtorZ, line 22: // TODO: Design and implement a proper abstraction for slot arithmetic. See https://github.com/pragma-org/amaru/pull/26/files#r1807394364

amaru/src/consensus/header_validation.rs
KtorZ: > @abailly: Unrelated comment, but is Hash<32> a good type? Don't we want it to be more Hash<Blake2b_256> or Hash<Blake2b<32>>?

At this point in Pallas, `Hash` exclusively refers to Blake2b hash digests. So the type is somewhat unambiguous. It will as soon as we introduce need for another hash algorithm. In which case I rather just swap `Hash` for `Blake2bDigest`.

PR #30, opened 2024-11-28
Create and persist stable db snapshots at epoch boundaries

KtorZ:
This PR introduces mainly two things:

1. The ability to create snapshots at every epoch boundary. This heavily piggyback on [RocksDB's native Checkpoints feature](https://github.com/facebook/rocksdb/wiki/Checkpoints) which provides a quick and efficient to take database snapshots without much copy (hard-linking as much as possible between snapshots).

2. It also adds the logic to collapse stake pool updates at the epoch boundary. Indeed, we've been so far collecting pool updates coming from published certificates, in an append-only fashion without ever applying them when they should take effect. This is now done at the epoch boundary, just before the snapshot is taken. 

Regarding (2), I've been trying to come up with an API design that (a) remains simple for the consumer (i.e. the ledger state) and (b) doesn't leak details of the concrete key/value store implementation to the ledger state. Said differently, I've been doing my best to preserve the ledger state holding on a abstract `Store<E>`.

```rs
pub struct State<S, E>
where
    S: Store<Error = E>,
{
    stable: Arc<Mutex<S>>,
    volatile: VolatileDB,
}
```

and not on a concrete `RocksDB`; this in order to keep a clear boundary between the ledger state and the stable store. However, this has required some finesse since updating pools require iterating over all pool entries in the stable store and conditionally apply updates. In an attempt to not bloat the `Store` trait API with a myriad of functions get/set/update (which will rapidly get out of hands for every underlying collection to come), I've tried to provide a Rust-idiomatic interface based on Iterators. 

The core idea is thus to provide an iterator to the ledger state of some item that can be borrowed either immutably or mutably. When a mutable borrow occurs, the store should persist the update automatically. This is achieved through a couple of indirections, but the overall approach remains quite "simple" (I think...) as the store's API boils down to:

```rs
    fn with_pools(&self, with: impl Fn(IterPools<'_, '_>)) -> Result<(), Self::Error>;
```

where `IterPools` is a callback argument that provides an iterator for mutable pool entries with tracked mutations:

```rs
pub type IterPools<'a, 'b> = IterBorrow<'a, 'b, Option<PoolParamsUpdates>>;

pub type IterBorrow<'a, 'b, T> =
    Box<dyn Iterator<Item = Box<dyn BorrowMut<T> + 'a>> + 'b>);
```

> [!NOTE]
> The `Option<_>` is used to delete entries. When iterating, all items are necessarily `Some`, but they may be mutated to `None` to indicate they should be removed from the store entirely. 

The difficulty, of course, is tie all the pieces together and ensure a safe API that automatically reflects mutations on the iterators back to the store. This is why the top-level API takes a callback, and does not simply return an iterator. By taking a callback, we ensures that the iterator only lives for the duration of that callback and we can update the store once the callback returns. This is akin to a [bracket-style resource acquisition pattern](https://wiki.haskell.org/Bracket_pattern).

 On the plus side, the overall approach is abstract over any type `T` that can be serialised and deserialised to the database. So it'll be easy to re-use for entries other than pools.


amaru/src/ledger/state.rs
```diff
@@ -168,6 +192,70 @@ impl<'a, E: std::fmt::Debug> State<'a, E> {
...
+
+        if let Some(new_params) = update {
+            pool.current_params = new_params;
+        }
+
+        // Regardless, always prune future params from those that are now-obsolete.
+        pool.future_params
+            .retain(|(_, epoch)| epoch > &current_epoch);
+    }
+}
```
KtorZ: Should add some logging here, to visualize pool updates at the epoch boundary.

PR #31, opened 2024-12-02
Setup OpenTelemetry & experiment with Jaeger

KtorZ:


AndrewWestberg:
Not really my area of expertise, but I like the idea of better code observability.

KtorZ:
I am leaning towards open telemetry because it's (1) easy to produce from Tokio's tracing library and (2) widely compatible with a good range of tools (including the usual Grafana / Prometheus). Jaeger here is another example, that is easy to run locally and provide good observability already. 

I still need to tweak a bit the setup and the events we emit but it's already being quite useful. 

Ideally, we should rapidly decide on a holistic heuristic for tracing that we can apply to all sub-components and easily document for downstream consumers. There's no point having metrics if people don't know what to do with them.

wolf31o2:
We're using otel in Dingo.

abailly:
Thanks a lot for treading that path, I am 200% in support to add o11y very early in the system and OpenTelemetry looks like a standard we want to follow. In addition, I would like to: 1/ ensure all traces are emitted as structured data (probably JSON, until the volume becomes so high we need something else but I doubt it for a single node) an 2/ document and standardise the emitted traces so that third party tools can have a reliable source to base their analysis on instead of having to reverse engineer source code.

FWIW, here is an interesting article from the person who coined _observability_ as a word about using log events stream as a unified source of truth: https://www.honeycomb.io/blog/one-key-difference-observability1dot0-2dot0

KtorZ:
>  @abailly: 1/ ensure all traces are emitted as structured data (probably JSON, until the volume becomes so high we need something else but I doubt it for a single node) 

Open telemetry has our back here, since the format is serializable and structured. For now it's using gRPC to transfer telemetry to the various services, but it's trivial to configure to return JSON if needed. I can walk you through the setup if you're interested.

> @abailly: 2/ document and standardise the emitted traces so that third party tools can have a reliable source to base their analysis on instead of having to reverse engineer source code.

Yup. That's something I would like to discuss further eventually. I don't know of any good tool to document traces / spans / metrics; so I was thinking to just keep it simple and have some kind of structured markdown. But the bits I want to discuss is how do we want to organize that. The default targets for traces that comes with the `tracing` crate is sub-optimal IMO because it's tied to the folder/module structure -- which means that the semantic of traces changes as we reorganize the code and, it's also tied to our _internal layout_, which isn't necessarily what we want to communicate to consumer of the app. So anyway, I'll probably propose something which we can use as a starting point for discussion.

Also, I was chatting with @wolf31o2 the other day, and we agreed that it's probably a good idea as a starter to replicate (at least some of) the metrics that the node currently emits. For traces, it's a bit more delicate since we have a quite different architecture overall; so I am more in favor of having our own (documented) set of traces, and possibly, provide an adaptor from our traces to the Haskell's node traces -- as a best effort.

PR #32, opened 2024-12-05
feat: Create initial mempool interface

AndrewWestberg:
This *incomplete* implementation of the mempool interface is opened to provide feedback and criticism for further refinement.

### DIFF
--- a/amaru/src/mempool/mod.rs
+++ b/amaru/src/mempool/mod.rs
@@ -39,6 +39,9 @@ pub trait MempoolTx<M>: Ord + PartialEq {
     fn tx(&self) -> &Tx;
 }
 
+/// An ordered queue of transactions.
+pub type TxQueue<T> = BTreeSet<T>;
+
 /// The mempool is a queue of transactions that are waiting to be validated and applied to an
 /// ephemeral ledger state.
 pub struct Mempool<T, M>
@@ -47,7 +50,7 @@ where
 {
     /// The queue of transactions. The ordering of transactions are defined by the Ord trait of the
     /// type T and left to the implementer.
-    transactions: Arc<RwLock<BTreeSet<T>>>,
+    transactions: Arc<RwLock<TxQueue<T>>>,
 
     _marker: std::marker::PhantomData<M>,
 }
--- a/amaru/src/mempool/mod.rs
+++ b/amaru/src/mempool/mod.rs
@@ -26,7 +26,7 @@ where
     ephemeral: ledger::state::VolatileDB,
 }
 
-/// A transaction in the mempool.
+/// A transaction in the mempool with optional _M_ metadata type useful for ordering transactions.
 pub trait MempoolTx<M>: Ord + PartialEq {
     /// Any additional _metadata_ associated with the transaction. For example, it could include a timestamp
     /// of when it was received, a reference to the peer that sent it, or some priority value used
--- a/amaru/src/mempool/mod.rs
+++ b/amaru/src/mempool/mod.rs
@@ -1,0 +1,104 @@
+use crate::ledger;
+use pallas_primitives::conway::Tx;
+use std::collections::BTreeSet;
+use std::sync::{Arc, RwLock, RwLockReadGuard, RwLockWriteGuard};
+
+// State
+// ----------------------------------------------------------------------------
+
+/// The state of the mempool is split into two sub-components:
+///
+/// - A _ledger_state_ which contains a view of a ledger state, which is used to validate
+///   transactions.
+///
+/// - An _ephemeral_ state, which is maintained as a sequence of diff operations to be applied on
+///   top of the _ledger_state_. It should be re-calculated as transactions are added or removed
+///   from the mempool.
+pub struct State<S, E>
+where
+    S: ledger::store::Store<Error = E>,
+{
+    /// The current view of the ledger state.
+    ledger_state: ledger::state::State<S, E>,
+
+    /// The ephemeral state, which is applied on top of the ledger_state using transactions
+    /// in the mempool.
+    ephemeral: ledger::state::VolatileDB,
+}
+
+/// A transaction in the mempool.
+pub trait MempoolTx<M>: Ord + PartialEq {
+    /// Any additional _metadata_ associated with the transaction. For example, it could include a timestamp
+    /// of when it was received, a reference to the peer that sent it, or some priority value used
+    /// to order transactions.
+    fn metadata(&self) -> Option<M> {
+        None
+    }
+
+    /// The transaction itself.
+    fn tx(&self) -> &Tx;
+}
+
+/// The mempool is a queue of transactions that are waiting to be validated and applied to an
+/// ephemeral ledger state.
+pub struct Mempool<T, M>
+where
+    T: MempoolTx<M>,
+{
+    /// The queue of transactions. The ordering of transactions are defined by the Ord trait of the
+    /// type T and left to the implementer.
+    transactions: Arc<RwLock<BTreeSet<T>>>,
+
+    _marker: std::marker::PhantomData<M>,
+}
+
+impl<T: MempoolTx<M>, M> Default for Mempool<T, M> {
+    fn default() -> Self {
+        Mempool {
+            transactions: Arc::new(RwLock::new(BTreeSet::new())),
+            _marker: Default::default(),
+        }
+    }
+}
+
+impl<T: MempoolTx<M>, M> Mempool<T, M> {
+    pub fn new() -> Self {
+        Mempool::default()
+    }
+
+    pub fn insert(&self, tx: T) {
+        self.transactions.write().unwrap().insert(tx);
+    }
+
+    pub fn len(&self) -> usize {
+        self.transactions.read().unwrap().len()
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.transactions.read().unwrap().is_empty()
+    }
+
+    pub fn iter(&self) -> RwLockReadGuard<'_, BTreeSet<T>> {
+        self.transactions.read().unwrap()
+    }
+
+    pub fn iter_mut(&self) -> RwLockWriteGuard<'_, BTreeSet<T>> {
+        self.transactions.write().unwrap()
+    }
+
+    pub fn clear(&self) {
+        self.transactions.write().unwrap().clear();
+    }
+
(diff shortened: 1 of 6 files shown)


amaru/src/mempool/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+    /// The current view of the ledger state.
+    ledger_state: ledger::state::State<S, E>,
+
+    /// The ephemeral state, which is applied on top of the ledger_state using transactions
+    /// in the mempool.
+    ephemeral: ledger::state::VolatileDB,
+}
+
+/// A transaction in the mempool.
+pub trait MempoolTx<M>: Ord + PartialEq {
```
abailly, line 30: What is `M` here?

amaru/src/mempool/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+pub struct State<S, E>
+where
+    S: ledger::store::Store<Error = E>,
+{
+    /// The current view of the ledger state.
+    ledger_state: ledger::state::State<S, E>,
+
+    /// The ephemeral state, which is applied on top of the ledger_state using transactions
+    /// in the mempool.
+    ephemeral: ledger::state::VolatileDB,
```
abailly, line 26: I am not sure why we represent the `ephemeral` state as a queue: transactions can be removed arbitrarily from the mempool and being able to "reset" the set at an earlier point _sequentially_ does not seem useful. Or perhaps I misunderstand how the `VolatileDB` is supposed to work?

amaru/src/mempool/mod.rs
abailly: Ah I see, the `Metadata`. Perhaps it would help to use the full word as the name of the type variable?

amaru/src/mempool/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+
+/// The mempool is a queue of transactions that are waiting to be validated and applied to an
+/// ephemeral ledger state.
+pub struct Mempool<T, M>
+where
+    T: MempoolTx<M>,
+{
+    /// The queue of transactions. The ordering of transactions are defined by the Ord trait of the
+    /// type T and left to the implementer.
+    transactions: Arc<RwLock<BTreeSet<T>>>,
```
abailly: It's a bit odd the comment says this is a "queue of transactions" whereas the type is a "set", but I understand the intention of using a `BTreeSet` here. Perhaps a type alias would help? 
Also, the `Arc/RwLock` decorator should probably be documented somewhere, perhaps by saying the data structure is threadsafe?

amaru/src/mempool/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+impl<T: MempoolTx<M>, M> Default for Mempool<T, M> {
+    fn default() -> Self {
+        Mempool {
+            transactions: Arc::new(RwLock::new(BTreeSet::new())),
+            _marker: Default::default(),
+        }
+    }
+}
+
+impl<T: MempoolTx<M>, M> Mempool<T, M> {
```
abailly, line 67: I think what would help me (and others) better review this design is to provide some kind of model embodying the expected properties and behaviour of the mempool as tests. This might strengthen the case for the data structures used or on the contrary suggests we might use something else, and would expose more clearly the interaction between mempool and other parts of the system.
For example, how does removing a transaction plays with the ledger state? When a tx is added/removed, what happens to the `State`? In general, what are the invariant(s) we want the mempool to guarantee b/w State and the queue of transactions it holds? 
This does not require real transactions and does not need to be super involved, but I think it would help to expose potential shortcomings of the design, or validate it.

amaru/src/mempool/mod.rs
AndrewWestberg: The ephemeral state is where we apply the mempool txns on top of the existing ledger state. This is the same object that is used internally in the ledger state to keep part of it volatile and subject to rollbacks. I feel like the same thing could be utilized in a mempool.  So, a LedgerState has both stable and volatile transactions.  A MempoolState has a stable, volatile, and ephemeral transactions.

So, let's say a rollback occurs on-chain in the volatile section, that should trigger a rebuilding of the ephemeral transactions.

It could also be that a transaction is removed from the mempool by a user, or something is re-ordered. This should also trigger a rebuild of the ephemeral transactions.

When we go to make a block, we have a nice ordered list of transactions in the ephemeral state we can take a bite of and submit as a block.

amaru/src/mempool/mod.rs
AndrewWestberg: `Arc/RwLock` does not need to be documented as this is one of the most common patterns seen in async rust code for thread safety. People typically use either `Arc/Mutex` or `Arc/RwLock`. The former causes exactly one reader or one writer to access the data at a time. The later allows for a single writer, but multiple concurrent readers. Since we will have different mempools we are maintaining at once, each of them having simultaneous read access to the transactions makes the most sense.

I'll create a type alias for the BTreeSet as that will add clarity to the code.

amaru/src/mempool/mod.rs
abailly: My understanding of the purpose of this structure from my discussions with @KtorZ is that it's sequential and meant for fast and easy handling of rollbacks: Should you have a rollback to some previous transaction, you  retrieve the cached state from the queue then reapply from that point on. This is useful in the context of the ledger/consensus because there might be a large number of blocks and transactions involved.

My understanding is that the behaviour of the mempool is different: It's perfectly possible that given a list of transactions say. `[tx1, tx2 ... tx_10]` in the mempool, you would need to remove say `tx_5` and `tx_8`, which would mean that you would rollback to `tx_4` and then need to reapply the remaining txs. Given the expected size of the mempool, and the fact you don't need to redo phase 2 validations but only phase 1 which is very fast, I would argue that this `VolatileState` is a premature optimisation.

amaru/src/mempool/mod.rs
abailly: I wasn't clear in my comment (😂 ): I meant not to document the implementation, but the behaviour the implementation provides, eg. that `transactions` is thread-safe.

amaru/src/mempool/mod.rs
AndrewWestberg: I can move forward with it and start implementing, but what I wanted to avoid by making this a `Draft` PR was doing a ton of work and then people criticizing the design late in the game. I wanted to get early feedback on it to make sure it's on the right track.

I agree that it would make more sense if there were an example implementation where you could see what happens on block rollbacks for example.  This will be the next phase for this PR, but I wanted to make sure I'm at least somewhat in the ballpark before moving forward.

amaru/src/mempool/mod.rs
AndrewWestberg: I'm happy to use whatever we decide to do as a codebase. If we want more verbose generic parameters, that's fine, but we should probably do it project-wide to lessen confusion.  It's fairly idiomatic rust to just use a single capital letter, but only if it's not adding confusion.  For now, I'll update the doc comment.

abailly: I agree one letter type arguments is much more common, but it makes me sad :)

amaru/src/mempool/mod.rs
abailly: > doing a ton of work and then people criticizing the design late in the game

I can relate. I would not go for a full blown implementation, but having a small model to play with and understanding the dynamics of the proposed design would help, ie. a couple of tests showing how the interfaces are used, how the state evolves, what invariants we need to maintain...

amaru/src/mempool/mod.rs
```diff
@@ -0,0 +1,107 @@
...
+{
+    /// The current view of the ledger state.
+    ledger_state: ledger::state::State<S, E>,
+
+    /// The ephemeral state, which is applied on top of the ledger_state using transactions
+    /// in the mempool.
+    ephemeral: ledger::state::VolatileDB,
+}
+
+/// A transaction in the mempool with optional _M_ metadata type useful for ordering transactions.
```
abailly, line 29: ❤️

amaru/src/mempool/mod.rs
KtorZ: Indeed; I am not sure the volatile state is best suited for storing the transient mempool. Mostly also because it's a VecDeque, which has fast access to both ends of the vector but not much to arbitrary elements. 

For a mempool, we might need something with fast access to arbitrary transactions and possibly with extra metadata such as their size and execution units. 


To be frank, I find the weighted RoseTree from the Haskell implementation not bad; with good amortised performances. Another option could be a weighted Patricia trie; which should work quite well with transaction ids as keys.

amaru/src/mempool/mod.rs
```diff
@@ -0,0 +1,107 @@
...
+        self.transactions.write().unwrap().extend(txs);
+    }
+
+    pub fn remove(&self, tx: &T) {
+        self.transactions.write().unwrap().remove(tx);
+    }
+
+    pub fn contains(&self, tx: &T) -> bool {
+        self.transactions.read().unwrap().contains(tx)
+    }
```
KtorZ, line 106: I think this is a lovely interface for a mempool implementor. Yet, for the block-forging interface, we might need something less granular and more tailored to forging-specific operation; otherwise we might not be able to provide a fast implementation for those specific operations.

For example, we will likely need to have the ability to get a bunch of transactions that satisfy specific criteria w.r.t block size and execution units. I don't think it should be the responsibility of the caller to decide on how that batch of transactions is made. 

Thoughts?

amaru/src/mempool/mod.rs
abailly: Given we don't really know, this is a cache and optimisation, and this might be heavily dependent on performance measures, I would leave that out for the moment and go for the simplest solution: 
* Keep 2 states an "anchor" and a "tip"
* Every time txs are added, update the "tip"
* Every time txs are removed, revalidate all txs and define new "tip" 
* Every time underlying chain changes, change "anchor", revalidate txs and define new "tip"

amaru/src/mempool/mod.rs
abailly: Actually, you also need an interface to notify the pool its base state has changed.

abailly: A diagram would be useful

KtorZ:
We'll revisit this in due time, but I am closing the PR for now as @AndrewWestberg isn't around to drive it much and we are focusing on slightly different areas at the moment.

PR #39, opened 2024-12-18
feat: implement slot arithmetic

rrruko:
this is a draft PR for slot arithmetic to help align on implementation as per last week's meeting.

this implements slot-to-time, time-to-slot, slot-to-epoch, and epoch boundary functions. these functions depend on the current state of the node. in particular, the start and end times, the slot lengths, and the epoch lengths of all known eras are required. this information is captured by an `EraHistory` type roughly corresponding to the `Interpeter` type used in [`Ouroboros.Consensus.HardFork.History.Qry`](https://github.com/IntersectMBO/ouroboros-consensus/blob/592e11c8ce941d10aaf185ca58796f25e6f403f0/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/History/Qry.hs).

just like in `Qry.hs`, these functions require that the latest era in the era history argument ends at the epoch of the safe zone (see [`Ouroboros.Consensus.HardFork.Combinator.State.Infra`](https://github.com/IntersectMBO/ouroboros-consensus/blob/592e11c8ce941d10aaf185ca58796f25e6f403f0/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/HardFork/Combinator/State/Infra.hs#L160-L257)), so care must be taken by the caller to ensure that an up-to-date era history is supplied. the end of the era history expresses that we can't definitively predict times beyond that point.

to elaborate on the details, as far as i understand:
- an era transition is considered to have been announced if the block confirming the transition is 'stable', i.e. it is at least k blocks behind the tip.
- at any given time, if an era transition has not been announced, it is guaranteed not to happen in the next 3k/f slots (this is the "safe zone").
- because era transitions only happen at epoch boundaries, an era transition is guaranteed not to happen until the next epoch after the end of the safe zone.
- this means that we can safely use the current era's parameters until the start of the epoch after the end of the safe zone.

the expectations for this deliverable have been kept pretty loose so far, so i'm open to any feedback.


crates/amaru/src/time.rs
KtorZ, line 1: Let's put that in another crate actually 🫡

crates/amaru/src/time.rs
```diff
@@ -0,0 +1,274 @@
+#[derive(PartialEq, Eq)]
+pub struct Bound {
+    pub bound_time: u64, // Milliseconds
+    pub bound_slot: u64,
+    pub bound_epoch: u64,
```
KtorZ:
```suggestion
    pub time: u64, // Milliseconds
    pub slot: u64,
    pub epoch: u64,
```

crates/amaru/src/time.rs
```diff
@@ -0,0 +1,274 @@
...
+            let epoch_number = era.start.bound_epoch + epochs_elapsed;
+            return Ok(epoch_number)
+        }
+    }
+    return Err(TimeHorizonError::PastTimeHorizon)
+}
+
+pub struct EpochBounds {
+    pub start_slot: u64,
+    pub end_slot: u64,
```
KtorZ:
```suggestion
    pub start: Slot,
    pub end: Slot,
```

And define somewhere `type Slot = u64`, that better convey the intent IMO, even though it's semantically the same.

crates/amaru/src/time.rs
```diff
@@ -0,0 +1,274 @@
...
+    pub bound_time: u64, // Milliseconds
+    pub bound_slot: u64,
+    pub bound_epoch: u64,
+}
+
+#[derive(PartialEq, Eq)]
+pub struct EraParams {
+    pub epoch_size: u64,
+    pub slot_length: u64, // Milliseconds
+}
```
KtorZ: Given that `epoch_size` and `slot_length` can't be null, it might be worth making those fields private and defining a smart constructor enforcing those invariants.

crates/amaru/src/time.rs
```diff
@@ -0,0 +1,274 @@
...
+
+// The start is inclusive and the end is exclusive. In a valid EraHistory, the
+// end of each era will equal the start of the next one.
+#[derive(PartialEq, Eq)]
+pub struct Summary {
+    pub start: Bound,
+    pub end: Bound,
```
KtorZ: If i recall correctly, eras don't necessarily have an upper-bound (the last one in particular). At least, it used to be like that but I recall it was changed around Babbage. Not sure how anymore. To be confirmed 😅 ...

crates/amaru/src/time.rs
```diff
@@ -0,0 +1,274 @@
...
+#[derive(PartialEq, Eq)]
+pub enum TimeHorizonError {
+    PastTimeHorizon,
+    InvalidEraHistory,
+}
+
+// The last era in the provided EraHistory must end at the time horizon for accurate results. The
+// horizon is the end of the epoch containing the end of the current era's safe zone relative to
+// the current tip. Returns number of milliseconds elapsed since the system start time.
+pub fn slot_to_relative_time(eras: &EraHistory, slot: u64) -> Result<u64, TimeHorizonError> {
```
KtorZ: Seems like all functions in this module are defined on `&EraHistory`, so perhaps consider making them methods of that type?

crates/amaru/src/time.rs
```diff
@@ -0,0 +1,274 @@
...
+    pub start: Bound,
+    pub end: Bound,
+    pub params: EraParams,
+}
+
+// A complete history of eras that have taken place.
+#[derive(PartialEq, Eq)]
+pub struct EraHistory {
+    pub eras: Vec<Summary>,
+}
```
KtorZ: Note: we'll eventually have to persist this on disk; so defining CBOR (minicbor) encoding and decoding for it would be useful. JSON (via serde) is also arguably useful, though priority is on CBOR.

crates/amaru/src/time.rs
```diff
@@ -0,0 +1,274 @@
...
+        }
+        let t1 = slot_to_relative_time(&eras, 172800);
+        match t1 {
+            Ok(t) => {
+                assert_eq!(t, 172800000);
+            }
+            _ => {
+                panic!("expected no error");
+            }
+        }
```
KtorZ:
```suggestion
        assert_eq!(t1, Ok(172800000));
```

KtorZ: Both more concise and will provide a better error message showing the actual unexpected value.

crates/amaru/src/time.rs
```diff
@@ -0,0 +1,274 @@
...
+        let bounds1 = epoch_bounds(&eras, 10);
+        match bounds1 {
+            Err(TimeHorizonError::PastTimeHorizon) => {
+            }
+            _ => {
+                panic!("expected error");
+            }
+        }
```
KtorZ:
```suggestion
        assert_eq!(epoch_bounds(&eras, 10), Err(TimeHorizonError::PastTimeHorizon))
```

crates/amaru/src/time.rs
```diff
@@ -0,0 +1,274 @@
...
+        let e3 = slot_to_epoch(&eras, 864001);
+        match e3 {
+            Err(TimeHorizonError::PastTimeHorizon) => {
+            }
+            _ => {
+                panic!("expected error");
+            }
+        }
+    }
```
KtorZ: Same for all those 😅, pardon my laziness to suggest changes.

KtorZ (commented):
Nice start, thanks for the early PR!

crates/time/src/lib.rs
```diff
@@ -0,0 +1,230 @@
...
+        }
+        Some(EraParams {
+            epoch_size,
+            slot_length,
+        })
+    }
+}
+
+// The start is inclusive and the end is exclusive. In a valid EraHistory, the
+// end of each era will equal the start of the next one.
```
Quantumplation, line 102: Is this another constraint it makes sense to validate / enforce in an `EraHistory` constructor @KtorZ?

i.e. just pass in a `Vec<(Bound, Params)>`, and then use the `start` of the next one as the `end` of the previous one, with the last one taking a `None`?

crates/time/src/lib.rs
```diff
@@ -0,0 +1,319 @@
+use minicbor::{Encode, Decode};
+
+type Slot = u64;
+
+#[derive(Clone, PartialEq, Eq)]
+pub struct Bound {
+    pub time: u64, // Milliseconds
```
abailly: suggestion: either rename to `time_ms` or have a `type Millis = u64` to remove the comment

crates/time/src/lib.rs
```diff
@@ -0,0 +1,319 @@
...
+    pub slot: Slot,
+    pub epoch: u64,
```
abailly, line 9: I think I would also have a `type Epoch = u64` but maybe overkill?

crates/time/src/lib.rs
```diff
@@ -0,0 +1,319 @@
...
+        self.epoch.encode(e, ctx)?;
+        e.end()?;
+        Ok(())
+    }
+}
+
+#[derive(Clone, PartialEq, Eq)]
+pub struct EraParams {
+    epoch_size: u64,
+    slot_length: u64, // Milliseconds
```
abailly, line 56: Same remark than above

crates/time/src/lib.rs
```diff
@@ -0,0 +1,319 @@
...
+        self.slot.encode(e, ctx)?;
+        self.epoch.encode(e, ctx)?;
+        e.end()?;
+        Ok(())
+    }
+}
+
+#[derive(Clone, PartialEq, Eq)]
+pub struct EraParams {
+    epoch_size: u64,
```
abailly: I would make the unit explicit, either through the type or name

crates/time/src/lib.rs
```diff
@@ -0,0 +1,319 @@
...
+                        epoch: 2,
+                    },
+                    params: params.clone(),
+                },
+            ],
+        };
+        let t0 = eras.slot_to_relative_time(172801);
+        assert_eq!(t0, Err(TimeHorizonError::PastTimeHorizon));
+        let t1 = eras.slot_to_relative_time(172800);
+        assert_eq!(t1, Ok(172800000));
```
abailly: I am not a big fan of testing several things in a single test: When the test fails, it's hard to know quickly what's failed, you would have to look at the source code of the test to know exactly what broke, which is annoying.

I would suggest to:
* not prefix test functions with `test_`,  we know they are tests because there's an annotation just above
* have more explicit test names (eg. `slot_to_time_fails_given_its_past_upper_bound`)
* have each function test a single "thing" (possibly using parameterised tests to avoid duplication)

crates/time/src/lib.rs
```diff
@@ -0,0 +1,319 @@
...
+        assert_eq!(eras.epoch_bounds(1).unwrap().end, 172800);
+        assert_eq!(eras.epoch_bounds(10), Err(TimeHorizonError::PastTimeHorizon));
+    }
+
+    #[test]
+    fn test_slot_to_epoch() {
+        let params = EraParams::new(86400, 1000).unwrap();
+        let eras = EraHistory {
+            eras: vec![
+                Summary {
```
abailly: You could remove duplication by lifting this definition to be a constant within the test module perhaps?

crates/time/src/lib.rs
```diff
@@ -0,0 +1,319 @@
...
+        let e1 = eras.slot_to_epoch(86399);
+        assert_eq!(e1, Ok(0));
+        let e2 = eras.slot_to_epoch(864000);
+        assert_eq!(e2, Ok(10));
+        let e3 = eras.slot_to_epoch(864001);
+        assert_eq!(e3, Err(TimeHorizonError::PastTimeHorizon));
+    }
+
+    #[test]
+    fn test_encode_era_history() {
```
abailly: One would typically want to have unit tests for encoding/decoding and a roundtrip property which is quite powerful to ensure whatever is encoded can be decoded and find edge cases.

abailly (commented):
It's one of those parts of the system where I think we need test vectors to ensure computations are identical across implementations. That's something I could contribute if you think that's useful (but not next week :) )

crates/time/Cargo.toml
```diff
@@ -12,5 +12,9 @@ readme = "README.md"
 rust-version = "1.81.0"
 
 [dependencies]
+quickcheck = "1.0.3"
```
KtorZ: We're already using `proptest` in other parts of the codebase; Although I am reasonably fond of Haskell's QuickCheck, I didn't really find the Rust version to be as convenient / practical to use. 

In particular, QuickCheck gets most of its value when values are properly shrunk, though here it seems you've completely skipped shrinking altogether 😅 ... `proptest` is a bit more Rust-idiomatic (although the API is very macro-based), and comes with integrated shrinking. 

Is there any reason to stick with QuickCheck that I miss here? If not, let's switch to proptest (and possibly start some kind of _contribution guidelines_ with these elements ... I am sorry to come late with this).

KtorZ (approved):
I have only rapidly skimmed through the module and the test, but it looks okay. We'll confirm and iterate once put in use at a later stage.

Thanks also @Quantumplation for jumping in and switching tests to prop_test! ; I feel we could write more interesting properties than the roundtrip since the generator isn't just a structural one but seems to be trying to generating semantically valid era history!

KtorZ:
Although, we have the same issue as with #48... @rrruko didn't sign his commits so we cannot merge this into Amaru.

Quantumplation:
@KtorZ for now i'll rebase and sign the commits myself

PR #42, opened 2024-12-20
Track stake distribution across epochs

KtorZ:


abailly (approved):
Ouch  :) 

> +1,431,355 −572,246 

I am not going to review this, I trust you.

KtorZ:
That's mostly the snapshots data though. The actual changes aren't that large, though the diff on the import command is perhaps hard to read.

Quantumplation:
I imagine that's going to be a recurring problem for this kind of work; Random idea: what if we stored snapshot tests in a separate but closely related repo, perhaps with a git submodule, to keep diffs less intimidating?

KtorZ:
We could have them as git sub modules I suppose. Those tests are ignored by default anyway because they require a synchronised node ledger database to run. So they can easily be taken out of the main source code.

KtorZ:
As discussed off-band, merging this, so we can move on with other stuff on top. It isn't complete (still missing rewards calculation), but current `master` isn't either so it's not like the ledger state is _more wrong_ 🤷. Actually gets us one step closer.

PR #43, opened 2024-12-21
fix: release mutex sooner

jeluard:
Release the `mutex` locking ledger access onc not needed anymore. On my machine this reduce `forward` time from couple hundreds ms to a couple hundreds μs.
Note that a better but more involved solution might be to reduce the need to interact with the lock per block. This fix allows to move forward while I investigate a potentially better option.

*Before*

Capture d’écran 2024-12-20 à 20 46 24

*After*

Capture d’écran 2024-12-20 à 20 45 18

### DIFF
--- a/crates/amaru/src/consensus/mod.rs
+++ b/crates/amaru/src/consensus/mod.rs
@@ -94,6 +94,9 @@ impl gasket::framework::Worker<Stage> for Worker {
                 let ledger = stage.ledger.lock().await;
                 assert_header(&header, &stage.epoch_to_nonce, &*ledger)?;
 
+                // Make sure the Mutex is released as soon as possible
+                drop(ledger);
+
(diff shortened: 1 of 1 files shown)


abailly:
There will be "a lot" of interactions with the ledger state which is a central structure in the system, it would be great to investigate lock-free solutions to avoid having to troubleshoot deadlocks further down the road. I don't have any solidified idea, but perhaps something like a ring-buffer with a queue of changes to the ledger state applied from a single writer could be interesting?

abailly (approved):

jeluard:
>  it would be great to investigate lock-free solutions to avoid having to troubleshoot deadlocks further down the road

Definitely. I don't have yet a clear picture of the various components involved lifecycle and interactions, but planning to look into that.

PR #44, opened 2024-12-21
Update config.yml to allow free-form issues

abailly:
I don't see the point of having too stringent requirements on issues at this stage.

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,4 +1,4 @@
-blank_issues_enabled: false
+blank_issues_enabled: true
 contact_links:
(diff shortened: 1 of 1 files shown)


wolf31o2 (approved):

KtorZ:
I suspect that what you want from issue now by doing this change is probably what I rather have us do in _discussions_? 

My intent with constraining issues as such was to keep issues exclusively for bug reports, while discussions would be more of a free form format with category/topics.

abailly:
> I suspect that what you want from issue now by doing this change is probably what I rather have us do in _discussions_?

No. I just want to be able to use issues to build a reasonably useful roadmap and clarify things that need to be done.

PR #45, opened 2024-12-27
Add a friendly panic hook

Quantumplation:
A block producing node, and Amaru, is considered "mission critical" software, for the following reasons:
- The network is relying on the availability and correctness of each alternative implementation in proportion to its stake
- The stake pool operators are (i.e. eventually will be) relying on the continuing functioning of the block producer as a form of income

For that reason, the node should never just hard "crash"; even in error cases, the node should respond to these bad conditions in a more graceful way, perhaps eventually even including raising an alarm via any metrics / alerting systems built into the node.

Therefore, we install a panic hook; if a crash does arise in the wild, we'd *love* a bug report, so we can address it more appropriately. For that reason, we want to make the process to open the issue as welcoming as possible.

Here's what it looks like for an explicit panic:
image

Here's what it looks like for an implicit panic, like divide by zero:
image

Here's what it looks like for a contextual panic, like `.expect()`:
image

I included a commit in the history with an explicit panic; if you want to test it yourself, you can checkout `e9bfdf5931ff6b093e830bf0252c6be29eaa7372` and run `cargo run daemon --peer-address 127.0.0.1`.

Note that I borrowed a lot of code from Aiken; For this PR, I just included the needed utility functions in the `panic.rs` file, but we might want to pull some of these into their own files / modules, as they are in Aiken.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -58,7 +58,6 @@ dependencies = [
  "opentelemetry_sdk",
  "ouroboros",
  "ouroboros-praos",
- "owo-colors",
  "pallas-addresses",
  "pallas-codec",
  "pallas-crypto",
@@ -69,7 +68,6 @@ dependencies = [
  "proptest",
  "rocksdb",
  "serde",
- "strip-ansi-escapes",
  "thiserror 2.0.3",
  "tokio",
  "tokio-util",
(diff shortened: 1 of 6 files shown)


KtorZ (approved):
Thanks, I simply got rid of owo-colors for now; I am not sure we want to introduce colors and other ANSI symbols just yet as they've proven to be slightly painful to work with in Aiken. If we do so, I'd be looking for something more holistic that allows turning it on and off more easily.

PR #47, opened 2024-12-28
Fix tracing on process exit; instrument Import command

Quantumplation:
I was playing around with instrumentation to get a feel for it, and test out my fix to the Grafana / tempo setup, and I wanted something that was a bit more nested than what we had in the sync, so I instrumented the import command.

This command is likely temporary, but served as a good test-bed / example for tracing. Indeed, it allowed me to discover a bug in capturing trace information when the process exits before the tracing provider has had a chance to export it from the process.

Now, Grafana consistently reports the full traces for an import, including in the presence of errors, etc.

image

See the commit messages for more details; in particular, I'm not attached to this specific instrumentation, so feel free to suggest changes if you think it's overboard.

This branch is based on top of my pi/fix-tempo branch, so also includes that fix.

### DIFF
--- a/crates/amaru/src/bin/amaru/cmd/import.rs
+++ b/crates/amaru/src/bin/amaru/cmd/import.rs
@@ -14,9 +14,10 @@ use indicatif::{ProgressBar, ProgressStyle};
 use miette::{Diagnostic, IntoDiagnostic};
 use pallas_codec::minicbor as cbor;
 use std::{collections::HashMap, fs, iter, path::PathBuf};
-use tracing::info;
+use tracing::{info, info_span};
 
 const BATCH_SIZE: usize = 5000;
+const EVENT_TARGET: &str = "amaru::import";
 
 #[derive(Debug, Parser)]
 pub struct Args {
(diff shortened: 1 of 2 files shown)


crates/amaru/src/bin/amaru/cmd/import.rs
```diff
@@ -286,14 +404,28 @@ pub async fn run(args: Args) -> miette::Result<()> {
...
+        db.with_pools(|iterator| {
+            for (_, pool) in iterator {
+                pools::Row::tick(pool, epoch + 1)
+            }
+        })
+        .into_diagnostic()?;
+    }
 
     Ok(())
 }
```
KtorZ, line 431: I appreciate the effort in this file but I still want to cry a little because I greatly reworked that file in #42.

KtorZ:
(I've cherry-picked your first commit).

crates/amaru/src/bin/amaru/cmd/import.rs
Quantumplation: Hah that's fine we can discard the changes; like I said, it was mostly an interesting test before for me to make sure tracing was working etc.

Quantumplation: I'd still keep the shutdown fix though

KtorZ: As said, cherry-picked already 🫡

PR #55, opened 2025-01-13
Import ouroboros repository as a crate

abailly:
The ouroboros repository was meant to contain the "pure" ouroboros consensus part of Amaru in order to support more easily concurrent development of the various components. We merge them back as it seems simpler to experiment with various designs in a monorepo.

## Release Notes

- **New Features**
    - Added new Ouroboros crate with comprehensive consensus, cryptographic, and validation modules
    - Introduced support for Verifiable Random Function (VRF) and Key Encapsulation Scheme (KES) functionality
    - Implemented block validation and ledger state management components

- **Dependency Updates**
    - Updated dependencies in Amaru crate
    - Replaced external Ouroboros dependencies with local path reference

- **Infrastructure**
    - Added new workspace member for Ouroboros crate
    - Configured Criterion benchmarking settings
    - Enhanced project-wide error handling and type definitions

- **Testing**
    - Added comprehensive test suites for validation and cryptographic components
    - Introduced benchmarking capabilities for block validation performance


crates/amaru/src/consensus/header_validation.rs
```diff
@@ -1,7 +1,7 @@
 use crate::{consensus::ValidateHeaderEvent, ledger::kernel::epoch_from_slot, sync::PullEvent};
 use gasket::framework::*;
+use ouroboros::consensus::BlockValidator;
```
jeluard, line 3: We might consider some clippy rules to avoid those kind of changes, if that's necessary.

abailly: Not sure I understand what you mean @jeluard? What kind of changes?

jeluard: The change in `use` ordering. Nitpick :)

rvcas:
Shall I do the same for uplc?

KtorZ:
> Shall I do the same for uplc?

Yes. Please. Try to preserve commit history if you can (`git filter-branch` may be handy). So that you can not only flex later, but that'll be helpful in the future when digging back this stuff.

abailly:
@KtorZ The reason why this PR is still in draft is that it's red: I need to fix clippy warnings, but more importantly I need to fix my local dev env to ensure I don't miss those warnings

jeluard:
@abailly This PR might help to catch all CI warnings: https://github.com/pragma-org/amaru/pull/56

jeluard (approved):

PR #56, opened 2025-01-13
chore: rely on cargo aliases

jeluard:
Introduce cargo `aliases` so that the same commands can be easily shared between local dev env and CI.
I introduced a `a` prefix (fo `amaru` ?) to distinguish those, feel free to suggest a better alternative :)

Can be used as:

```shell
cargo atest --target wasm32-unknown-unknown
```

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,6 @@
+# Find the full list of available options here: https://doc.rust-lang.org/cargo/reference/config.html
+
+[alias]
+acfmt = "fmt --all -- --check"
+aclippy = "clippy --workspace --all-targets --all-features -- -D warnings"
+atest = "test --workspace --all-targets --all-features "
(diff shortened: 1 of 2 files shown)


.cargo/config.toml
```diff
@@ -0,0 +1,6 @@
+# Find the full list of available options here: https://doc.rust-lang.org/cargo/reference/config.html
+
+[alias]
+acfmt = "fmt --all -- --check"
+aclippy = "clippy --workspace --all-targets --all-features -- -D warnings"
+atest = "test --workspace --all-targets --all-features "
```
KtorZ: words are cheap, I suggest we use longer aliases just the sake of clarity? 

```
fmt-amaru
clippy-amaru
test-amaru
```

?

jeluard: Sounds better!

KtorZ (approved):

PR #58, opened 2025-01-13
chore: add raspberry PI support for CI

jeluard:
Add CI support for  `aarch64` on linux. This allows to ensure that `amaru` can build for recent raspberry (including `zero`).

Running on a rapsberry 5:
demo

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -27,6 +27,10 @@ jobs:
             target: aarch64-apple-darwin
           - runner: windows-latest
             target: x86_64-pc-windows-msvc
+          - runner: ubuntu-latest
+            target: aarch64-unknown-linux-gnu
+            optional: true
+            cross-compile: true
     timeout-minutes: 30
     runs-on: ${{ matrix.environments.runner }}
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/CI.yml
```diff
@@ -28,6 +28,11 @@ jobs:
           - runner: windows-latest
             target: x86_64-pc-windows-msvc
             optional: true
+          - runner: ubuntu-latest
+            target: aarch64-unknown-linux-gnu
+            optional: true
+            build-only: true
```
KtorZ: Debatable; If we're going to support another architecture, we might as well test on it as well. That should _in principle_ always be fine, but I wouldn't necessarily disable tests on those.

jeluard: Good point! As the `raspberry` target relies on cross-compilation (as opposed to being compiled on a real hardware), `cargo test` fails (as it runs the real binary).
An option could be to have a `self-hosted` runner here if we want to have `raspberry` as a first-class citizen.

jeluard: Actually looks like [cross](https://github.com/cross-rs/cross) does the heavy lifting here and allows to run `cross-test` (via `docker`). Looks simpler.

KtorZ: > As the raspberry target relies on cross-compilation 

Ho, didn't know that :( ... That's a bummer, and indeed, diminishes a lot the value we would get from testing. So I suppose it's fine, but let's capture this as a comment/justification in the workflow configuration. Because I'll forget that conversion tomorrow already.

jeluard: Actually [cross](https://github.com/cross-rs/cross) solves this and tests can be run too. I made the necessary changes in a more recent commit.

I also validated that the generated binary runs fine on a real raspberry when using `musl` and target `aarch64-unknown-linux-musl`. Maybe we should directly use the musl `target` for `CI` too, as it's most probably the one we should use to ship binaries.

KtorZ: > validated that the generated binary runs fine on a real raspberry

Yolo. Love it. 

Goal for next month: keep a running Amaru node in the office on your raspberry.

PR #59, opened 2025-01-14
chore: use fancy miette reporting

jeluard:
Use miette [fancy reporting](https://docs.rs/miette/latest/miette/#installing). This helps improve the error displayed in the console and remove this note: 
`NOTE: If you're looking for the fancy error reports, install miette with the `fancy` feature, or write your own and hook it up with miette::set_hook().`

An alternative is to not use `fancy` reporting, but find a way to disable this `NOTE`.

**BEFORE**

```shell
Error: Diagnostic { message: "No such file or directory (os error 2)" }
````

**AFTER**

```shell
Error: 
  × No such file or directory (os error 2)
````


rvcas:
Nice, I was wondering why the feature wasn't already turned on

KtorZ (approved):

PR #63, opened 2025-01-16
Minor cleanups related to ledger

jeluard:
While preparing the work for https://github.com/pragma-org/amaru/discussions/53 I took the opportunity to reduce some object allocation and do some minor cleanups.

### DIFF
--- a/crates/amaru/src/ledger/state.rs
+++ b/crates/amaru/src/ledger/state.rs
@@ -4,10 +4,9 @@ pub mod diff_set;
 
 use crate::ledger::{
     kernel::{
-        self, block_point, epoch_from_slot, output_lovelace, Certificate, Epoch, Hash, Hasher,
-        Lovelace, MintedBlock, Point, PoolId, PoolParams, PoolSigma, StakeCredential,
-        TransactionInput, TransactionOutput, CONSENSUS_SECURITY_PARAM, STABILITY_WINDOW,
-        STAKE_CREDENTIAL_DEPOSIT,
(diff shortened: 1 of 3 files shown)


crates/amaru/src/ledger/state.rs
```diff
@@ -188,64 +184,71 @@ impl<S: Store<Error = E>, E: std::fmt::Debug> State<S, E> {
 }
 
 /// Process a given block into a series of ledger-state diff (a.k.a events) to apply.
-fn apply_block(span: &Span, block: MintedBlock<'_>) -> VolatileState<()> {
+fn apply_block(parent: &Span, block: MintedBlock<'_>) -> VolatileState<()> {
     let failed_transactions = FailedTransactions::from_block(&block);
+    let transaction_bodies = block.transaction_bodies.to_vec();
+    let total_count = transaction_bodies.len();
+    let failed_count = failed_transactions.inner.len();
```
KtorZ: Pardon my Haskell here; I avoided using `.len` to avoid traversing the vector twice, which is why  those numbers were counted during the traversals. But I just double-checked and turns out `Vec` actually holds onto its length, so `len()` is merely just reading a variable. So, all good 👍

KtorZ:
Note: I rebased on top of `main` since my latest PR #54 brought a lot of conflicts in the `ledger/state.rs`. Also, this created a new clippy warning, which I fixed by introducing a new intermediate type `StoreUpdates`; although in the long run, we'll probably want to rework the store interface.

KtorZ (approved):

PR #64, opened 2025-01-19
Track withdrawals in ledger state + important fixes

KtorZ:
This PR solves the remaining discrepancy in the rewards calculations before governance ones kicks in. In particular, it now correctly reset rewards account on withdrawals (which are always necessary withdrawing the full amount available on the account). 

Also, it includes a fix where accounts supposed to receive multiple rewards (either because a reward account is being used for multiple pools, or because it used both as a reward account for a pool and delegated to another -- thus receiving member rewards). 

The snapshots are now therefore correct all the way from epoch 163 to epoch 172 💪 ! Starting from epoch 173, things start to diverge again because this is the first epoch where a governance proposal gets refunded; which is quite visible from the diff (we're missing exactly 100k ADA): 

```diff
 "40b3c98d3b5f5a7748829d612e9e1ccd5f429970abc9abfeae867af9": {
-   "lovelace": 101001639743,
+   "lovelace": 1001639743,
    "pool": "pool1rccstu3l9ty3k0a5cd06fl3szsss9r34dcg5j38fqgq9kvng0tg"
},
``` 

## Release Notes

- **New Features**
	- Added support for withdrawal transactions in the ledger state management.
	- Introduced a new function to reset reward counters for multiple accounts.

- **Improvements**
	- Enhanced reward calculation logic to accumulate rewards correctly.
	- Updated database save methods to accommodate withdrawals.

- **Technical Updates**
	- Modified state tracking mechanisms to include withdrawal information.
	- Added new telemetry for monitoring stake distribution and rewards computation.

The release introduces more robust handling of blockchain transactions, particularly focusing on withdrawal processes and reward management.

### DIFF
--- a/crates/amaru/src/ledger/rewards.rs
+++ b/crates/amaru/src/ledger/rewards.rs
@@ -84,10 +84,10 @@ certain mutations are applied to the system.
 
 use crate::ledger::{
     kernel::{
-        encode_bech32, output_lovelace, output_stake_credential,
-        reward_account_to_stake_credential, Epoch, Hash, Lovelace, PoolId, PoolParams,
-        StakeCredential, ACTIVE_SLOT_COEFF_INVERSE, MAX_LOVELACE_SUPPLY, MONETARY_EXPANSION,
-        OPTIMAL_STAKE_POOLS_COUNT, PLEDGE_INFLUENCE, SHELLEY_EPOCH_LENGTH, TREASURY_TAX,
+        encode_bech32, expect_stake_credential, output_lovelace, output_stake_credential, Epoch,
+        Hash, Lovelace, PoolId, PoolParams, StakeCredential, ACTIVE_SLOT_COEFF_INVERSE,
+        MAX_LOVELACE_SUPPLY, MONETARY_EXPANSION, OPTIMAL_STAKE_POOLS_COUNT, PLEDGE_INFLUENCE,
+        SHELLEY_EPOCH_LENGTH, TREASURY_TAX,
     },
     store::{columns::*, Store},
 };
@@ -765,16 +765,8 @@ impl RewardsSummary {
 
         let rewards_leader = pool.leader_rewards(rewards_pot, owner_stake, total_stake);
 
-        let credential = reward_account_to_stake_credential(&pool.parameters.reward_account)
-            .unwrap_or_else(|| {
-                panic!(
-                    "unexpected malformed reward account: {:?}",
-                    &pool.parameters.reward_account
-                )
-            });
-
         accounts
-            .entry(credential)
+            .entry(expect_stake_credential(&pool.parameters.reward_account))
             .and_modify(|rewards| *rewards += rewards_leader)
             .or_insert(rewards_leader);
 
--- a/crates/amaru/src/ledger/rewards.rs
+++ b/crates/amaru/src/ledger/rewards.rs
@@ -681,7 +681,10 @@ impl RewardsSummary {
         if let Some(PoolRewards { pot, .. }) = pools.get(&st.pool) {
             let member_rewards = pool.member_rewards(&credential, *pot, st.lovelace, total_stake);
             if member_rewards > 0 {
-                accounts.insert(credential, member_rewards);
+                accounts
+                    .entry(credential)
+                    .and_modify(|rewards| *rewards += member_rewards)
+                    .or_insert(member_rewards);
             }
             member_rewards
         } else {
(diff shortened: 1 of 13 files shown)


KtorZ:
@coderabbitai full review

crates/amaru/src/ledger/rewards.rs
```diff
@@ -727,17 +730,18 @@ impl RewardsSummary {
...
-        );
+        let credential = reward_account_to_stake_credential(&pool.parameters.reward_account)
+            .unwrap_or_else(|| {
+                panic!(
+                    "unexpected malformed reward account: {:?}",
+                    &pool.parameters.reward_account
+                )
+            });
+
+        accounts
```
jeluard, line 768: It could be useful to extract those 3 lines in a function with a meaningful name, so that we don't need a comment to explicit the behavior and there is no future risk of having both functions diverge.

crates/amaru/src/ledger/state/transaction.rs
```diff
@@ -51,10 +52,21 @@ pub fn apply<T>(
...
     apply_certificates(&span, &mut state.pools, &mut state.accounts, certificates);
 
+    let withdrawals = transaction_body
+        .withdrawals
+        .unwrap_or_else(|| NonEmptyKeyValuePairs::Def(vec![]));
+    span.record("transaction.withdrawals", withdrawals.len());
+    state
+        .withdrawals
+        .extend(withdrawals.iter().map(|(account, _)| {
+            reward_account_to_stake_credential(account).unwrap_or_else(|| {
```
jeluard, line 65: Could be replaced with `expect`?
Now I'm wondering if we want to fail in those cases? And what happens if a panic is thrown here?

crates/amaru/src/ledger/store/columns/accounts.rs
```diff
@@ -104,6 +104,29 @@ pub mod rocksdb {
         Ok(())
     }
 
+    /// Reset rewards counter of many accounts.
+    pub fn reset<DB>(
+        db: &Transaction<'_, DB>,
+        rows: impl Iterator<Item = super::Key>,
```
jeluard, line 110: Usage of iterator looks a bit fragile to me here as iterator can only be traversed once. This could lead to confusion if some future changes consume the iterator beforehand.

KtorZ: My idea behind using an iterator here is double:

1. Avoid committing to a specific data-structure in this interface (since it needs not to)
2. Actually convey the idea that we only need to traverse (once) the data. 

Surely, if the caller partially consume the iterator before passing it then, it's going to be missing entries; but that's a caller's fault / problem (or own desire; as perhaps it is what the caller specifically wants); not an API issue IMO 🤔

crates/amaru/src/ledger/state/transaction.rs
KtorZ: I came to avoid using `expect` over time because it doesn't usually allow to provide good error messages with context. It boils down to the same thing in terms of behaviour (the program panics), but only provides a string of context. Whereas `unwrap_or_else` + `panic!` typically allows to embed more context with the panic, which is preferable. 

Having said that; another reason why I try to stick with using `panic!` is because it makes it easier to identify later on. For now, we don't necessarily try to handle behaviours that are slightly outside of what we expect. Though we may very likely revise that and ensures that we cannot crash too easily. So I sprinkle those `panic!` like TODOs 😅 ... as they become easy to find and revisit later (either by introducing an error at the value-level to carry across, or by replacing them with `unreachable!` in places where we can bring good evidence of truly unreachable paths).

crates/amaru/src/ledger/rewards.rs
KtorZ: Fair point! Will do.

crates/amaru/src/ledger/state/transaction.rs
```diff
@@ -51,10 +52,21 @@ pub fn apply<T>(
...
+                panic!("invalid reward account found in transaction ({transaction_id}): {account}")
+            })
+        }));
```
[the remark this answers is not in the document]
KtorZ: This is already ensured by the serialisation of withdrawals themselves. The on-the-wire format doesn't allow duplicate (as this counts as invalid serialisation). This is then reflected at the type-level in Rust. 

So while we may use vectors when passing around internal representations, we can guarantee that they are unique due to their origin (minted blocks)

crates/amaru/src/ledger/store/columns/accounts.rs
jeluard: Makes sense.
Looks like the [rust guideline](https://rust-lang.github.io/api-guidelines/flexibility.html#functions-minimize-assumptions-about-parameters-by-using-generics-c-generic) for functions consuming collections is to use [IntoIterator](https://doc.rust-lang.org/std/iter/trait.IntoIterator.html). Maybe something to consider.

jeluard (approved):

PR #71, opened 2025-01-22
Improvements to telemetry

jeluard:
Some small telemetry improvements:
* clarify error message in case of a communication error with the telemetry backend
* allow to disable telemetry export via the global CLI option `--disable-telemetry-export` (telemetry is still used for internal needs like logging)
* fix import idempotency 

## Release Notes

- **New Features**
  - Added option to disable telemetry export via command-line argument

- **Improvements**
  - Enhanced snapshot import logging
  - Improved error handling for tracing teardown
  - Updated snapshot creation process to handle existing directories more robustly

- **Documentation**
  - Added clarification about idempotent snapshot generation method

These updates provide more user control and improved system reliability during snapshot and telemetry operations.

### DIFF
--- a/crates/amaru/src/bin/amaru/main.rs
+++ b/crates/amaru/src/bin/amaru/main.rs
@@ -54,7 +54,9 @@ async fn main() -> miette::Result<()> {
     result
 }
 
-pub fn setup_tracing(disable_telemetry_export: bool) -> (TracerProvider, SdkMeterProvider, Counter<u64>) {
+pub fn setup_tracing(
+    disable_telemetry_export: bool,
+) -> (TracerProvider, SdkMeterProvider, Counter<u64>) {
     use opentelemetry::{metrics::MeterProvider, trace::TracerProvider as _, KeyValue};
     use opentelemetry_sdk::{metrics::Temporality, Resource};
     use tracing_subscriber::{prelude::*, *};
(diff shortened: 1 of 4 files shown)


crates/amaru/src/bin/amaru/main.rs
```diff
@@ -29,28 +29,34 @@ enum Command {
...
+    if let Err(report) = teardown_tracing(tracing, metrics) {
+        eprintln!("Failed to teardown tracing: {report}");
+    }
 
     result
 }
 
-pub fn setup_tracing() -> (TracerProvider, SdkMeterProvider, Counter<u64>) {
+pub fn setup_tracing(
+    disable_telemetry_export: bool,
```
abailly, line 58: I am not a huge fan of `bool` arguments in general, they are fragile, opaque, hard to interpret at call or use site, prone to errors...
Using a trait and 2 different implementations would be cleaner, eg.

```
trait LoggingSystem {
  fn setup();
  fn teardown();
}

struct SendToOTLP {
}

struct SendToStdout {
}

...
```
you get the idea I think :)

crates/amaru/src/ledger/store/rocksdb/mod.rs
```diff
@@ -247,6 +245,12 @@ impl Store for RocksDB {
                 })?;
             }
 
+            let path = self.dir.join(snapshot.to_string());
```
abailly, line 248: Nitpicking but: is this related to `Improvements to telemetry` ?

abailly (changes requested):
Some suggestion to avoid `bool` arguments

crates/amaru/src/bin/amaru/main.rs
jeluard: Makes sense, although I would suggest a simpler struct container for now. It looks to me the whole telemetry/logging setup should be refactored more globally.

crates/amaru/src/ledger/store/rocksdb/mod.rs
jeluard: Ah right, stumbled on this while testing multiple `import` in a row,  looked like something simple enough I could sneak in :)

crates/amaru/src/bin/amaru/main.rs
abailly: Agreed. I trust your judgment, as long as we mark this as temporary and subject to further refactoring. IMO this whole OTLP thing should be completely externalised, with Amaru logging JSON to stdout by default and the environment doing "the right thing" to feed those into whatever o11y infrastructure is there.

abailly (approved):

PR #73, opened 2025-01-23
feat: format logs as Json by default

abailly:
And allow user to select 'pretty' formatting through environment variables if they so choose.
I could not avoid the duplication although I tried for quite some time factoring types, parameters, bounds and what not to the point the compiler had to dump types to files!

## Summary by CodeRabbit

- **New Features**
	- Added support for configurable log formatting.
	- Introduced an environment variable to control log output style.

- **Documentation**
	- Updated monitoring README with instructions for log format customisation.

- **Improvements**
	- Enhanced tracing configuration to support JSON and pretty-print log formats.
	- Enabled dynamic log formatting at runtime based on environment settings.

### DIFF
--- a/crates/amaru/src/bin/amaru/main.rs
+++ b/crates/amaru/src/bin/amaru/main.rs
@@ -57,6 +57,7 @@ pub fn setup_tracing() -> (TracerProvider, SdkMeterProvider, Counter<u64>) {
 
     const AMARU_LOG: &str = "AMARU_LOG";
     const AMARU_DEV_LOG: &str = "AMARU_DEV_LOG";
+    const AMARU_LOG_FORMAT: &str = "AMARU_LOG_FORMAT";
 
     // Enabling filtering from env var; but disable gasket low-level traces regardless. We use the
     // env directive filtering here instead of the .or() / .and() Rust API provided on EnvFilter so
@@ -91,10 +92,6 @@ pub fn setup_tracing() -> (TracerProvider, SdkMeterProvider, Counter<u64>) {
         )
         .build();
     let opentelemetry_tracer = opentelemetry_provider.tracer(SERVICE_NAME);
-    let opentelemetry_layer = tracing_opentelemetry::layer()
-        .with_tracer(opentelemetry_tracer)
-        .with_filter(filter(AMARU_LOG));
-
     // Metrics
     // NOTE: We use the http exporter here because not every OTLP receivers (in particular Jaeger)
     // support gRPC for metrics.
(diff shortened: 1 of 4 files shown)


crates/amaru/src/bin/amaru/main.rs
```diff
@@ -120,15 +117,38 @@ pub fn setup_tracing() -> (TracerProvider, SdkMeterProvider, Counter<u64>) {
...
+                    .with_span_events(fmt::format::FmtSpan::ENTER | fmt::format::FmtSpan::EXIT)
+                    .with_filter(filter(AMARU_DEV_LOG)),
+            )
+            .with(
+                tracing_opentelemetry::layer()
+                    .with_tracer(opentelemetry_tracer)
+                    .with_filter(filter(AMARU_LOG)),
+            )
+            .init(),
+    }
```
KtorZ, line 151: I find having both JSON and OTL somewhat redundant. Though I agree that the current status quo (wall of "pretty" logs on stdout) isn't also quite acceptable. 

I talked with @jeluard today about this as he showed me what the main Polkadot client does (which I very much like): on startup, it displays various informations about the node's configuration, and a variety of status check (e.g. was the database found, what was the tip, etc...). Information that basically fit on a screen; and then succinct logs about synchronization status. 

So in essence, it's a dev/ops dashboard. I find that to be a good default. Now, it should be simple to switch to JSON logs on stdout instead, if that is what one's prefer. But then, I don't see the point of enabling OLT in this case. Concretely, my suggestion is:

- Default to a simple "dashboard" on stderr (either by logging stuff, or by actually using a quick TUI for this), enabling traces & spans via an open telemetry backend. 

- Allow disabling open telemetry, and switching to returning all traces on stdout via a command-line flag (e.g. `--json`); keeping (or not) the dashboard/logs on stderr. 

WDYT?

wolf31o2: Logs and telemetry serve different purposes, especially for me.

Logs are meant for human consumption, which is why we have "pretty" logs, but they can be structured instead of unstructured to allow for indexing and classification, especially at scale. Logs are also very much "opt-in" in that a developer needs to create log messages to produce information they're seeking. However, the information is customizable and very rich.

Telemetry is about measurements and statistics. While we can send specific spans for specific events, it's usually tied to performance and tracing of requests and flows. Generally, this gets hooked into at a low level and is "automatic" such as being shoved into a service handler. It's point data and limited to simple measurements and metadata about them.

One is an information stream of application events generally for humans while the other is an information stream of application measurements generally for machines.

They're generally considered two separate pillars of observability. IMO, they should be handled separately (and maybe full telemetry push isn't enabled by default)

Quantumplation:
I think the default should be pretty printed logs, with an environment variable to switch them to structured; it is far easier to consistently set environment variables as part of a deployment pipeline / IAC / systemd setup, than it is for a user to remember to set the environment variable when they're just getting started running the node, or when a developer is debugging some issue.

crates/amaru/src/bin/amaru/main.rs
KtorZ: @wolf31o2 i fully agree (and I believe that we're saying the same thing? Stdout = useful stuff to humans, OTLP = useful stuff to monitoring systems). 

The reality being that: we don't have much logs. We have mainly traces,  spans and metrics (i.e. telemetry) which is why printing them on stdout (in either pretty text or json) isn't really helpful nor it is human readable. There are just too many traces to be any meaningful to humans, unless aggressively filtered. 

So I guess we could perhaps start by separating logs from traces a bit better in the codebase, and have only logs be emitted on stdout while routing traces through the open telemetry backend.

abailly: I obviously disagree, but I feel I am in the minority here, so I will just go with the flow. In my opinion, based on my experience:

* logs, traces and spans are ultimately one and a same thing, _events_ a system emits signalling something happened with enough metadata to reconstruct the causality graph (ingress request ids, timestamps, component ids...)
* those events _must_ be structured to be useful for humans and machines (until we have precise enough LLMs that can "understand" unstructured events)
* should a human wants to look at the log as they are output and display some nice text, it's obviously much easier to produce text from structured data than the converse. More importantly, this text is essentially contextual to the observer and therefore should not be something baked into the system
* it is not the business of the system to know how those events will be consumed, which means the system should do the simplest thing that can possibly work: output the stream of events to its `stdout` stream and let downstream consumers do what they want with it. This has the nice consequence it removes the need for complicated configuration for the system and puts the burden of selecting, filtering, transforming those events on the observer (the fact a typical [cardano-node config](https://github.com/IntersectMBO/cardano-node/blob/ca38b9d5acc08d625548470bc716deeae5e75dd6/configuration/cardano/mainnet-config.json#L19) has about 80% of its content dedicated to controlling logs is IMNSHO ridiculous and source of complexity and headaches)

Now, like I said, I will defer to the majority as long as I can output the stream of JSON-formatted logs to my `stdout` and not have to bother about errors about OTLP.

abailly: I would like to quote an industry-wide recognised expert on this:

> In 2018, Peter Bourgon wrote a [blog post](https://peter.bourgon.org/blog/2017/02/21/metrics-tracing-and-logging.html) proposing that “observability has three pillars: metrics, logs, and traces.” Ben Sigelman did a masterful job of unpacking why [metrics, logs, and traces are just telemetry](https://softwareengineeringdaily.com/2021/02/04/debunking-the-three-pillars-of-observability-myth/). However, lots of people latched on to the three pillars language: 
> Vendors, because they (coincidentally!) had metrics products, logging products, and tracing products to sell. 
Engineers, because it described their daily reality.
> Since then, the industry has been stuck in a weird space, where the language used to describe the problems and solutions has evolved, but the solutions themselves are largely the same ones as five or ten years ago. They’ve improved, of course—massively improved—but structurally, they’re variations on the same old pre-aggregated metrics.

https://www.honeycomb.io/blog/time-to-version-observability-signs-point-to-yes

abailly:
Superseded by #84

PR #75, opened 2025-01-24
Naive block selection

abailly:
This is unfortunately a big PR which will probably be hard to review, so here is a summary of what it does:

* It introduces a first (and naive) `chain_selection` logic and make it a core part of the `consensus` module
* It adds a `ChainStore` tied to the consensus module to store raw headers (and later blocks) indexed by their hash. The implementation is based on RocksDB
* It adds a `import-chain-db` command to be able to retrieve headers' data from a cardano-node. This is needed because the consensus starts from some `tip` which must be known
* It adds a generic `Header` interface with a concrete `ConwayHeader` implementation to bypass issues with using `MultiEraHeader` which introduces lifecycle annotations which make things painful

There are way too many things it _doesn't do_ to list them, but the most important ones I would like to tackle as immediate follow-up to this PR are:

* Enable multiple upstream peers, using a basic `topology.json` file while we wait for P2P
* Store block bodies, which is required to ...
* ... expose `ChainSync` and `BlockFetch` to downstream peers so we can connect Amaru as a server for other nodes

Another line of work we should be able to pursue with this PR is connecting with @stevana's work on simulation testing, either at the network interfaces layer or better embedded as a library.

**Disclaimer**: I am still learning Rust so feel free to suggest improvements related to the feature of the languages or ecosystem I should be using.

## Summary by CodeRabbit

- **New Features**
  - Introduced a new command that allows users to import the chain database from a remote peer.
  - Added an option for specifying a custom on-disk storage path for the blockchain data.

- **Documentation**
  - Updated the demo instructions to clarify prerequisites (such as access to a pre-prod node) and highlight potential delays.
  - Added comprehensive guidance about consensus workflows, including chain selection and testing approaches.

- **Improvements**
  - Streamlined peer session management and synchronisation processes.
  - Enhanced dependency management for better performance and stability.


jeluard:
Shall we wait for this PR to be based on `main` branch before reviewing?

abailly:
I will rebase today so I would say yes but I don’t expect much impacts. Plus this is a large PR and I guess the idea is more to agree on the overall direction than the details

crates/amaru/Cargo.toml
```diff
@@ -46,6 +46,7 @@ opentelemetry-otlp = { version = "0.27.0", features = [
     "http-proto",
     "reqwest-client",
 ] }
+rand = "0.8"
```
jeluard: Is `rand` only used for tests by chance?

README.md
```diff
@@ -31,13 +34,23 @@ cargo run --release -- import \
   --snapshot 70070379.d6fe6439aed8bddc10eec22c1575bf0648e4a76125387d9e985e9a3f8342870d.cbor
 ```
 
-2. Setup observability backends:
+2. Import chain data from remote peer
```
jeluard, line 37: I would assume this import is temporary-ish, as we probably would came up with a single snapshot format containing everything?
Would this data be available in the `cardano-node` data folder that is signed by `mithril`?

crates/amaru/src/consensus/chain_selection.rs
```diff
@@ -0,0 +1,421 @@
...
+
+impl<H> ChainSelector<H>
+where
+    H: Header + Clone + Debug + PartialEq,
+{
+    /// Roll forward the chain with a new header from given peer.
+    ///
+    /// The function returns the result of the chain selection process, which might lead
+    /// to a new tip, a switch to a fork, no change, or some change in status for the peer.
+    #[instrument(skip(self, header), fields(slot = header.slot(), hdr = header.hash().to_string()))]
```
jeluard: We might want to agree on this syntax for instrumentation.

crates/amaru/src/consensus/store.rs
```diff
@@ -0,0 +1,31 @@
+use super::header::Header;
+use miette::Diagnostic;
+use pallas_crypto::hash::Hash;
+use std::fmt::Display;
+use thiserror::Error;
+
+pub mod rocksdb;
+
+#[derive(Error, Diagnostic, Debug)]
```
jeluard: We probably want to have a common strategy for defining errors, e.g. use `thiserror`

crates/amaru/src/consensus/store/rocksdb.rs
```diff
@@ -0,0 +1,70 @@
+use super::{ChainStore, StoreError};
+use crate::consensus::header::Header;
+use pallas_crypto::hash::Hash;
+use rocksdb::{OptimisticTransactionDB, Options};
+use std::path::PathBuf;
+use tracing::instrument;
+
+pub struct RocksDBStore {
```
jeluard: `ledger` crate also has a `store` implemented via `rocksdb`. Would make sense to discuss and see if it's worth having some synergies.

crates/ouroboros/src/validator/mod.rs
```diff
@@ -1,6 +1,7 @@
+use serde::{Deserialize, Serialize};
 use thiserror::Error;
 
-#[derive(Error, Debug)]
+#[derive(Error, Debug, PartialEq, Serialize, Deserialize)]
```
jeluard, line 18: What requires those new derive attributes?

abailly:
I wish we had merged this PR _before_ restructuring crates 🙄

crates/amaru/src/consensus/store/rocksdb.rs
abailly: Yes, but not now.

crates/amaru/src/consensus/chain_selection.rs
abailly: Could you elaborate?

README.md
abailly: While this particular import is definitely temporary, I am not sure we don't want to have ways to separate importing the raw chain date from the ledger. Mithril bundles it all in the cardano-node db snapshot, but anyhow the ledger snapshot bundled there is not suitable for use by Amaru for now.

crates/amaru/Cargo.toml
abailly: Good question. I think it was in `dev-dependencies` and I moved it there because I moved some code back and forth between tests and prod.

crates/ouroboros/src/validator/mod.rs
abailly: the snapshots we record on disk, precisely this line: https://github.com/pragma-org/amaru/blob/ed049ed26036de350b20edc683ea0c9558a10e76/crates/ouroboros/src/consensus/test.rs#L117

abailly:
@jeluard I can move the consensus code to its own crate, like what you did for ledger, do we want to do it with this PR?

crates/amaru/src/consensus/chain_selection.rs
jeluard: I mean use `instrument` instead of manually creating spans as a standard practise. So maybe a candidate for an AR?

crates/ouroboros/src/validator/mod.rs
jeluard: Ok, looks a bit weird to have to change Error definitions for a specific test. Maybe we could avoid storing errors somehow?

jeluard:
> I wish we had merged this PR _before_ restructuring crates 🙄

Oups right! I guess that's the kind of changes you want sooner rather than later.. I had some of my PRs broken too :D

> @jeluard I can move the consensus code to its own crate, like what you did for ledger, do we want to do it with this PR?

I believe @KtorZ had some opinions regarding this? Now that's probably the right time to make that call.

crates/ouroboros/src/validator/mod.rs
abailly: I am not a great fan of snapshot tests so perhaps we could do that. In general however, I don't see much of a problem to have an error that's serialisable. Rather the opposite actually.

crates/amaru/src/consensus/chain_selection.rs
abailly: sure, we should strive to remove the o11y clutter from the code.

crates/amaru/src/consensus/store/rocksdb.rs
abailly: Seems like coderabbit agrees with you :)
Screenshot 2025-02-01 at 22 44 13

crates/amaru/src/bin/amaru/main.rs
```diff
@@ -34,6 +34,9 @@ enum Command {
 
     /// Import the ledger state from a CBOR export produced by the Haskell node.
     Import(cmd::import::Args),
```
KtorZ, line 36: TODO: rename for consistency (something like: ImportLedger)

crates/amaru/src/bin/amaru/cmd/import_chain_db.rs
```diff
@@ -0,0 +1,189 @@
...
+    match next {
+        NextResponse::RollForward(content, tip) => {
+            let header = ConwayHeader::from_cbor(&content.cbor).unwrap();
+            let hash = header.hash();
+
+            db.put(&hash, &header).map_err(|_| WorkerError::Panic)?;
+            *count += 1;
+            let slot = header.slot();
+            let tip_slot = tip.0.slot_or_default();
+            let progress = (slot - start) as f64 / (tip_slot - start) as f64;
```
KtorZ: TODO: use the same progress utils that we already use for the ledger import command

crates/amaru/src/bin/amaru/cmd/daemon.rs
```diff
@@ -43,6 +43,10 @@ pub struct Args {
     #[arg(long, default_value = super::DEFAULT_LEDGER_DB_DIR)]
     ledger_dir: PathBuf,
 
+    /// Path of the chain on-disk storage.
+    #[arg(long, default_value = super::DEFAULT_CHAIN_DATABASE_PATH)]
+    chain_database_path: PathBuf,
```
KtorZ, line 49: TODO: rename to align with import and ledger dir

crates/amaru/src/consensus/header.rs
```diff
@@ -0,0 +1,312 @@
...
+    fn block_height(&self) -> u64;
+
+    /// Slot number of the header
+    fn slot(&self) -> u64;
+
+    /// Point to this header
+    fn point(&self) -> Point;
+
+    /// Encode to CBOR
+    fn to_cbor(&self) -> Vec<u8>;
```
KtorZ: Feels like this should be a separate trait / dependent trait; like for decode.

crates/amaru/src/consensus/mod.rs
```diff
@@ -73,6 +88,117 @@
...
+        match result {
+            chain_selection::ChainSelection::NewTip(hdr) => {
+                self.forward_block(&hdr).await?;
+
+                self.block_count.inc(1);
+                self.track_validation_tip(point);
+            }
+            chain_selection::ChainSelection::RollbackTo(_) => {
+                panic!("RollbackTo should never happen on a RollForward")
+            }
```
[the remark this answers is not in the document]
abailly: There should be a better way, possibly through typing, to express that fact and avoid having to deal with an error that should never happen.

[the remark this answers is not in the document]
abailly: Exactly what I was thinking of, but we'll do it later :)

abailly:
@jeluard @KtorZ OK, I will move this to a new crate, merge with ouroboros crate which is only used there AFAIK, and merge the PR to have a first baseline for consensus.

crates/amaru/src/bin/amaru/cmd/import_chain_db.rs
abailly: I don't like those fancy terminal progress bars but I agree this make things consistent :-D

crates/amaru/src/consensus/header.rs
abailly: True. I suspect there's already something existing?

KtorZ: Correct. The `minicbor` crate we use has `Decode` and `Encode` traits. You can find some usage of those as trait constraints here for example: https://github.com/pragma-org/amaru/blob/8aa6446f47bbd8fae39e3a56d1a6797f90867aa6/crates/ledger/src/iter/borrow.rs#L61-L73

Not so trivial as we need to introduce higher-ranked lifetimes, but manageable.

abailly: I had a look there while writing this code, and at the end thought it would be simpler to have a `to_cbor/from_cbor` method in the `Header` trait 😅

PR #82, opened 2025-02-01
chore: make sure amaru crate name are uniques

jeluard:
All `amaru` crates now have unique names. This ensures they do not collide with existing crates and fixes some issues with tooling conflating both.


jeluard:
@abailly @KtorZ  Not clear to me why those tests are failing, maybe you have some ideas? This PR only renames some of `amaru` crates.

jeluard:
Tests are ok now.

PR #83, opened 2025-02-01
chore: introduce security CI

jeluard:
Added some workflow that audits every new dependencies for:

* security advisories
* license compatibility with amaru

To have the license check pass I had to whitelist the following licenses:

* MIT
* ISC
* Apache-2.0
* BSD-2-Clause
* BSD-3-Clause
* Unicode-3.0
* BlueOak-1.0.0

Those should be double-checked, and associated dependencies adapted to match removed licenses.

## Summary by CodeRabbit

- **Chores**  
  • Introduced an automated security audit workflow and updated dependency configurations to strengthen overall project integrity and security.  
  • Updated linting and advisory checks configuration for improved clarity and usability.

- **Refactor**  
  • Streamlined the system performance tracking process with enhanced CPU and memory monitoring for improved efficiency on non-Windows platforms.  
  • Removed unnecessary error handling in metric collection for a simpler control flow.

### DIFF
--- a/crates/amaru/src/bin/amaru/metrics.rs
+++ b/crates/amaru/src/bin/amaru/metrics.rs
@@ -33,7 +33,7 @@ pub fn track_system_metrics(metrics: SdkMeterProvider) -> JoinHandle<()> {
 
 mod internals {
     use opentelemetry::{
-        metrics::{Counter, Gauge, MeterProvider},
+        metrics::{Gauge, MeterProvider},
         KeyValue,
     };
     use opentelemetry_sdk::metrics::SdkMeterProvider;
@@ -42,7 +42,7 @@ mod internals {
     pub struct SystemCounters {
         total_memory: Gauge<u64>,
         free_memory: Gauge<u64>,
-        cpu_usage: Counter<f64>,
+        cpu_usage: Gauge<f64>,
     }
 
     pub fn make_system_counters(metrics: SdkMeterProvider) -> SystemCounters {
@@ -61,9 +61,9 @@ mod internals {
             .build();
 
         let cpu_usage = meter
-            .f64_counter("cpu.percent")
+            .f64_gauge("cpu.percent")
             .with_description("the cpu usage in percent, measured once per second")
-            .with_unit("ms")
+            .with_unit("%")
             .build();
 
         SystemCounters {
@@ -90,7 +90,7 @@ mod internals {
         for cpu in usages {
             counters
                 .cpu_usage
-                .add(cpu.1 as f64, &[KeyValue::new("cpu_name", cpu.0)]);
+                .record(cpu.1 as f64, &[KeyValue::new("cpu_name", cpu.0)]);
         }
     }
 }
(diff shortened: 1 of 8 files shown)


crates/ouroboros/Cargo.toml
```diff
@@ -1,7 +1,14 @@
 [package]
 name = "ouroboros"
-version = "0.1.0"
-edition = "2021"
+version.workspace = true
```
jeluard, line 3: Align with other crates

jeluard:
@Quantumplation You might have some ideas on which metrics library would be a good fit to replace `sys_metrics` ?

jeluard:
I understand that the `rustsec` failure is due to a lack of `GITHUB_TOKEN` access when the job is run in a PR.

I would suggest to address the `sys_metrics` issue in a followup PR and merged this one as is.

Quantumplation:
In that case we should probably just write our own low level system metrics library; everything else in the ecosystem I could find is either worse supported across platforms, or very unmaintained

KtorZ:
Seems like `sys_metrics` is mostly just a wrapper around `libc` (which explains why it doesn't work on Windows). So we might as well start from one of the other existing Rust lib, fork and maintain it?

jeluard:
@Quantumplation Did you consider https://github.com/GuillaumeGomez/sysinfo ? Looks like it could do the job?

jeluard:
I did a first migration to `sysinfo`. haven't checked with `grafana` though.

crates/amaru/src/bin/amaru/metrics.rs
```diff
@@ -92,55 +62,47 @@ mod internals {
...
             .with_description("the 1m average load, measured once per second")
             .build();
 
-        let user_time = meter
-            .u64_counter("cpu.time")
-            .with_description("the total cpu time spent in user processes")
+        let cpu_usage = meter
+            .f64_counter("cpu.percent")
+            .with_description("the cpu usage in percent, measured once per second")
             .with_unit("ms")
```
KtorZ:
```suggestion
            .with_unit("%")
```

Note also that is seems the number we're getting can greatly go above 100% though. The documentation says the following:

> Notice that it might be bigger than 100 if run on a multi-core machine. If you want a value between 0% and 100%, divide the returned value by the number of CPUs. 

Although, I only have 9 CPUs; and so even if considering CPU's threads, I don't see how numbers above 3000 can make sense 😅 (especially when Amaru is mostly idle):

image

crates/amaru/src/bin/amaru/metrics.rs
```diff
@@ -92,55 +62,47 @@ mod internals {
...
         let cpu_load = meter
             .f64_gauge("cpu.utilization")
             .with_description("the 1m average load, measured once per second")
             .build();
 
-        let user_time = meter
-            .u64_counter("cpu.time")
-            .with_description("the total cpu time spent in user processes")
+        let cpu_usage = meter
+            .f64_counter("cpu.percent")
```
KtorZ: Shouldn't those be gauges instead?

PR #85, opened 2025-02-02
feat: handle connection to multiple upstream peers

abailly:
This small PR adds the ability for an Amaru node to connect to multiple upstream peers. This is a temporary solution to test and demo the chain selection process, while we wait for @scarmuega's P2P handling component.

**Note**: This PR sits on top of #75


abailly:
@coderabbitai review

KtorZ:
@abailly I didn't look much further but I am now getting an error when using the command as I used to (single local `--peer-address`):

```
Error:   × error connecting bearer
  ╰─▶ Connection refused (os error 61)
```

I've checked against `main`, which still works fine.

abailly:
I find it surprising, I tried (after a rebase) to run:

```
% AMARU_DEV_LOG=info cargo run --release -- daemon \
  --peer-address=clermont:3001 --peer-address=preprod-node.world.dev.cardano.org:30000 \
  --network=preprod  | grep chain_selection
```

(without otlp stack) and it "works": It does not crash and shows progress and headers from both peers defined.

crates/consensus/src/consensus/mod.rs
```diff
@@ -86,11 +90,16 @@ impl Stage {
...
+    async fn forward_block(&mut self, peer: &Peer, header: &dyn Header) -> Result<(), WorkerError> {
         let point = header.point();
         let block = {
-            let mut peer_session = self.peer_session.lock().await;
-            let client = (*peer_session).blockfetch();
+            // FIXME: should not crash if the peer is not found
+            let peer_session = self
+                .peer_sessions
+                .get(peer)
+                .expect("Unknown peer, bailing out");
```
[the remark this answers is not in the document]
abailly: Thanks for the tip but this will be implemented later

abailly:
@KtorZ Were you able to go further?

KtorZ:
Not yet, but about to try!

crates/consensus/src/consensus/header_validation.rs
```diff
@@ -38,8 +38,13 @@ pub fn assert_header<'a>(
             FixedDecimal::from(5u64) / FixedDecimal::from(100u64);
         let c = (FixedDecimal::from(1u64) - active_slots_coeff).ln();
         let block_validator = BlockValidator::new(header, cbor, ledger, epoch_nonce, &c);
-        block_validator.validate().or_panic()?;
+        block_validator
+            .validate()
+            .map_err(|e| {
+                warn!("fail to validate header {}: {:?}", header.hash(), e);
```
KtorZ, line 44:
```suggestion
                warn!(header.hash = ?header.hash(), validation.error = ?e, "failed_to_validate");
```

A "best-practice" I've noted from using the `tracing` library (but which apply to telemetry in general) is to leverage _fields_ as much as possible, and avoid embedding fields/payloads in the log message. So said differently, embrace _structured logs/traces_ ! This is what allows us to have nice JSON logs out-of-the-box 😛 

Also by convention for now, I've always used snake_cased names for all traces & spans; really more out of paranoia than out of any legit reason.

abailly: Totally agreed, that's why I voluntarily wrote this temporary kludge as a textual log entry so that we spot it more easily and change it to something consistent with the rest. 

In general, I think we should only ever need `instrument!`, any other kind of logs is probably unneeded, or is begging for a function to instrument :)

KtorZ: > I think we should only ever need instrument!

That's a good rule of thumb regarding how we slice the code, indeed.

PR #88, opened 2025-02-03
chore: ledger refactoring in preparation for governance work

jeluard:
Refactoring of the current `state` handling to prepare for `governance` support inclusion.

The general direction of this refactoring is to:

* decouple `store` usage from reward calculation so that they become stateless
* introduce a higher-level data structure that smoothen integration between stable and volatile storage
* introduce a basic abstract `Error` struct; no more crate specific errors (e.g. `rocksdb`)

Some extra specifics:

* remove non-storage related functions from `Store` trait
* introduce `AnchoredVolatileState` to clearly separate anchored from non-anchored `VolatileState`

Replaces https://github.com/pragma-org/amaru/pull/76

## Summary by CodeRabbit

- **Refactor**
  - Internal improvements have streamlined state management and error handling, resulting in greater system robustness and clearer error messages.
  
- **Chores**
  - Dependency and build configuration updates have been implemented to enhance overall maintenance and ensure smoother operations.
  
- **Bug Fixes**
  - Enhanced error reporting now provides more detailed feedback during failure events, aiding in quicker diagnosis and resolution.

These updates contribute to a more stable and reliable experience for end users.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,6 +61,7 @@ proptest = "1.5.0"
 insta = "1.41.1"
 envpath = "0.0.1-beta.3"
 criterion = "0.5.1"
+test-case = "3.3.1"
 
 # build-dependencies
 built = "0.7.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,7 +39,7 @@ pallas-traverse = "0.32.0"
 rand = "0.8"
 rayon = "1.10"
 rocksdb = "0.22.0"
-serde = "1.0"
+serde = { version = "1.0", default-features = false }
 serde_json = { version = "1.0.128", default-features = false }
 tempfile = "3.15.0"
 thiserror = "2.0.3"
--- a/crates/ledger/src/lib.rs
+++ b/crates/ledger/src/lib.rs
@@ -12,12 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crate::{
-    kernel::{Hash, Hasher, MintedBlock, Point},
-    state::BackwardErr,
-};
+use crate::kernel::{Hash, Hasher, MintedBlock, Point};
 use gasket::framework::*;
 use pallas_codec::minicbor as cbor;
+use state::BackwardError;
 use std::sync::Arc;
 use store::Store;
 use tokio::sync::Mutex;
@@ -53,16 +51,16 @@ pub mod rewards;
 pub mod state;
 pub mod store;
 
-pub struct Stage<T>
+pub struct Stage<S>
 where
-    T: Store,
+    S: Store,
 {
     pub upstream: UpstreamPort,
     pub downstream: DownstreamPort,
-    pub state: Arc<Mutex<state::State<T, T::Error>>>,
+    pub state: Arc<Mutex<state::State<S>>>,
 }
 
-impl<T: Store> gasket::framework::Stage for Stage<T> {
+impl<S: Store> gasket::framework::Stage for Stage<S> {
     type Unit = ValidateBlockEvent;
     type Worker = Worker;
 
@@ -75,8 +73,8 @@ impl<T: Store> gasket::framework::Stage for Stage<T> {
     }
 }
 
-impl<T: Store> Stage<T> {
-    pub fn new(store: T) -> (Self, Point) {
+impl<S: Store> Stage<S> {
+    pub fn new(store: S) -> (Self, Point) {
         let state = state::State::new(Arc::new(std::sync::Mutex::new(store)));
 
         let tip = state.tip().into_owned();
@@ -143,7 +141,7 @@ impl<T: Store> Stage<T> {
 
         let result = match state.backward(&point) {
             Ok(_) => BlockValidationResult::RolledBackTo(point),
-            Err(BackwardErr::UnknownRollbackPoint(_)) => {
+            Err(BackwardError::UnknownRollbackPoint(_)) => {
                 BlockValidationResult::InvalidRollbackPoint(point)
             }
         };
@@ -157,14 +155,14 @@ impl<T: Store> Stage<T> {
 pub struct Worker {}
 
 #[async_trait::async_trait(?Send)]
-impl<T: Store> gasket::framework::Worker<Stage<T>> for Worker {
-    async fn bootstrap(_stage: &Stage<T>) -> Result<Self, WorkerError> {
+impl<S: Store> gasket::framework::Worker<Stage<S>> for Worker {
+    async fn bootstrap(_stage: &Stage<S>) -> Result<Self, WorkerError> {
         Ok(Self {})
     }
 
     async fn schedule(
         &mut self,
-        stage: &mut Stage<T>,
+        stage: &mut Stage<S>,
     ) -> Result<WorkSchedule<ValidateBlockEvent>, WorkerError> {
         let unit = stage.upstream.recv().await.or_panic()?;
         Ok(WorkSchedule::Unit(unit.payload))
@@ -173,7 +171,7 @@ impl<T: Store> gasket::framework::Worker<Stage<T>> for Worker {
     async fn execute(
         &mut self,
         unit: &ValidateBlockEvent,
-        stage: &mut Stage<T>,
+        stage: &mut Stage<S>,
     ) -> Result<(), WorkerError> {
         let result = match unit {
             ValidateBlockEvent::Validated(point, raw_block) => {
--- a/crates/ledger/src/rewards.rs
+++ b/crates/ledger/src/rewards.rs
@@ -155,10 +155,15 @@ impl StakeDistribution {
         let mut accounts = db
             .iter_accounts()?
             .filter_map(|(credential, account)| {
-                account.delegatee.map(|pool| (credential, AccountState {
-                    pool,
-                    lovelace: account.rewards,
-                }))
+                account.delegatee.map(|pool| {
+                    (
+                        credential,
+                        AccountState {
+                            pool,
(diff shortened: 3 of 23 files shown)


crates/ledger/src/lib.rs
```diff
@@ -45,16 +43,17 @@ pub mod rewards;
 pub mod state;
 pub mod store;
 
-pub struct Stage<T>
+pub struct Stage<HS, S>
 where
-    T: Store,
+    HS: HistoricalStores,
```
KtorZ: Given the prevalence of the word "snapshot" in the codebase and specs for referring to these, should we consider calling it simply `Snapshots`?

crates/ledger/src/store.rs
```diff
@@ -17,23 +17,76 @@ pub mod columns;
...
+    /// Provide an access to iterate over accounts, similar to 'with_pools'.
+    fn with_accounts(&self, with: impl FnMut(accounts::Iter<'_, '_>)) -> Result<(), StoreError>;
 
+    /// Provide an iterator over slot leaders, similar to 'with_pools'. Note that slot leaders are
+    /// stored as a bounded FIFO, so it only make sense to use this function at the end of an epoch
+    /// (or at the beginning, before any block is applied, depending on your perspective).
+    fn with_block_issuers(&self, with: impl FnMut(slots::Iter<'_, '_>)) -> Result<(), StoreError>;
+
+    /// Provide an access to iterate over utxo, similar to 'with_pools'.
+    fn with_utxo(&self, with: impl FnMut(utxo::Iter<'_, '_>)) -> Result<(), StoreError>;
```
KtorZ: We could potentially be (i.e. in a subsequent iteration) be smarter here since we won't need to ever _modify_ items in a `ViewStore`. So the with-pattern here with the `BorrowMut` proxies is a bit overkill. Nor do we necessarily needs exactly the same information about those db items (e.g. we may be fine with just the Ada value of UTxO, although now I am perhaps over-fitting to the rewards calculation...).

KtorZ: (reviewed further down the PR): I see; I thought `ViewStore` represented the immutable snapshots, but no, it's capturing some "query layer" on top of the store; 

Then, I am not sure to fully understand the boundary between `Store` and `ViewStore` given that there are still accessor methods in `Store` (e.g. `pool`, `resolve_inputs` ...) ?

crates/ledger/src/state/volatile_db.rs
```diff
@@ -126,37 +133,25 @@ impl VolatileCache {
...
-            pools: Default::default(),
-            accounts: Default::default(),
-            withdrawals: Default::default(),
-            fees: 0,
-        }
-    }
+pub struct AnchoredVolatileState {
+    pub anchor: (Point, PoolId),
+    pub state: VolatileState,
 }
```
KtorZ, line 148: (genuine question) Any particular reason to prefer this over a generic (other than aesthetics)?

crates/ledger/src/store.rs
jeluard: Right, currently I just migrated what was strictly needed for the rewards calculation. How about we add the other read-only fns to the trait?

crates/ledger/src/state/volatile_db.rs
jeluard: My understanding is that `anchor` will never be something else? Removing the generics clarify that, and simplify a bit the syntax.

crates/ledger/src/lib.rs
jeluard: Sounds good! Do we want to change `ViewStore` to `Snapshot` then?

crates/ledger/src/store.rs
KtorZ: That would make the frontier clearer to me 😅 ? At first, I thought the split was Read-only / Read-write, but no; then I thought it's database-creation vs accesses, but no. 

So perhaps it's about having a `ReadWriteStore` and `ReadOnlyStore`?

Cargo.toml
```diff
@@ -39,7 +39,7 @@ pallas-traverse = "0.32.0"
 rand = "0.8"
 rayon = "1.10"
 rocksdb = "0.22.0"
-serde = "1.0"
+serde = { version = "1.0", default-features = false }
 serde_json = { version = "1.0.128", default-features = false }
```
KtorZ, line 43: Nitpick: shall we move _features_ down to packages and keep only versions in the top-level Cargo.toml? (if we truly agree with this strategy, I'll amend the CONTRIBUTING.md to capture it).

crates/ledger/src/lib.rs
KtorZ: _possibly_; the other option is "ViewStore" / "ReadOnlyStore"; so in one case we name based on _capabilities_ and in the other based on the _responsibilities_. Can't decide which one is better, so whatever you prefer 😅 ...

Cargo.toml
jeluard: Looks useful to me to have reasonable features defined as default at the top level. It's not always straightforward to have them right, as some dependencies ship everything and the kitchen sink as default. This helps rely on dependencies without having to go through this process each time. `serde` or `rocksdb` are good example of this (`rocksdb` enables by default multiple compression algorithms while we only ever want to use one probably).

crates/ledger/src/rewards.rs
```diff
@@ -151,58 +151,55 @@ impl StakeDistribution {
...
-
-        let mut pools: BTreeMap<PoolId, PoolState> = BTreeMap::new();
-        db.with_pools(|rows| {
-            for (pool, row) in rows {
-                if let Some(row) = row.borrow() {
-                    pools.insert(
+    pub fn new(db: &impl Snapshot) -> Result<Self, StoreError> {
+        // TODO: Avoid creating this intermediate map, and directly create the keys/scripts ones.
+        // Then, when looking for _active accounts_ delegated to pools, we can prune those not
+        // referenced anywhere.
```
KtorZ:
```suggestion
```

crates/ledger/src/rewards.rs
```diff
@@ -151,58 +151,55 @@ impl StakeDistribution {
...
+        let mut accounts = db
+            .iter_accounts()?
+            .filter_map(|(credential, account)| {
+                account.delegatee.map(|pool| (pool, credential, account))
```
KtorZ:
```suggestion
                account.delegatee.map(|pool| (credential, AccountState {
                      pool,
                      lovelace: account.rewards,
                }))
```

crates/ledger/src/rewards.rs
```diff
@@ -151,58 +151,55 @@ impl StakeDistribution {
...
+            })
+            .map(|(pool, credential, account)| {
+                (
+                    credential,
+                    AccountState {
```
KtorZ:
```suggestion
```

crates/ledger/src/rewards.rs
```diff
@@ -151,58 +151,55 @@ impl StakeDistribution {
...
-                            parameters: row.current_params.clone(),
-                        },
-                    );
-                }
-            }
-        })?;
+                        lovelace: account.rewards,
+                    },
+                )
+            })
```
KtorZ:
```suggestion
```

crates/ledger/src/rewards.rs
```diff
@@ -151,58 +151,55 @@ impl StakeDistribution {
...
+
+        let credentials = db.iter_utxos()?.filter_map(|(_, out)| {
+            output_stake_credential(&out).map(|credential| (out, credential))
+        });
+        credentials.for_each(|(output, credential)| {
+            let value = output_lovelace(&output);
+            accounts
+                .entry(credential)
+                .and_modify(|account| account.lovelace += value);
+        });
```
KtorZ:
```suggestion
        db.iter_utxos()?.for_each(|(_, output)| {
            if let Some(credential) = output_stake_credential(&output) {
                let value = output_lovelace(&output);
                accounts
                    .entry(credential)
                    .and_modify(|account| account.lovelace += value);
            }
        });
```

So that we only traverse the structure once; there are potentially millions of credentials so it matters quite a lot.

crates/ledger/src/store.rs
```diff
@@ -64,50 +108,41 @@ pub trait Store: Send + Sync {
...
 
     /// Provide an iterator over slot leaders, similar to 'with_pools'. Note that slot leaders are
     /// stored as a bounded FIFO, so it only make sense to use this function at the end of an epoch
     /// (or at the beginning, before any block is applied, depending on your perspective).
-    fn with_block_issuers(&self, with: impl FnMut(slots::Iter<'_, '_>)) -> Result<(), Self::Error>;
+    fn with_block_issuers(&self, with: impl FnMut(slots::Iter<'_, '_>)) -> Result<(), StoreError>;
 
     /// Provide an access to iterate over utxo, similar to 'with_pools'.
-    fn with_utxo(&self, with: impl FnMut(utxo::Iter<'_, '_>)) -> Result<(), Self::Error>;
+    fn with_utxo(&self, with: impl FnMut(utxo::Iter<'_, '_>)) -> Result<(), StoreError>;
```
KtorZ, line 145: I believe we can likely remove that method entirely; I don't think we ever need to iter over UTxOs other than for rewards calculation 🤔.

crates/ledger/src/rewards.rs
jeluard: ?

crates/ledger/src/rewards.rs
jeluard: The first `map` is nested inside the `filter_map`, those are not 2 maps in a row.

KtorZ:
(merged manually)

PR #89, opened 2025-02-04
chore: CI speedups

jeluard:
Some CI improvements:

* only run `target` specifics per target (i.e. run `fmt` only once)
* lib only target (e.g. `wasm`) do not run on final `amaru` CLI

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -17,8 +17,21 @@ env:
   CARGO_PROFILE_RELEASE_CODEGEN_UNITS: 1
 
 jobs:
+  sanity:
+    timeout-minutes: 30
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: Swatinem/rust-cache@v2
+        with:
+          prefix-key: ${{ runner.os }}
+      - name: Check format
+        run: cargo fmt-amaru
+      - name: Run clippy
+        run: cargo clippy-amaru
   build:
     name: Build on ${{ matrix.environments.runner }} with target ${{ matrix.environments.target }}
+    needs: sanity
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


jeluard:
Compilation failure of the `musl` target is addressed in #90

jeluard:
`musl` target has been disabled for now has it's been broken for some days (unrelated to this PR). It was silently broken because it was incorrectly set `optional`. I couldn't find the reason of this regression yet.

.github/workflows/CI.yml
```diff
@@ -17,36 +17,42 @@ env:
...
+      - uses: Swatinem/rust-cache@v2
+        with:
+          prefix-key: ${{ runner.os }}
+      - name: Check format
+        run: cargo fmt-amaru
+      - name: Run clippy
+        run: cargo clippy-amaru
   build:
     name: Build on ${{ matrix.environments.runner }} with target ${{ matrix.environments.target }}
+    needs: sanity
```
KtorZ: Or not? If we're really after a speed up, we could have both happen in parallel; if the `sanity` fails, it'll cancel the build anyway and there's in fact no artifact produced by the `sanity` that is needed for the build & test to start.

.github/workflows/CI.yml
```diff
@@ -58,6 +64,30 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v4
+      - uses: Swatinem/rust-cache@v2
+        with:
+          prefix-key: ${{ runner.os }}-${{ matrix.environments.target }}
+      - run: rustup target add ${{ matrix.environments.target }}
+      - name: Run tests
+        run: |
+          set +e
+          cargo test --locked --all-features -p ${{ matrix.environments.package }} --target ${{ matrix.environments.target }}
```
KtorZ: I am not sure I follow here? Why build the ledger separately from the rest (and has a separate job altogether)? Couldn't the same be achieved by adding packages to the matrix in the previous job? And more importantly, would the previous job _already build and test_ the ledger crate?

.github/workflows/CI.yml
jeluard: Makes sense!

.github/workflows/CI.yml
jeluard: `ledger` would indeed be tested by the previous job but some target (e.g. `wasm`) won't realistically compile the full `amaru`, failing the whole job. So the idea is to restrict to some identified packages those specific targets.

I can see if there is some way to combine both.

PR #99, opened 2025-02-06
feat: introduce new stage for block validation

abailly:
This commit breaks down the consensus monolithic pipeline in 2 stages, with ledger block validation in between. While the last stage currently does nothing beyond logging the result of the ledger validation, it is intended it will be responsible for:

1. storing the blocks on disk
2. updating the node's chain candidate depending on ledger validation result
3. forward new selected chain to downstream peers

## Summary by CodeRabbit

- **New Features**
  - Introduced a dedicated stage to streamline block processing, improving data flow and validation.
  - Added new event types and validation results to enhance block validation processes.

- **Refactor**
  - Updated the consensus process to focus on complete block validation rather than header-only checks.
  - Enhanced the ledger’s validation and rollback handling for clearer error reporting and improved reliability.

### DIFF
--- a/crates/consensus/src/chain_forward.rs
+++ b/crates/consensus/src/chain_forward.rs
@@ -22,6 +22,8 @@ use tracing::{error, info, warn};
 
 pub type UpstreamPort = gasket::messaging::InputPort<BlockValidationResult>;
 
+pub const EVENT_TARGET: &str = "amaru::consensus::chain_forward";
+
 /// Forwarding stage of the consensus where blocks are stored and made
 /// available to downstream peers.
 ///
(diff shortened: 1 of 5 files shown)


crates/consensus/src/chain_forward.rs
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    async fn execute(
+        &mut self,
+        unit: &BlockValidationResult,
+        _stage: &mut ForwardStage,
+    ) -> Result<(), WorkerError> {
+        match unit {
+            BlockValidationResult::BlockValidated(point) => {
+                info!(name: "block_validated", target: "amaru::consensus::chain_forward", slot = ?point.slot_or_default(), hash = ?point_hash(point).to_string());
```
KtorZ:
```suggestion
  pub const EVENT_TARGET: &str = "amaru::consensus::chain_forward";

                info!(name: "block_validated", target: EVENT_TARGET, slot = ?point.slot_or_default(), hash = ?point_hash(point).to_string());
```

crates/ledger/src/lib.rs
```diff
@@ -92,88 +165,31 @@ impl<T: Store> gasket::framework::Worker<Stage<T>> for Worker {
...
-        }
+            ValidateBlockEvent::Rollback(point) => stage.rollback_to(point.clone()).await,
+        };
 
-        Ok(())
+        Ok(stage.downstream.send(result.into()).await.or_panic()?)
     }
 }
 
+#[instrument(skip(bytes), fields(block.size = bytes.len()))]
```
KtorZ:
```suggestion
#[instrument(level = Level::DEBUG, skip(bytes), fields(block.size = bytes.len()))]
```

abailly: Note I was about to do the same for `validate_block` but we pass the `span` to called functions so it's more involved, I will do that in another PR perhaps?

crates/consensus/src/chain_forward.rs
```diff
@@ -0,0 +1,94 @@
...
+
+pub const EVENT_TARGET: &str = "amaru::consensus::chain_forward";
+
+/// Forwarding stage of the consensus where blocks are stored and made
+/// available to downstream peers.
+///
+/// TODO: currently does nothing, should store block, update chain state, and
+/// forward new chain downstream
+
+#[derive(Stage)]
```
jeluard, line 33: At some point it might make sense to have all gasket plumbing extracted out of `ledger` and `consensus` so that they remain lean and workflow agnostic.

crates/consensus/src/consensus/chain_selection.rs
```diff
@@ -96,7 +96,7 @@ pub enum ChainSelection<H: Header> {
     /// The current best chain has switched to given fork starting at
     /// given hash, from given Peer.
     ///
-    /// FIXME: The peer should not be needed here, as the fork should be
+    /// FIXME: The peer should not be neederd here, as the fork should be
```
jeluard:
```suggestion
    /// FIXME: The peer should not be needed here, as the fork should be
```

jeluard (approved):

crates/consensus/src/chain_forward.rs
abailly: Definitely. I think @stevana's work will be a sting in our back to force us to do that :)

crates/consensus/src/consensus/chain_selection.rs
abailly: You don't like my portemanteau between `need` and `nerd` :-D ?

PR #100, opened 2025-02-06
Rolling stake distributions

KtorZ:
This PR ensures that we now keep the last 2 stake distributions readily available in plain memory. The most recent one is meant to be used by the consensus to validate the leader schedule, whereas the oldest one intervenes in the rewards calculation mid-epoch. 

The rewards calculation consumes the stake distribution, and pushes a new one from the previous epoch. 

It also fixes a couple of small things I noticed along the way regarding withdrawals, traces or diagrams 😬 ...

## Summary by CodeRabbit

- **New Features**
	- Upgraded rewards processing now utilises updated distribution data for more accurate and consistent summaries.
  
- **Refactor**
	- Reorganised rewards and state management across core systems for improved clarity and reliability.
	- These updates streamline backend operations, ensuring a smoother and more dependable user experience.

### DIFF
--- a/crates/consensus/src/consensus/mod.rs
+++ b/crates/consensus/src/consensus/mod.rs
@@ -111,6 +111,7 @@ impl Stage {
             .or_panic()
     }
 
+    #[instrument(level = Level::INFO, skip(self))]
     async fn switch_to_fork(
         &mut self,
         peer: &Peer,
--- a/crates/ledger/src/state.rs
+++ b/crates/ledger/src/state.rs
@@ -29,7 +29,7 @@ use crate::{
 };
 use std::{
     borrow::Cow,
-    collections::BTreeSet,
+    collections::{BTreeSet, VecDeque},
     sync::{Arc, Mutex},
 };
 use tracing::{debug, debug_span, info_span, Span};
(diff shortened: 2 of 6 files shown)


crates/consensus/src/consensus/mod.rs
```diff
@@ -111,6 +111,7 @@ impl Stage {
             .or_panic()
     }
 
+    #[instrument(level = Level::INFO, skip(self))]
```
abailly, line 114: This probably has nothing to do here?

abailly (approved):

crates/consensus/src/consensus/mod.rs
KtorZ: No no, it does. I wanted to see this one as log while working on this 😅 ... as surprising as it may seem. 

I separated the change in a single commit: 3df064bf70e635a08b57a284d37fc85842bc92bf

crates/stores/src/rocksdb/mod.rs
```diff
@@ -209,11 +210,11 @@ impl Store for RocksDB {
                 pools::add(&batch, add.pools)?;
                 accounts::add(&batch, add.accounts)?;
 
+                accounts::reset(&batch, withdrawals)?;
```
jeluard, line 213: Any reason for this change?

crates/ledger/src/state.rs
```diff
@@ -61,10 +61,38 @@ where
...
+        let latest_epoch = db.most_recent_snapshot();
+        let mut stake_distributions = VecDeque::new();
+        for epoch in latest_epoch - 2..=latest_epoch - 1 {
+            stake_distributions.push_front(
+                db.stake_distribution(epoch)
+                    .unwrap_or_else(|e| panic!("unable to get current stake distribution: {e:?}")),
+            );
+        }
+
+        drop(db);
```
jeluard, line 94: `db` should be dropped right at the end of the fn, so looks unnecessary?

crates/stores/src/rocksdb/mod.rs
KtorZ: Yes! See: f4f4a78eea11dfcb6a05e4ed2086efc7a7b77cfa

crates/ledger/src/state.rs
KtorZ: Somehow it was not, and the borrow checker was complaining when trying to return the `stake_distributions` as part of the struct below.

PR #101, opened 2025-02-06
chore: cleanups to improve feedback

abailly:
## Summary by CodeRabbit

- **Refactor**  
  - Improved internal data handling for header management and chain selection to enhance overall clarity and consistency.

- **Chore**  
  - Enhanced logging during consensus events to provide more detailed contextual information for improved monitoring and debugging.  
  - Updated method names for clarity in header storage and retrieval processes.

### DIFF
--- a/crates/consensus/src/consensus/chain_selection.rs
+++ b/crates/consensus/src/consensus/chain_selection.rs
@@ -15,9 +15,7 @@
 use super::{header::Header, peer::Peer, Point};
 use pallas_crypto::hash::Hash;
 use std::{collections::HashMap, fmt::Debug};
-use tracing::{debug, instrument, Level};
-
-const EVENT_TARGET: &str = "amaru::consensus::chain_selection";
+use tracing::{instrument, Level};
 
 /// A fragment of the chain, represented by a list of headers
 /// and an anchor.
@@ -99,7 +97,12 @@ pub enum ChainSelection<H: Header> {
     /// FIXME: The peer should not be needed here, as the fork should be
     /// comprised of known blocks. It is only needed to download the blocks
     /// we don't currently store.
-    SwitchToFork(Peer, Point, Vec<H>),
+    SwitchToFork {
+        peer: Peer,
+        rollback_point: Point,
+        tip: H,
+        fork: Vec<H>,
+    },
 }
 
 /// Builder pattern for `ChainSelector`.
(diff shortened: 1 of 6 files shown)


crates/consensus/src/consensus/chain_selection.rs
```diff
@@ -207,7 +209,8 @@ where
     /// If the chain of the peer is still the longest, the function will return a
     /// `RollbackTo` result, otherwise it will return a `NewTip` result with the new
     /// tip of the chain.
-    #[instrument(level = Level::DEBUG, skip(self))]
+    #[instrument(level = Level::DEBUG, skip(self),
+                 fields(peer = peer.name, point = point.to_string()))]
```
KtorZ:
```suggestion
    #[instrument(level = Level::DEBUG, skip(self), fields(peer = peer.name, %point))]
```

crates/consensus/src/consensus/chain_selection.rs
```diff
@@ -243,7 +247,7 @@ where
         best
     }
 
-    #[instrument(level = Level::DEBUG, skip(self))]
+    #[instrument(level = Level::DEBUG, skip(self), fields(peer = peer.name, point = point.to_string()))]
```
KtorZ:
```suggestion
    #[instrument(level = Level::DEBUG, skip(self), fields(peer = peer.name, %point))]
```

crates/consensus/src/consensus/mod.rs
```diff
@@ -111,7 +113,10 @@ impl HeaderStage {
             .or_panic()
     }
 
-    #[instrument(level = Level::INFO, skip(self))]
+    #[instrument(level = Level::DEBUG, skip_all,
```
KtorZ: Why the switch back to DEBUG here? I thought that it's a significant enough (and occasional) event that it deserves being treated as log, no ?

crates/consensus/src/consensus/chain_selection.rs
abailly: Seems like code formatter does not format inside macros

crates/consensus/src/consensus/mod.rs
abailly: I thought this was a mistake as we/you changed all `instrument` to `DEBUG`.

abailly: I really think we should not care about those levels in the code, but I will revert back to `INFO`

KtorZ: Funny enough, I went for `DEBUG` in ADR-007 because i didn't see that `tracing` had a `TRACE` level. So I will be switching those `DEBUG` to `TRACE` in an upcoming commit. 

So that anything DEBUG/INFO will be treated as log. It'll be less confusing. 

Still for this one, I really think it shouldn't be a trace, but a log. A debug one perhaps but as you said, I don't think we should care about those levels for logging. Every log as INFO is good IMO.

abailly: I am having a hard time making a difference between logs and traces, they are one and the same thing. the difference happens downstream with the use you make of that information: you can ignore timing and treat the entry or exit of a function as a log, being interested in the discrete event this represents, or you can care about the duration, or perhaps the fact it raised and error, and/or the call tree it produced, with or without timings. 

IMNSHO, those differences we make up-front are probably short-sighted and only breeds confusion and complexity in the code base. But I value more alignment within the team than opinion so I will make sure to read and follow the policy we agreed on in the ADR 👍

KtorZ: We can summarize the difference simply:

- logs = for humans
- traces = for programs

As a human, I cannot process outputs on stdout when my terminal is flooded with thousands of lines per second. So of course, I can redirect the output, or export it to a sub-system for analysis/observability (which I believe is what you are advocating for).   

But, it's also nice to have some readily available output meant for immediate human-consumption. Something which I need not to spin up an telemetry backend with a collector and a trace visualizer to make sense of (which I believe, you have also been advocating for 😅 ...).

PR #102, opened 2025-02-06
Re-introduce header validation using own ledger state

KtorZ:
- :round_pushpin: **feat: rework the consensus/ledger interface around header validation**
    The interface has been renamed and slightly reworked to better reflect
  the reality.

  The major change it brings in particular is the moving of the Mutex
  from the whole ledger down to simply the vector of stake distribution.
  This reduces drastically the contention between the two components,
  since the ledger only access or modifies the stake distribution once
  per epoch; whereas the consensus needs it for every blocks.

- :round_pushpin: **fix: pass active_slot_coefficient down to the header validation**
    Somehow, we were pre-computing 'c', which is now computed as part of the validate method itself.

- :round_pushpin: **fix: temporarily disable validate_kes_signature verification**
    None of the signature verification are valid... somehow. Either it is because the verification has changed in Conway, or it's because we aren't passing the raw data in the right format. I suspect a missing CBOR tag or something like that...

## Summary by CodeRabbit

- **New Features**
  - Enhanced exposure of stake distribution data for improved visibility.
  - Introduced new constants for KES period management.
  - Added a new struct for pool summaries and a trait for interacting with stake distribution.

- **Refactor**
  - Optimised ledger state management and consensus processing to reduce complexity and boost performance.
  - Simplified error handling for pool-related validations.
  - Streamlined the handling of ledger state within the BlockValidator.
  - Updated method signatures and parameter types to improve type safety and clarity.

- **Bug Fixes**
  - Adjusted span event logging to provide more granular logging output during execution.

- **Tests**
  - Updated and streamlined testing methodologies using refined mock implementations for greater reliability.


crates/consensus/src/consensus/mod.rs
```diff
@@ -143,12 +143,16 @@ impl HeaderStage {
         &mut self,
         peer: &Peer,
         point: &Point,
-        raw_header: &RawBlock,
+        raw_header: &[u8],
     ) -> Result<(), WorkerError> {
-        let header: ConwayHeader = minicbor::decode(raw_header)
+        let header: babbage::MintedHeader<'_> = minicbor::decode(raw_header)
```
jeluard, line 158: Works as conway header share the same structure with babbage header, bu looks a bit confusing. Maybe `MintedHeader` should be made available in `conway` too?

Similarly, we probably need some abstraction here to deal with eventual future header changes introduced by future eras.

KtorZ: That was the intention, but unfortunately, for some reasons, MintedHeader isn't publicly exposed in Pallas from the `conway` module :/ ... I think it's an oversight, but for now, I went with this.

PR #110, opened 2025-02-12
Initial Scaffolding for Block Validation

yHSJ:
This PR introduces `rules` to the Cardano ledger used to perform block and transaction validations when updating ledger state. The block validation is in an MVP state, only containing a few validations on the block as a proof of concept. It does include some very rudimentary tests that rely on a specific block to prove functionality instead of rigorously testing the logic. 

## Summary by CodeRabbit

- **New Features**
  - Enhanced block validation with comprehensive checks for block sizes and resource usage to improve network stability.
  - Introduced configurable protocol parameters that streamline performance and resource management.
  - Improved block processing with more robust error handling for a smoother transaction experience.
  - Added new modules for validating block rules, including body size, execution units, and header size.

- **Chores**
  - Performed housekeeping updates to tidy project configuration.
  - Updated `.gitignore` to include JetBrains IDE files and backup files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,8 +25,6 @@ chain.db
 # A directory to store snapshots without committing them
 snapshots/
 
-# Default databse location for the chain store
-chain.db
 
 # RustRover
 #  JetBrains specific template is maintained in a separate JetBrains.gitignore that can
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,9 @@ chain.db
 # A directory to store snapshots without committing them
 snapshots/
 
+# Default databse location for the chain store
+chain.db
+
 # RustRover
 #  JetBrains specific template is maintained in a separate JetBrains.gitignore that can
 #  be found at https://github.com/github/gitignore/blob/main/Global/JetBrains.gitignore
--- a/crates/amaru-ledger/src/protocol_parameters.rs
+++ b/crates/amaru-ledger/src/protocol_parameters.rs
@@ -1,0 +1,262 @@
+use amaru_kernel::{Coin, Epoch, ExUnits, RationalNumber};
+
+#[derive(Clone)]
+pub struct ProtocolParameters {
+    pub max_block_body_size: u32,
+    pub max_tx_size: u32,
+    pub max_header_size: u32,
+    pub max_tx_ex_units: ExUnits,
+    pub max_block_ex_units: ExUnits,
+    pub max_val_size: u32,
+    pub max_collateral_inputs: u32,
+
+    pub min_fee_a: u32,
+    pub min_fee_b: u32,
+    pub key_deposit: Coin,
+    pub pool_deposit: Coin,
+    pub coins_per_utxo_byte: Coin,
+    pub prices: Prices,
+    pub min_fee_ref_script_coins_per_byte: RationalNumber,
+    pub max_ref_script_size_per_tx: u32,
+    pub max_ref_script_size_per_block: u32,
+    pub ref_script_cost_stride: u32,
+    pub ref_script_cost_multiplier: RationalNumber,
+
+    pub max_epoch: Epoch,
+    pub a_0: RationalNumber,
+    pub n_opt: u32,
+    pub tau: RationalNumber,
+    pub rho: RationalNumber,
+    pub collateral_percentage: u32,
+    pub cost_models: CostModels,
+
+    pub pool_thresholds: PoolThresholds,
+    pub drep_thresholds: DrepThresholds,
+    pub cc_min_size: u32,
+    pub cc_max_term_length: u32,
+    pub gov_action_lifetime: u32,
+    pub gov_action_deposit: Coin,
+    pub drep_deposit: Coin,
+    pub drep_activity: Epoch,
+}
+
+#[derive(Debug, Clone)]
+pub struct Prices {
+    pub mem: RationalNumber,
+    pub step: RationalNumber,
+}
+
+#[derive(Debug, Clone)]
+pub struct CostModels {
+    pub plutus_v1: Vec<i64>,
+    pub plutus_v2: Vec<i64>,
+    pub plutus_v3: Vec<i64>,
+}
+
+#[derive(Debug, Clone)]
+pub struct PoolThresholds {
+    pub q1: RationalNumber,
+    pub q2a: RationalNumber,
+    pub q2b: RationalNumber,
+    pub q4: RationalNumber,
+    pub q5e: RationalNumber,
+}
+
+#[derive(Debug, Clone)]
+pub struct DrepThresholds {
+    pub p1: RationalNumber,
+    pub p2a: RationalNumber,
+    pub p2b: RationalNumber,
+    pub p3: RationalNumber,
+    pub p4: RationalNumber,
+    pub p5a: RationalNumber,
+    pub p5b: RationalNumber,
+    pub p5c: RationalNumber,
+    pub p5d: RationalNumber,
+    pub p6: RationalNumber,
+}
+
+impl Default for ProtocolParameters {
+    // This default is the protocol parameters on Preprod as of epoch 197
+    fn default() -> Self {
+        Self {
+            min_fee_a: 44,
+            min_fee_b: 155381,
+            max_block_body_size: 90112,
+            max_tx_size: 16384,
+            max_header_size: 1100,
+            max_tx_ex_units: ExUnits {
+                mem: 14000000,
+                steps: 10000000000,
+            },
+            max_block_ex_units: ExUnits {
+                mem: 62000000,
+                steps: 20000000000,
+            },
+            max_val_size: 5000,
+            max_collateral_inputs: 3,
+            key_deposit: 2000000,
+            pool_deposit: 500000000,
+            coins_per_utxo_byte: 4310,
+            prices: Prices {
+                mem: RationalNumber {
+                    numerator: 577,
+                    denominator: 10000,
+                },
+                step: RationalNumber {
+                    numerator: 721,
+                    denominator: 10000000,
+                },
+            },
+            min_fee_ref_script_coins_per_byte: RationalNumber {
+                numerator: 15,
+                denominator: 1,
+            },
+            max_ref_script_size_per_tx: 200 * 1024, //Hardcoded in the haskell ledger (https://github.com/IntersectMBO/cardano-ledger/blob/3fe73a26588876bbf033bf4c4d25c97c2d8564dd/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ledger.hs#L154)
+            max_ref_script_size_per_block: 1024 * 1024, // Hardcoded in the haskell ledger (https://github.com/IntersectMBO/cardano-ledger/blob/3fe73a26588876bbf033bf4c4d25c97c2d8564dd/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Bbody.hs#L91)
+            ref_script_cost_stride: 25600, // Hardcoded in the haskell ledger (https://github.com/IntersectMBO/cardano-ledger/blob/3fe73a26588876bbf033bf4c4d25c97c2d8564dd/eras/conway/impl/src/Cardano/Ledger/Conway/Tx.hs#L82)
+            ref_script_cost_multiplier: RationalNumber {
+                numerator: 12,
+                denominator: 10,
+            }, // Hardcoded in the haskell ledger (https://github.com/IntersectMBO/cardano-ledger/blob/3fe73a26588876bbf033bf4c4d25c97c2d8564dd/eras/conway/impl/src/Cardano/Ledger/Conway/Tx.hs#L85)
+            max_epoch: 18,
+            a_0: RationalNumber {
+                numerator: 3,
+                denominator: 10,
+            },
+            n_opt: 500,
+            tau: RationalNumber {
+                numerator: 2,
+                denominator: 10,
+            },
+            rho: RationalNumber {
+                numerator: 3,
+                denominator: 1000,
+            },
+            collateral_percentage: 150,
+            cost_models: CostModels {
+                plutus_v1: vec![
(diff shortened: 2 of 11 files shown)


.gitignore
```diff
@@ -25,6 +25,9 @@ chain.db
 # A directory to store snapshots without committing them
 snapshots/
 
+# Default databse location for the chain store
+chain.db
```
jeluard: It's already there

crates/amaru-ledger/src/rules.rs
```diff
@@ -0,0 +1,193 @@
...
...6f762d218d9d2ca3e67c27f0a3d82cdd1bab4afa7f3f5d3ecb10c6449300c01b55e5d83f6cefc6a12382577fc7f3de09146b5f9d78f48113622ee923c3484e53bff74df65895ec0ddd43bc9f00bf330681811d5d20d0e30eed4e0d4cc2c75d1499e05572b13fb4e7b0dabf6e36d1988b47fbdecffc01316885f802cd6c60e044bf50a15418530d628cffd506d4eb0db6155be94ce84fbf6529ee06ec78e9c3009c0f5504978dd150926281a400d90102828258202e6b2226fd74ab0cadc53aaa18759752752bd9b616ea48c0e7b7be77d1af4bf400825820d5dc99581e5f479d006aca0cd836c2bb7ddcd4a243f8e9485d3c969df66462cb00018182583900bbe56449ba4ee08c471d69978e01db384d31e29133af4546e6057335061771ead84921c0ca49a4b48ab03c2ad1b45a182a46485ed1c965411b0000000ba4332169021a0002c71d14d9010281841b0000000ba43b7400581de0061771ead84921c0ca49a4b48ab03c2ad1b45a182a46485ed1c965418400f6a2001bffffffffffffffff09d81e821bfffffffffffffffe1bfffffffffffffffff68275687474703a2f2f636f73746d646c732e74657374735820931f1d8cdfdc82050bd2baadfe384df8bf99b00e36cb12bfb8795beab3ac7fe581a100d9010281825820794ff60d3c35b97f55896d1b2a455fe5e89b77fb8094d27063ff1f260d21a67358403894a10bf9fca0592391cdeabd39891fc2f960fae5a2743c73391c495dfdf4ba4f1cb5ede761bebd7996eba6bbe4c126bcd1849afb9504f4ae7fb4544a93ff0ea080").expect("Failed to decode Conway3.block hex");
```
jeluard, line 84: Would make sense to have those loaded from separate files

crates/amaru-ledger/src/rules.rs
```diff
@@ -0,0 +1,193 @@
...
+
+use crate::protocol_parameters::ProtocolParameters;
+
+pub enum RuleViolation {
+    BlockBodySizeTooBig(BlockBodySizeTooBig),
+    BlockHeaderSizeTooBig(BlockHeaderSizeTooBig),
+    TooManyExUnitsBlock(TooManyExUnits),
+}
+
+pub enum BlockValidationError {
```
jeluard, line 21: We use `thiserror` for errors. Here you probably would want to have RuleViolation a `kind`.
This would also probably remove the need for a bunch of `map_err` and `From` implementations.

jeluard (changes requested):

crates/amaru-ledger/src/rules.rs
Quantumplation: That's the plan, but for now, we discussed with Arnaud and KtorZ and felt it was better to merge while we explore generating some golden tests loaded from files

crates/amaru-ledger/src/protocol_parameters.rs
```diff
@@ -0,0 +1,262 @@
...
+    pub p1: RationalNumber,
+    pub p2a: RationalNumber,
+    pub p2b: RationalNumber,
+    pub p3: RationalNumber,
+    pub p4: RationalNumber,
+    pub p5a: RationalNumber,
+    pub p5b: RationalNumber,
+    pub p5c: RationalNumber,
+    pub p5d: RationalNumber,
+    pub p6: RationalNumber,
```
KtorZ:
```suggestion
    /// a.k.a. p1
    pub no_confidence: RationalNumber,
    /// a.k.a. p2a
    pub committee: RationalNumber,
    /// a.k.a. p2b
    pub committee_under_no_confidence: RationalNumber,
    /// a.k.a. p3
    pub constitution: RationalNumber,
    /// a.k.a. p4
    pub hard_fork: RationalNumber,
    pub protocol_parameters: ProtocolParametersThresholds,
    /// a.k.a. p6
    pub treasury_withdrawal: RationalNumber,
}

#[derive(Debug, Clone)]
pub struct ProtocolParametersThresholds {
    /// a.k.a. p5a
    pub network_group: RationalNumber,
    /// a.k.a. p5b
    pub economic_group: RationalNumber,
    /// a.k.a. p5c
    pub technical_group: RationalNumber,
    /// a.k.a. p5d
    pub governance_group: RationalNumber,
}
```

Awe x.x ... let's use proper identifier please. For two reasons: 

1. This is extremely error-prone here, since we have a wall of `RationalNumber` that all share the same type; it is easy to mess up the order between two fields. 

2. It is completely unreadable, including to yourself in a month from now. While some Agda spec or mathematical model may named them as such; this isn't math here, it's code and words are cheap. Any field is one tab away from auto-completion, and the compiler takes care of eliminating field names during compilation. 

If you still want to preserve the "mathematical" name, leave them as comments 👍 .

crates/amaru-ledger/src/protocol_parameters.rs
```diff
@@ -0,0 +1,262 @@
...
+}
+
+#[derive(Debug, Clone)]
+pub struct PoolThresholds {
+    pub q1: RationalNumber,
+    pub q2a: RationalNumber,
+    pub q2b: RationalNumber,
+    pub q4: RationalNumber,
+    pub q5e: RationalNumber,
+}
```
KtorZ:
```suggestion
pub struct PoolThresholds {
    /// a.k.a. `q1`
    pub no_confidence: RationalNumber,
    /// a.k.a. `q2a`
    pub committee: RationalNumber,
    /// a.k.a. `q2b`
    pub committee_under_no_confidence: RationalNumber,
    /// a.k.a. `q4`
    pub hard_fork: RationalNumber,
    /// a.k.a. `q5e`
    pub security_group: RationalNumber,
}
```

crates/amaru-ledger/src/protocol_parameters.rs
KtorZ, line 1: This module shall probably live in the `amaru-kernel`.

crates/amaru-ledger/src/rules.rs
```diff
@@ -0,0 +1,193 @@
...
+impl Into<BlockValidationError> for Vec<Option<RuleViolation>> {
+    fn into(self) -> BlockValidationError {
+        BlockValidationError::RuleViolations(
+            self.into_iter()
+                .filter(|r| r.is_some())
+                .map(|r| r.unwrap().into())
+                .collect(),
+        )
+    }
+}
```
KtorZ:
```suggestion
impl From<Vec<Option<RuleViolation>>> for BlockValidationError {
    fn from(violations: Vec<Option<RuleViolation>>) -> Self {
        BlockValidationError::RuleViolations(violations.into_iter().flatten().collect())
    }
}
```

Two things here:

1. Prefer implementing instances of `From` instead of `Into`, it's more versatile. And it is actually suggested by Rust-analyzer, which hints to the fact that you're not using it? I would actually recommend to configure your editor to pick that up! 

2. Every time you find yourself doing `filter` and then `map`, you _probably_ want to use `.filter_map`. In this case however, `filter_map` would lead to `.filter_map(|r| r)`, for which we can simply shortcut to `.flatten()`. This works because `Option` is in fact, an iterator which behaves like a vector with zero (None) or one (Some) element.

crates/amaru-ledger/src/rules.rs
```diff
@@ -0,0 +1,193 @@
...
+) -> Result<(Hash<32>, MintedBlock), BlockValidationError> {
+    let (block_header_hash, minted_block) = parse_block(raw_block)?;
+    let block: Block = minted_block.clone().into();
+
+    block_header_size_valid(minted_block.header.raw_cbor(), &protocol_params)
+        .map_err(|err| BlockValidationError::RuleViolations(vec![err.into()]))?;
+    block_body_size_valid(&minted_block, &protocol_params)
+        .map_err(|err| BlockValidationError::RuleViolations(vec![err.into()]))?;
+
+    let ex_units = match block.transaction_witness_sets {
```
KtorZ, line 46:
```suggestion
    // TODO: rewrite this to use iterators defined on `Redeemers` and `MaybeIndefArray`, ideally 
    // upstreaming them to Pallas.
    let ex_units = match block.transaction_witness_sets {
```

crates/amaru-ledger/src/rules.rs
```diff
@@ -0,0 +1,193 @@
...
+    //     let state = validate_transaction(
+    //         body,
+    //         witness_set,
+    //         is_valid,
+    //         ledger_state,
+    //         protocol_params,
+    //         fast_fail,
+    //     )?;
+    // }
+
```
KtorZ:
```suggestion
```

It's git. Do not commit commented-out code. Either commit it and revert-it so that it's in the history, or just don't commit it 😅

crates/amaru-ledger/src/rules.rs
```diff
@@ -0,0 +1,193 @@
...
+//     //         return Err(BlockValidationError::Composite(
+//     //             error.into(),
+//     //             Box::new(vec![].into()),
+//     //         ));
+//     //     }
+//     // };
+
+//     todo!()
+// }
+
```
KtorZ:
```suggestion
```

crates/amaru-ledger/src/rules.rs
```diff
@@ -0,0 +1,193 @@
...
+fn parse_block(bytes: &[u8]) -> Result<(Hash<32>, MintedBlock<'_>), BlockValidationError> {
```
KtorZ, line 70:
```suggestion
#[instrument(level = Level::TRACE, skip(bytes), fields(block.size = bytes.len()))]
fn parse_block(bytes: &[u8]) -> Result<(Hash<32>, MintedBlock<'_>), BlockValidationError> {
```

I am all for moving code around but hey! Don't remove the traces!

crates/amaru-ledger/src/rules.rs
```diff
@@ -0,0 +1,193 @@
...
+    raw_block: &[u8],
+    protocol_params: ProtocolParameters,
+) -> Result<(Hash<32>, MintedBlock), BlockValidationError> {
+    let (block_header_hash, minted_block) = parse_block(raw_block)?;
+    let block: Block = minted_block.clone().into();
+
+    block_header_size_valid(minted_block.header.raw_cbor(), &protocol_params)
+        .map_err(|err| BlockValidationError::RuleViolations(vec![err.into()]))?;
+    block_body_size_valid(&minted_block, &protocol_params)
+        .map_err(|err| BlockValidationError::RuleViolations(vec![err.into()]))?;
```
KtorZ: I still find this mapping into a singleton vector slightly awkward. Not a blocker but it just feels like something that is introduced _too early_. There's no need to collect multiple errors at this point.

crates/amaru-ledger/src/rules/block/body_size.rs
```diff
@@ -0,0 +1,93 @@
...
+    .is_err()
+    {
+        // TODO: handle appropriately
+        panic!("failed to encode block invalid_transactions in CBOR");
+    }
+
+    let block_body_size = tx_bodies_raw.len()
+        + tx_witness_sets_raw.len()
+        + auxiliary_data_set_raw.len()
+        + invalid_tranasctions_raw.len();
```
KtorZ: This really ought to be `block.raw_cbor().len()` or something like that. We shouldn't be re-serializing and cloning structures here. 

Although, the Haskell implementation seems a little bit contrived about how the block size is computed  (using some "abstract sizing" based on types) . We need to ensure the calculations are the same and not only that an arbitrarily chosen block is below the limit. 

If anything, leave a `// FIXME` at the top of this function with details about what's up and need to be done.

crates/amaru-ledger/src/rules.rs
jeluard: I understand this can be removed altogether by introducing a proper error hierarchy?

KtorZ: I agree.

crates/amaru-ledger/src/protocol_parameters.rs
yHSJ: Heh, yes. Honestly, the reason I went with the names from the spec is because I couldn't find any other names for them used anywhere. Just threw it together to get it working and was gonna ask for standard names (then forgot)

crates/amaru-ledger/src/protocol_parameters.rs
yHSJ: Makes sense, I agree! Will move it there.

crates/amaru-ledger/src/rules/block/body_size.rs
yHSJ: Totally agreed. It seems the block sizing is everything but the header, so I can't use the whole raw CBOR, but yes. This is sloppy and risky.

KtorZ: Could be total size minus header size, though (modulo some CBOR overhead for the wrapping array)?

yHSJ: The check in the Haskell code is only validating that the purported size in the header is within the bounds, not that the actual block size is valid. I would think the actual check to make sure the purported size is accurate would happen during serialization in Pallas (though it doesn't seem to happen currently).

I'll update this to check the header instead of the actual bytes.

crates/amaru-ledger/src/rules.rs
yHSJ: We're going to improve error handling in a future PR, so this is out of scope.

PR #114, opened 2025-02-16
feat: echo service to test integration with maelstrom and simulation-testing

abailly:
This PR introduces a new crate `amaru-sim` containing a binary `echo` whose purpose is to implement the [echo](https://github.com/jepsen-io/maelstrom/blob/main/doc/workloads.md#workload-echo) protocol from maelstrom, in order to experiment and validate both blackbox testing with maelstrom  and deterministic testing with [simulation-testing](https://github.com/pragma-org/simulation-testing) package on something simpler than ouroboros.

This prepares the (re-)introduction of a PR to test Amaru itself and more precisely chain selection logic. Note that it does no use [maelstrom-node](https://docs.rs/crate/maelstrom-node/0.1.6) which has a problematic license and instead favours simpler and explicit specialised messages.

## Summary by CodeRabbit

- **New Features**
  - Introduced an asynchronous echo command offering structured JSON logging and dynamic message processing, including periodic text transformation.
  - Added a new messaging system with `Envelope` and `Message` structures for enhanced message handling.
  - Implemented a new public `echo` module for echoing functionality.
  - Added new dependencies to support asynchronous operations and message handling.

- **Refactor**
  - Updated exit handling and re-organised application modules to streamline command execution and improve overall workflow.
  - Enhanced visibility of `pipeline` and `point` modules, and introduced a new public `exit` module.


crates/amaru-sim/src/bin/echo/main.rs
```diff
@@ -0,0 +1,25 @@
+// Copyright 2024 PRAGMA
```
stevana:
```suggestion
// Copyright 2025 PRAGMA
```

crates/amaru-sim/src/echo/message.rs
```diff
@@ -0,0 +1,90 @@
+// Copyright 2024 PRAGMA
```
stevana:
```suggestion
// Copyright 2025 PRAGMA
```

crates/amaru-sim/src/echo/mod.rs
```diff
@@ -0,0 +1,24 @@
+// Copyright 2024 PRAGMA
```
stevana:
```suggestion
// Copyright 2025 PRAGMA
```

crates/amaru-sim/src/lib.rs
```diff
@@ -0,0 +1,15 @@
+// Copyright 2024 PRAGMA
```
stevana:
```suggestion
// Copyright 2025 PRAGMA
```

crates/amaru-sim/src/echo/message.rs
```diff
@@ -0,0 +1,90 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Clone, Debug, Eq, PartialEq)]
+pub struct Envelope {
+    pub id: u64,
```
stevana: Is this needed, given that all messages already have a msg_id?

crates/amaru-sim/src/echo/run.rs
```diff
@@ -0,0 +1,295 @@
+// Copyright 2024 PRAGMA
```
stevana:
```suggestion
// Copyright 2025 PRAGMA
```

crates/amaru-sim/src/echo/run.rs
```diff
@@ -0,0 +1,295 @@
...
+            .send(msg.into())
+            .await
+            .map_err(|_| WorkerError::Panic)?;
+        Ok(())
+    }
+}
+
+#[derive(Stage)]
+#[stage(name = "do_echo", unit = "Action", worker = "EchoWorker")]
+struct DoEcho {
```
stevana: There are three, largely orthogonal, things going on in this module:

  1. Gasket related things (down/upstream);
  2. Node related things (node_id and count);
  3. The echo "business logic".

It could worth to try to isolate them from each other.

stevana (approved):
Looks good!

crates/amaru-sim/src/echo/message.rs
abailly: Unfortunately yes, I am unsure why but this is part of the messages' schema 🤷 
```
INFO [2025-02-17 09:03:23,004] n1 stderr - maelstrom.process {"timestamp":"2025-02-17T08:03:23.004123Z","level":"INFO","fields":{"message":"input '{\"id\":76,\"src\":\"c2\",\"dest\":\"n1\",\"body\":{\"echo\":\"Please echo 107\",\"type\":\"echo\",\"msg_id\":38}}\n'"},"target":"amaru_sim::echo::run","span":{"name":"execute"},"spans":[{"stage":"read","name":"stage"},{"name":"execute"}]}
```

crates/amaru-sim/src/echo/run.rs
abailly: Agreed. The stage derivation is somewhat annoying, I will see if I can move it out of the way.

crates/amaru-sim/src/echo/message.rs
stevana: Hmm, you're right... That's odd.

For what it's worth, the `id` field isn't listed in the protocol:  https://github.com/jepsen-io/maelstrom/blob/main/doc/protocol.md#messages .

I didn't even notice that it was there in my implementation (I assume the JSON deserialiser simply ignored it).

abailly: I know, that's also why I said I found implementing the protocol a bit confusing

stevana: Does serde complain that there are fields you didn't specify in the target datatype of the deserialisation, or how did you even notice this? Can you tell it to ignore fields you don't have in you Message datatype?

abailly: I have removed the field and it works fine

abailly:
@stevana I have separated the "pure" echo logic (`service.rs`) from the boilerplate (`io.rs`) 

@KtorZ @jeluard I have tried to use exclusive async IOs but this adds some dependencies, not sure we want that, happy to change to something else if you have suggestions. In this particular case the need is to read and write single lines to stdin/stdout. Also I have removed the dependency on EPL license `maelstrom-node`

stevana (approved):

KtorZ (approved):

PR #121, opened 2025-02-20
Rework & trim ouroboros crate

KtorZ:
- :round_pushpin: **chore: tidy up ouroboros::vrf module.**
    - Removed some aliases, adding noise and little benefits;
  - Removed unnecessary unwrap / panic;
  - Removed internal field names, adding noise;
  - Used more consistent naming for errors;
  - Introduced a new 'Input' type to host input challenge related
    interactions;
  - Re-organized module's internal, grouping functions by types;
  - Avoided use internal API and rely on higher level From/Into
    abstractions for conversions.

- :round_pushpin: **chore: rework ouroboros header validation to be more granular**
    As discussed, this changes from the 'BlockValidator' approach to a
  group of small functions that can be invoked independently.

  The 'praos::header' module also now provide a 'assert_all' method to
  bundle all assertion together, although the execution of the
  assertion in parallel has been moved upstream (to amaru-consensus).
  The goal is move that upstream again to 'amaru', but for now at least,
  it removes this logic from the ouroboros crate.

  Note that I have also completely ditched the serde instances, as
  they're completely unnecessary here and only make things more
  complicated underneath. We can revise that decision later, likely
  re-introducing them behind a feature flag.

  Also, the validation errors are now split per assertion kind, and own
  typed arguments, instead of being stringly typed.

## Summary by CodeRabbit

- **New Features**
  - Introduced concurrent header validation, boosting performance during consensus checks.
  - Added new peer management and raw header representation interfaces to enhance error reporting.

- **Refactor**
  - Organised and streamlined module imports and dependency configurations for improved code clarity.
  - Updated cryptographic key and VRF mechanisms to simplify interactions and improve readability.

- **Chores**
  - Removed outdated dependencies and configuration entries.
  - Adjusted test interfaces to align with the new concurrent validation and refactored error handling.


crates/amaru-consensus/src/consensus/header_validation.rs
```diff
@@ -32,12 +32,14 @@ pub fn assert_header<'a>(
...
 
-        let block_validator = BlockValidator::new(header, ledger, epoch_nonce, &active_slots_coeff);
-
-        block_validator.validate().or_panic()
+        praos::header::assert_all(header, ledger, *epoch_nonce, &active_slot_coeff)
+            .and_then(|assertions| {
+                use rayon::prelude::*;
+                assertions.into_par_iter().try_for_each(|assert| assert())
+            })
+            .or_panic()
```
KtorZ: Note: parallel running of the assertions has been moved here from downstream dependency. Ideally, as discussed, this should end up in the amaru crate. But, baby steps.

crates/amaru-consensus/src/consensus/mod.rs
```diff
@@ -40,6 +43,12 @@ pub mod header_validation;
 pub mod nonce;
 pub mod store;
 
+#[derive(Clone)]
+pub enum PullEvent {
+    RollForward(Peer, Point, Vec<u8>, Span),
+    Rollback(Peer, Point, Span),
+}
```
KtorZ: Moved this here from the ouroboros crate; it makes slightly more sense since this is the actual place where this is used.

crates/ouroboros/src/consensus/validator.rs
KtorZ, line 1: Renamed as `ouroboros::praos::header`

crates/amaru-consensus/src/consensus/header_validation.rs
abailly: I am not sure I understand why you want to expose the details of header validation at the level of the `amaru` exe. Or do you mean the decision to evaluate sequentially or concurrently?

crates/ouroboros/src/praos/header.rs
```diff
@@ -0,0 +1,428 @@
...
+            (
+                pool.vrf,
+                FixedDecimal::from(pool.stake) / FixedDecimal::from(pool.active_stake),
+            )
+        })
+        .ok_or(AssertHeaderError::UnknownPool { pool })?;
+
+    Ok(vec![
+        Box::new(move || {
+            AssertKnownLeaderVrfError::new(
```
abailly, line 92: It's a bit weird the assertions should be named `XXXError`, why not the other way round, eg. name the assertions `XXX`? Especially as some of those assertions can return different error (sub) types.

abailly (approved):
Definitely much better, esp. with the structured errors. I have some doubts about the assertion types naming scheme but perhaps just stylistic differences.

crates/amaru-consensus/src/consensus/header_validation.rs
KtorZ: I mean the latter indeed. And I am open to suggestions as for how you would imagine that without leaking internals or exposing "Amaru" to the header validation directly.

jeluard (approved):

PR #128, opened 2025-02-26
feat: keep track of DReps

jeluard:
This is a first step towards full **governance** support. This PR goes straight to the point without introducing significant refactoring that are left for when the global picture will be better understood.

What's done:

* simplify certificate logic handling so that we only deal with atomic certificate (complex ones are flattened)
* improve `DiffBind` error handling
* simplify some abstractions
* keep track of DReps registration/unregistration/update
* keep track of delegations registration/unregistration/update
* keep track of last DRep interaction `epoch` (to prepare for DRep `expiry` handling)
* persist relevant changes

Note that `voting_dreps` is kept separate as it is expected that following PRs will significantly alter `votes` handling.

Note that generalized error handling is for now left out of this PR as long as there is no global strategy to tackle this.

## Summary by CodeRabbit

- **New Features**
	- Enhanced delegation management and voting credentials tracking.
	- Introduced a governance summary to provide structured insights into delegation data.
	- Added a demo synchronisation mode to run the system until a target epoch.
	- New functionality for managing DRep entries in the database, including import and update operations.
	- Added methods for iterating over DRep data in memory and RocksDB implementations.
	- New constant and struct for standardising DRep registration timestamps.
	- New Bash script for monitoring daemon execution with specified parameters.
	- Updates to trace events for improved monitoring capabilities.

- **Refactors**
	- Streamlined transaction processing and ledger state updates for improved reliability.
	- Improved error handling and logging for delegation and certificate operations.

- **Tests**
	- Expanded test coverage to ensure robustness of the new features, including property-based tests for new data structures.

- **Documentation**
	- Updated monitoring logs with new events for delegation registration, updates, and removals.


crates/amaru-ledger/src/state/transaction.rs
```diff
@@ -204,78 +207,103 @@ impl InputsOutputs<true> for MintedTransactionBody<'_> {
...
-                | Certificate::StakeVoteDeleg(credential, pool, ..) => {
-                    trace!(name: "certificate.stake.delegation", target: EVENT_TARGET, parent: parent, credential = ?credential, pool = %pool);
-                    accounts.bind(credential, Some(pool));
-                }
-                Certificate::StakeRegDeleg(credential, pool, ..)
-                // FIXME: register DRep delegation
-                | Certificate::StakeVoteRegDeleg(credential, pool, ..) => {
-                    trace!(name: "certificate.stake.registration", target: EVENT_TARGET, parent: parent, credential = ?credential);
-                    trace!(name: "certificate.stake.delegation", target: EVENT_TARGET, parent: parent, credential = ?credential, pool = %pool);
-                    accounts.register(
```
jeluard, line 231: @KtorZ this is now split into 2 operations (register then bind) so that the last argument of `register` is always `None`. Unclear about the implication: do we want to remove this param and simplify `DiffBind` ?

Looks to me that this simplification is the right direction, even at the cost of removing the opportunity for those kind of optimizations.

crates/amaru-ledger/src/state/diff_bind.rs
```diff
@@ -55,3 +55,47 @@ impl<K: Ord, J: Clone, V> DiffBind<K, J, V> {
...
+    }
+
+    #[test]
+    fn bind_then_register_none() {
+        let mut diff_bind = DiffBind::default();
+        diff_bind.bind(1, Some("c"));
+        diff_bind.register(1, "a", None);
+        assert_eq!(true, diff_bind.unregistered.is_empty());
+        assert_eq!(true, diff_bind.registered.contains_key(&1));
+        assert_eq!(Some(&(None, Some("a"))), diff_bind.registered.get(&1));
```
jeluard: @KtorZ I would assume that this behavior of `register` value overriding `bind` implies we always expect `register` to come first?
If that's true, how about we make sure `bind` fails in case it's called first?

monitoring/README.md
```diff
@@ -178,11 +178,11 @@ Filters can be provided as a sequence of `,`-separated values. Right-most filter
 
 | name                               | severity | description                                          |
 | ---                                | ---      | ---                                                  |
-| `certificate.stake.registration`   | `debug`  | A new stake credential registration was processed    |
-| `certificate.stake.delegation`     | `debug`  | A new stake delegation was processed                 |
-| `certificate.stake.deregistration` | `debug`  | A new stake credential de-registration was processed |
-| `certificate.pool.registration`    | `debug`  | A new stake pool registration was processed          |
-| `certificate.pool.retirement`      | `debug`  | A new stake pool retirement was processed            |
+| `certificate.stake.registration`   | `trace`  | A new stake credential registration was processed    |
```
jeluard, line 181: @KtorZ we should investigate ways to generate those docs to ensure they don't go out of sync

crates/amaru-ledger/src/state/diff_bind.rs
KtorZ: Yes and no. We do expect register to come first, but it's also possible to bind without registering (because the registration might have happened in an earlier block!) 

What is actually an error is calling register after bind. That is simply not allowed. 

The difficulty here is that we are in-between state management and validation rules. Plus, we operate on a partial view of the system here since there might exist previous registrations or delegations in either of the volatile or stable dbs.

crates/amaru-ledger/src/state/transaction.rs
KtorZ: That shouldn't matter indeed; and we could write a quick property test to convince ourselves of it (if it's not already the case).

register(None) >> bind(pool)

==

register(Some(pool))

monitoring/README.md
KtorZ: Indeed, as discussed last week 😅? 

Now I don't remember if you were part of that conversation about having custom macros for tracing and generating doc? 

We haven't captured it as an EDR yet :s

jeluard: Cool! I will keep that in mind!

crates/amaru-ledger/src/state/diff_bind.rs
jeluard: > (because the registration might have happened in an earlier block!)

Of course!
It makes me think that we should have a consistent error strategy here. It looks to me that some failing scenario might generate errors at different time depending on if happening in a single or multiple blocks (i.e. during initial block processing or delayed at the storage layer). e.g. registering the same pool multiple times.

crates/amaru-ledger/src/state/transaction.rs
```diff
@@ -85,197 +94,213 @@ pub fn apply(
     span.exit();
 }
 
-/// A trait to extract inputs, outputs and fees from a transaction; based on whether it is a failed
-/// transaction or not. It is meant to provide a unified interface that makes the parallel between
-/// the two cases.
-trait InputsOutputs<const IS_FAILED: bool> {
```
jeluard, line 91: This felt like unnecessary abstraction so I removed it, open to rollback

crates/amaru-ledger/src/state/diff_bind.rs
```diff
@@ -32,14 +32,28 @@ impl<K: Ord, J, V> Default for DiffBind<K, J, V> {
...
+        if self.registered.contains_key(&k) {
+            return Err(Error::KeyAlreadyUnregistered);
+        }
         self.unregistered.remove(&k);
         self.registered.insert(k, (j, Some(v)));
+        Ok(())
     }
 
-    pub fn bind(&mut self, k: K, j: Option<J>) {
-        assert!(!self.unregistered.contains(&k));
```
jeluard, line 42: We could use `clippy` to prevent assert usage vi this [lint](https://rust-lang.github.io/rust-clippy/master/index.html#/disallowed_macros) unfortunately it can't be configured to be usable in tests too. See https://github.com/rust-lang/rust-clippy/issues/13790

crates/amaru-stores/src/rocksdb/columns/dreps.rs
```diff
@@ -0,0 +1,76 @@
...
+                .map_err(|err| StoreError::Internal(err.into()))?;
+        }  else if let Some(deposit) = deposit {
+            let row = Row {
+                anchor,
+                deposit,
+            };
+            db.put(key, as_value(row))
+                .map_err(|err| StoreError::Internal(err.into()))?;
+        } else {
+            error!(
```
jeluard, line 59: @KtorZ Replicating the behavior from `accounts` now it appears to me that both always have a deposit? This logic derives from `DiffBind` storing values as `Option`.

crates/amaru-stores/src/rocksdb/columns/dreps.rs
```diff
@@ -0,0 +1,76 @@
...
+
+/// Register a new DRep.
+pub fn add<DB>(
+    db: &Transaction<'_, DB>,
+    rows: impl Iterator<Item = (Key, Value)>,
+) -> Result<(), StoreError> {
+    for (credential, (anchor, deposit)) in rows {
+        let key = as_key(&PREFIX, &credential);
+
+        // In case where a registration already exists, then we must only update the underlying
```
jeluard, line 37: @KtorZ Looks to me that there would be some values in having a separated `update` methods in stores.

crates/amaru-stores/src/rocksdb/mod.rs
```diff
@@ -304,12 +306,14 @@ impl Store for RocksDB {
                 utxo::add(&batch, add.utxo)?;
                 pools::add(&batch, add.pools)?;
                 accounts::add(&batch, add.accounts)?;
+                dreps::add(&batch, add.dreps)?;
```
jeluard: All those separate changes to add new entities looks a bit brittle. There's tradeoff to be made if there's a need to introduce some more genericity, considering that the number of new entities is capped.

crates/amaru-ledger/src/state/diff_bind.rs
KtorZ: I would definitely treat any oddities like that in the storage layer as _errors_. Anything that the store receives should have been validated and the storage present our last safety net to check for consistency. 

Yet, yes, you are right. Some errors are just "local to the transaction", while others depend on the ongoing state. In principle, those errors shall be what the ledger rules enforce and check. So, as we process the state, we might _also_ yield validation errors unless we manage to treat the state changes as a _non faillible_ builder API (driven by the ledger rules). 

In which case, the validation rules become in fact the only place where we return validation failures whereas other places (state management or storage) are plain panics (wrapped in value errors, yes, but with a different semantic than the validation failures).

crates/amaru-stores/src/rocksdb/columns/dreps.rs
KtorZ: So technically:

- DReps always have an explicit deposits
- Accounts *may* have an explicit deposits; historically, they had not. 

Although if I recall correctly how this was implemented in Amaru, we are _defaulting_ to the protocol parameters' deposit value when the deposit is unspecified. So from a storage perspective, they should indeed always have a deposit. 

That's a simplification we could do on the DiffBind structure; I am honestly not quite sure why it's still an option. Maybe trying to remove it will re-surface the reason why ^^"

crates/amaru-stores/src/rocksdb/mod.rs
KtorZ: It's an awkward trade-off indeed. On the one hand, it's straightforward to read, write and modify. There's nothing inherently complicated with this code and it doesn't really change _that_ often. On the other hand, it's easy to forget to extend or change when we add new columns.

But the effect of forgetting anything here is immediately visible; you just won't get anything out of the app. So the risk of "shipping it wrong" is rather low. 

Which is why I am not _so interested_ in making this generic or more abstract. There are enough small subtleties between those methods to make a generic abstraction over it potentially way more complex than spelling them out. 

I think two areas we could look into making generic / more abstract to avoid silly mistakes and catastrophic accident are:

1. The conversion from key objects to their prefixed rocksdb bytes identifier. Right now, you _must_ call `as_key` very religiously, otherwise you may end up storing things with a wrong key without actually noticing the damage until it's perhaps too late. This screams for a `AsKey` trait that we require on each key type. 

2. Most of those function calls, as well as few others on epoch boundary MUST happen in a db transaction. Currently, they happen in _multiple_ db transactions so the system is "easily" corruptible. So rather than hiding away the construction and commit of the db transaction, the API should make the transaction management explicit to the caller. We can still _abstract away_ the RocksDB internal but we most likely need to reverse this responsibility here.

crates/amaru-ledger/src/store.rs
```diff
@@ -101,11 +101,15 @@ pub trait Store: Snapshot + Send + Sync {
...
             impl Iterator<Item = (accounts::Key, accounts::Value)>,
+            impl Iterator<Item = (dreps::Key, dreps::Value)>,
+            impl Iterator<Item = (delegations::Key, delegations::Value)>,
         >,
         remove: Columns<
             impl Iterator<Item = utxo::Key>,
             impl Iterator<Item = (pools::Key, Epoch)>,
             impl Iterator<Item = accounts::Key>,
+            impl Iterator<Item = dreps::Key>,
+            impl Iterator<Item = delegations::Key>,
```
jeluard: Unused but required by type

crates/amaru-stores/src/rocksdb/columns/dreps.rs
```diff
@@ -0,0 +1,82 @@
...
+            if let Some(epoch) = epoch {
+                row.last_interaction = epoch;
+            }
+            db.put(key, as_value(row))
+                .map_err(|err| StoreError::Internal(err.into()))?;
+        } else if let Some(deposit) = deposit {
+            let row = Row {
+                anchor,
+                deposit,
+                last_interaction: epoch.unwrap(), // Cannot fail; this is a new registration.
```
jeluard: This is unfortunate. Will be probably improved in further PR working on governance.

crates/amaru-ledger/src/store/columns/delegations.rs
KtorZ, line 1: Any reason why this isn't stored next to pool delegations inside the `accounts` column? I believe they fulfill the same role and relate to the same entities. So it would make more sense to extend the `accounts` column?

crates/amaru-ledger/src/store/columns/dreps.rs
```diff
@@ -0,0 +1,69 @@
...
+
+use amaru_kernel::{cbor, Anchor, Epoch, Lovelace, StakeCredential};
+use iter_borrow::IterBorrow;
+
+pub const EVENT_TARGET: &str = "amaru::ledger::store::dreps";
+
+/// Iterator used to browse rows from the Accounts column. Meant to be referenced using qualified imports.
+pub type Iter<'a, 'b> = IterBorrow<'a, 'b, Key, Option<Row>>;
+
+pub type Value = (Option<Anchor>, Option<Lovelace>);
```
KtorZ: I don't think the DRep's deposit can ever be `None`? Unless I am missing something?

crates/amaru-stores/src/rocksdb/mod.rs
```diff
@@ -305,11 +313,27 @@ impl Store for RocksDB {
...
+                let dreps = add
+                    .dreps
+                    // Each new DRep has an associated epoch
+                    .map(|(key, value)| (key, value, Some(epoch)))
+                    // Voting DReps get their epoch extended
+                    .chain(
+                        voting_dreps
+                            .into_iter()
+                            .map(|drep| (drep, (None, None), Some(epoch))),
+                    );
```
KtorZ: Worth updating via a different method? I presume this is why you need the deposit to be an `Option`? Something like `tick_drep` or whatever that would indicate we're not creating a new drep but simply updating its lifetime.

crates/amaru-ledger/src/store/columns/dreps.rs
jeluard: That's indirectly due to `DiffBind#registered` type (`BTreeMap<K, (Option<J>, Option<V>)>`). I didn't want to update this structure as part of this PR but that would make sense to reconsider this once the global picture is known.

crates/amaru-ledger/src/store/columns/delegations.rs
jeluard: _Edit: This was based on a misunderstanding of the comment_

`delegations` are a map that can be updated frequently irregardless of the `account` lifecycle. Storing them as part of an `account` would imply having to read/deserialize, update the map, serialize and finally update the whole key per delegation change.

I described a bit how I think the workflow could be achieved with this proposed layout (tradeoff being that there are some extra computations needed at startup).

Overall it looks like we could benefit some more structure for storage depending on overall entities lifecycle (what is updated when, what has to be computed when).

crates/amaru-stores/src/rocksdb/mod.rs
jeluard: It makes sense especially given in the future we could decide to update those only at the epoch boundary.

crates/amaru-ledger/src/store/columns/accounts.rs
```diff
@@ -12,22 +12,27 @@
...
 use iter_borrow::IterBorrow;
 
 pub const EVENT_TARGET: &str = "amaru::ledger::store::accounts";
 
 /// Iterator used to browse rows from the Accounts column. Meant to be referenced using qualified imports.
 pub type Iter<'a, 'b> = IterBorrow<'a, 'b, Key, Option<Row>>;
 
-pub type Value = (Option<PoolId>, Option<Lovelace>, Lovelace);
+pub type Value = (
+    Option<(Option<PoolId>, Option<DRep>)>,
```
jeluard: Required by DiffBind feature to bind a previous value.

KtorZ: Wat

crates/amaru-ledger/src/state.rs
```diff
@@ -84,6 +85,16 @@ where
     stake_distributions: Arc<Mutex<VecDeque<StakeDistribution>>>,
 }
 
+fn parse_voting_procedures(voting_procedures: &VotingProcedures) -> HashSet<StakeCredential> {
+    voting_procedures
+        .iter()
+        .filter_map(|(k, _)| match k {
+            Voter::DRepKey(hash) => Some(StakeCredential::AddrKeyhash(*hash)),
```
KtorZ, line 91:
```suggestion
            Voter::DRepKey(hash) => Some(StakeCredential::AddrKeyhash(*hash)),
            Voter::DRepScript(hash) => Some(StakeCredential::ScriptHash(*hash)),
```

KtorZ: Dreps can also be scripts.

crates/amaru-ledger/src/state.rs
```diff
@@ -84,6 +85,16 @@ where
...
+            _ => None,
```
```suggestion
            Voter::ConstitutionalCommitteeKey(..) | Voter::ConstitutionalCommitteeScript(..) | StakePoolKey(..) => None,
```

That's probably something to put on a style guide but, I've been bitten by that enough in the past to avoid _reckless wildcards_ in places like this. It is very possible that we have to support new tpe of DReps or credentials in the future, and it would be nice to let the compiler have our back if this enum grows bigger.

crates/amaru-ledger/src/state.rs
```diff
@@ -84,6 +85,16 @@ where
     stake_distributions: Arc<Mutex<VecDeque<StakeDistribution>>>,
 }
 
+fn parse_voting_procedures(voting_procedures: &VotingProcedures) -> HashSet<StakeCredential> {
```
```suggestion
fn select_stake_credentials(voting_procedures: &VotingProcedures) -> HashSet<StakeCredential> {
```

There's no parsing going on here whatsoever 😅!

crates/amaru-ledger/src/state.rs
```diff
@@ -352,6 +365,17 @@ impl<S: Store> State<S> {
             let transaction_id = Hasher::<256>::hash(transaction_body.raw_cbor());
             let transaction_body = transaction_body.unwrap();
 
+            let voting_procedures: Option<
+                NonEmptyKeyValuePairs<Voter, NonEmptyKeyValuePairs<GovActionId, VotingProcedure>>,
+            > = transaction_body.voting_procedures.clone();
+            let voting_dreps = voting_procedures
+                .as_ref()
```
```suggestion
            let voting_dreps = transaction_body
                .voting_procedures
                .as_ref()
```

Unnecessary clone?

crates/amaru-ledger/src/state.rs
```diff
@@ -352,6 +365,17 @@ impl<S: Store> State<S> {
...
+                .map(parse_voting_procedures)
```
```suggestion
                .map(select_stake_credentials)
```

crates/amaru-ledger/src/state.rs
```diff
@@ -352,6 +365,17 @@ impl<S: Store> State<S> {
...
+                .unwrap_or_default();
+            state.voting_dreps.extend(voting_dreps);
+
+            // Calculate votes for all pending  voting procedures per block ?
```
```suggestion
            // TODO: Calculate votes for all pending  voting procedures per block ?
```

crates/amaru-ledger/src/state/diff_bind.rs
```diff
@@ -32,26 +32,82 @@ impl<K: Ord, J, V> Default for DiffBind<K, J, V> {
...
+    KeyAlreadyRegistered,
+    #[error("Key is already unregistered")]
+    KeyAlreadyUnregistered,
+}
+
 impl<K: Ord, J: Clone, V> DiffBind<K, J, V> {
-    pub fn register(&mut self, k: K, v: V, j: Option<J>) {
+    pub fn register(&mut self, k: K, v: V, j: Option<J>) -> Result<(), Error> {
+        if self.registered.contains_key(&k) {
+            return Err(Error::KeyAlreadyUnregistered);
```
KtorZ, line 105:
```suggestion
            return Err(Error::KeyAlreadyRegistered);
```

?

crates/amaru-ledger/src/store/columns/accounts.rs
jeluard: I improved it a bit just for you

crates/amaru-ledger/src/state.rs
jeluard: Needed because `transaction_body.voting_procedures` introduces a partial move, and `transaction_body` is passed by value to `transaction::apply`.

KtorZ: It compiles just fine on my end 😅... `as_ref()` is doing the trick, taking only a reference and thus not moving the voting procedure out of the body whatsoever.

crates/amaru-ledger/src/state/diff_bind.rs
KtorZ: I suppose it's the eternal discussion between user errors and exceptions. Here, the asset was to signify an API misuse. It's a developer mistake to run into this problem; and it indicates that it is your job to check that you cannot bind on an unregistered element.

In this particular case, it's arguable. As we discussed, the line between state management and validation rules is really blurry. In a way, returning an error now makes this instrumental for implementation validation rules; whereas the assertion assumes that the invocations are valid.

crates/amaru-ledger/src/state/transaction.rs
KtorZ: Felt cute at the time; completely unnecessary I concede. Two functions is more boring. We like boring.

crates/amaru-ledger/src/state.rs
jeluard: My bad, you are right. Introducing the local `voting_procedures` was indeed making the clone necessary.

crates/amaru-ledger/src/state.rs
jeluard: Good point! There is also a [clippy rule](https://rust-lang.github.io/rust-clippy/master/index.html#wildcard_enum_match_arm) for that. Might be worth adding. We have a couple of those in the codebase still.

crates/amaru-ledger/src/state/transaction.rs
```diff
@@ -85,197 +96,220 @@ pub fn apply(
...
+            },
+            Certificate::UpdateDRepCert(credential, anchor) => {
+                trace!(name: "drep.update", target: EVENT_TARGET, parent: parent, credential = ?credential, anchor = ?anchor);
+                dreps.bind_left(credential, Option::from(anchor)).unwrap();
+            },
+            Certificate::VoteDeleg(credential, drep) => {
+                trace!(name: "vote.delegation", target: EVENT_TARGET, parent: parent, credential = ?credential);
+                accounts.bind_right(credential, Some(drep)).unwrap();
+            },
+            // Ignore complex type certificates as they have been made useless via `flatten_certificate`
```
KtorZ: This kind of thing makes me uneasy because we have a statically typed language and we could enforce this through types. At the same time, this feels a bit daunting. Another option could be to flatten inline and recursively call ourselves? 

For example, we could imagine defining `apply_certificate` instead of `apply_certificates`

```rs
fn apply_certificate(
    parent: &Span,
    pools: &mut DiffEpochReg<PoolId, PoolParams>,
    accounts: &mut DiffBind<StakeCredential, PoolId, DRep, Lovelace>,
    dreps: &mut DiffBind<StakeCredential, Anchor, Empty, Lovelace>,
    certificate: Certificate,
)
```

and use it as:

```rs
        Certificate::StakeVoteDeleg(credential, pool, drep) => {
            let drep_deleg = Certificate::VoteDeleg(credential.clone(), drep);
            apply_certificate(parent, pools, accounts, dreps, drep_deleg);
            let pool_deleg = Certificate::StakeDelegation(credential, pool);
            apply_certificate(parent, pools, accounts, dreps, pool_deleg);
        }
        Certificate::StakeRefDeleg(credential, pool) => { .. }
```

while, at call-site, it merely becomes:

```rs
    certificates.into_iter().for_each(|certificate| {
        apply_certificate(
            &span,
            &mut state.pools,
            &mut state.accounts,
            &mut state.dreps,
            certificate,
        )
    });
```

This way, we keep `apply_certificate` _complete_, and we avoid too much repetition also between each certificate handling?

crates/amaru-ledger/src/store/columns/accounts.rs
```diff
@@ -53,9 +54,10 @@ impl<C> cbor::encode::Encode<C> for Row {
         e: &mut cbor::Encoder<W>,
         ctx: &mut C,
     ) -> Result<(), cbor::encode::Error<W::Error>> {
-        e.array(3)?;
+        e.array(4)?;
         e.encode_with(self.delegatee, ctx)?;
         e.encode_with(self.deposit, ctx)?;
+        e.encode_with(self.drep.clone(), ctx)?;
```
KtorZ:
```suggestion
        e.encode_with(self.drep.as_ref(), ctx)?;
```

crates/amaru-ledger/src/state/transaction.rs
jeluard: Makes sense

crates/amaru-ledger/src/state/transaction.rs
```diff
@@ -85,197 +98,210 @@ pub fn apply(
...
+            let reg = Certificate::Reg(credential.clone(), coin);
+            apply_certificate(parent, pools, accounts, dreps, reg);
+            let pool_deleg = Certificate::StakeDelegation(credential.clone(), pool);
+            apply_certificate(parent, pools, accounts, dreps, pool_deleg);
+            let drep_deleg = Certificate::VoteDeleg(credential, drep);
+            apply_certificate(parent, pools, accounts, dreps, drep_deleg);
+        }
+        Certificate::VoteRegDeleg(credential, drep, coin) => {
+            let drep_deleg = Certificate::Reg(credential.clone(), coin);
+            apply_certificate(parent, pools, accounts, dreps, drep_deleg);
```
KtorZ:
```suggestion
            let reg = Certificate::Reg(credential.clone(), coin);
            apply_certificate(parent, pools, accounts, dreps, reg);
```

crates/amaru-ledger/src/state.rs
```diff
@@ -236,12 +258,27 @@ impl<S: Store> State<S> {
...
+                    voting_dreps,
                 )
                 .and_then(|()| {
                     db.with_pots(|mut row| {
                         row.borrow_mut().fees += fees;
                     })
                 })
+                .and_then(|()| {
+                    // Go over all accounts and remove the DRep if it's unregistered
+                    db.with_accounts(|iterator| {
```
jeluard: This require iterating over all accounts which is a performance issue. It might make sense to introduce de-normalization here.

KtorZ: Indeed. I was thinking about the following options:

1. Have DReps keep track of their delegators; although this could be problematic for very large dreps. We could potentially save space and make this relatively efficient by using a Patricia Trie. 

2. Attach a date (slot) to certificates. So that, when building the stake distribution, we can compare the dates of registration of a drep with the ongoing delegations.

jeluard: Option 2. sounds like a good plan. It reduces the overall performance hit. Additionally we might clear those older delegations from state async by triggering some tasks when they are discovered.

crates/amaru-ledger/src/governance.rs
```diff
@@ -13,20 +13,27 @@ impl DRepsSummary {
...
         let delegations = db
             .iter_accounts()?
             .filter_map(|(credential, account)| {
-                account.drep.and_then(|drep| {
-                    if !dreps.contains(&drep) {
-                        return None;
+                account.drep.and_then(|(drep, since)| {
+                    let registered_at = dreps.get(&drep);
+                    if let Some(registered_at) = registered_at {
+                        if since > *registered_at {
```
KtorZ: Now is where we need to carefully check things. I would assume `>=` is okay from a slot-standpoint, because the registration can happen in the same transaction. But that means we need to include the certificate position as part of the ordering. Because certificates are ordered, and so long as the DRep is registered beforehand, then the delegation shall be authorized in the same transaction.

crates/amaru-ledger/src/store/columns/dreps.rs
```diff
@@ -12,22 +12,23 @@
...
-pub type Value = (Option<Anchor>, Option<Lovelace>, Epoch);
+pub type Value = (Option<Anchor>, Option<Lovelace>, Slot, Epoch);
 
 pub type Key = StakeCredential;
 
 #[derive(Debug, Clone, PartialEq, Default)]
 pub struct Row {
     pub deposit: Lovelace,
     pub anchor: Option<Anchor>,
+    pub registered_at: Slot,
```
KtorZ: We're going to need three indices here in the end: Slot, transaction position within the block, and certification position within the transaction. The ordering must be total.

KtorZ: Since we these indices will carry a quite precise semantic, then a dedicated type is indicated (e.g. `CertificatePointer`, where we define and test the ordering behaviour accordingly).

crates/amaru-kernel/src/lib.rs
```diff
@@ -115,6 +116,9 @@ pub static PLEDGE_INFLUENCE: LazyLock<Ratio<BigUint>> =
 /// The optimal number of stake pools target for the incentives, a.k.a k
 pub const OPTIMAL_STAKE_POOLS_COUNT: usize = 500;
 
+/// Epoch duration after which inactive DReps are considered expired.
+pub const DREP_EXPIRY: u64 = 20;
```
yHSJ, line 120: In my work for block validation, I've added a `ProtocolParamters` struct, so we don't have to hard code these values.

https://github.com/pragma-org/amaru/blob/main/crates/amaru-kernel/src/protocol_parameters.rs#L4

crates/amaru-ledger/src/governance.rs
```diff
@@ -0,0 +1,106 @@
+use crate::store::{columns::dreps::Row, Snapshot, StoreError};
+use amaru_kernel::{encode_bech32, DRep, StakeCredential};
+use serde::ser::SerializeStruct;
+use std::collections::BTreeMap;
+
+#[derive(Debug)]
+// TODO: Also add DRep information to this summary
+pub struct DRepsSummary {
```
yHSJ, line 8: Does it make more sense to keep each of these "slices" of the ledger state `amaru-kernel::store` for organization sake?

crates/amaru-ledger/src/governance.rs
```diff
@@ -0,0 +1,106 @@
...
+        });
+
+        s.serialize_field("keys", &keys)?;
+        s.serialize_field("scripts", &scripts)?;
+
+        s.end()
+    }
+}
+
+/// Serialize a (registerd) DRep to bech32, according to [CIP-0129](https://cips.cardano.org/cip/CIP-0129).
```
yHSJ, line 68: What's going on with this commented code?

crates/amaru-ledger/src/governance.rs
yHSJ: I see now that this follows the structure we have for other pieces of the ledger, so maybe it's a conversation for another time.

yHSJ (commented):
I'm making an effort to be more engaged in the review process of PRs. It's worth noting that I have little context on Amaru as a whole (and my rust is a work in progress), so take my thoughts with a grain of salt. But, gotta start somewhere!

crates/amaru-ledger/src/governance.rs
KtorZ: It's called documentation tests! 

See -> https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html

crates/amaru-ledger/src/governance.rs
KtorZ: Those are store-independent, though. Yet, you're right that the _slices_ will probably end up being quite close to those; if not those exactly for some of them.

crates/amaru-kernel/src/lib.rs
KtorZ: Yes and no. Unfortunately, _just using the new ProtocolParameters struct_ doesn't really change the problem that those values need to come from _somewhere_. 

Ideally (i.e. moving forward), we shall:

1. Extract them from the initial snapshots.
2. Update them according to governance settings.  

We could start with (1) rather soon, that's a good isolated issue which can make use of the new struct indeed.

crates/amaru-ledger/src/governance.rs
yHSJ: Wow, learn something new every day!

This is really cool, thanks for introducing me

crates/amaru-kernel/src/lib.rs
yHSJ: Of course, but I'm not sure why that doesn't remove the need for these constants? Anywhere we are using these constants we can substitute the usage of the ProtocolParameters. When calling those functions, we can pass in temporary hard coded values (like these constants), until we actually handle correctly storing and updating the pps.

Wouldn't that make it easier to update in the future (fewer points of change)?

KtorZ: Yes, that's definitely step 0. Although, right now, it doesn't get us closer to our immediate goal w.r.t governance: we still need to instantiate all those values at the beginning with hard-wired values; and we aren't solving any _known unknown_. Threading the protocol parameters through the app is something we know we have to do, but the focus _right now_ isn't about solving the things we know. But figuring out the things we don't. Hence the global constants.

PR #129, opened 2025-02-27
Initial Transaction Validations

yHSJ:
This PR introduces some initial transaction validations for the block validation

## Summary by CodeRabbit

- **New Features**
	- Updated the public transaction interface with new utility types, including `Bytes`, `AuxiliaryData`, and `PseudoTransactionOutput`.
	- Enhanced block validation by incorporating detailed transaction validation checks, including specific error reporting for invalid transactions and metadata.
	- Organised transaction validations into modular components, introducing new modules for managing non-disjoint reference inputs, transaction metadata, and output size.
	- Simplified output size validation logic for improved readability and maintainability.
	- Renamed and restructured block body size validation to focus on direct size comparisons.
	- Introduced a trait-based approach for accessing Lovelace values, enhancing code clarity.

- **Chores**
	- Added a new workspace dependency for better project management.


crates/amaru-ledger/src/rules/transaction/disjoint_ref_inputs.rs
```diff
@@ -0,0 +1,36 @@
+use amaru_kernel::{TransactionBody, TransactionInput};
+
+use crate::rules::TransactionRuleViolation;
+
+pub struct NonDisjointRefInputs {
+    pub intersection: Vec<TransactionInput>,
+}
+
+impl Into<TransactionRuleViolation> for NonDisjointRefInputs {
```
yHSJ, line 9: I attempted to implement the `From` trait here instead, but I ran into a strange `rustc` error I'm still working on debugging.

crates/amaru-ledger/src/rules.rs
```diff
@@ -43,29 +63,87 @@ pub fn validate_block(
...
+    let witness_sets = block.transaction_witness_sets.to_vec();
+
+    let invalid_transactions = match block.invalid_transactions {
+        Some(invalid_transactions) => invalid_transactions.to_vec(),
+        None => Vec::new(),
+    };
+
+    for (i, (transaction, raw_transaction)) in
+        (0u32..).zip(transactions.iter().zip(raw_transactions))
+    {
```
jeluard: First `zip` might be replaced with `transactions.iter().enumerate` ?

crates/amaru-ledger/src/rules/transaction/output_size.rs
```diff
@@ -0,0 +1,53 @@
...
+                    amaru_kernel::alonzo::Value::Coin(lovelace) => lovelace,
+                    amaru_kernel::alonzo::Value::Multiasset(lovelace, _) => lovelace,
+                },
+                amaru_kernel::PseudoTransactionOutput::PostAlonzo(post_alonzo) => {
+                    match post_alonzo.value {
+                        amaru_kernel::Value::Coin(lovelace) => lovelace,
+                        amaru_kernel::Value::Multiasset(lovelace, _) => lovelace,
+                    }
+                }
+            };
```
KtorZ: See: https://github.com/pragma-org/amaru/blob/main/crates/amaru-kernel/src/lib.rs#L396

crates/amaru-ledger/src/rules/transaction/output_size.rs
```diff
@@ -0,0 +1,53 @@
...
+
+            lovelace <= bytes.len() as u64 * coins_per_utxo_byte
+        })
+        .collect::<Vec<_>>();
+
+    if outputs_too_small.is_empty() {
```
KtorZ, line 32: It would be better to use `any` instead of `filter` on the iterator. So that we only yield what's necessary, and the entire traversal can be short-circuited in case an invalid case is found.

crates/amaru-ledger/src/rules/transaction/output_size.rs
```diff
@@ -0,0 +1,53 @@
...
+    }
+}
+
+pub fn validate_output_size(
+    transaction: &TransactionBody,
+    protocol_parameters: &ProtocolParameters,
+) -> Result<(), OutputTooSmall> {
+    let coins_per_utxo_byte = protocol_parameters.coins_per_utxo_byte;
+    let outputs_too_small = transaction
+        .clone()
```
KtorZ: Hugh. If anything, you should clone the `outputs`,  (`.iter().cloned()`) but not the entire transaction. Still, I don't think validation rules should be cloning anything. 

Rather, we should introduce a lifetime on the error which is bound to the transaction body.

crates/amaru-ledger/src/rules/transaction/metadata.rs
```diff
@@ -0,0 +1,53 @@
...
+        TransactionRuleViolation::InvalidTransactionMetadata(self)
+    }
+}
+
+// impl From<InvalidTransactionMetadata> for TransactionRuleViolation {
+//     fn from(value: InvalidTransactionMetadata) -> Self {
+//         TransactionRuleViolation::InvalidTransactionMetadata(value)
+//     }
+// }
+
```
KtorZ:
```suggestion
```

crates/amaru-ledger/src/rules/transaction/metadata.rs
```diff
@@ -0,0 +1,53 @@
...
+pub fn validate_metadata(
+    transaction: &TransactionBody,
+    auxilary_data: Option<AuxiliaryData>,
+) -> Result<(), InvalidTransactionMetadata> {
+    match (transaction.auxiliary_data_hash.clone(), auxilary_data) {
```
KtorZ: Same here, the clone should be avoided. Introducing lifetime to all errors is likely the way to go.

crates/amaru-ledger/src/rules/transaction/disjoint_ref_inputs.rs
```diff
@@ -0,0 +1,36 @@
...
+        TransactionRuleViolation::NonDisjointRefInputs(self)
+    }
+}
+
+// impl From<NonDisjointRefInputs> for TransactionRuleViolation {
+//     fn from(value: NonDisjointRefInputs) -> Self {
+//         TransactionRuleViolation::NonDisjointRefInputs(value)
+//     }
+// }
+
```
KtorZ:
```suggestion
```

crates/amaru-ledger/src/rules.rs
yHSJ: enumerate returns a usize, not a u32. Since I need to check the presence of a u32 in an array with `i`, I thought it was safer to use `zip` and cast from u32 to usize where needed.

crates/amaru-ledger/src/rules/transaction/output_size.rs
yHSJ: I had that, but in the _current_ error handling I decided that this was better as it allowed me to provide context.

crates/amaru-ledger/src/rules/transaction/output_size.rs
yHSJ: Yeah, this is something I feel is pretty weak in my Rust abilities -- avoiding cloning with better, and perhaps more complicated, idiomatic solutions. I'll add a TODO to clean up these clones.

KtorZ: Is there any ways we can help with that? Perhaps we can do some pairing to _remove those clones_ and see where this leads us?

crates/amaru-ledger/src/rules.rs
KtorZ: I think it's reasonable to expect less than 4 billions transactions in a block 😄 ... A comment to justify that sounds acceptable.

crates/amaru-ledger/src/rules/transaction/output_size.rs
yHSJ: Yeah, that could absolutely be helpful. I can also work with Pi or another Sundae person on that, to avoid taking you off your tasks.

crates/amaru-ledger/src/rules/transaction/disjoint_ref_inputs.rs
jeluard: Looks like worth looking into it. Please share the rustc error if it can be helpful.

crates/amaru-ledger/src/rules/transaction/output_size.rs
```diff
@@ -0,0 +1,45 @@
...
+) -> Result<(), OutputTooSmall> {
+    let coins_per_utxo_byte = protocol_parameters.coins_per_utxo_byte;
+    let outputs_too_small = transaction
+        .outputs
+        .clone()
+        .into_iter()
+        .filter(|output| {
+            let mut bytes = Vec::new();
+            cbor::encode(output, &mut bytes)
+                .unwrap_or_else(|_| panic!("Failed to serialize output"));
```
jeluard: It would make sense to handle this error via proper error handling now. It would force us to figure out what's the proper way to handle this case and design a proper API. 
We probably don't want to crash the node here.

i understand it's a PITA, but it's one we want to solve sooner rather than later IMO

jeluard: If you rebase on `main` you should be good to go thanks to https://github.com/pragma-org/amaru/pull/146

crates/amaru-ledger/src/rules.rs
```diff
@@ -39,33 +59,93 @@ pub fn validate_block(
...
-    .collect::<Vec<_>>();
+    let ex_units = block
+        .transaction_witness_sets
+        .clone()
+        .to_vec()
+        .into_iter()
+        .flat_map(|witness_set| {
+            witness_set
+                .redeemer
+                .into_iter()
```
rrruko:
```suggestion
        .iter()
        .flat_map(|witness_set| {
            witness_set
                .redeemer
                .iter()
```
you can avoid cloning here

crates/amaru-ledger/src/rules.rs
```diff
@@ -39,33 +59,93 @@ pub fn validate_block(
...
+    let raw_transactions = minted_block.transaction_bodies.clone().to_vec();
+
+    let witness_sets = block.transaction_witness_sets.to_vec();
+
+    let invalid_transactions = match block.invalid_transactions {
+        Some(invalid_transactions) => invalid_transactions.to_vec(),
+        None => Vec::new(),
+    };
+
+    // using `zip` here instead of enumerate as it is safer to cast from usize to u32 than u32 to usize
```
rrruko: nitpick: this comment is the wrong way around. the safe cast is the u32 -> usize cast. the code is correct

crates/amaru-ledger/src/rules.rs
```diff
@@ -119,27 +199,4 @@ mod tests {
             })
         )
     }
-
-    #[test]
-    #[allow(clippy::wildcard_enum_match_arm)]
-    fn validate_block_body_size_too_big() {
```
rrruko, line 125: is it ok to delete this test?

crates/amaru-ledger/src/rules/block/body_size.rs
```diff
@@ -1,58 +1,52 @@
...
         Ok(())
     }
 }
 
-#[cfg(test)]
-mod tests {
-    use super::*;
-    use amaru_kernel::{cbor, Block, MintedBlock};
-    #[test]
-    fn test_block_body_size_valid() {
```
rrruko, line 46: is it ok to delete this test?

crates/amaru-ledger/src/rules.rs
yHSJ: that's what I meant, good catch

yHSJ:
Closed in favor of https://github.com/pragma-org/amaru/pull/145 instead

PR #130, opened 2025-02-28
Compute VRF nonces while following headers

KtorZ:
Instead of a relying on a pre-fetched nonce table.

> [!NOTE]
> This is a Draft PR because it's still lack one important element to _work_: bootstrapping. Yes, we need yet-another-snapshot since nonces information are not part of the `NewEpochState`, nor can they be simply derived from headers. It's a rolling computation which needs to be extracted from the "extended ledger state" (which lives in the consensus). 
>
> I've made this information "queriable" [in Ogmios](https://github.com/CardanoSolutions/ogmios/commit/b11967a3455a8e96bd62baae66e06cd1b3d2fb50) but I'll add a few command-line options to the import command in order to specify the required value to bootstrap Amaru's internal state. 

## Summary by CodeRabbit

- **New Features**
  - Introduced an additional command for importing VRF nonce data, complementing the refined ledger state and block header import commands.
  - Enhanced the log filtering and error reporting in the command-line interface for improved clarity during operations.

- **Documentation**
  - Updated user instructions and examples in the README, now clearly detailing the revised commands for importing ledger snapshots, header data, and nonce information.


crates/amaru-consensus/src/consensus/chain_selection.rs
```diff
@@ -275,23 +275,79 @@ where
...
+    }
+
+    proptest! {
+        #[test]
+        fn prop_roundtrip_cbor(hdr in any_test_header()) {
+            let bytes = to_cbor(&hdr);
+            let hdr2 = from_cbor::<FakeHeader>(&bytes).unwrap();
+            assert_eq!(hdr, hdr2);
+        }
+    }
```
KtorZ, line 337: Note: this was moved from the `consensus::header::test` module here; while the definition of `TestHeader` has been moved to `ouroboros_trait` and renamed `FakeHeader`.

crates/amaru-consensus/src/consensus/header_validation.rs
```diff
@@ -159,24 +146,40 @@ impl Consensus {
...
+        )
         .entered();
 
-        let header: babbage::MintedHeader<'_> = minicbor::decode(raw_header)
+        let header: MintedHeader<'_> = minicbor::decode(raw_header)
             .map_err(|_| ConsensusError::CannotDecodeHeader(point.clone()))?;
 
+        let header_hash = Hasher::<256>::hash(header.header_body.raw_cbor());
+
+        // first make sure we store the header
```
KtorZ, line 169: My bad, this comment shouldn't have been moved.

crates/ouroboros/src/praos/nonce.rs
```diff
@@ -0,0 +1,50 @@
+// Copyright 2024 PRAGMA
```
abailly, line 1: Super minor nitpick but shouldn't this be 2025 ?

crates/ouroboros/src/praos/nonce.rs
```diff
@@ -0,0 +1,50 @@
...
+///
+/// Return `None` if header has no parent (i.e. never because all our blocks have parents in
+/// Amaru).
+pub fn from_candidate<H: IsHeader>(header: &H, candidate: &Nonce) -> Option<Nonce> {
+    Some(Hasher::<256>::hash(
+        &[&candidate[..], &header.parent()?[..]].concat(),
+    ))
+}
+
+/// Evolve the currenet nonce with combining it with the current rolling nonce and the
```
abailly:
```suggestion
/// Evolve the current nonce by combining it with the current rolling nonce and the
```

crates/amaru-consensus/src/consensus/store.rs
```diff
@@ -37,8 +39,152 @@ impl Display for StoreError {
...
+        })?;
+
+        let (parent_epoch, parent) =
+            self.get_nonces(&parent_hash)
+                .ok_or_else(|| NoncesError::UnknownParent {
+                    header: header.hash(),
+                    parent: parent_hash,
+                })?;
+
+        // Compute the new evolving nonce by combining it with the current one and the header's VRF
```
abailly, line 126: Shouldn't this comment be in the definition of `evolve` function?

crates/amaru-consensus/src/consensus/store.rs
```diff
@@ -37,8 +39,152 @@ impl Display for StoreError {
...
+            epoch,
+            Nonces {
+                evolving,
+
+                // On epoch changes, compute the new active nonce by combining:
+                //   1. the (now stable) candidate; and
+                //   2. the previous epoch's last block's parent header hash.
+                //
+                // If the epoch hasn't changed, then our active nonce is unchanged.
+                active: if epoch > parent_epoch {
```
abailly: This inline definition is large and complicated enough to warrant a comment: this seems to call for a dedicated function or variable

crates/amaru-consensus/src/consensus/store.rs
abailly, line 1: I feel I miss tests here, instead of comments

crates/amaru-consensus/src/consensus/store.rs
```diff
@@ -37,8 +39,152 @@ impl Display for StoreError {
...
+                } else {
+                    parent.active
+                },
+
+                // Unless we are within the randomness stability window, we also update the candidate. This
+                // means that outside of the stability window, we always have:
+                //
+                //   evolving == candidate
+                //
+                // They only diverge for the last blocks of each epoch; The candidate remains stable while
```
abailly, line 161: Rather than rather long comment, would it not make sense to have unit or property tests here?

abailly: A pure tested function containing all the possible transitions for the nonce would seem to be more readable and maintainable, as there a somewhat significant of cases we want to ensure covered/

crates/amaru-consensus/src/consensus/header_validation.rs
```diff
@@ -47,42 +43,34 @@ const EVENT_TARGET: &str = "amaru::consensus";
...
 }
 
 impl Consensus {
     pub fn new(
         peer_sessions: Vec<PeerSession>,
         ledger: Box<dyn HasStakeDistribution>,
-        store: Arc<Mutex<dyn ChainStore<ConwayHeader>>>,
-        chain_selector: Arc<Mutex<ChainSelector<ConwayHeader>>>,
-        epoch_to_nonce: HashMap<Epoch, Hash<32>>,
+        store: Arc<Mutex<dyn ChainStore<Header>>>,
```
abailly, line 79: Nice we have this dependency removed 🎉

crates/ouroboros-traits/src/has_stake_distribution/mod.rs
```diff
@@ -1,3 +1,46 @@
...
 pub mod mock;
-mod traits;
-pub use traits::*;
+
+#[derive(Debug)]
+pub struct PoolSummary {
+    /// The blake2b-256 hash digest of the pool's VRF public key.
+    pub vrf: VrfKeyhash,
+    /// Total stake, in Lovelace, delegated to registered pools.
+    pub active_stake: Lovelace,
```
abailly, line 24: Is this not a value that's common to _all_ pools? If so, why put it into each pool structure?

crates/ouroboros-traits/src/has_stake_distribution/mod.rs
```diff
@@ -1,3 +1,46 @@
...
+    fn slot_to_kes_period(&self, slot: u64) -> u64;
+
+    /// Get the maximum number of KES evolutions from the ledger state
+    fn max_kes_evolutions(&self) -> u64;
+
+    /// Get the latest opcert sequence number we've seen for a given issuer_vkey
+    ///
+    /// TODO: This should most probably live within the consensus, and not the ledger, similar to
+    /// the tracking of the epoch nonce.
+    fn latest_opcert_sequence_number(&self, pool: &PoolId) -> Option<u64>;
```
abailly, line 45: Indeed

crates/ouroboros-traits/src/is_header/mod.rs
```diff
@@ -0,0 +1,76 @@
...
+pub mod fake;
+
+pub const HASH_SIZE: usize = 32;
+
+/// Interface to a header for the purpose of chain selection.
+pub trait IsHeader: cbor::Encode<()> + for<'d> cbor::Decode<'d, ()> {
+    /// Hash of the header
+    ///
+    /// This is used to identify the header in the chain selection.
+    /// Header hash is expected to be unique for each header, eg.
```
abailly, line 26: Is this comment really necessary? A cryptographic hash is expected to uniquely identify a piece of data anyway, headers are not special in that respect.

crates/ouroboros/src/praos/nonce.rs
```diff
@@ -0,0 +1,50 @@
...
+            &current[..],
+            &Hasher::<256>::hash(header.extended_leader_output().as_slice())[..],
+        ]
+        .concat(),
+    )
+}
+
+/// Check whether a header is within the stability of the current epoch, and return its epoch for
+/// convenience.
+pub fn randomness_stability_window<H: IsHeader>(header: &H) -> (Epoch, bool) {
```
abailly, line 49: Perhaps not as explicit but more useful from a code perspective?

```suggestion
pub fn randomness_stability_window<H: IsHeader>(header: &H) -> Optional<Epoch> {
```

crates/ouroboros-traits/src/praos/mod.rs
```diff
@@ -0,0 +1,65 @@
...
+/// ─────────────╵┼────────────⛶──────────────╷─ ─ ─ ─ ─ ─ ─┼───── ─ ─ ─ ─ ─ ─ ─
+/// [e-1]          [e]         │              │              [e+1]
+///                            │              │
+///                            │              └─ candidate nonce for [e] is now fixed.
+///                            │
+///                            │
+///                            │          ┌─ 🔎 ────────────────────┐
+///                            └─────────>│   ┏━━━┓  ┏━━━┓  ┏━━━┓   │
+///                                       │..━┫ η ┣━━┫ η ┣━━┫ η ┣━..│
+///                                       │   ┗━╷━┛  ┗━╷━┛  ┗━╷━┛   │
```
jeluard: Just for this this PR is worth it :D

abailly (commented):
LGTM, I think I would like to see some tests for the nonce computation/storage

crates/ouroboros-traits/src/praos/mod.rs
abailly: I was about to say this is cute but an actual drawing would be better 😛

crates/amaru/src/bin/amaru/cmd/daemon.rs
```diff
@@ -102,32 +101,10 @@ pub async fn run_pipeline(pipeline: gasket::daemon::Daemon, exit: CancellationTo
 }
 
 fn parse_args(args: Args) -> Result<Config, Box<dyn std::error::Error>> {
-    // TODO: Figure out from ledger + consensus store
-    let root = args.data_dir.join(args.network.to_string());
-
-    let nonces = read_csv(&root.join("nonces.csv"), nonce::from_csv)?;
```
jeluard, line 108: Shall the `nonces.csv` file be removed then?

crates/ouroboros/src/praos/nonce.rs
KtorZ: Can be both; I technically moved and entirely rewrote the nonce module that was created and licensed starting from 2024. So the 2024 terms still apply ^^"

crates/ouroboros-traits/src/has_stake_distribution/mod.rs
KtorZ: (note: I didn't alter this module in any ways, just moved it up one layer)

It isn't really that the value is in each "pool structure", it is in the return type of `getPool`. It is simply for convenience since this information is readily available (and required) when fetching. 

We could also just separate both having `get_pool` and `get_active_stake`, but that means doing double work behind the scene. One could say "premature optimization", one could also say "wrong function name". But calling is "get_pool_info_for_validation" felt somewhat wrong.

crates/ouroboros-traits/src/is_header/mod.rs
KtorZ: I don't know, you wrote it 🙈  ( https://github.com/pragma-org/amaru/commit/91f7905fe70c226323b933b45fc436acad844170#diff-ae8f32ce58b881f1f7c46be7c48fc53d4e7e79ee308c4d8a451b4437fd2f7f76R5-R10 )

Felt rude to remove it.

crates/ouroboros-traits/src/praos/mod.rs
KtorZ: What do you mean, this is an actual drawing 😄 !

crates/ouroboros/src/praos/nonce.rs
KtorZ: Hmmm. Semantically different actually. You need both the epoch and, the knowledge of whether the header is in the randomness stability window.

crates/amaru-consensus/src/consensus/store.rs
KtorZ: Most likely. The code was inline initially, and I later moved it into a function. The comment somehow stayed there. Will move.

crates/amaru-consensus/src/consensus/header_validation.rs
KtorZ: Well, don't get excited 😅  ... since I renamed `Header` to `IsHeader`, this `Header` actually still refers to a `ConwayHeader`! So, nothing has changed really.

crates/amaru/src/bin/amaru/cmd/daemon.rs
KtorZ: Yes, will do in an upcoming commit.

crates/ouroboros/src/praos/nonce.rs
abailly: I see, I thought you only care about the `Epoch` when stability window is true

crates/amaru-consensus/src/consensus/store.rs
KtorZ: d432c63ae416fb6ed6b8c244e1f98a6e69f4baf6

KtorZ: Spoiler: there were bugs.

PR #131, opened 2025-02-28
chore: use regular riscv target

jeluard:
Use regular `riscv` target, not `risc0` specific.


jeluard:
@KtorZ looks like targeting `riscv32im-unknown-none-elf` is more complex than  anticipated :/

KtorZ:
I thought we were cool on keeping the RISC-0 one ?

jeluard:
I forgot what we decided apparently. SGTM!

PR #134, opened 2025-03-05
chore: simplify DiffBind API by only generalizing to 2 relations.

KtorZ:


crates/amaru-ledger/src/store/columns/accounts.rs
```diff
@@ -20,7 +20,7 @@ pub const EVENT_TARGET: &str = "amaru::ledger::store::accounts";
 /// Iterator used to browse rows from the Accounts column. Meant to be referenced using qualified imports.
 pub type Iter<'a, 'b> = IterBorrow<'a, 'b, Key, Option<Row>>;
 
-pub type Value = ((Option<PoolId>, Option<DRep>), Option<Lovelace>, Lovelace);
+pub type Value = (Option<PoolId>, Option<DRep>, Option<Lovelace>, Lovelace);
```
KtorZ, line 23: That's probably a good time to introduce a proper record here too 😅 ... those tuples are getting our of hands.

jeluard (approved):

PR #136, opened 2025-03-06
chore: no match wildcard

jeluard:
Prevent using wildcard in `match`.

Instead of:

```rust
match value {
  Value::1 => {},
  _ => {}
}
```

do

```rust
match value {
  Value::1 => {},
  Value::2 | Value::3 => {}
}
```

This prevents newly added enum element to be missed in fall in the wildcard handling.

## Summary by CodeRabbit

- **Chores**
  - Updated linting settings to flag broad pattern matching in enumerations.
  - Added attributes to suppress non-critical warnings in various functions, ensuring robust internal quality without impacting the app's functionality.


crates/ouroboros/src/praos/header.rs
```diff
@@ -270,6 +270,7 @@ impl AssertLeaderStakeError {
         let c = (&FixedDecimal::from(1u64) - active_slot_coeff).ln();
         let x = -(leader_relative_stake * &c);
         let ordering = x.exp_cmp(1000, 3, &recip_q);
+        #[allow(clippy::wildcard_enum_match_arm)]
         match ordering.estimation {
             ExpOrdering::LT => Ok(()),
             _ => Err(Self::InsuficientLeaderStake),
```
KtorZ:
```suggestion
        match ordering.estimation {
            ExpOrdering::LT => Ok(()),
            ExpOrdering::GT | ExpOrdering::UNKNOWN => Err(Self::InsuficientLeaderStake),
```

PR #140, opened 2025-03-10
feat: basic amaru consensus simulator

abailly:
This PR introduces `amaru-sim`, an executable that's used to simulate a network and test amaru, essentially the consensus part for now. This is a work-in-progress as the executable is not yet complete: It will crash upon startup and there's a couple of features missing to be able to connect it to https://github.com/pragma-org/simulation-testing/tree/main/moskstraumen but I given it has no impact on the actual behaviour of amaru, I thought I'd rather get feedback earlier rather than later.

## Summary by CodeRabbit

- **Bug Fixes**
	- Corrected a typographical error in internal documentation.
- **New Features**
	- Introduced a simulation application with new executables, offering blockchain simulation capabilities.
	- Added support for fake ledger operations and stake distribution.
	- Launched an enhanced echo service with refined messaging.
	- Added a new stake distribution configuration in the test data.
	- Introduced a README for the Amaru Simulator, detailing testing strategies and protocol specifications.
- **Refactor**
	- Improved type safety and debugging support.
	- Updated dependency configurations and project settings for better performance and maintainability.
- **Chores**
	- Added copyright notices and licensing information to relevant files.
	- Removed unnecessary dependencies to streamline the project.


simulation/amaru-sim/src/bin/amaru-sim/simulator.rs
```diff
@@ -0,0 +1,189 @@
...
+use tracing::trace;
+
+#[derive(Debug, Parser)]
+#[clap(name = "Amaru Simulator")]
+#[clap(bin_name = "amaru-sim")]
+#[clap(author, version, about, long_about = None)]
+pub struct Args {
+    /// Upstream peer addresses to synchronize from.
+    ///
+    /// This option can be specified multiple times to connect to multiple peers.
```
stevana, line 46: How is this supposed to work in the tests? Clients that make the requests don't really have an address?

Also how come downstream peers are not specified? Will other nodes using a node as upstream peer make it forward messages to them?

abailly: I need to update the comment to say these `--peer-address` are actually just names.
Downstream peers are not specified because they would _connect to us_ so in the real system you would specify a `--listen` address

stevana: So this flag is only there to maintain compatibility with the real binary, or does it actually do something? If it's the latter, then what is it supposed to be set to in the tests?

simulation/amaru-sim/src/bin/amaru-sim/bytes.rs
```diff
@@ -0,0 +1,53 @@
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
```
jeluard, line 15: Missing license header?

simulation/amaru-sim/src/bin/amaru-sim/sync.rs
```diff
@@ -0,0 +1,198 @@
+use amaru_consensus::{consensus::wiring::PullEvent, peer::Peer};
```
jeluard, line 15: Missing license header?

simulation/amaru-sim/src/bin/amaru-sim/bytes.rs
```diff
@@ -0,0 +1,53 @@
...
+
+#[derive(Debug, PartialEq)]
+pub struct Bytes {
+    pub bytes: Vec<u8>,
```
jeluard, line 19: Naive question: wasn't a simple type alias doing most of the job?

simulation/amaru-sim/src/bin/amaru-sim/simulator.rs
```diff
@@ -0,0 +1,190 @@
...
+    let sync = bootstrap(args);
+
+    let exit = amaru::exit::hook_exit_token();
+
+    run_pipeline(gasket::daemon::Daemon::new(sync), exit.clone()).await;
+}
+
+fn define_gasket_policy() -> gasket::runtime::Policy {
+    let retries = gasket::retries::Policy {
+        max_retries: 20,
```
jeluard, line 74: <is this policy with specifics retries actually meaningful for the simulation?

simulation/amaru-sim/src/bin/amaru-sim/simulator.rs
abailly: The names are actually important, they are used internally by `chain-selector` component to track upstream peers' state. As the node is responsible to connect to those peers and until we have p2p and infer those from the ledger, I think passing them here makes sense.

simulation/amaru-sim/src/bin/amaru-sim/sync.rs
abailly: These were created before I configured Emacs to inject automatically the header upon file creation

simulation/amaru-sim/src/bin/amaru-sim/bytes.rs
abailly: I am probably too much infected by Haskell and `newtype`s :) If you think a `type` is better I am totally fine changing that.

simulation/amaru-sim/src/bin/amaru-sim/simulator.rs
abailly: Probably not when we will actually _simulate_ the runtime but right now we are mostly testing externally. I would like to have a good plan for simulating gasket and async runtime, happy to discuss this together with @stevana

PR #144, opened 2025-03-13
chore: improve tracing handling

jeluard:
Clarifies `tracing` usage by advocating to leverage smaller functions and `instrument`.
Remove `parent` from spans as they should be inherited, this removes the need to pass too much spans around, cleaning the API.
The one thing to enable this is to make sure `parent` spans are kept when crossing threads. `gasket` handles this internally so without modifying `gasket` this has to be done manually on `amaru` side when crossing stages.

* instrument driven (no `_span` usage), always use `Level::TRACE` and `skip_all`
* no more `EVENT_TARGET`
* one `span` per transaction
* event can be attached to the current span

TLDR for devs:

* only use `instrument`
* new `fields` can be added dynamically to the current span (created by the closest `instrument`) but they have to be listed in `instrument`
* `events` can be added via `trace!` (and family) macros
* do not abuse `instrument` as it clogs the system and make it harder to understand the flow

A typical block processing now looks like the following:

Capture d’écran 2025-03-18 à 10 44 54

## Summary by CodeRabbit

- **New Features**
	- Introduced updated block processing and state management to deliver improved system stability and observability.
	- Added new methods for managing rewards and state within the ledger.
	- Enhanced logging capabilities in the `roll_back` method for better traceability.
	- Added a new method for ticking pools in the store, improving operational efficiency.
	- Simplified the `PullEvent::Rollback` variant for clearer event handling.

- **Refactor**
	- Streamlined internal interfaces by removing redundant parameters and consolidating processing logic, resulting in smoother operations.
	- Improved method signatures to enhance usability without consuming instances.
	- Adjusted tracing instrumentation to improve logging clarity by omitting parameters.
	- Refactored methods in the `Stage` and `Worker` structs for better span management.

- **Chores**
	- Added refined methods for reward adjustments, state updates, and metric resets, enhancing overall system performance and monitoring.
	- Included a new dependency for enhanced tracing capabilities in the consensus package.
	- Removed an unused dependency from the shared examples.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -57,7 +57,7 @@ dependencies = [
  "sysinfo",
  "tempfile",
  "test-case",
- "thiserror 2.0.11",
+ "thiserror 2.0.12",
  "tokio",
  "tokio-util",
  "tracing",
@@ -87,9 +87,10 @@ dependencies = [
  "rayon",
  "rocksdb",
  "tempfile",
- "thiserror 2.0.11",
+ "thiserror 2.0.12",
  "tokio",
  "tracing",
+ "tracing-opentelemetry",
 ]
 
 [[package]]
@@ -119,7 +120,7 @@ dependencies = [
  "num",
  "proptest",
  "serde",
- "thiserror 2.0.11",
+ "thiserror 2.0.12",
  "tracing",
 ]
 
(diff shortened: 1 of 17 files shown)


abailly (commented):
Looks much better!

abailly (approved):

KtorZ (approved):
Nice. 

I've done another pass to remove some redundant fields from spans, and to split in even smaller pieces the surroundings of the volatile state management.

PR #146, opened 2025-03-14
chore: show that to_cbor errors are actually Infallible when buffer is Vec.

KtorZ:
Related to: https://github.com/pragma-org/amaru/pull/129#discussion_r1995135296


crates/amaru-kernel/src/lib.rs
```diff
@@ -191,11 +190,11 @@ pub type Nonce = Hash<32>;
 // CBOR conversions
 // ----------------------------------------------------------------------------
 
-#[allow(clippy::panic)]
+#[allow(clippy::unwrap_used)]
 pub fn to_cbor<T: cbor::Encode<()>>(value: &T) -> Vec<u8> {
     let mut buffer = Vec::new();
-    cbor::encode(value, &mut buffer)
-        .unwrap_or_else(|e| panic!("unable to encode value to CBOR: {e:?}"));
+    let result: Result<(), cbor::encode::Error<Infallible>> = cbor::encode(value, &mut buffer);
```
jeluard, line 196: Doesn't that depend on implementations of `cbor::Encode` ? Or does that imply we guarantee amaru implementations (and pallas?) do not return `Err` ?

KtorZ: It doesn't. It depends on the type of buffer we use. For an allocated slice, the encoder can hit an "end of slice" error when trying to write into it; but a vector will always auto-expand and any encoder on a vec is therefore Infallible -- as guaranteed by the types here.

jeluard (approved):

PR #147, opened 2025-03-14
Conformance tests scripts

KtorZ:
Here are the scripts and dataset I've been using for generating the `.snap` conformance tests for the ledger. Doing this, I realized that the `epoch-state` and `rewards-provenance` were obtained from a fork of the node, and aren't easily accessible through the _standard tooling_. 

I do think however it is possible to make them available; modulo some updates on Ogmios which I'll try out in the upcoming days. Still, it's worth documenting and versioning the overarching scripts for fetching and generating those datasets. 

## Summary by CodeRabbit

- **New Features**
  - Introduced comprehensive data enhancements for conformance testing, including structured delegation and stake pool information.
  - Added configuration changes to support accurate snapshot generation.
  
- **Documentation**
  - Published a detailed guide outlining prerequisites, installation steps, and usage instructions for conformance tests.

- **Chores**
  - Updated version control settings to properly flag generated files and ignore unnecessary dependencies.


KtorZ:
@abailly indeed, just as a reference: I am uploading these since they are the ones I've been using so far (ie it's nothing inherently new). It's not necessarily meant to be "the reference" but simply a way to get to those snapshots. 

I assume we could rewrite them to be produced directly from the db-server. I simply didn't have the time/energy to dive in.

abailly:
I need to find the time then 👍

PR #145, opened 2025-03-13
Introduce thiserror and initial transaction validation logic

yHSJ:
This PR:
 - introduces `thiserror` for block validation error handling
 - introduces some initial transaction validation logic

## Summary by CodeRabbit

- **New Features**
  - Expanded public APIs with enhanced validations for transactions and blocks, including checks on metadata, output sizes, and reference inputs.
  - Introduced a new trait for streamlined access to Lovelace values across various data structures.
  - Added a new module for transaction handling, including functions for validating transaction metadata and outputs.

- **Bug Fixes**
  - Improved error handling and reporting across validation modules to ensure clearer feedback and more robust compliance verification.

- **Refactor**
  - Simplified access to Lovelace values through a trait-based approach, enhancing object-oriented design.

PR #148, opened 2025-03-14
BlockValidationContext and Signature Validation

yHSJ:
This PR is a work in progress but will introduce a new data structure `BlockValidationContext` that contains all the ledger state slices (subsets of ledger state) necessary to validate a block.

The flow during block validation will (eventually) be:
- Receive block
- Fetch required data for block validation and construct `BlockValidationContext`
- Validate block (returns a mutated `BlockValidationContext`)
- Construct diffs
- "Apply" block

This PR should only be merged after: https://github.com/pragma-org/amaru/pull/145

## Summary by CodeRabbit

- **New Features**
  - Enhanced transaction processing with improved address and key validations.
  - Upgraded block and transaction validation, providing more detailed error feedback and refined witness checks.
  - Introduced new modules for bootstrap and key witness validations.
  - Added a new context management structure to improve validation state handling.

- **Refactor**
  - Streamlined block validation workflows and context management, improving reliability and error reporting.

- **Chores**
  - Added a new cryptographic dependency to support enhanced functionality.
  - Reformatted dependency specifications for clarity in configuration files.


crates/amaru-ledger/src/rules/transaction/signature.rs
```diff
@@ -0,0 +1,86 @@
...
+    for witness in vkey_witnesses {
+        // TODO: rust-ier way to handle this? I really don't to clone every key and signature for every single witness in every tx...
+        let vkey_bytes: [u8; 32] =
+            witness
+                .vkey
+                .deref()
+                .clone()
+                .try_into()
+                .unwrap_or_else(|v: Vec<_>| {
+                    panic!("Invalid vkey length {} (expected 32 bytes)", v.len())
```
jeluard: Make sure you don't introduce new panics. You probably can use the new errors you introduced, decide what to do when an error is found, then propagate.

yHSJ: Yeah, much of this is a WIP, I've just put some panics in places where I haven't yet decided how to best handle the error

crates/amaru-kernel/src/lib.rs
```diff
@@ -502,6 +538,56 @@
...
+            }
+        }
+        Address::Byron(_byron_address) => {
+            // TODO: handle byron addresses
+            panic!("Uh oh, byron address discovered")
+        }
+
+        Address::Stake(_) => None,
+    }
+}
```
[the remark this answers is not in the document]
yHSJ: Good catch rabbit, forgot I left this as a TODO. Time to finally learn more about Byron addresses.

crates/amaru-kernel/src/lib.rs
```diff
@@ -533,6 +538,54 @@ pub fn to_ex_units(witness_set: WitnessSet) -> ExUnits {
...
+            PseudoTransactionOutput::Legacy(transaction_output) => {
+                Address::from_bytes(&transaction_output.address)
+            }
+            PseudoTransactionOutput::PostAlonzo(modern) => Address::from_bytes(&modern.address),
+        }
+    }
+}
+
+pub trait HasKeyHash {
+    fn get_key_hash(&self) -> Option<Hash<28>>;
```
SupernaviX: nit: maybe this should be named `key_hash`? since that matches the naming conventions of pallas, and of the other helper trait in here

crates/amaru-ledger/Cargo.toml
```diff
@@ -22,6 +22,7 @@ thiserror.workspace = true
 iter-borrow.workspace = true
 amaru-kernel.workspace = true
 amaru-ouroboros-traits.workspace = true
+sha3 = "0.10.8"
```
SupernaviX: nit: this version should probably be managed at the workspace level like other deps

crates/amaru-ledger/src/rules.rs
```diff
@@ -22,6 +26,9 @@ pub enum BlockValidationError {
     RuleViolations(Vec<RuleViolation>),
     #[error("Cascading rule violations: root: {0:?}, resulting error(s): {1:?}")]
     Composite(RuleViolation, Box<BlockValidationError>),
+    // TODO: This error shouldn't exist, it's a placeholder for better error handling in less straight forward cases
+    #[error("Unnammed error: {0}")]
+    Unnamed(String),
```
SupernaviX:
```suggestion
    #[error("Unnamed error: {0}")]
    Unnamed(String),
```
this is 100% nit, but I'd call it something like `UncategorizedError`, just so it was clear that it wasn't a validation error like "this block is missing a name"

crates/amaru-ledger/src/rules.rs
```diff
@@ -100,11 +123,12 @@ pub fn validate_block(
...
-        let witness_set = match witness_sets.get(i as usize) {
-            Some(witness_set) => witness_set,
-            None => continue,
-        };
+        let witness_set = witness_sets
+            .get(i as usize)
+            .ok_or(BlockValidationError::Unnamed(format!(
+                "Missing witness set for transaction index {}",
+                i
+            )))?;
```
SupernaviX: What's the rationale for making this an "unnamed" error instead of a validation error? It looks like it could happen if someone submitted a block which had different numbers of bodies and witness sets.

The cddl says "the length of transaction_bodies and transaction_witness_sets must be the same"; if you validate that the lengths match before this for loop, you could avoid this error by including witness_sets in the iterator:

```rust
    for (i, transaction, witness_set) in izip!(0u32.., transactions, witness_sets) {
```

crates/amaru-ledger/src/rules/context.rs
```diff
@@ -0,0 +1,62 @@
...
+            });
+
+            for (output, index) in transaction_body.outputs.iter().zip(0u64..) {
+                self.utxo_slice.insert(
+                    TransactionInput {
+                        transaction_id: tx_hash,
+                        index,
+                    },
+                    // TODO: Can we not clone here?
+                    output.clone().into(),
```
SupernaviX: doesn't look avoidable

crates/amaru-ledger/src/rules/transaction/vkey_witness.rs
```diff
@@ -0,0 +1,204 @@
...
+        return Err(TransactionRuleViolation::MissingRequiredVkeyWitnesses { missing_key_hashes });
+    }
+
+    let invalid_witnesses = vkey_witnesses
+        .iter()
+        .filter(|witness| {
+            match validate_witness(witness, transaction_body.original_hash().as_slice()) {
+                Ok(is_valid) => !is_valid,
+                Err(e) => {
+                    eprintln!("Failed to validate witness: {:?}", e);
```
SupernaviX: Same as in `bootstrap_witness`,  I don't think this log line is helpful

crates/amaru/src/pipeline.rs
```diff
@@ -75,9 +75,12 @@ impl<S: Store> Stage<S> {
...
 
-        let (block_header_hash, block) =
-            rules::validate_block(&raw_block[..], ProtocolParameters::default())
-                .unwrap_or_else(|e| panic!("Failed to validate block: {:?}", e));
+        let (block_header_hash, block) = validate_block(
+            &raw_block[..],
+            ProtocolParameters::default(),
+            &mut BlockValidationContext::default(),
+        )
+        .unwrap_or_else(|e| panic!("Failed to validate block: {:?}", e));
```
SupernaviX, line 83: I know this was here before, but is panicking the right behavior here? `validate_block` returns an error when validation fails, so it'll happen pretty often.

crates/amaru-ledger/src/rules/transaction/bootstrap_witness.rs
```diff
@@ -0,0 +1,113 @@
...
+        );
+    }
+
+    let invalid_witnesses = bootstrap_witnesses
+        .iter()
+        .filter(|witness| {
+            match validate_witness(witness, transaction_body.original_hash().as_slice()) {
+                Ok(is_valid) => !is_valid,
+                Err(e) => {
+                    eprintln!("Failed to validate witness: {:?}", e);
```
SupernaviX: This log line won't be very helpful in the wild.
 - Using `eprintln!` instead of one of the tracing macros like `warn!` means it won't get picked up by a log aggregator
 - The error is a `TryFromSliceError`, so it'll only log `"could not convert slice to array"` (and not mention whether it was the vkey or the signature that was an invalid length)
 - This message isn't getting included in the validation result anyway.

crates/amaru-ledger/src/rules.rs
```diff
@@ -52,6 +59,21 @@ pub enum TransactionRuleViolation {
...
+    MissingRequiredBootstrapWitnesses {
+        missing_bootstrap_roots: Vec<Hash<28>>,
+    },
+    #[error("Invalid bootstrap witnesses: {invalid_witnesses:?}")]
+    InvalidBootstrapWitnesses {
+        invalid_witnesses: Vec<BootstrapWitness>,
+    },
+    // TODO: This error shouldn't exist, it's a placeholder for better error handling in less straight forward cases
+    #[error("Unnamed error: {0}")]
+    Unnanmed(String),
```
SupernaviX:
```suggestion
    #[error("Unnamed error: {0}")]
    Unnamed(String),
```
this is 100% nit, but I'd call it something like `UncategorizedError`, just so it was clear that it wasn't a validation error like "this block is missing a name"

crates/amaru-ledger/src/rules.rs
yHSJ: This is one of those things that's a bit unclear how to properly handle. I would generally say that anything defined in the CDDL should be enforced at deserialization, but the Pallas type itself doesn't have this enforced right now. So, in theory, it's not an error that should ever be seen at this stage in the process of validation, but at the moment it is.

crates/amaru-ledger/src/rules/transaction/vkey_witness.rs
```diff
@@ -0,0 +1,204 @@
...
+    Certificate, HasAddress, HasKeyHash, Hash, Hasher, KeepRaw, MintedTransactionBody, NonEmptySet,
+    OriginalHash, PublicKey, Signature, VKeyWitness, Voter,
+};
+
+pub fn validate_vkey_wintesses(
+    transaction_body: &KeepRaw<'_, MintedTransactionBody<'_>>,
+    vkey_witnesses: &Option<NonEmptySet<VKeyWitness>>,
+    utxo_slice: &UtxoSlice,
+) -> Result<(), TransactionRuleViolation> {
+    let empty_vec = vec![];
```
jeluard, line 14: All those `empty_vec` look weird. Isn't `unwrap_or_default` useful here?

crates/amaru-ledger/Cargo.toml
jeluard: We want to assess this lib for security and all. Can't we find the necessary bits in one of the lib we already use?

crates/amaru-ledger/src/rules.rs
yHSJ: I like this a lot more

crates/amaru-ledger/src/rules/transaction/vkey_witness.rs
yHSJ: I just wanted to avoid swallowing the error and providing no context as to why it failed.

crates/amaru-ledger/src/rules.rs
```diff
@@ -164,7 +202,14 @@ mod tests {
...
 
-        assert!(validate_block(bytes.as_slice(), ProtocolParameters::default()).is_ok())
+        let results = validate_block(
+            bytes.as_slice(),
+            ProtocolParameters::default(),
+            &mut BlockValidationContext::default(),
+        );
+
+        println!("{:?}", results);
```
jeluard:
```suggestion
```

crates/amaru/src/pipeline.rs
yHSJ: Yeah, handling this error *is* definitely something we need to do, but I think it's out of scope for this error

crates/amaru-ledger/src/rules/transaction/vkey_witness.rs
```diff
@@ -0,0 +1,204 @@
...
+
+    let withdrawal_pkhs = transaction_body
+        .withdrawals
+        .as_ref()
+        .map(|withdrawals| {
+            withdrawals
+                .iter()
+                .filter_map::<Hash<28>, _>(|(reward_account, _)| {
+                    if reward_account[0] & 0b00010000 == 0 {
+                        Some(Hash::from(&reward_account[1..29]))
```
jeluard, line 57: Magic numbers?

crates/amaru/src/pipeline.rs
jeluard: Right let's try to retrofit the error here.

crates/amaru-ledger/src/rules/transaction/bootstrap_witness.rs
yHSJ: These are all very good points. I think while I want to avoid swallowing errors, this is not the best way to handle it.

crates/amaru-ledger/Cargo.toml
yHSJ: No, I could not find anything related to sha3 in any of our current dependencies. I'll double check though

crates/amaru-ledger/src/rules/transaction/vkey_witness.rs
jeluard: You might want to introduce `tracing` then, via `instrument`.

crates/amaru-ledger/src/rules/transaction/vkey_witness.rs
yHSJ: Yeah, I really don't like it. I can't use `unwrap_or_default` though because there is no default implemented for `&Vec<T>`

crates/amaru-ledger/src/rules.rs
```diff
@@ -52,6 +59,21 @@ pub enum TransactionRuleViolation {
     },
     #[error("Invalid transaction metadata: {0}")]
     InvalidTransactionMetadata(#[from] InvalidTransactionMetadata),
+    #[error("Missing required signatures: pkhs {missing_key_hashes:?}")]
+    MissingRequiredVkeyWitnesses { missing_key_hashes: Vec<Hash<28>> },
+    #[error("Invalid vkey witnesses: {invalid_witnesses:?}")]
+    InvalidVkeyWitnesses { invalid_witnesses: Vec<VKeyWitness> },
+    #[error("Missing required signatures: bootstrap roots {missing_bootstrap_roots:?}")]
```
KtorZ, line 66: You probably don't want to debug-display these, but perhaps have them displayed hex-encoded? Fortunately, Pallas' Hash already implement `Display` and turns them into hex-encoded strings so that simplifies the work.

crates/amaru-ledger/src/rules.rs
```diff
@@ -52,6 +59,21 @@ pub enum TransactionRuleViolation {
     },
     #[error("Invalid transaction metadata: {0}")]
     InvalidTransactionMetadata(#[from] InvalidTransactionMetadata),
+    #[error("Missing required signatures: pkhs {missing_key_hashes:?}")]
+    MissingRequiredVkeyWitnesses { missing_key_hashes: Vec<Hash<28>> },
+    #[error("Invalid vkey witnesses: {invalid_witnesses:?}")]
```
KtorZ, line 64: Same here; generally speaking, we very likely want to avoid `Debug` in error's display. It would be better to yield structured errors in the long run. But for now, at least a non-rust specific display is best.

crates/amaru-ledger/src/rules.rs
```diff
@@ -133,18 +159,31 @@ pub fn validate_block(
     Ok((block_header_hash, block))
 }
 
-pub fn validate_transaction(
-    transaction_body: &MintedTransactionBody<'_>,
-    _witness_set: &MintedWitnessSet<'_>,
-    auxiliary_data: Option<&AuxiliaryData>,
-    _is_valid: bool,
+pub fn validate_transaction<'a>(
+    transaction_body: &KeepRaw<'_, MintedTransactionBody<'_>>,
```
KtorZ, line 163: Seems like there's an extra KeepRaw here?

crates/amaru/src/pipeline.rs
KtorZ: > I know this was here before, but is panicking the right behavior here?

At this point it is since we do not expect any invalid block. Eventually, as we move towards being a proper relay, it won't.

crates/amaru-ledger/src/rules/transaction/vkey_witness.rs
```diff
@@ -0,0 +1,196 @@
...
+        .unwrap_or_default();
+
+    let required_vkey_hashes = [
+        spend_pkhs.as_slice(),
+        required_signers.as_slice(),
+        withdrawal_pkhs.as_slice(),
+        vote_pkhs.as_slice(),
+        certificate_pkhs.as_slice(),
+    ]
+    .concat();
```
KtorZ, line 151: You'd be better off:

1. Creating a mutable `BTreeSet` at the beginning
2. Add elements to that set as they're found. 

This would also enable more similar patterns as we move forward (e.g. if we also need to collect script credentials for another validation) and would be much more efficient for both constructing it, and querying it afterwards (since a lookup is then performed in logarithmic time).

crates/amaru-ledger/src/rules/transaction/vkey_witness.rs
```diff
@@ -0,0 +1,196 @@
...
+
+    let invalid_witnesses = vkey_witnesses
+        .iter()
+        .filter(|witness| {
+            match validate_witness(witness, transaction_body.original_hash().as_slice()) {
+                Ok(is_valid) => !is_valid,
+                Err(_) => true,
+            }
+        })
+        .cloned()
```
KtorZ, line 178: We probably don't need to clone and report the entire witness here; rather just the indice and/or the corresponding verification key. So I'd suggest using `filter_map` here instead.

crates/amaru-ledger/src/rules/transaction/vkey_witness.rs
```diff
@@ -0,0 +1,196 @@
...
+        withdrawal_pkhs.as_slice(),
+        vote_pkhs.as_slice(),
+        certificate_pkhs.as_slice(),
+    ]
+    .concat();
+
+    let empty_vec = vec![];
+    let vkey_witnesses = vkey_witnesses.as_deref().unwrap_or(&empty_vec);
+    let vkey_hashes = vkey_witnesses
+        .iter()
```
KtorZ, line 156: We're already traversing the witnesses down below for checking signature. So it would be better to check for required vkey hashes there.

crates/amaru-ledger/src/rules/transaction/vkey_witness.rs
```diff
@@ -0,0 +1,196 @@
...
+        .map(|witness| Hasher::<224>::hash(&witness.vkey))
+        .collect::<Vec<_>>();
+
+    // Are we worried about efficiency here? this is quadratic time
+    let missing_key_hashes: Vec<_> = required_vkey_hashes
+        .into_iter()
+        .filter(|hash| !vkey_hashes.contains(hash))
+        .collect();
```
KtorZ, line 164: See: https://doc.rust-lang.org/std/collections/struct.BTreeSet.html#method.difference

crates/amaru-ledger/src/rules/transaction/vkey_witness.rs
```diff
@@ -0,0 +1,196 @@
...
+                        relays: _,
+                        pool_metadata: _,
+                    } => Some(*operator),
+                    Certificate::PoolRetirement(hash, _) => Some(*hash),
+                    Certificate::Reg(stake_credential, coin) => {
+                        if coin == &0 {
+                            None
+                        } else {
+                            stake_credential.key_hash()
+                        }
```
KtorZ, line 114: That looks suspicious?

crates/amaru-ledger/src/rules/transaction/bootstrap_witness.rs
```diff
@@ -0,0 +1,110 @@
...
+            let prefix: &[u8] = &[131, 0, 130, 0, 88, 64];
+
+            let mut sha_hasher = Sha3_256::new();
+            sha_hasher.update(prefix);
+            sha_hasher.update(bootstrap_witness.public_key.deref());
+            sha_hasher.update(bootstrap_witness.chain_code.deref());
+            sha_hasher.update(bootstrap_witness.attributes.deref());
+
+            let sha_digest = sha_hasher.finalize();
+            Hasher::<224>::hash(&sha_digest)
```
KtorZ, line 67: This deserves to live in its own function and be tested independently.

crates/amaru-ledger/src/rules/context.rs
```diff
@@ -0,0 +1,61 @@
...
+    pub utxo_slice: UtxoSlice,
+    // TODO: add more slices as needed
+}
+
+impl BlockValidationContext {
+    pub fn update(
+        &mut self,
+        transaction_body: &KeepRaw<'_, MintedTransactionBody<'_>>,
+        transaction_is_valid: bool,
+    ) {
```
KtorZ, line 20: I am not sure about this API; I think we're going to need something more _granular_ as you need to update the context while validating the transaction (e.g. add an available credential after processing a 'Reg' certificate).

Also, for this to work, the implementation should likely come from elsewhere and not be implemented by the context itself. The context defines the API, but the caller is going to implement the context API.

crates/amaru-ledger/src/rules.rs
yHSJ: Yes, that is intentional. I need to work with the CBOR of the transaction body as a whole in some validations (for hashing).

crates/amaru-ledger/src/rules/transaction/vkey_witness.rs
yHSJ: It does, it was based off of the comment found [here](https://github.com/IntersectMBO/cardano-ledger/blob/81637a1c2250225fef47399dd56f80d87384df32/eras/conway/impl/src/Cardano/Ledger/Conway/TxCert.hs#L698). We don't have an `Option` for coin here though, so not sure how that case would be handled

KtorZ:
Re-opened https://github.com/pragma-org/amaru/pull/155 because we cannot push on your branch being on `SundaeSwap-finance`, which makes it quite annoying to work from. 

I rebase the PR on latest `main`.

crates/amaru-ledger/src/rules.rs
KtorZ: Which ones 🤔 ?

crates/amaru-ledger/src/rules/transaction/vkey_witness.rs
KtorZ: It's worth porting that comment :)

crates/amaru-ledger/src/rules.rs
yHSJ: Signature validation is the one I needed it for, so far. User signs the hash of the tx body.

KtorZ: That's just the transaction id, we already compute this one as part of the state management. It's fine, I am working on unifying the two parts of the codebase here because the duplication is getting out of hands 😅

crates/amaru-ledger/src/rules/transaction/vkey_witness.rs
yHSJ: This is much nicer, thank you for the suggestion!

PR #149, opened 2025-03-17
actor-based structure of downstream server

rkuhn:
Posting this mainly for discussion of the approach.

## Summary by CodeRabbit

- **New Features**
  - Added support for configurable network listening address, maximum downstream peers, and network magic number for incoming connections.
  - Enhanced block forwarding stage to accept and manage network peers, enabling client protocol management and event forwarding.
  - Introduced client protocol handling for chain synchronization, including rollback and forward operations.
  - Added a command-line option to specify the listening address for the daemon.

- **Documentation**
  - Added detailed documentation and diagrams explaining the architecture and concurrency model of the chain forwarding stage.
  - Expanded README with instructions for running a Cardano node using Docker.

- **Chores**
  - Updated dependencies and development dependencies to support new features and protocol handling.
  - Improved .gitignore rules for Cardano node support files and snapshots.

- **Tests**
  - Introduced comprehensive tests for chain synchronization logic, header traversal, and protocol integration.

- **Refactor**
  - Refactored Makefile for flexible configuration of ledger and chain directories, and improved database handling commands.


crates/amaru-consensus/src/chain_forward.rs
```diff
@@ -85,9 +164,18 @@ impl gasket::framework::Worker<ForwardStage> for Worker {
                     hash = %Hash::<32>::from(point),
                 );
 
+                // FIXME: block height should be part of BlockValidated message
+                let store = stage.store.lock().await;
+                if let Some(header) = store.load_header(&Hash::from(point)) {
+                    self.tip = Tip(to_pallas_point(point), header.block_height());
+
+                    self.clients
+                        .send(ClientMsg::Op(ClientOp::Forward(header.clone())));
```
rkuhn: unnecessary clone

crates/amaru-consensus/src/chain_forward/client_protocol.rs
```diff
@@ -0,0 +1,245 @@
...
+            }
+        } else {
+            // parent terminated
+            return Ok(());
+        }
+    }
+}
+
+fn to_header_content(_header: Header) -> HeaderContent {
+    todo!()
```
rkuhn: here I’ll need help, no clue how to construct that header content

crates/amaru-consensus/src/chain_forward.rs
```diff
@@ -107,15 +195,75 @@ impl gasket::framework::Worker<ForwardStage> for Worker {
...
-            BlockValidationResult::RolledBackTo(point) => {
+            Unit::Block(BlockValidationResult::RolledBackTo(point)) => {
                 info!(
                     target: EVENT_TARGET,
                     slot = point.slot_or_default(),
                     hash = %Hash::<32>::from(point),
                     "rolled_back_to"
                 );
+
+                // FIXME: block height should be part of BlockValidated message
```
abailly: Good point

crates/amaru-consensus/src/chain_forward.rs
```diff
@@ -107,15 +195,75 @@ impl gasket::framework::Worker<ForwardStage> for Worker {
...
+                    self.tip = Tip(to_pallas_point(point), header.block_height());
+
+                    self.clients
+                        .send(ClientMsg::Op(ClientOp::Backward(to_pallas_point(point))));
+                }
+
+                Ok(())
+            }
+            Unit::Peer(peer) => {
+                // FIXME: pallas design bug that we only get &Unit and thus cannot take values from it without internal mutability
```
abailly: Do you mean `gasket` instead of `pallas`?

crates/amaru-consensus/src/chain_forward/client_protocol.rs
```diff
@@ -0,0 +1,245 @@
...
+        // TODO: Implement block fetch
+        server.send_no_blocks().await?;
+    }
+
+    Ok(())
+}
+
+enum TxSubmissionMsg {}
+
+async fn tx_submission(
```
abailly: We don't need `TxSubmission` at this stage so probably better to not clutter the code with it

crates/amaru-consensus/src/chain_forward/client_state.rs
```diff
@@ -0,0 +1,119 @@
...
+                current_header = header;
+            }
+            None => return None, // Broken chain
+        }
+    }
+
+    None // Reached genesis without finding any matching point
+}
+
+pub(super) fn hash_point(point: &Point) -> Hash<32> {
```
abailly: there's a `impl From<&Point> for Hash<32>` in `amaru_kernel` IIRC

simulation/amaru-sim/src/bin/amaru-sim/simulator.rs
```diff
@@ -116,7 +116,7 @@ pub fn bootstrap(args: Args) -> Vec<Tether> {
 
     let mut consensus_stage = HeaderStage::new(consensus);
 
-    let mut block_forward = chain_forward::ForwardStage::new(chain_ref.clone());
+    let mut block_forward = chain_forward::ForwardStage::new(chain_ref.clone(), 1);
```
abailly: You can ignore that, simulator code has changed drastically

crates/amaru-consensus/src/chain_forward/client_state.rs
```diff
@@ -0,0 +1,119 @@
...
+/// Find headers between points in the chain store.
+/// Returns None if none of the points in `points` lies in the past of `start_point`.
+/// Otherwise returns Some(headers) where headers is a list of headers leading from
+/// the first found point in the past of `start_point` matching a point from `points`
+/// up to `start_point`.
+pub(super) async fn find_headers_between(
+    store: &Arc<Mutex<dyn ChainStore<Header>>>,
+    start_point: &Point,
+    points: &[Point],
+) -> Option<(Vec<ClientOp>, Tip)> {
```
abailly: I would extract the intersection logic from this module and put it closer to `chain_selection` but that can happen later

crates/amaru-consensus/src/chain_forward/client_protocol.rs
abailly: `HeaderContent` is basically just the CBOR serialiazation of the header
```
pub struct HeaderContent {
    pub variant: u8,
    pub byron_prefix: Option<(u8, u64)>,
    pub cbor: Vec<u8>,
}
```

so something like 
```
let header_content = HeaderContent { 
    variant: 1,
    byron_prefix: None,
   cbor: to_cbor(&header)
} 
```

should work

abailly (commented):
Overall it makes sense to me but there's a lot of ground to tread and I think what would help me understand better the design is a diagram. Adding some design notes  in the `README` of the crate as part of this PR would be great. 
Also, I'd assume you will add tests after having gathered consensus (ah! ah!) on the design?

crates/amaru-consensus/src/chain_forward/client_state.rs
```diff
@@ -0,0 +1,119 @@
...
+        match store.load_header(&parent_hash) {
+            Some(header) => {
+                if points.iter().any(|p| hash_point(p) == parent_hash) {
+                    // Found a matching point, return the collected headers
+                    return Some((
+                        headers,
+                        Tip(to_pallas_point(&header.point()), header.block_height()),
+                    ));
+                }
+                headers.push(ClientOp::Forward(header.clone()));
```
abailly: This is where we'll need to revise the design later to only use indices/pointers to track each followers chain, because otherwise the intersection process could accumulate the whole chain as `Op`s to send to the follower.

crates/amaru-consensus/src/chain_forward/client_state.rs
```diff
@@ -0,0 +1,119 @@
...
+                current_header = header;
+            }
+            None => return None, // Broken chain
+        }
+    }
+
+    None // Reached genesis without finding any matching point
```
abailly: `Genesis` is a valid intersection point though, if no other matches.

crates/amaru-consensus/src/chain_forward/client_state.rs
```diff
@@ -0,0 +1,119 @@
...
+    match point {
+        Point::Origin => Hash::from([0; 32]),
+        Point::Specific(_slot, hash) => Hash::from(hash.as_slice()),
+    }
+}
+
+pub(super) fn to_pallas_point(point: &amaru_kernel::Point) -> Point {
+    match point {
+        amaru_kernel::Point::Origin => Point::Origin,
+        amaru_kernel::Point::Specific(_slot, hash) => Point::Specific(0, hash.clone()),
```
abailly: shouldn't you use `_slot` instead of hardcoded `0`?

abailly (commented):
Some more comments

crates/amaru-consensus/src/chain_forward.rs
rkuhn: yes, indeed

crates/amaru-consensus/src/chain_forward/client_state.rs
rkuhn: that’s a different `Point`, though

crates/amaru-consensus/src/chain_forward/client_state.rs
rkuhn: d'oh — that’s a new kind of bug caused by pressing `tab` to get the first line and then not adequately checking the second line that also materialized

crates/amaru-consensus/src/chain_forward/client_protocol.rs
rkuhn: thanks! my question was mostly, which variant do we want to produce?

crates/amaru-consensus/src/chain_forward/client_protocol.rs
rkuhn: yeah, I just wanted to sketch the whole range, we don’t need details here

rkuhn (commented):
@abailly I added a README with a diagram, laying out the basic reasoning behind the approach.

crates/amaru-consensus/src/chain_forward/client_protocol.rs
abailly: It should be 1 AFAIK

crates/amaru-consensus/Cargo.toml
```diff
@@ -30,6 +30,7 @@ amaru-kernel.workspace = true
 amaru-ledger.workspace = true
 amaru-ouroboros.workspace = true
 amaru-ouroboros-traits.workspace = true
+acto = { version = "0.7.2", features = ["tokio"] }
```
KtorZ: Please declare the dependency version in the top-level Cargo.toml, following the existing pattern 🙏

crates/amaru-consensus/src/chain_forward.rs
```diff
@@ -13,67 +13,146 @@
...
+    },
+    task::JoinHandle,
+};
 use tracing::{error_span, info, trace_span, warn};
 
 pub type UpstreamPort = gasket::messaging::InputPort<BlockValidationResult>;
 
 pub const EVENT_TARGET: &str = "amaru::consensus::chain_forward";
 
+pub const LISTEN_ADDRESS: &str = "0.0.0.0:3001";
```
KtorZ:
```suggestion
// FIXME: should be default for an incoming configuration, eventually. 
pub const LISTEN_ADDRESS: &str = "0.0.0.0:3001";
```

crates/amaru-consensus/src/chain_forward.rs
```diff
@@ -13,67 +13,146 @@
...
 
-pub struct Worker {}
+pub enum Unit {
+    Peer(RefCell<Option<PeerServer>>),
+    Block(BlockValidationResult),
+}
+
+pub struct Worker {
+    server: JoinHandle<()>,
+    rx: Receiver<PeerServer>,
```
KtorZ:
```suggestion
    incoming_peers: Receiver<PeerServer>,
```

crates/amaru-consensus/src/chain_forward.rs
```diff
@@ -85,9 +164,16 @@ impl gasket::framework::Worker<ForwardStage> for Worker {
                     hash = %Hash::<32>::from(point),
                 );
 
+                // FIXME: block height should be part of BlockValidated message
+                let store = stage.store.lock().await;
+                if let Some(header) = store.load_header(&Hash::from(point)) {
+                    self.tip = Tip(to_pallas_point(point), header.block_height());
+                    self.clients.send(ClientMsg::Op(ClientOp::Forward(header)));
+                }
```
KtorZ: Indeed!

crates/amaru-consensus/src/chain_forward.rs
```diff
@@ -107,15 +193,74 @@ impl gasket::framework::Worker<ForwardStage> for Worker {
...
+                if let Some(header) = store.load_header(&Hash::from(point)) {
+                    self.tip = Tip(to_pallas_point(point), header.block_height());
+                    self.clients
+                        .send(ClientMsg::Op(ClientOp::Backward(to_pallas_point(point))));
+                }
+
+                Ok(())
+            }
+            Unit::Peer(peer) => {
+                // FIXME: gasket design bug that we only get &Unit and thus cannot take values from it without internal mutability
```
KtorZ: Ah. Hence the RefCell, I see.

crates/amaru-consensus/src/chain_forward/client_protocol.rs
```diff
@@ -0,0 +1,249 @@
...
+    fn from(value: txsubmission::Error) -> Self {
+        ClientError::TxSubmission(value)
+    }
+}
+
+impl From<keepalive::ServerError> for ClientError {
+    fn from(value: keepalive::ServerError) -> Self {
+        ClientError::KeepAlive(value)
+    }
+}
```
KtorZ: Can't we just get those _for free_ using the `#[from]` annotation in the `ClientError` enum?

crates/amaru-consensus/src/chain_forward.rs
```diff
@@ -107,15 +193,74 @@ impl gasket::framework::Worker<ForwardStage> for Worker {
...
+                let addr = peer
+                    .accepted_address()
+                    .map(|a| a.to_string())
+                    .unwrap_or_default();
+
+                let client = cell.spawn_supervised(&addr, {
+                    let store = store.clone();
+                    move |cell| client_protocols(cell, peer, store, tip)
+                });
+                clients.insert(client.id(), client);
```
KtorZ: Might be worth make the clients a bounded (configurable) structure? We probably don't want to just _accept_ as many connections as there are requests. We shall protect ourselves from potential adversaries spamming our node.

crates/amaru-consensus/src/chain_forward/client_protocol.rs
```diff
@@ -0,0 +1,249 @@
...
+
+pub async fn client_protocols(
+    mut cell: ActoCell<ChainSyncMsg, impl ActoRuntime, Result<(), ClientError>>,
+    server: PeerServer,
+    store: Arc<Mutex<dyn ChainStore<Header>>>,
```
KtorZ: It might be worth thinking about introducing some read-only interface for the ChainStore eventually? Here, we'll have to lock and wait on the mutex every time clients wants to interact with the database -- also blocking the main thread from progressing. 

If anything, serving any of the chain-sync client should have a lower priority than the main job. So we need some kind of priority Mutex that ensures the main job is served _first_ when possible. Clients are secondary though since they don't need to write anything to the store in principle, we could have them work from some read-only structure and enable (1) more parallelism in the serving of the clients, and (2) avoid blocking the main job needlessly.

crates/amaru-consensus/src/chain_forward.rs
rkuhn: Yes, of course, as soon as I get it working in principle.

crates/amaru-consensus/Cargo.toml
rkuhn: sure, this was temporary while I wasn’t sure whether `acto` is an acceptable dependency

crates/amaru-consensus/src/chain_forward/client_state.rs
rkuhn: Thinking some more about it: this is not related to chain selection, it is only related to computing the intersection in chain sync. So I’d like to keep it here.

crates/amaru-consensus/src/chain_forward/client_state.rs
rkuhn: This requires some more work and thought: should Amaru support syncing an empty client from Genesis, or should it in that case use a better protocol that would ship a snapshot?

crates/amaru-consensus/src/chain_forward/client_protocol.rs
rkuhn: yes, you’re right

crates/amaru-consensus/src/chain_forward/client_state.rs
abailly: For the time being, let's assume we can sync the whole chain from genesis, which is the base protocol. In practice, we could respond to an intersect request not including `Genesis`

KtorZ:
What's up with this PR? Is it abandoned / superseded?

rkuhn:
So far I didn’t get any thumbs-up — how does this work here?

abailly:
> So far I didn’t get any thumbs-up — how does this work here?

I think the rule is that if 1/ the CI is green and 2/ there's been reviews without any blocking comments, you should be good to go. If this works, I would like to merge it so that at least we have _some_ ability to propagate headers downstream.

rkuhn:
aye aye; rebased across the move, will now add some basic client limits in place and then make CI green

crates/amaru-consensus/src/chain_forward/client_protocol.rs
rkuhn: Yes, I created #194 to track this.

crates/amaru-consensus/src/chain_forward.rs
rkuhn: simple approach has been implemented, to be expanded in subsequent PRs.

rkuhn:
Alright, I pushed this as far as I can without diving into rabbit holes. Any ideas about the remaining CI failure?

abailly:
I agree the message is not very explicit

```
Error: Open(IO(Os { code: 2, kind: NotFound, message: "No such file or directory" }))
```

abailly:
> Alright, I pushed this as far as I can without diving into rabbit holes. Any ideas about the remaining CI failure?

I think the problem is that you've set the directories where ledger and chain are restored explicitly in the `Makefile`, but the `demo.sh` script still uses the default paths defined in the code. Either revert the `Makefile` to use default directories (I think that's better if only because it reduces the surface of the change), or pass the directories to `demo.sh`

crates/amaru/src/stages/consensus/chain_forward/client_protocol.rs
```diff
@@ -0,0 +1,250 @@
...
+                    tracing::debug!("sending roll backward");
+                    server.send_roll_backward(point.0, tip).await?;
+                }
+                None => {
+                    tracing::debug!("sending await reply");
+                    server.send_await_reply().await?;
+                    let ActoInput::Message(Some((op, tip))) = cell.recv().await else {
+                        return Ok(());
+                    };
+                    match op {
```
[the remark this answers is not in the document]
rkuhn: @abailly This remark from the code rabbit is uncanny: it does recognise how `acto` works (which is not the most popular library) while implying that dropping the `chainsync` server won’t be noticed by the client on the other side. Is closing the stream not signalled by the muxer?

crates/amaru/src/stages/consensus/chain_forward/test_infra.rs
```diff
@@ -0,0 +1,351 @@
...
+            .spawn_actor(
+                "test",
+                |mut cell: ActoCell<ForwardEvent, AcTokioRuntime>| async move {
+                    while let ActoInput::Message(msg) = cell.recv().await {
+                        port_tx.send(msg).await.unwrap();
+                    }
+                },
+            )
+            .me;
+        let (block_tx, block_rx) = mpsc::channel(1);
```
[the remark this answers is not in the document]
rkuhn: This is a great catch, thanks! Who taught this rabbit about The Running Man, though?

crates/amaru/src/stages/consensus/chain_forward/tests.rs
```diff
@@ -0,0 +1,172 @@
...
+    let (ops, Tip(p, h)) = find_headers_between(&store, &tip, &points).unwrap();
+    assert_eq!(
+        (ops.len() as u64, p, h),
+        (
+            store.get_height(TIP_47) - store.get_height(WINNER_47),
+            peer,
+            store.get_height(WINNER_47)
+        )
+    );
+}
```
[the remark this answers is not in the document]
rkuhn: indeed, will fix the docs

crates/amaru/src/stages/mod.rs
```diff
@@ -86,7 +89,14 @@ pub fn bootstrap(
...
         .collect::<Vec<_>>();
     let chain_store = RocksDBStore::new(config.chain_dir.clone(), era_history)?;
-    let chain_selector = make_chain_selector(tip, &chain_store, &peer_sessions)?;
+
+    #[allow(clippy::expect_used)]
+    let header: Header = chain_store
+        .load_header(&Hash::from(&tip))
+        .expect("Tip not found");
+    let our_tip = Tip(header.pallas_point(), header.block_height());
+
```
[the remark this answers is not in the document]
rkuhn: Good catch: a fresh node should start out from origin, of course.

crates/amaru/src/stages/consensus/chain_forward/test_infra.rs
```diff
@@ -0,0 +1,351 @@
...
+            match msg {
+                NextResponse::RollForward(header, tip) => {
+                    ops.push(ClientMsg::Forward(from_cbor(&header.cbor).unwrap(), tip))
+                }
+                NextResponse::RollBackward(point, tip) => ops.push(ClientMsg::Backward(point, tip)),
+                NextResponse::Await => break,
+            }
+        }
+        ops.try_into().unwrap()
+    }
```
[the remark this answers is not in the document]
rkuhn: No, this is test infrastructure, panic is exactly the right thing to do.

crates/amaru/src/stages/consensus/chain_forward.rs
```diff
@@ -85,30 +258,134 @@
...
+            Unit::Peer(peer) => {
+                // FIXME: gasket design bug that we only get &Unit and thus cannot take values from it without internal mutability
+                let peer = peer.borrow_mut().take();
+                if let Some(peer) = peer {
+                    self.clients
+                        .send(ClientMsg::Peer(peer, self.our_tip.clone()));
+                } else {
+                    tracing::error!(target: EVENT_TARGET, "Unit::Peer was empty in execute");
+                }
+                Ok(())
```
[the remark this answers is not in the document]
rkuhn: Gasket has made this API choice because it allows scheduled work units to be retries. This is not used here, but changing the gasket API seems inadequate in this case.

crates/amaru/src/stages/consensus/chain_forward.rs
```diff
@@ -12,68 +12,241 @@
...
+        downstream: Option<ActoRef<ForwardEvent>>,
+        store: Arc<Mutex<dyn ChainStore<Header>>>,
+        network_magic: u64,
+        listen_address: &str,
+        max_peers: usize,
+        our_tip: Tip,
+    ) -> Self {
+        #[allow(clippy::expect_used)]
+        let runtime =
+            AcTokio::new("consensus.forward", 1).expect("failed to create AcTokio runtime");
```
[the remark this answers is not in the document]
rkuhn: We’re currently working in the mode that fatal errors tear down the node with a clear message — which is exactly what a Rust `panic!()` is made for. Ideally, this would be handled using `Result` passing and shutting down gracefully or in central places, but the codebase is just not there yet. So for now the `.expect()` should stay.

PR #152, opened 2025-03-19
fix: fix chain selection for children of genesis

abailly:
Due to discrepancies between the representation of parents b/w Rust and Haskell, the chain selection would not work for blocks immediately after genesis and therefore would ignore whole chain if it starts from genesis. By fixing the FakeHeader representation to be more faithful to actual header, we make the problem surface and can fix it by handling this edge case.

## Summary by CodeRabbit

- **Refactor**
  - Enhanced chain selection and header processing to achieve clearer and more robust management of hierarchical relationships.
  - Streamlined the header representation to more intuitively handle optional parent data.

- **Tests**
  - Updated and expanded test cases to verify consistency in header generation and sequencing, ensuring improved reliability.


crates/amaru-consensus/src/consensus/chain_selection.rs
```diff
@@ -62,13 +62,20 @@ impl<H: IsHeader + Clone> Fragment<H> {
...
-                return FragmentExtension::Extend;
+                FragmentExtension::Extend
             }
+            None => match self.tip() {
+                Tip::Genesis => {
+                    self.headers.push(header.clone());
+                    FragmentExtension::Extend
+                }
+                Tip::Hdr(_) => FragmentExtension::Ignore,
+            },
```
rkuhn, line 76: I’d prefer to keep the decision logic forcibly consistent with the performed action:
```rs
        let ret = match header.parent() {
            Some(parent) if parent == self.tip().hash() => FragmentExtension::Extend,
            None if self.tip() == Tip::Genesis => FragmentExtension::Extend,
            _ => FragmentExtension::Ignore,
        };

        if ret == FragmentExtension::Extend {
            self.headers.push(header.clone());
        }

        ret
```
(need some `#[derive(PartialEq)]` to work)

PR #153, opened 2025-03-20
chore: make sure Store doesn't have to be shareable

jeluard:
`gasket` requires stages to be `Send + Sync`. Move some constraints so that `Store` itself (via `State`) doesn't have to impose shareability on all implementations. This can be useful in e.g. `wasm` context.


crates/amaru/src/pipeline.rs
```diff
@@ -15,14 +15,14 @@ pub type DownstreamPort = gasket::messaging::OutputPort<BlockValidationResult>;
 
 pub struct Stage<S>
 where
-    S: Store,
+    S: Store + Send + Sync,
```
rkuhn, line 18: The `Sync` bound is not needed: the Stage isn't accessed simultaneously from multiple threads.

PR #154, opened 2025-03-21
Introduce some useful test macros

KtorZ:
## Summary by CodeRabbit

- **Refactor**
	- Organised internal code for improved clarity and maintenance.
- **Tests**
	- Streamlined property-based tests ensuring robust data serialisation and deserialisation.
	- Introduced new testing modules to expand coverage.
- **Chores**
	- Removed outdated test components and refined module imports for simplified upkeep.

These changes enhance code quality and testing reliability without impacting any user-facing functionality.

### DIFF
--- a/crates/amaru-kernel/src/macros.rs
+++ b/crates/amaru-kernel/src/macros.rs
@@ -28,3 +28,36 @@ macro_rules! prop_cbor_roundtrip {
         prop_cbor_roundtrip!(prop_cbor_roundtrip, $ty, $strategy);
     };
 }
+
+/// Easily create a hash from a hex-encoded literal string. Useful for testing.
+///
+/// Requires:
+/// - hex
+///
+/// Usage:
+///
+/// # ```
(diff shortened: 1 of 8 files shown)


crates/amaru-kernel/src/macros.rs
```diff
@@ -0,0 +1,63 @@
...
+///
+/// Usage:
+///
+/// # ```
+/// # let my_hash32: Hash<32> = hash!("a7c4477e9fcfd519bf7dcba0d4ffe35a399125534bc8c60fa89ff6b50a060a7a"),
+/// # let my_hash28: Hash<28> = hash!("a7c4477e9fcfd519bf7dcba0d4ffe35a399125534bc8c60fa89ff6b5"),
+/// # ```
+#[macro_export]
+macro_rules! hash {
+    ($str:literal $(,)?) => {{
```
rkuhn, line 45: might also include the expected hash length as a parameter here and then check the length against twice that value, also use `Hash::<SIZE>` below so calling code doesn’t have to add the type constraint again

PR #155, opened 2025-03-21
Block validation context and signature validation

KtorZ:
Originally from #148 

## Summary by CodeRabbit

- **New Features**
	- Introduced a new cryptographic dependency to increase security.
	- Added enhanced validation routines for improved transaction integrity.
	- Extended public interfaces to support refined address and key hash handling.
	- Introduced new modules for validating bootstrap and vkey witnesses.
	- Added new traits for managing key hashes required for transaction validation.
	- Implemented new structures for block preparation and validation contexts.

- **Refactor & Style**
	- Reorganised dependency configurations for better readability.
	- Updated validation processes to utilise an enhanced state management context.
	- Improved error reporting mechanisms throughout the validation workflows.


KtorZ:
@yHSJ: I've wrapped up the PR, ensuring some early integration between the ledger validations and the newly introduce _context_. A few important things for you to note:

1. There's now [a preparation step](https://github.com/pragma-org/amaru/blob/eeb898e2dbf87ab8eaa7616a32bb1e815600c1de/crates/amaru-ledger/src/rules.rs#L154-L181), ahead of the validation. This step is used to figure out what data are going to be needed by the validation context. At this point, since the rules only requires information about inputs (for key witness validations), then we only construct a context for those. 

2. The preparation of the validation context is entirely decoupled from the [actual pre-fetching of the underlying data](). Right now, we define two types Preparation/Validation context: 
    - A [fake](https://github.com/pragma-org/amaru/blob/eeb898e2dbf87ab8eaa7616a32bb1e815600c1de/crates/amaru-ledger/src/rules/context/fake.rs) one, that is suitable for testing and requires the fixture data to be provided upfront. Used for example [here](https://github.com/pragma-org/amaru/blob/eeb898e2dbf87ab8eaa7616a32bb1e815600c1de/crates/amaru-ledger/src/rules.rs#L309-L327) or [here](https://github.com/pragma-org/amaru/blob/eeb898e2dbf87ab8eaa7616a32bb1e815600c1de/crates/amaru-ledger/src/rules.rs#L371-L379).
    - A [simple](https://github.com/pragma-org/amaru/blob/eeb898e2dbf87ab8eaa7616a32bb1e815600c1de/crates/amaru-ledger/src/rules/context/simple.rs) one, that is currently being used during normal operation. Calls to `require` will mark data as needed, and I've introduced a step for [pre-fetching the underlying data and constructing a valid simple validation context](https://github.com/pragma-org/amaru/blob/eeb898e2dbf87ab8eaa7616a32bb1e815600c1de/crates/amaru/src/pipeline.rs#L65-L98). 

    For now, the `simple` is extremely basic, and implement just what's needed to get the integration tests pass. But now, a next step (from my end) will be to fully port the current state management to this new approach, very likely just re-using what we have now. 

3. Then, the validation code now successfully uses the provided context to [(a) resolve inputs from it](https://github.com/pragma-org/amaru/blob/eeb898e2dbf87ab8eaa7616a32bb1e815600c1de/crates/amaru-ledger/src/rules/transaction/vkey_witness.rs#L24) and [(b) produce outputs which may be needed by further validations](https://github.com/pragma-org/amaru/blob/eeb898e2dbf87ab8eaa7616a32bb1e815600c1de/crates/amaru-ledger/src/rules.rs#L290-L298).

    At this point, we do not _consume_ inputs, so double-spending is effectively possible :). We shall do this at the end of every transaction, of course. 

4. With this last changes, we're able to run the integration tests and actually validate signatures on PreProd as part of the ledger validations 🎉! Note that, this wasn't the case before since we would simply ignore any such validation of any unknown input... 
    
    Generally speaking, the code in the rules crucially misses tests. At the very least now, the integration tests will catch a few things, but they aren't sufficient -- they should remain only a last rampart. 

    I've added [an example of a validation yielding a `MissingVKeyWitness`](https://github.com/pragma-org/amaru/blob/eeb898e2dbf87ab8eaa7616a32bb1e815600c1de/crates/amaru-ledger/src/rules.rs#L421-L455) and moving forward, I'd expect more of those for each possible validation failure -- and this, before we introduce any new rules (and that'll give us some time in the meantime to finish merging the state & validations!). 

 5. I've also reworked some errors, to make the rules a bit more composable. The idea is to structure errors in the same way we structure rules, and instead of having "low-level rules" yields plain `TransactionRuleViolation` which will rapidly grow quite large; they can yield an error that's more restricted to the thing they're validating. 

    So far, I've done it in two (three if we count the unification of bootstrap/vkey witnesses verification) places:
    - The [`InvalidVKeyWitnesses`](https://github.com/pragma-org/amaru/blob/eeb898e2dbf87ab8eaa7616a32bb1e815600c1de/crates/amaru-ledger/src/rules.rs#L90-L92) now takes an array of `InvalidVKeyWitness` errors which are produced by [verify_ed25519_signature](https://github.com/pragma-org/amaru/blob/eeb898e2dbf87ab8eaa7616a32bb1e815600c1de/crates/amaru-ledger/src/rules/transaction/vkey_witness.rs#L122-L144). 

    - The [`InvalidOutputs`](https://github.com/pragma-org/amaru/blob/eeb898e2dbf87ab8eaa7616a32bb1e815600c1de/crates/amaru-ledger/src/rules.rs#L76-L78) now takes an array of `InvalidOutput` errors, which are produced by single outputs validations. We only have one for now [regarding the min ada value](https://github.com/pragma-org/amaru/blob/eeb898e2dbf87ab8eaa7616a32bb1e815600c1de/crates/amaru-ledger/src/rules/transaction/outputs/inherent_value.rs#L20-L38)

    Remark that I am no longer cloning the underlying data as part of the error, but instead prefer reporting errors using an indice/position of the corresponding element within the transaction. To avoid repeating this over and over, I've introduced [a little helper structure](https://github.com/pragma-org/amaru/blob/eeb898e2dbf87ab8eaa7616a32bb1e815600c1de/crates/amaru-ledger/src/rules.rs#L118-L128) that makes [the construction of such errors straighforward](https://github.com/pragma-org/amaru/blob/eeb898e2dbf87ab8eaa7616a32bb1e815600c1de/crates/amaru-ledger/src/rules/transaction/outputs/mod.rs#L29-L30). 

6. Finally, you might notice that I keep re-ordering arguments in functions. From example,

    going from:

    ```rust
    pub fn validate_block(
        raw_block: &[u8],
        context: &mut impl BlockValidationContext,
        protocol_params: ProtocolParameters,
    ) 
   ```

   to

   ```rust
   pub fn validate_block(
       context: &mut impl BlockValidationContext,
       protocol_params: ProtocolParameters,
       raw_block: &[u8],
   ) 
   ```

   This is just my "functional programming brain" at play here, where we tend to put parameters to a function first, and keep the main operand last. This is particular useful in languages that support currying since you can generate new specialized functions more easily by partially applying parameters different parameters. 

    In a language like Rust, this is debatable, although I still like the semantic context that comes with it (the actual _thing_ the function operates on comes last). So I'll most likely keep doing this and that's a style I invite you to follow 😅.

So, as TODOs before introducing new rules; I do expect:

- [ ] More unit tests covering all existing validation rules (and violations). If you need help for generating test blocks or arbitrary data for some property tests, I am sure we can find someone -- if not me -- to help with that! 

- [ ] Some tests specifically the [`to_root`](https://github.com/pragma-org/amaru/blob/eeb898e2dbf87ab8eaa7616a32bb1e815600c1de/crates/amaru-kernel/src/lib.rs#L616-L629) function, with various bootstrap witness parts.

crates/amaru-kernel/src/lib.rs
```diff
@@ -387,6 +396,28 @@ impl<'b, C> cbor::decode::Decode<'b, C> for CertificatePointer {
...
+/// # let public_key: [u8; ed25519::PublicKey::SIZE] = into_sized_array(vkey, |error, expected| {
+/// #     InvalidVKeyWitness::InvalidKeySize { error, expected }
+/// # })?;
+/// # ```
+pub fn into_sized_array<const SIZE: usize, E, T>(
+    bytes: T,
+    into_error: impl Fn(TryFromSliceError, usize) -> E,
+) -> Result<[u8; SIZE], E>
+where
+    T: Deref<Target = Bytes>,
```
rkuhn, line 412: why not `AsRef<[u8]>`?

crates/amaru-kernel/src/lib.rs
```diff
@@ -534,7 +565,70 @@ pub fn to_ex_units(witness_set: WitnessSet) -> ExUnits {
...
+                StakePayload::Script(_) => None,
+            },
+        }
+    }
+}
+
+/// Construct the bootstrap root from a bootstrap witness
+pub fn to_root(witness: &BootstrapWitness) -> Hash<28> {
+    // CBOR header for data that will be encoded
+    let prefix: &[u8] = &[131, 0, 130, 0, 88, 64];
```
rkuhn, line 619: If we leave these magical arrays in the codebase, then they should be written for easy parsing:

- `0x83` starts an array of length 3
  - `0` is the integer value zero
  - `0x82` starts a nested array of length 2
    - `0` is the integer value zero
    - `0x58, 64` starts a byte string of length 64, but the content is missing

From the code I was able to infer that `public_key` and `chain_code` are 32 bytes each, so they are concatenated for this hashing purpose. It follows that `attributes` needs to be a proper CBOR item to complete the outer array.

Perhaps a better way would be to use [`cbor-data`](https://docs.rs/cbor-data) for constructing these values:

```rs
let cbor = CborBuilder::new()
    .with_max_definite_size(None) // always use definite size
    .encode_array(|builder| {
        builder.encode_u64(0);
        builder.encode_array(|builder| {
            builder.encode_u64(0);
            builder.write_bytes_chunked([&witness.public_key, &witness.chain_code]);
        });
        builder.write_canonical(witness.attributes.as_slice()).expect("valid CBOR item");
    });

sha_hasher.update(cbor.as_slice());
```

(the above allocates twice; this could be mitigated by using a thread-local [scratch space](https://docs.rs/cbor-data/latest/cbor_data/struct.CborBuilder.html#method.with_scratch_space))

crates/amaru-ledger/src/rules/context.rs
```diff
@@ -0,0 +1,106 @@
...
+pub trait BlockValidationContext:
+    PotsSlice + UtxoSlice + PoolsSlice + AccountsSlice + DRepsSlice
+{
+}
+
+/// The BlockPreparationContext is a collection of interfaces needed to prepare a block
+pub trait BlockPreparationContext<'a>:
+    PrepareUtxoSlice<'a> + PreparePoolsSlice<'a> + PrepareAccountsSlice<'a> + PrepareDRepsSlice<'a>
+{
+}
```
rkuhn, line 34: When using these small slices it is usually nice to add a blanket implementation for anything that implements all the slices.

crates/amaru-ledger/src/rules/context.rs
```diff
@@ -0,0 +1,106 @@
...
+pub trait UtxoSlice {
+    fn lookup(&self, input: &TransactionInput) -> Option<&TransactionOutput>;
+    fn consume(&mut self, input: &TransactionInput);
+    fn produce(&mut self, input: TransactionInput, output: TransactionOutput);
+}
+
+/// An interface to help constructing the concrete UtxoSlice ahead of time.
+pub trait PrepareUtxoSlice<'a> {
+    fn require_input(&'_ mut self, input: &'a TransactionInput);
+}
```
rkuhn, line 51: This is a matter of taste and opinion: I only tend to split up traits when the individual pieces shall be used independently from each other. This PR so far doesn’t seem to show such independence, instead using the preparation context trait.

crates/amaru-ledger/src/rules/transaction/outputs/mod.rs
```diff
@@ -0,0 +1,41 @@
...
+use amaru_kernel::{
+    protocol_parameters::ProtocolParameters, MintedTransactionOutput, TransactionOutput,
+};
+
+mod inherent_value;
+
+pub fn execute<'a>(
+    protocol_parameters: &ProtocolParameters,
+    outputs: impl Iterator<Item = &'a MintedTransactionOutput<'a>>,
+    yield_output: &mut impl FnMut(u64, TransactionOutput),
```
rkuhn, line 25: this would be a perfect opportunity for `&mut dyn FnMut` instead, reducing monomorphisation cost and binary size

rkuhn (commented):
I’m trying to get more into the codebase by reviewing some PRs. Unfortunately, I typically won't be able to do this in real-time during the week since I can’t switch to focused reading on short notice. I’d appreciate your insights in order to learn the lay of the land.

crates/amaru-kernel/src/lib.rs
KtorZ: Because I unfortunately don't get to choose the input type here. The concrete type at call-site only have a `Deref` instance. Ideally, it should also have AsRef, but I add that to the long list of things we need to contribute to Pallas.

crates/amaru-ledger/src/rules/context.rs
KtorZ: What would that look like, and why is it nice? (genuine question)

crates/amaru-ledger/src/rules/context.rs
KtorZ: Good point, though the PR mostly introduces some preparatory work, but the traits are indeed used independently!

crates/amaru-ledger/src/rules/transaction/outputs/mod.rs
KtorZ: But I do not want a dynamic dispatch here e_e ; there are only two cases where this function is going to be called and so, I would hope this to be a mostly-zero-cost-abstraction!

crates/amaru-kernel/src/lib.rs
rkuhn: Hmm, okay, but that’s a really weird API choice: `Deref`’s purpose is purely for convenient wrapper types, to delegate calls to the wrapped value — the docs make it quite clear that it is for a type that transparently behaves like the other type, but _is not that type_ (there’s no blanket impl, so the above wouldn’t accept `Byes` as argument type). It is not very useful as a type bound because it has no semantic meaning. `AsRef<T>` on the other hand exists only for the purpose of using values in other places that need a specific reference type to be projected.

crates/amaru-ledger/src/rules/context.rs
rkuhn:
```rs
impl<'a, T> BlockPreparationContext<'a> for T
where T: PrepareUtxoSlice<'a> + PreparePoolsSlice<'a> + ...
{}
```

The nicety lies in implementations of that particular bunch of traits automatically getting an impl for the combined trait. Otherwise that type would manually have to say “hey, I also support BlockPreparationContext”.

crates/amaru-ledger/src/rules/transaction/outputs/mod.rs
rkuhn: Yeah, your stance sounds entirely reasonable and it is what everybody does. No single instance within a codebase can be blamed individually, but the tendency to just use `impl T` instead of `dyn T` then leads to long compilation times and huge binaries. Huge binaries also lead to slower code because of limited CPU caches.

Perhaps keep in mind that almost all other modern languages only every allow you to do dynamic dispatch — the JVM is extremely good at it because the Java language requires that to be the case. It wouldn’t hurt the Rust ecosystem to default to `dyn` and only use `impl` after benchmarking that it actually improves the result.

But this is not a hill I will die on. I’m just trying to spread the word.

crates/amaru-kernel/src/lib.rs
KtorZ: I do not disagree; merely composing with what is given 🤭. 

I wish Pallas implemented AsRef; but hey, we can contribute!

Quantumplation: Small things like this are great candidates to put someone from Sundae on.

crates/amaru-ledger/src/rules/transaction/outputs/mod.rs
KtorZ: Words have been heard! I am definitely biased w.r.t compilation time. I come from Haskell. Rust just feels blazing fast to me 😅! But I can certainly agree with your reasoning. Admittedly using `impl T` because of _performance reasons_ which is probably not quite backed by any solid argument. In the end, I don't have any number either to back this claim. I've just _been told_ by other Rust developers to avoid dynamic dispatch when possible. I guess as always... both have trade-offs. Thanks for explaining!

PR #157, opened 2025-03-23
Complete basic consensus simulator

abailly:
This PR completes work to connect `simulation-testing` package to Amaru to test consensus and, indirectly ledger. 
Main changes are:

* breakdown header and block validation pipeline to add a `block_fetch` stage that takes care of fetching requested block given header is valid. This allows the simulator to be built with the same `header_validation` code without having to simulate block fetching which shall be taken care of later
* allow test to control various parameters for header validation through configuration files (eg. nonces, starting header hash, etc.)
* "correct" (manually tested) handling of maelstrom-style messages sent by the tester

TODO:  
* configure network parameters. Tests can only currently run by manually editing `epoch_from_slot` and `next_epoch_first_slot` to match what's in the generated test headers


crates/amaru-kernel/src/lib.rs
```diff
@@ -152,6 +152,12 @@ impl Point {
             Point::Specific(slot, _) => *slot,
         }
     }
+    pub fn pretty_point(&self) -> String {
```
jeluard: Better to implement `Display`?

crates/amaru-kernel/src/lib.rs
```diff
@@ -152,6 +152,12 @@ impl Point {
...
+        match self {
+            Point::Origin => "origin".to_string(),
+            Point::Specific(_slot, hash) => String::from_utf8(hex::encode(hash).into()).unwrap(),
```
jeluard: `unwrap` shouldn't be used

simulation/amaru-sim/src/simulator/mod.rs
```diff
@@ -151,7 +151,7 @@ async fn run_simulator(
                 let events = match mk_message(msg, span) {
                     Ok(event) => match event {
                         PullEvent::RollForward(peer, point, raw_header, _span) => {
-                            info!("got forward event {:?}", &point);
+                            info!("got forward event {:?}", &point.pretty_point());
```
jeluard: Those kind of `info` do not quite align with the `tracing` philosophy. Is it required by the simulator @stevana ?

crates/amaru-kernel/src/lib.rs
stevana: There's several unwraps in the code already. Perhaps there should be a linter to catch these?

jeluard: Definitely :) There is one already, your commit doesn't pass the [sanity check](https://github.com/pragma-org/amaru/actions/runs/14036257998/job/39294877094)

simulation/amaru-sim/src/simulator/mod.rs
stevana: I didn't add the `info!`, I merely tried to make it's output actually readable (the output was gibberish without my change).

abailly: I will remove it, it's useless anyway

PR #158, opened 2025-03-23
Unify ledger rules and state management

KtorZ:
It's a chunky one but, it showcase that how the ledger context/slices approach can be fully integrated with the state management and that, it works 🎉! 

Please make sure to start any new work on top of this branch @jeluard and @yHSJ since it merges both worlds and while there's still a separation between the state and the rules, we're likely going to need slightly more alignment between chunks of work. [The comment from #155](https://github.com/pragma-org/amaru/pull/155#issuecomment-2745382656) still applies.

I did preserve all existing tests, and left the VolatileState untouched -- although we spoke several time about possibly adjusting it with @jeluard. The input pre-fetching is also not ideal as it happens one-by-one, and there are still a few clones here and there that have been necessary. Overall though, that puts us in a nice spot.

## Summary by CodeRabbit

- **New Features**
  - Upgraded transaction processing with new modules for fee management, certificate handling, voting procedures, and withdrawals, resulting in more robust and secure operations.
  - Enhanced ownership credential handling to improve delegation and registration flows, offering a smoother user experience.

- **Refactor**
  - Extensive improvements to state management and error reporting, optimising validation and processing workflows for increased overall system reliability.


crates/amaru-ledger/src/context.rs
```diff
@@ -33,12 +44,68 @@ pub trait PreparationContext<'a>:
 {
 }
 
+// Errors
+// -------------------------------------------------------------------------------------------------
+
+#[derive(thiserror::Error, Debug)]
+pub enum RegisterError<ROLE, K> {
+    #[error("unknown entity: {0:?}")] // TODO: Use Display
+    AlreadyRegistered(PhantomData<ROLE>, K),
```
Quantumplation, line 53: I'm confused by the pairing of the enum variant name and the error message; copy paste error?

jeluard:
The nodejs example now fail as the block validation now performs some validation expecting some stats e.g. inputs.

KtorZ:
We probably don't even need to touch the store, but only the validation context. 

I'd suggest we do just that; instead of passing an empty set of UTxO, we provide the ones required for that block validation.

crates/amaru-ledger/src/context.rs
KtorZ: Seems like it. Good catch

PR #163, opened 2025-03-25
feat: allow testnets with magic in network names

abailly:
## Summary by CodeRabbit

- **New Features**
	- Introduced support for test networks with a dedicated variant, allowing users to specify a network via an associated numeric identifier. Enhanced parsing and display of network names ensure a smoother user experience when dealing with test configurations.

- **Refactor**
	- Updated network configuration by sourcing a key network transition value from an external definition rather than a fixed internal value, streamlining the network parameter management process.
	- Simplified argument parsing for the network field, removing custom validation logic for input values.

### DIFF
--- a/crates/amaru-kernel/src/network.rs
+++ b/crates/amaru-kernel/src/network.rs
@@ -12,21 +12,25 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#[derive(Debug, Clone, Copy)]
+/// Epoch number in which the PreProd network transitioned to Shelley.
+pub const PREPROD_SHELLEY_TRANSITION_EPOCH: usize = 4;
+
+#[derive(Debug, PartialEq, Clone, Copy)]
 pub enum NetworkName {
     Mainnet,
     Preprod,
     Preview,
+    Testnet(u32),
 }
 
 impl std::fmt::Display for NetworkName {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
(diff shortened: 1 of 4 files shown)


abailly:
@KtorZ not sure about using `testnet<1234>` to represent Testnets, WDYT?

crates/amaru-kernel/src/network.rs
```diff
@@ -38,21 +42,55 @@ impl std::str::FromStr for NetworkName {
...
+    }
+
+    proptest! {
+        #[test]
+        fn prop_can_parse_pretty_print_network_name(network in any_network()) {
+            let name = format!("{}", network);
+            assert_eq!(
+                FromStr::from_str(&name),
+                Ok(network),
+            )
```
KtorZ, line 93: 👌

crates/amaru-kernel/src/network.rs
```diff
@@ -12,21 +12,25 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#[derive(Debug, Clone, Copy)]
+/// Epoch number in which the PreProd network transitioned to Shelley.
+pub const PREPROD_SHELLEY_TRANSITION_EPOCH: usize = 4;
```
KtorZ, line 16: Any reason this was moved?

KtorZ (approved):

crates/amaru-kernel/src/network.rs
abailly: I anticipated the need to group network-related parameters/constant (eg. `EraHistory`), I can move it back if you prefer

PR #164, opened 2025-03-25
feat: use era history to compute slot/epoch

abailly:
This PR introduces the use of `EraHistory` from `slot-arithmetic` crate into Amaru to provide epoch and slot computations. So far, those computations were hardcoded but in order to be able to test Amaru we need to be able to configure the system with different testing environments. This is a first step in that direction to validate the approach.

## Summary by CodeRabbit

- **New Features**
	- Enhanced network time conversion with a refined method for calculating epochs using detailed era history.
	- Expanded project capabilities through the integration of new workspace dependencies, including `serde`, `thiserror`, and `slot-arithmetic`.
	- New methods and fields added to various structs to incorporate era history into state management and database operations.
	- Introduced functionality for era history retrieval in test modules.
- **Bug Fixes**
	- Improved error handling for era history file operations.
- **Tests**
	- Introduced new validations to ensure accurate slot-to-epoch conversions for the PreProd network.

### DIFF
--- a/crates/amaru-kernel/src/lib.rs
+++ b/crates/amaru-kernel/src/lib.rs
@@ -495,24 +495,6 @@ pub fn relative_slot(slot: u64) -> u64 {
     slot - shelley_previous_slots - BYRON_TOTAL_SLOTS as u64
 }
 
-/// Get the first slot of the next epoch (i.e. the slot coming straight after the last slot of the
-/// epoch).
-///
-/// ```
-/// use amaru_kernel::next_epoch_first_slot;
-/// assert_eq!(next_epoch_first_slot(3), 86400);
-/// assert!(next_epoch_first_slot(114) <= 48038412);
-/// assert!(next_epoch_first_slot(114) > 48038393);
-/// assert!(next_epoch_first_slot(150) <= 63590410);
-/// assert!(next_epoch_first_slot(150) > 63590393);
-/// ```
-// TODO: Design and implement a proper abstraction for slot arithmetic. See https://github.com/pragma-org/amaru/pull/26/files#r1807394364
-pub fn next_epoch_first_slot(current_epoch: u64) -> u64 {
-    (BYRON_TOTAL_SLOTS as u64)
-        + (SHELLEY_EPOCH_LENGTH as u64)
-            * (1 + current_epoch - PREPROD_SHELLEY_TRANSITION_EPOCH as u64)
-}
-
 /// TODO: Ideally, we should either:
 ///
 /// - Have pallas_traverse or a similar API works directly from base objects instead of KeepRaw
--- a/crates/amaru-kernel/src/lib.rs
+++ b/crates/amaru-kernel/src/lib.rs
@@ -484,15 +484,6 @@ pub fn encode_bech32(hrp: &str, payload: &[u8]) -> Result<String, Box<dyn std::e
     Ok(bech32::encode::<bech32::Bech32>(hrp, payload)?)
 }
 
-/// Calculate the epoch number corresponding to a given slot on the PreProd network.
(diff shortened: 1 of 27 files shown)


crates/amaru-kernel/src/network.rs
```diff
@@ -12,9 +12,194 @@
...
+        },
+    },
+];
+
+pub fn preprod_era_history() -> &'static EraHistory {
+    static PREPROD_ERA_HISTORY: OnceLock<EraHistory> = OnceLock::new();
+    PREPROD_ERA_HISTORY.get_or_init(|| EraHistory {
+        eras: PREPROD_ERAS.to_vec(),
+    })
+}
```
KtorZ: Curious 🤔. Intuitively, I'd have just opted for:

```rust
pub static ERA_HISTORY: LazyLock<EraHistory> = LazyLock::new(|| EraHistory { 
  eras: PREPROD_ERAS.to_vec(),
})
```

Doesn't matter much as this code won't last long.

crates/amaru-kernel/src/network.rs
```diff
@@ -67,11 +252,71 @@ impl NetworkName {
...
+
+impl std::error::Error for EraHistoryFileError {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        match self {
+            Self::FileOpenError(err) => Some(err),
+            Self::JsonParseError(err) => Some(err),
+        }
+    }
+}
```
KtorZ:
```suggestion
/// Error type for era history file operations
#[derive(Debug, thiserror::Error)]
pub enum EraHistoryFileError {
    #[error("Failed to open era history file: {0}")]
    FileOpenError(#[from] std::io::Error),
    #[error("Failed to parse era history JSON: {0}")]
    JsonParseError(#[from] serde_json::Error),
}
```

crates/amaru-kernel/src/lib.rs
```diff
@@ -485,15 +485,18 @@ pub fn encode_bech32(hrp: &str, payload: &[u8]) -> Result<String, Box<dyn std::e
 
 /// Calculate the epoch number corresponding to a given slot on the PreProd network.
 // TODO: Design and implement a proper abstraction for slot arithmetic. See https://github.com/pragma-org/amaru/pull/26/files#r1807394364
+#[allow(clippy::unwrap_used)]
 pub fn epoch_from_slot(slot: u64) -> u64 {
-    let shelley_slots = slot - BYRON_TOTAL_SLOTS as u64;
-    (shelley_slots / SHELLEY_EPOCH_LENGTH as u64) + PREPROD_SHELLEY_TRANSITION_EPOCH as u64
+    // FIXME: currently hardwired to preprod network
+    preprod_era_history().slot_to_epoch(slot).unwrap()
```
KtorZ: For completeness, would it make sense to also adjust `relative_slot` and `next_epoch_first_slot` ?

KtorZ (approved):

crates/amaru-kernel/src/lib.rs
jeluard: Also, can it realistically fail? Do we want to propagate the error?

KtorZ: It can if we ever request a slot before the start of the shelley era; Which can totally happen is someone submits a transaction with such a lower bound.

abailly: or a slot past the `EraHistory` horizon, which is the reason for the `unwrap()`

jeluard: Then I think it's worth bubbling the error up. `unwrap` should only be used when there is no way it can happen for soundness reason, and that somehow the type system doesn't help with expressing this.

abailly: Yes, that's what I am working on now

PR #165, opened 2025-03-25
feat: add database transactional support across epoch boundaries

jeluard:
Introduce a `TransactionalContext` abstraction used to perform multiple atomic changes to the underlying `Store`. Introduce `ReadOnlyStore` and `Snapshot` to clarify semantic and not force dummy implementation due to incorrect abstractions. 
Finally `HistoricalStores` breaks the direct link between `Store` and `Snapshot` allowing to manipulate both independently. This also opens potential `Snapshot` implementations leveraging different backends.

# Potential future improvements

Read-only access is kept at the `Store` level. It might make sense to funnel all accesses via the `TransactionContext` so that it potentially helps with implementation of other stores and reduce potential visibility issues related to isolation levels.

# Notes about testing failure scenarios

* delegate to antithesis/simulation testing?
* allow some fault-injection and kill the db?
* rely on traces and kill the db?

## Summary by CodeRabbit

- **New Features**
  - Added Makefile targets for cleaning databases and running end-to-end tests from scratch.
  - Enabled multi-threaded column family support in RocksDB for improved performance.

- **Refactor**
  - Separated historical snapshot access from live database operations for better modularity.
  - Introduced explicit transactional contexts for all database mutations, enhancing atomicity and consistency.
  - Updated trait and struct definitions to support new abstractions for snapshots and transactional operations.
  - Updated ledger stage and state initialization to incorporate historical snapshot management.
  - Refactored import commands to use transactional contexts for all database operations.
  - Enhanced snapshot handling and transaction management in RocksDB implementation for improved stability.
  - Improved store trait implementations for in-memory and RocksDB stores with transactional support.
  - Expanded test infrastructure to use snapshot-based stores for rewards testing.

- **Bug Fixes**
  - Improved management of epoch transitions and snapshot creation, ensuring correct reward calculations and state updates.

- **Chores**
  - Enhanced test configuration and command aliases to increase test coverage and reliability.
  - Adjusted CI workflow to avoid failure on Codecov upload errors and updated cache versioning.


crates/amaru-stores/src/rocksdb/mod.rs
```diff
@@ -337,23 +268,85 @@ fn with_prefix_iterator<
...
-            .map(|bytes| cbor::decode(&bytes))
-            .transpose()
-            .map_err(|err| StoreError::Tip(TipErrorKind::Undecodable(err)))?
-            .ok_or(StoreError::Tip(TipErrorKind::Missing))
+    #[instrument(
+        level = Level::TRACE,
+        skip_all,
+    )]
+    fn reset_fees(&self) -> Result<(), StoreError> {
+        self.with_pots(|mut row| {
```
jeluard, line 353: Those 4 fns should be default fns on the `Transaction` trait. Failed so far due to intricacies with `iter_borrow` .

crates/amaru-stores/src/rocksdb/mod.rs
```diff
@@ -354,18 +284,109 @@ fn with_prefix_iterator<
...
-        Self::for_epoch_with(self.dir.as_path(), epoch)
+impl TransactionalContext<'_> for RocksDBTransactionalContext<'_> {
+    fn commit(self) -> Result<(), StoreError> {
+        let res = self
+            .transaction
+            .commit()
+            .map_err(|err| StoreError::Internal(err.into()));
+        self.db.transaction_committed();
+        res
     }
```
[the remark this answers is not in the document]
KtorZ: While the commitable fixed suggested by the rabbit doesn't seem right, I think the overall remark is? We should ensure the transaction is either dropped or committed; if it's dropped/rolled back; let's raise an `error!`.

crates/amaru-stores/src/rocksdb/mod.rs
KtorZ: I would perhaps more challenge the idea of them being in the store to begin with? I rather keep the store interface minimal, and have those defined as _helpers_ on top.

jeluard: So just regular functions taking a `Transaction` as parameter? And not default trait method? Makes sense.

crates/amaru-stores/src/rocksdb/mod.rs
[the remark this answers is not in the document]
jeluard: `RocksDBTransactionalContext ` for reasons similar to [this one](https://stackoverflow.com/questions/34278607/can-not-move-out-of-type-which-defines-the-drop-trait-e0509) so I went ahead and introduced an `OngoingTransaction` struct that encapsulates the check logic. As `Drop` cascades to all fields of a struct it will be called accordingly.
I also `panic` when an unsupported scenario is detected as it is a developper bug: no `Transaction` should be left open, either `commit` or `rollback` explicitly.

KtorZ (approved):

PR #166, opened 2025-03-26
Transaction Validation Tests

yHSJ:
This PR introduces some initial unit tests for the existing transaction rules.

These unit tests are based on real world blocks that may contain some interesting cases


crates/amaru-ledger/src/rules/transaction/inputs.rs
```diff
@@ -96,3 +96,318 @@ where
...
...c0ffd8799f581cba4ab50bdecca85162f3b8114739bc5ba3aaa6490e2b1d15ad0f9c661a00021833ffd8799f581c25aa4132c7ce7d8f96ee977cd921cba7681891d114d088449d1d63b21a000221a5ffd8799f581c5309fa786856c1262d095b89adf64fe8a5255ad19142c9c537359e411a0001ee15ffff1925dbffff82581d605309fa786856c1262d095b89adf64fe8a5255ad19142c9c537359e411a00a229b4021a00175e18031a043724da081a043724620b5820ee8f2ca9b9a3e8cdc86ab2421fe721753bfd9bb2bfbf483a8e0a35d4b7ab28e20d8182582047a890217e4577ec3e6d5db161a4aa524a5cce3302e389ccb22b5662146f52ab020e81581c5309fa786856c1262d095b89adf64fe8a5255ad19142c9c537359e411082581d605309fa786856c1262d095b89adf64fe8a5255ad19142c9c537359e411a00827b8d111a00370c3f128782582047a890217e4577ec3e6d5db161a4aa524a5cce3302e389ccb22b5662146f52ab00825820b6dcfdc52df91296d060b4fee05c05340c51e89b16ff7033a53a35d2cd7949b0008258205c6d1dba0f74e55143a751d3f92433bcfbaa63d20b909f5dde8246db9f38144c00825820f0b861a084a983537de82d769ddf50876ef1bbb7b504690f45bc65e65d4de8f200825820878296eb7662c6a6e600f48941b3eae1e582bc05aacd0d635fcc96448a5e4e2600825820efd8a48f69e5270cb73b438067e1b566d39758329a45e5d6d36755f2c732618400825820537de728655d2da5fc21e57953a8650b25db8fc84e7dc51d85ebf6b8ea16559601").expect("failed to decode hex");
```
KtorZ: It'd be better to move those into test data files, and rely on some macro like `include_bytes!` or something making use of it to:

1. Ease the readability of the supporting test code. 
2. Make the test vectors easier to find and share with others. 

See for example:

- https://github.com/pragma-org/amaru/tree/main/crates/amaru-consensus/tests/data/headers
- https://github.com/pragma-org/amaru/blob/main/crates/amaru-consensus/src/lib.rs#L51-L64
- https://github.com/pragma-org/amaru/blob/main/crates/amaru-consensus/src/consensus/store.rs#L206-L214

crates/amaru-ledger/src/rules/transaction/inputs.rs
```diff
@@ -96,3 +96,318 @@ where
...
+                                "47a890217e4577ec3e6d5db161a4aa524a5cce3302e389ccb22b5662146f52ab",
+                                2
+                            ),
+                            fake_output!(
+                                "605309fa786856c1262d095b89adf64fe8a5255ad19142c9c537359e41"
+                            ),
+                        ),
+                    ]),
+                },
+            ),
```
KtorZ: It feels to me that these would be better provided as external JSON or CBOR files 😅, next to the serialized transactions.

crates/amaru-ledger/src/lib.rs
```diff
@@ -36,3 +36,43 @@ pub mod rules;
...
+    macro_rules! fake_output {
+        ($address:expr) => {
+            TransactionOutput::PostAlonzo(PostAlonzoTransactionOutput {
+                address: Bytes::from(hex::decode($address).expect("Invalid hex address")),
+                value: Value::Coin(0),
+                datum_option: None,
+                script_ref: None,
+            })
+        };
+    }
```
KtorZ: Not sure to see what benefits does the macro provides here over a function call 🤔 ?

yHSJ: No real difference afaik, I just wanted to make it accessible from outside the `rules`, but I can just make it a public function.

crates/amaru-kernel/src/lib.rs
```diff
@@ -750,4 +750,40 @@ mod test {
...
+                hex::decode("A5A8B29A838CE9525CE6C329C99DC89A31A7D8AE36A844EEF55D7EB9").expect("Failed to decode root")
+            ),
+            (
+                // 17d3497083063ef23d25280a64657a0efa0b53cc369cc5ef430941f5cc91c94e
+                cbor::decode(hex::decode("8458202B13CBA22AAD6A2B775FDA17F85E7BC74A5CEDB21DDE1D2BE14EC23BE2C9B3C2584006883DEDFFF9C6AF6B4422A81D4D6904C8372E4571D8610007473653722CC07841D5FFA7E427818DD6B9323DE629CF452C5EA08FEA90B2E33E577E26AD8BEF0F58209422C0D6AA61D186C40A0D562911311578463F91C26FA56F3FECD8BBEC0E3A1544A1024101").expect("failed to decode bootstrap witness").as_slice()).expect("failed to decode bootstrap witness cbor"),
+                hex::decode("59F44FD32EE319BCEA9A51E7B84D7C4CB86F7B9B12F337F6CA9E9C85").expect("Failed to decode root")
+            ),
+        ];
+
+        cases.into_iter().for_each(|(witness, expected_root)| {
```
KtorZ: Did you mean to create _multiple test cases_ 😄? Note that the [test_case](https://docs.rs/test-case/latest/test_case/) crate is useful for that.

crates/amaru-ledger/src/context/assert.rs
```diff
@@ -237,3 +322,77 @@ impl WitnessSlice for AssertValidationContext {
...
+// Deserialization of Validation Context JSON files
+impl<'de> serde::Deserialize<'de> for AssertValidationContext {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        use serde::de::Error;
+
+        let mut map: serde_json::Map<String, serde_json::Value> =
+            serde::Deserialize::deserialize(deserializer)?;
```
KtorZ: Hmmm. That's one way of doing it, but quite not the idiomatic serde's way which relies on a visitor design pattern. See for example: https://serde.rs/deserialize-struct.html

Although, for _testing_, it's probably sufficient (and much simpler) to use macro annotations on the host structure. I am not sure what the manually written instance provides here over an annotations-derived one?

KtorZ: One thing that is perhaps not obvious with serde, is that serde isn't a JSON-specific serialisation library. It's a framework for writing generic structured codecs, which may be instantiated using various encodings. So, technically, a `Deserialize` instance can be used to deserialise from JSON, but also from CBOR, or MsgPack, or, whatever encoding format is supported. 

So the instance itself shall not explicitly depend on the underlying data being JSON. Otherwise, you're probably better-off using a JSON-specific library (which isn't out of the question here 😅). 

We get little benefits from serde overall, since most of our binary serialisation is done via minicbor, which is a cbor-specific library. But there's a world where we _could have_ relied on serde entirely and decide at call-site which of JSON or CBOR to generate; from a same `Deserialize` instance definition.  

The main reason we don't do this is three folds:

1. Writing generic serde instances by hand is a pain in the a**, and can be quite convoluted since you need to make detour that makes it generic and can't simply _assume_ JSON or CBOR. 
2. We have very concrete serialisation requirements when it comes to Cardano structure. So we can't usually rely on a library to make serialisation choices for us. For example, there are places where we might require an indefinite CBOR array, and can't let that decision be abstracted away.  
3. Its performances are quite terrible (which usually goes hand in hand with highly generic code). So writing specialized CBOR serialisers becomes necessary since they are on our critical execution path.

crates/amaru-ledger/src/rules/transaction/inputs.rs
```diff
@@ -96,3 +96,69 @@ where
...
+            load_preperation_context_from_file("tests/data/transactions/preprod/d731b9832921c0cf9294eea0da2de215d0e9afd36126dc6af9af7e8d6310282a/context.json").expect("failed to load prepreation context from file")
+            ),
+            (cbor::decode(tx_6961d536a1f4d09204d5cfe3cc42949a0e803245fead9a36fad328bf4de9d2f4)
+                .expect("failed to decode tx body cbor"),
+            load_preperation_context_from_file("tests/data/transactions/preprod/6961d536a1f4d09204d5cfe3cc42949a0e803245fead9a36fad328bf4de9d2f4/context.json").expect("failed to load prepreation context from file")
+
+            )
+        ];
+
+        cases.into_iter().for_each(|(tx, ctx)| {
```
KtorZ: See: https://github.com/SundaeSwap-finance/amaru/pull/1#discussion_r2026509087

crates/amaru-ledger/src/rules/transaction/vkey_witness.rs
```diff
@@ -86,3 +86,280 @@ pub fn execute(
...
+    fn invalid_signature_length() {
+        // The following test relies on a real tranasction found on Preprod (44762542f8e2f66da2fa0d4fdf2eb82cc1d24ae689c1d19ffd7e57d038f50bca)
+        // The witness set is modified to include an invalid signature (due to length)
+        let mut ctx: AssertValidationContext = load_validation_context_from_file("tests/data/transactions/preprod/44762542f8e2f66da2fa0d4fdf2eb82cc1d24ae689c1d19ffd7e57d038f50bca/invalid-signature-length/context.json").expect("failed to laod validation context");
+        let tx_bytes = include_bytes!("../../../tests/data/transactions/preprod/44762542f8e2f66da2fa0d4fdf2eb82cc1d24ae689c1d19ffd7e57d038f50bca/tx.cbor");
+        let transaction_body: KeepRaw<'_, MintedTransactionBody<'_>> =
+            cbor::decode(tx_bytes).expect("Failed to cbor decode transaction body");
+        let witness_set_bytes = include_bytes!("../../../tests/data/transactions/preprod/44762542f8e2f66da2fa0d4fdf2eb82cc1d24ae689c1d19ffd7e57d038f50bca/invalid-signature-length/witness.cbor");
+        let witness_set: WitnessSet =
+            cbor::decode(witness_set_bytes).expect("Failed to cbor decode witness set");
```
KtorZ: So, this is where a cute macro would be useful to avoid repeating the same boilerplate code over and over.

crates/amaru-ledger/src/context/assert.rs
KtorZ: See: https://github.com/SundaeSwap-finance/amaru/pull/2

KtorZ: An alternative could also to just entirely rely on CBOR for test vectors also. For non-standard elements (e.g. the validation context), we can write down a bit of CDDL to document the expected schema.

crates/amaru-ledger/src/lib.rs
```diff
@@ -36,3 +36,212 @@ pub mod rules;
...
+
+    pub(crate) fn with_tracing<F, R>(test_fn: F) -> R
+    where
+        F: FnOnce(&TestingTraceCollector) -> R,
+    {
+        let mut collector = TestingTraceCollector::new();
+        collector.clear();
+        let collector_clone = collector.clone();
+        let format_layer = fmt::layer()
+            .with_writer(move || collector_clone.clone())
```
KtorZ, line 166: What is going on here 😅

crates/amaru-ledger/src/lib.rs
```diff
@@ -36,3 +36,212 @@ pub mod rules;
...
+    pub(crate) use include_transaction_body;
+    pub(crate) use include_witness_set;
+
+    // -----------------------------------------------------------------------------
+    // Tracing for Tests
+    // -----------------------------------------------------------------------------
+
+    #[derive(Clone)]
+    pub(crate) struct TestingTraceCollector {
+        pub lines: Arc<Mutex<Vec<String>>>,
```
KtorZ, line 127: Is the mutex actually needed 🤔 ?

crates/amaru-ledger/src/lib.rs
yHSJ: I forgot I didn't want to commit this 😅 

I was having issues with `with_writer` and ownership (as you can see)

PR #171, opened 2025-03-28
chore: use newtype pattern instead of raw types for core concepts

abailly:
This PR provides 2 things:

* An _Engineering Decision Records_ to discuss general use of `newtype` pattern in Amaru codebase
* An implementation of the pattern for type `Slot` within `slot-arithmetic` crate as a first example. 

I have not attempted to implement it for other types where it would make sense (eg. `Epoch` ) nor to replace other definitions of `type Slot` 

## Summary by CodeRabbit

- **Refactor**
	- Enhanced the system’s internal logic by transitioning slot values from primitive integers to a dedicated, type-safe representation. This update improves consistency and robustness in calculations and state management.
- **Documentation**
	- Introduced a new architectural document outlining the approach to reduce reliance on primitive types, establishing clearer guidelines for domain-specific data handling.

These improvements bolster system reliability and provide a clearer foundation for future enhancements.

### DIFF
--- a/crates/amaru-ledger/src/state.rs
+++ b/crates/amaru-ledger/src/state.rs
@@ -180,7 +180,7 @@ impl<S: Store> State<S> {
         let start_slot = now_stable.anchor.0.slot_or_default();
         let current_epoch = self
             .era_history
-            .slot_to_epoch(start_slot)
+            .slot_to_epoch(start_slot.into())
             .map_err(|e| StateError::ErrorComputingEpoch(start_slot, e))?;
 
         // Note: the volatile sequence may contain points belonging to two epochs. We diligently
@@ -200,7 +200,7 @@ impl<S: Store> State<S> {
         let anchor_slot = now_stable.anchor.0.slot_or_default();
         let epoch = self
             .era_history
-            .slot_to_epoch(anchor_slot)
+            .slot_to_epoch(anchor_slot.into())
             .map_err(|e| StateError::ErrorComputingEpoch(anchor_slot, e))?;
         let StoreUpdate {
             point: stable_point,
@@ -280,10 +280,10 @@ impl<S: Store> State<S> {
         let next_state_slot = next_state.anchor.0.slot_or_default();
         let relative_slot = self
             .era_history
-            .slot_in_epoch(next_state_slot)
+            .slot_in_epoch(next_state_slot.into())
             .map_err(|e| StateError::ErrorComputingEpoch(next_state_slot, e))?;
 
-        if self.rewards_summary.is_none() && relative_slot >= STABILITY_WINDOW as u64 {
+        if self.rewards_summary.is_none() && relative_slot >= (STABILITY_WINDOW as u64).into() {
             self.rewards_summary = Some(self.compute_rewards()?);
         }
 
@@ -379,7 +379,7 @@ impl HasStakeDistribution for StakeDistributionView {
     #[allow(clippy::unwrap_used)]
     fn get_pool(&self, slot: Slot, pool: &PoolId) -> Option<PoolSummary> {
         let view = self.view.lock().unwrap();
-        let epoch = self.era_history.slot_to_epoch(slot).ok()? - 2;
+        let epoch = self.era_history.slot_to_epoch(slot.into()).ok()? - 2;
         view.iter().find(|s| s.epoch == epoch).and_then(|s| {
             s.pools.get(pool).map(|st| PoolSummary {
                 vrf: st.parameters.vrf,
--- a/crates/amaru-ledger/src/state.rs
+++ b/crates/amaru-ledger/src/state.rs
@@ -180,7 +180,7 @@ impl<S: Store> State<S> {
         let start_slot = now_stable.anchor.0.slot_or_default();
         let current_epoch = self
             .era_history
-            .slot_to_epoch(start_slot.into())
+            .slot_to_epoch(slot_arithmetic::Slot::new(start_slot))
             .map_err(|e| StateError::ErrorComputingEpoch(start_slot, e))?;
 
         // Note: the volatile sequence may contain points belonging to two epochs. We diligently
@@ -200,7 +200,7 @@ impl<S: Store> State<S> {
         let anchor_slot = now_stable.anchor.0.slot_or_default();
         let epoch = self
             .era_history
-            .slot_to_epoch(anchor_slot.into())
+            .slot_to_epoch(slot_arithmetic::Slot::new(anchor_slot))
             .map_err(|e| StateError::ErrorComputingEpoch(anchor_slot, e))?;
         let StoreUpdate {
             point: stable_point,
@@ -280,10 +280,12 @@ impl<S: Store> State<S> {
         let next_state_slot = next_state.anchor.0.slot_or_default();
         let relative_slot = self
             .era_history
-            .slot_in_epoch(next_state_slot.into())
+            .slot_in_epoch(slot_arithmetic::Slot::new(next_state_slot))
             .map_err(|e| StateError::ErrorComputingEpoch(next_state_slot, e))?;
 
-        if self.rewards_summary.is_none() && relative_slot >= (STABILITY_WINDOW as u64).into() {
+        if self.rewards_summary.is_none()
+            && relative_slot >= slot_arithmetic::Slot::new(STABILITY_WINDOW as u64)
+        {
             self.rewards_summary = Some(self.compute_rewards()?);
         }
 
@@ -379,7 +381,11 @@ impl HasStakeDistribution for StakeDistributionView {
     #[allow(clippy::unwrap_used)]
     fn get_pool(&self, slot: Slot, pool: &PoolId) -> Option<PoolSummary> {
         let view = self.view.lock().unwrap();
-        let epoch = self.era_history.slot_to_epoch(slot.into()).ok()? - 2;
+        let epoch = self
+            .era_history
+            .slot_to_epoch(slot_arithmetic::Slot::new(slot))
+            .ok()?
+            - 2;
         view.iter().find(|s| s.epoch == epoch).and_then(|s| {
             s.pools.get(pool).map(|st| PoolSummary {
                 vrf: st.parameters.vrf,
--- a/crates/amaru-ledger/src/state.rs
+++ b/crates/amaru-ledger/src/state.rs
@@ -180,7 +180,7 @@ impl<S: Store> State<S> {
         let start_slot = now_stable.anchor.0.slot_or_default();
         let current_epoch = self
             .era_history
-            .slot_to_epoch(slot_arithmetic::Slot::new(start_slot))
+            .slot_to_epoch(From::from(start_slot))
             .map_err(|e| StateError::ErrorComputingEpoch(start_slot, e))?;
 
         // Note: the volatile sequence may contain points belonging to two epochs. We diligently
@@ -200,7 +200,7 @@ impl<S: Store> State<S> {
         let anchor_slot = now_stable.anchor.0.slot_or_default();
         let epoch = self
             .era_history
-            .slot_to_epoch(slot_arithmetic::Slot::new(anchor_slot))
+            .slot_to_epoch(From::from(anchor_slot))
             .map_err(|e| StateError::ErrorComputingEpoch(anchor_slot, e))?;
         let StoreUpdate {
             point: stable_point,
@@ -280,12 +280,10 @@ impl<S: Store> State<S> {
         let next_state_slot = next_state.anchor.0.slot_or_default();
         let relative_slot = self
             .era_history
-            .slot_in_epoch(slot_arithmetic::Slot::new(next_state_slot))
+            .slot_in_epoch(From::from(next_state_slot))
             .map_err(|e| StateError::ErrorComputingEpoch(next_state_slot, e))?;
 
-        if self.rewards_summary.is_none()
-            && relative_slot >= slot_arithmetic::Slot::new(STABILITY_WINDOW as u64)
-        {
+        if self.rewards_summary.is_none() && relative_slot >= From::from(STABILITY_WINDOW as u64) {
             self.rewards_summary = Some(self.compute_rewards()?);
         }
 
@@ -381,11 +379,7 @@ impl HasStakeDistribution for StakeDistributionView {
     #[allow(clippy::unwrap_used)]
     fn get_pool(&self, slot: Slot, pool: &PoolId) -> Option<PoolSummary> {
         let view = self.view.lock().unwrap();
-        let epoch = self
-            .era_history
-            .slot_to_epoch(slot_arithmetic::Slot::new(slot))
-            .ok()?
-            - 2;
+        let epoch = self.era_history.slot_to_epoch(From::from(slot)).ok()? - 2;
         view.iter().find(|s| s.epoch == epoch).and_then(|s| {
             s.pools.get(pool).map(|st| PoolSummary {
                 vrf: st.parameters.vrf,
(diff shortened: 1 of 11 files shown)


crates/slot-arithmetic/src/lib.rs
```diff
@@ -12,10 +12,31 @@
...
+use derive_more::{Add, From, Sub};
 use minicbor::{Decode, Decoder, Encode};
 use serde::{Deserialize, Serialize};
 
-type Slot = u64;
+#[derive(Clone, Debug, Copy, PartialEq, From, Add, Sub, PartialOrd, Eq, Serialize, Deserialize)]
+#[repr(transparent)]
+pub struct Slot {
+    pub slot: u64,
+}
```
rkuhn:
```suggestion
pub struct Slot(64);
```
The wrapped value should not be directly accessible from the outside. This is more important when guaranteeing additional constraints, in which case no external mutation can be allowed (but a `.get()` or `.as_u64()` still makes sense).

crates/slot-arithmetic/src/lib.rs
```diff
@@ -12,10 +12,31 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use derive_more::{Add, From, Sub};
 use minicbor::{Decode, Decoder, Encode};
 use serde::{Deserialize, Serialize};
 
-type Slot = u64;
+#[derive(Clone, Debug, Copy, PartialEq, From, Add, Sub, PartialOrd, Eq, Serialize, Deserialize)]
```
rkuhn: I’d be skeptical regarding `Add` and `Sub` since these can overflow in release builds, and that doesn’t make sense to me for slot numbers. Do we really need these traits here?

crates/slot-arithmetic/src/lib.rs
```diff
@@ -215,51 +236,55 @@ impl EraHistory {
...
             if era.start.time_ms > time {
                 return Err(TimeHorizonError::InvalidEraHistory);
             }
             if era.end.time_ms >= time {
                 let time_elapsed = time - era.start.time_ms;
                 let slots_elapsed = time_elapsed / era.params.slot_length;
-                let slot = era.start.slot + slots_elapsed;
+                let slot = era.start.slot
+                    + Slot {
+                        slot: slots_elapsed,
```
rkuhn: This looks fishy: `slots_elapsed` is not a slot number, it is a distance in slot space. This is (and implies) the issues that drive the distinction between `Duration` and `Instant`, where two instants cannot be added, only subtracted (yielding a duration, of course).

I’d solve this by adding `Slot::offset_by(self, offset: u64) -> Self`, which should also validate that nothing wrong happens. Perhaps the offset should be restricted to u32, even.

engineering-decision-records/009-avoid-primitive-obsession.md
```diff
@@ -0,0 +1,59 @@
...
+1. **Creating domain-specific types**: Instead of using primitive types directly, we will create custom types that represent domain concepts, even when they wrap a single primitive value.
+
+2. **Using newtypes**: In Rust, we'll use the [newtype pattern](https://www.worthe-it.co.za/blog/2020-10-31-newtype-pattern-in-rust.html) to create zero-cost abstractions over primitive types. This provides type safety without runtime overhead.
+
+3. **Implementing appropriate traits**: Our domain-specific types will implement traits that make sense for their semantics, such as:
+   - `PartialEq`, `Eq`, `PartialOrd`, `Ord` for comparison
+   - `Display`, `Debug` for formatting
+   - `Serialize`, `Deserialize` for serialization
+   - [derive-more]() could be used for non-standard traits
+   - CBOR `Encode` and `Encode` need to be written manually
```
rkuhn: `serde` can automatically see through newtypes, perhaps we can teach minicbor to do the same?

engineering-decision-records/009-avoid-primitive-obsession.md
```diff
@@ -0,0 +1,59 @@
...
+
+4. **Encapsulating validation logic**: Domain-specific types will encapsulate validation logic to ensure they always represent valid values.
```
rkuhn, line 34: This implies that no direct access to mutate the wrapped value can be provided! Note how this also precludes deriving the arithmetic traits, for example.

engineering-decision-records/009-avoid-primitive-obsession.md
```diff
@@ -0,0 +1,59 @@
...
+- **Improved type safety**: The compiler will catch type errors that would otherwise manifest as runtime bugs.
+- **Better domain modeling**: Domain concepts are explicitly represented in the code, making the model clearer.
+- **Centralized validation**: Validation logic is centralized in the type definition, reducing duplication.
+- **Enhanced readability**: Code becomes more self-documenting as types clearly indicate their purpose.
+- **Easier refactoring**: Changes to domain concepts are localized to their type definitions.
+
+### Negative Outcomes
+
+- **Increased boilerplate**: Creating custom types requires more code than using primitives directly.
+- **Learning curve**: New contributors need to learn the domain-specific types rather than working with familiar primitives.
```
rkuhn: Assuming that people use IDEs this probably isn’t a negative point: the newtype allows us to provide targeted documentation right at your finger tips where a primitive type serves only to confuse or frustrate the programmer.

IOW: newtype allows newcomers to learn, smoothening the learning curve.

engineering-decision-records/009-avoid-primitive-obsession.md
```diff
@@ -0,0 +1,59 @@
...
+- **Potential performance overhead**: In some cases, wrapping primitives might introduce minor performance overhead (though this is usually negligible with Rust's zero-cost abstractions). Wherever possible this should be avoided by using the `#[repr(transparent)]` annotation to ensure compiled representation does not carry with it
```
rkuhn: This should instead be covered by presenting an example including `#[repr(transparent)]` above, which solves the problem. It wouldn’t be wise to trigger the nerd brain into thinking that newtypes “are not good enough for performance”. Even if there were a minuscule effect, we’re allocating all over the place, so it’d be quite some ways until we get to this point in the priority list.

engineering-decision-records/009-avoid-primitive-obsession.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Discussion points
+
+- We should establish guidelines for when a new type is warranted versus when using a primitive is acceptable.
+- We need to balance the benefits of domain-specific types with the additional complexity they introduce.
+- We should consider creating macros or helper functions to reduce the boilerplate associated with creating new types.
+- Integration with existing libraries that expect primitive types, or that already use their custom `newtypes` needs careful consideration.
```
rkuhn: This is best solved by a method like `Slot::as_u64(self) -> u64`. It should _not_ be tackled using `impl From`.

engineering-decision-records/009-avoid-primitive-obsession.md
```diff
@@ -0,0 +1,59 @@
...
+
+- **Increased boilerplate**: Creating custom types requires more code than using primitives directly.
+- **Learning curve**: New contributors need to learn the domain-specific types rather than working with familiar primitives.
+- **Potential performance overhead**: In some cases, wrapping primitives might introduce minor performance overhead (though this is usually negligible with Rust's zero-cost abstractions). Wherever possible this should be avoided by using the `#[repr(transparent)]` annotation to ensure compiled representation does not carry with it
+
+## Discussion points
+
+- We should establish guidelines for when a new type is warranted versus when using a primitive is acceptable.
+- We need to balance the benefits of domain-specific types with the additional complexity they introduce.
+- We should consider creating macros or helper functions to reduce the boilerplate associated with creating new types.
```
rkuhn: I think those bases are covered quite well already, perhaps a little something for minicbor.

engineering-decision-records/009-avoid-primitive-obsession.md
```diff
@@ -0,0 +1,59 @@
...
+### Negative Outcomes
+
+- **Increased boilerplate**: Creating custom types requires more code than using primitives directly.
+- **Learning curve**: New contributors need to learn the domain-specific types rather than working with familiar primitives.
+- **Potential performance overhead**: In some cases, wrapping primitives might introduce minor performance overhead (though this is usually negligible with Rust's zero-cost abstractions). Wherever possible this should be avoided by using the `#[repr(transparent)]` annotation to ensure compiled representation does not carry with it
+
+## Discussion points
+
+- We should establish guidelines for when a new type is warranted versus when using a primitive is acceptable.
+- We need to balance the benefits of domain-specific types with the additional complexity they introduce.
```
rkuhn: To my knowledge the only added complexity is the need to import them into the current module, which is tiny.

engineering-decision-records/009-avoid-primitive-obsession.md
```diff
@@ -0,0 +1,59 @@
...
+
+### Negative Outcomes
+
+- **Increased boilerplate**: Creating custom types requires more code than using primitives directly.
+- **Learning curve**: New contributors need to learn the domain-specific types rather than working with familiar primitives.
+- **Potential performance overhead**: In some cases, wrapping primitives might introduce minor performance overhead (though this is usually negligible with Rust's zero-cost abstractions). Wherever possible this should be avoided by using the `#[repr(transparent)]` annotation to ensure compiled representation does not carry with it
+
+## Discussion points
+
+- We should establish guidelines for when a new type is warranted versus when using a primitive is acceptable.
```
rkuhn, line 131: my proposal: A newtype is required for a type that semantically cannot support all operations offered by the primitive Rust type (which also includes permitting only a subset of the possible values). A newtype is recommended where the primitive type has a particular and well-established meaning that exists not only in the local code scope — because this implies that uses of the primitive type will overlap with other semantics that shall remain distinct. A newtype will typically not be used for external inputs that haven’t yet been parsed or validated (like `Vec<u8>` from files or network sockets).

crates/amaru-stores/src/rocksdb/mod.rs
```diff
@@ -423,7 +423,7 @@ impl Store for RocksDB {
                 dreps::tick(&batch, voting_dreps, {
                     let slot = point.slot_or_default();
                     self.era_history
-                        .slot_to_epoch(slot)
+                        .slot_to_epoch(slot.into())
```
rkuhn: This is why I don’t like `impl From` for these types: on github it is quite unclear what the conversion is here.

(I know what it does in this case, the point I want to make is more general)

rkuhn (commented):
Great initiative @abailly! Out future selves will thank ourselves for not using primitive types everywhere.

crates/slot-arithmetic/src/lib.rs
abailly: Agreed, just happened to use this a quick fix for another issue but it's solvable, I will revert to an `enum`

crates/slot-arithmetic/src/lib.rs
abailly: _really_ probably not, it makes some expressions in the `EraHistory` impl leaner. 
I will see how it looks if I remove them and replace them

engineering-decision-records/009-avoid-primitive-obsession.md
abailly: I will add a thorough example, good idea

crates/slot-arithmetic/src/lib.rs
rkuhn: I guess you meant tuple-struct, right?

engineering-decision-records/009-avoid-primitive-obsession.md
KtorZ: https://docs.rs/minicbor-derive/latest/minicbor_derive/index.html#cbortransparent ?

engineering-decision-records/009-avoid-primitive-obsession.md
```diff
@@ -0,0 +1,131 @@
...
+1. **Creating domain-specific types**: Instead of using primitive types directly, we will create custom types that represent domain concepts, even when they wrap a single primitive value.
+
+2. **Using newtypes**: In Rust, we'll use the [newtype pattern](https://www.worthe-it.co.za/blog/2020-10-31-newtype-pattern-in-rust.html) to create zero-cost abstractions over primitive types. This provides type safety without runtime overhead.
+
+3. **Implementing appropriate traits**: Our domain-specific types will implement traits that make sense for their semantics, such as:
+   - `PartialEq`, `Eq`, `PartialOrd`, `Ord` for comparison
+   - `Display`, `Debug` for formatting
+   - `Serialize`, `Deserialize` for serialization
+   - [derive-more]() could be used for non-standard traits
+   - CBOR `Encode` and `Encode` need to be written manually
```
KtorZ:
```suggestion
   - CBOR `Encode` and `Decode` need to be written manually
```

engineering-decision-records/009-avoid-primitive-obsession.md
```diff
@@ -0,0 +1,131 @@
...
+
+4. **Encapsulating validation logic**: Domain-specific types will encapsulate validation logic to ensure they always represent valid values.
+
+5. **Providing clear conversion methods**: When conversion between types is necessary, we'll provide explicit methods rather than relying on implicit conversions.
```
KtorZ: Not sure if I read this right but, I strongly dislike having methods like `new(u64) -> Self` and `to_u64(self) -> u64` in there; We should embrace `From<u64> for Slot` and `From<Slot> for u64`, as well as a couple of others `From<usize> for Slot` and `From<u32> for Slot`. 

Otherwise, I'd argue it defies some of goals we're trying to solve w.r.t readability and maintainability. I rather advocate, however, for avoiding `.into()` altogether and prefer the explicit version `Slot::from(...)` or `u64::from(...)` for safe coercions, and their `try_` equivalent for partial ones.

abailly: I often find with implicit conversions, eg. the affordance provided by `From`, hard to discover, but perhaps that's only me.

abailly: I nevertheless followed your suggestion and updated the proposal

crates/amaru-ledger/src/state.rs
```diff
@@ -180,7 +180,7 @@ impl<S: Store> State<S> {
         let start_slot = now_stable.anchor.0.slot_or_default();
```
jeluard, line 180: I would expect `slot_or_default` to return a `Slot` ? Or I guess that would be the next step?

crates/slot-arithmetic/src/lib.rs
```diff
@@ -15,7 +15,60 @@
...
+pub struct Slot(u64);
+
+#[derive(Debug, PartialEq, Eq, thiserror::Error)]
+pub enum SlotArithmeticError {
+    #[error("slot arithmetic overflow, substracting {1} from {0}")]
+    Underflow(u64, u64),
+}
+
+impl Slot {
+    pub const fn new(slot: u64) -> Self {
```
jeluard, line 29: Looks like `Slot(1)`, `Slot::new(1)` and `From::from(1)` are equivalent? Maybe we could align on `Slot(1)` that looks more idiomatic for those cases then?
I would expect most of those to disappear once we introduce the new `Slot` in most places.

crates/amaru-ledger/src/state.rs
abailly: Yes, I did not want the change to leak to far outside the `slot_arithmetic`

crates/slot-arithmetic/src/lib.rs
abailly: Ah, I forgot to delete `new`. `Slot(1)` is only possible locally as the "field" is private, so one must use `From`

jeluard (approved):

PR #173, opened 2025-03-28
Calculate DRep & SPO voting stake distribution

KtorZ:
### Entities & lifecycle

Cardano's governance spreads responsibilities over three groups: stake pool operators (SPOs), delegate representatives (DReps) and a constitutional committee. Both SPOs and DReps have an associated stake as voting power. It corresponds (roughly, we'll get into the subtleties in a moment) to the sum of the stake delegated to them from delegators at an epoch boundary.

Votes tally happens _at the beginning of an epoch_, after rewards calculation, snapshots taken and retired stake pools pruned. Because the construction of the stake distribution is a potentially expensive operation, the stake distribution of the epoch that just ended cannot be readily available at an epoch boundary. For this reason, the stake distribution is computed incrementally during the next epoch and tally/ratification always happen with one epoch of delay (e.g. the transition between epoch e+1 and e+2 uses the voting stake distribution from `e` for the tally).

Summarised in a little diagram:

```
                                    Pruning retired        Computing rewards[^1] using:
                                    pools                  │ - snapshot(e + 2) for
                                    │                      │     - pool performances
                    Stake is        │                      │     - treasury & reserves
                    delegated       │                      │ - snapshot(e) for:
                    │               │                      │     - stake distribution
                    │               │                      │     - pool parameters
                    │               │Using snapshot(e)     │
                    │               │for leader schedule   │                  Distributing rewards
                    │               ││                     │                  earned from (e)
                    │               ││                     │                  │
    snapshot(e)     │ snapshot(e+1) ││     snapshot(e + 2) │                  │snapshot(e + 3)
              ╽     ╽             ╽ ╽╽                   ╽ ╽                  ╽╽
━━━━━━━━━━━━╸╸╸╋━██━██━██━██━██━╸╸╸╋╸╸╸━██━██━██━██━██━╸╸╸╋╸╸╸━██━██━██━██━██━╸╸╋╸╸╸━██━██━██━>
     e                e + 1          ╿      e + 2 ╿         ╿       e + 3             e + 4
                                     │            │         │
                                     │            Cast vote │
                                     │                      │
                                     │                      Ratifying proposals
                                     │                      using voting power
                                     │                      of (e + 1)
                                     │
                                     Computing voting power for
                                     (e + 1) using state from
                                     beginning of (e + 2)
```

As a reminder, each step in the diagram actually happens on every epoch boundaries, but we only represent a single flow on the diagram. Technically, pools pruning happens at the beginning of every epoch, rewards are calculated every epoch, etc... it's an endless cycle of calculations referring to past states.

##### Consequences:

1. In addition to tracking the stake distribution of stake pools, we must also track the stake distribution of DReps. This corresponds to extra work to perform while computing the stake distribution, although much of this work can be mutualized with the existing stake distribution construction for SPOs. Plus, since the tally is deferred by an epoch, this means we can also construct the stake distribution asynchronously in a thread during the epoch while processing ongoing blocks.

2. The stake distribution for voting happens *after* retired pools are pruned, although this steps happens after the database snapshot occurs. This means that in order to maintain consistency, we must replay the pool ticking steps during the asynchronous stake distribution calculation, so that we can identify pools that retired at the epoch transition. Their stake is therefore always excluded from the voting round of the epoch prior to their retirement.

### Deposits

To prevent spam, governance proposals require a (substantial) deposit fixed by the protocol. The deposit is returned once the proposal expires, is ratified or is rejected. Interestingly, the value of those deposits also counts towards the active stake of both SPOs and DReps. This is unlike the rewards/consensus pools' stake distribution which doesn't include any sort of deposits (neither from proposals, nor from credentials). More precisely, the deposit is virtually credited to the return address defined in the proposal. If that stake address is itself delegated to a pool or a drep, it is counted as part of their respective voting stake.

##### Consequences:

1. In our pools stake distribution, we must track a secondary value in addition to the rewards/consensus stake distribution which we call `voting_stake`. It is sensibly equivalent to the `stake`, except that it may contain the extra deposits when relevant.

2. The deposits themselves aren't trivial to compute as we must first list out the _still valid_ proposals. We now perform this computation as part of the governance summary, also computing the DRep state mapping dreps to their expiry epoch.

3. Proposals that naturally expire must return their deposit to their corresponding reward account. In case the reward account doesn't or no-longer exists at the moment of the refund, the money is sent to the treasury instead.

## Summary by CodeRabbit

- **New Features**
	- Enhanced governance and delegation handling now supports additional stake addresses and new delegate statuses (e.g. abstain, no confidence).
	- Introduced refund processing for reward accounts, ensuring deposits are returned more efficiently.

- **Refactor**
	- Simplified data structures for delegates and proposals for a more streamlined and maintainable system.
	- Updated proposal handling to automatically process expiries during epoch transitions.

- **Chores**
	- Improved caching and query mechanisms to ensure optimised, up-to-date performance.


crates/amaru-ledger/src/summary/governance.rs
```diff
@@ -76,111 +104,44 @@ impl DRepsSummary {
...
-        s.serialize_field("delegations", &delegations)?;
-
-        let mut dreps = BTreeMap::new();
-        self.dreps.iter().for_each(|(drep, st)| {
-            if let Some(id) = into_drep_id(drep) {
-                dreps.insert(id, st);
-            }
-        });
-        s.serialize_field("dreps", &dreps)?;
+        let default_drep_state = || DRepState {
```
jeluard, line 129: Is this closure to avoid the clone?

jeluard (approved):

crates/amaru-ledger/src/summary/governance.rs
KtorZ: More like to avoid repeating the same code twice. Ideally, I'd just `DRepState::default()`. Not sure why it stayed as a local closure here.

PR #174, opened 2025-03-28
chore: refactor stages

jeluard:
Centralize `gasket` in `amaru` crate. Also refactor all stages in `amaru/stages`.


crates/amaru/src/stages/mod.rs
```diff
@@ -51,14 +53,27 @@ pub struct Config {
     pub network: NetworkName,
 }
 
+/// A session with a peer, including the peer itself and a client to communicate with it.
+#[derive(Clone)]
+pub struct PeerSession {
```
abailly, line 58: Is this really the right place for this? Seems to me this is a network-centric concern that perhaps pertains to its own module?

abailly (commented):
LGTM. As discussed on Discord, stages might be useful in their own crate to better decouple their definition as stages in a pipeline of validation/selection and actual implementation in `amaru` proper or `amaru_sim`

crates/amaru/src/stages/mod.rs
jeluard: Good question. It would maybe make sense to have a separate networking crate for this indeed.

PR #178, opened 2025-04-03
chore: added more errors

jeluard:
Make sure errors are propagated and remove some `panic`.

## Summary by CodeRabbit

- **Refactor**
	- Enhanced block processing and rollback event handling with improved error propagation and logging for better traceability.
- **Bug Fixes**
	- Refined transaction input validation to provide clearer, more descriptive error messages, reducing abrupt failures.
- **Chores**
	- Increased overall system robustness by exposing additional module functionality and enforcing stricter thread safety.

### DIFF
--- a/crates/amaru-ledger/src/context.rs
+++ b/crates/amaru-ledger/src/context.rs
@@ -47,7 +47,7 @@ pub trait PreparationContext<'a>:
 // Errors
 // -------------------------------------------------------------------------------------------------
 
-#[derive(thiserror::Error, Debug)]
+#[derive(thiserror::Error, Debug, Clone)]
 pub enum RegisterError<ROLE, K> {
     #[error("already registered entity: {0:?}")] // TODO: Use Display
     AlreadyRegistered(PhantomData<ROLE>, K),
@@ -61,7 +61,7 @@ impl<ROLE, K: fmt::Debug> From<diff_bind::RegisterError<K>> for RegisterError<RO
     }
 }
 
-#[derive(thiserror::Error, Debug)]
+#[derive(thiserror::Error, Debug, Clone)]
 pub enum UnregisterError<ROLE, K> {
     #[error("unknown entity: {0:?}")] // TODO: Use Display
     Unknown(PhantomData<ROLE>, K),
@@ -73,7 +73,7 @@ impl<ROLE, K: fmt::Debug> From<diff_bind::BindError<K>> for UnregisterError<ROLE
     }
 }
 
-#[derive(thiserror::Error, Debug)]
+#[derive(thiserror::Error, Debug, Clone)]
 pub enum DelegateError<S, T> {
     #[error("unknown source entity: {0:?}")] // TODO: Use Display
     UnknownSource(S),
@@ -88,7 +88,7 @@ impl<S: fmt::Debug, T: fmt::Debug> From<diff_bind::BindError<S>> for DelegateErr
     }
 }
 
-#[derive(thiserror::Error, Debug)]
+#[derive(thiserror::Error, Debug, Clone)]
 pub enum UpdateError<S> {
     #[error("unknown source entity: {0:?}")] // TODO: Use Display
     UnknownSource(S),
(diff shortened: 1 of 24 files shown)


crates/amaru-ledger/src/context.rs
```diff
@@ -47,7 +47,7 @@ pub trait PreparationContext<'a>:
 // Errors
 // -------------------------------------------------------------------------------------------------
 
-#[derive(thiserror::Error, Debug)]
+#[derive(thiserror::Error, Debug, Clone)]
```
KtorZ: Now that's suspicious e_e. 

I purposefully kept that one out because I don't think we should ever need to clone errors?

crates/amaru-ledger/src/rules/transaction/inputs.rs
```diff
@@ -90,7 +92,7 @@ where
                     };
                 };
             }
-            None => unimplemented!("failed to lookup input: {input:?}"),
+            None => Err(InvalidInputs::UnknownInput)?,
```
KtorZ: The error lacks a bit of context, we should at least provides the input's reference.

crates/amaru/src/stages/consensus/chain_forward.rs
```diff
@@ -87,6 +87,15 @@ impl gasket::framework::Worker<ForwardStage> for Worker {
 
                 Ok(())
             }
+            BlockValidationResult::BlockValidationFailed(invalid_block) => {
+                warn!(
```
KtorZ:
```suggestion
                trace!(
```

crates/amaru/src/stages/ledger.rs
```diff
@@ -93,13 +93,16 @@ impl<S: Store + Send> Stage<S> {
...
             block,
-        )
-        .unwrap_or_else(|e| panic!("Failed to validate block: {:?}", e))
-        .anchor(&point, issuer);
+        ) {
+            Ok(validation_result) => validation_result.anchor(&point, issuer),
+            Err(e) => {
+                return BlockValidationResult::BlockValidationFailed(e);
+            }
+        };
```
KtorZ:
```suggestion
        ) 
        .map_err(BlockValidationResult::BlockValidationFailed)?
        .anchor(&point, issuer);
```

jeluard: Can't `?` here, this function doesn't return `Result` not `Option`

crates/amaru-ledger/src/rules/transaction/inputs.rs
jeluard: 👍

crates/amaru-ledger/src/context.rs
jeluard: This is a side-effect of `InvalidBlock` being used in `BlockValidationResult`, itself passed across `gasket` stages that introduces this requirement.
The alternative would be to not pass `InvalidBlock` but a `Clone` friendly equivalent, now `InvalidBlock` is pretty complex and it looks useful to benefit from it down the line.

crates/amaru/src/stages/ledger.rs
KtorZ: Ho. That sucks.

crates/amaru-ledger/src/context.rs
jeluard: Another alternative would be to keep `Clone` frre the API and wrap the call at the `gasket` integration point. The fully structured error would be lost across `gasket` stages (e.g. stringified) but still be retained at the API level.

jeluard: I simplified things a bit, your point should be addressed now.

crates/amaru/src/stages/ledger.rs
jeluard: Refactored so that it's better handled.

crates/amaru-ledger/src/rules/transaction/inputs.rs
```diff
@@ -18,6 +18,8 @@ use thiserror::Error;
 
 #[derive(Debug, Error)]
 pub enum InvalidInputs {
+    #[error("Unknown input")]
```
KtorZ:
```suggestion
    #[error("Unknown input: {0.index}#{0.transaction_id}")]
```

KtorZ: Just noticed that the `NonDisjointRefInputs` below is wrong; it says:

```
                format!("{}#{}", input.transaction_id, input.transaction_id)
```

PR #179, opened 2025-04-04
feat: document ledger validation design decision.

KtorZ:


KtorZ:
@coderabbitai are you capable of remembering the design described in this document as well as decisions from other EDR and enhance your future suggestions using these considerations?

engineering-decision-records/010-ledger-validation-context.md
```diff
@@ -0,0 +1,321 @@
...
+```
+
+Where:
+
+- `st` designates the entire ledger state or a subset.
+- `cmd` designates a block, a transaction or part of a transaction.
+
+Cardano transactions are, in fact, akin to a collection of primitive operations that affect the state and a certain number of pre- and post-conditions. They are structured in fields and have a rather precise validation sequence order.
+
+The entire validation sequence can be [mapped as a flow chart](https://app.excalidraw.com/l/7Ao24g28S3o/872UZdalVsy), which reads top-down and left-to-right.
```
yHSJ, line 23: We should link to [here](https://cardano-scaling.github.io/cardano-blueprint/ledger/block-validation.html) instead of the Excalidraw, it's more permanent (and has other benefits that help the blueprint)

KtorZ: But the diagram on the blueprint is objectively worse 😅... it's harder to navigate and more importantly, it's missing all the links to the source code which add a lot of value.

yHSJ: I strongly agree with you 😛  
Mermaid + mdbook vs Excalidraw I guess?


My main concern with the Excalidraw is that it belongs to the sundae org and is inherently less "long-lived" than the blueprint. Also worth noting that the Excalidraw link gives the visitor edit permissions, which can cause problems...

I agree the links are vital. But aren't you suggesting improvements to the blueprint that should be added instead of reasons the Excalidraw is better? 😉 

I guess we should defer to @Quantumplation and @ch1bo on that

engineering-decision-records/010-ledger-validation-context.md
```diff
@@ -0,0 +1,321 @@
...
+### Abstract away state management from validations
+
+To decouple validations from state-management, we need to _at least_ abstract one or the other. After a few debates, our choice leaned towards abstracting the state management from the validations and keeping validation rules as the primary logic driver. We, therefore, provide an abstract interface as an eDSL available to the validation rules. The concrete implementation of those state operations is specified independently and chosen at the top of the stack by the caller.
+
+For example, we can capture the operation around UTxO management as follows:
+
+```rust
+pub trait UtxoStateManagement {
+    fn consume(&mut self, input: TransactionInput);
+    fn produce(&mut self, input: TransactionInput, output: TransactionOutput);
```
jeluard, line 60: `produce` doesn't convey the mutation imo. Maybe something like `append` or `include`?

engineering-decision-records/010-ledger-validation-context.md
```diff
@@ -0,0 +1,321 @@
...
+            .reference_inputs
+            .as_deref()
+            .map(|xs| xs.as_slice())
+            .unwrap_or(&[])
+            .iter();
+
+        inputs
+            .chain(reference_inputs)
+            .chain(collaterals)
+            .for_each(|input| context.require_input(input));
```
jeluard, line 240: We could consider having some data representation of those requirements instead of fn calls? 
So this function would return a list of e.g. `RerquiredTransactionInput` instead of calling `context.require_input`

engineering-decision-records/010-ledger-validation-context.md
```diff
@@ -0,0 +1,321 @@
...
+
+The current design will effectively process blocks in three passes:
+
+1. Parsing
+2. Preparation
+3. Validation
+
+It sounds _reasonable_ and more performant to combine the parsing & preparation steps and only requires two passes. Since the block is already being traversed for parsing, we could maintain an additional parsing context that takes care of marking elements to be fetched for validation. Besides, this approach seems well-geared towards the libraries that we already use for decoding CBOR (a.k.a [minicbor](https://docs.rs/minicbor/latest/minicbor/)) which provides [a mechanism](https://docs.rs/minicbor/latest/minicbor/fn.decode_with.html) for writing decoders that take an additional mutable context as parameter.
+
+This change would, however, require rewriting many of the decoders in Pallas to support decoding from such a custom context. Since the context wouldn't change the underlying serialization, we could reuse the existing decoders based on the unit (i.e., `()`) context.
```
jeluard, line 285: Sounds like digging the best tradeoff solution is a well-identified task that someone could tackle in isolation.

jeluard (commented):
I like the general direction and clarity of document! Definitely the way to go and an improvement over what we have now.

Later on I feel like we would want to iterate on a more data driven approach with no mutation and pure function calls returning data representation of operations.
This would open the door for concurrent execution, dynamic re-organization/simplification, fail-fast behavior.. 
Probably something we can prototype indenpendently.

engineering-decision-records/010-ledger-validation-context.md
KtorZ: I was under the impression that this type of more functional pattern were not quite idiomatic in Rust? Hence the method call. 

Fundamentally, they are equivalent to me; what would we gain from returning a structured type?

engineering-decision-records/010-ledger-validation-context.md
KtorZ: Indeed!

engineering-decision-records/010-ledger-validation-context.md
jeluard: I guess it depends what those functions do :) 
Data convey semantic while function are related to computation. Seems easier to me to use data to describe intent, then have implementation be free to interpret those (e.g. here de-duplication, reordering, caching). Strictly speaking you can achieve something similar with traits I guess.

It might be my clojure tropism ;) [This](https://ericnormand.me/article/data-functions-macros-why) is worth reading by the way.

yHSJ: Isn't this doing the same thing? The context owns the required inputs, so `require_input` for each input *is* constructing said list, if you want to think about it like that.

I don't come from a functional background, and I'm also not an incredibly experienced rust-acean, but this feels much more natural to me, since the context has to evolve during validation.

PR #181, opened 2025-04-06
JSON Testing Traces

yHSJ:
This PR changes the `TestingTraceCollector` to store JSON objects. This allows us to more easily support nesting fields such as:
```json
{
    "name": "require_witness",
    "credential": {
        "hash": "69d992b3a4f5426fd336ac054ba738e977c22fe1e4731a7697f130ab",
        "type": "verification_key"
    }
}
```

instead of:
```json
{
   "name":"require_witness",
   "credential.hash":"69d992b3a4f5426fd336ac054ba738e977c22fe1e4731a7697f130ab",
   "credential.type":"verification_key"
}
```

This is a minor change but feels more idiomatic if this is to become an ecosystem wide testing standard.

## Summary by CodeRabbit

- **New Features**
  - Introduced JSON-based tracing collection and assertion utilities for enhanced trace validation.
- **Refactor**
  - Removed legacy string-based trace collection and verification mechanisms.
  - Updated tests to use the new JSON tracing approach with simplified trace assertions.
  - Standardized credential data in expected trace files to a nested object format.
- **Chores**
  - Added `tracing-json` dependency and adjusted related project configurations for JSON tracing support.


yHSJ:
I have a lot of unwraps in the json tracing for mutexes and options, will have to go through and make sure they're safe, or handle them correctly

crates/tracing-json/src/lib.rs
```diff
@@ -0,0 +1,222 @@
...
+    actual_traces: Vec<json::Value>,
+    expected_traces: Vec<json::Value>,
+) -> Result<(), InvalidTrace> {
+    if actual_traces.len() != expected_traces.len() {
+        return Err(InvalidTrace::TraceLengthMismatch {
+            expected: expected_traces.len(),
+            actual: actual_traces.len(),
+        });
+    }
+
```
KtorZ: Small DevX note: I'd probably put this check _after_ the other. If some traces are missing, it's may be due to something else and comparing/visualizing earlier traces seem more helpful.

crates/tracing-json/src/lib.rs
```diff
@@ -0,0 +1,222 @@
...
+        .zip(expected_traces.into_iter())
+        .enumerate()
+    {
+        if actual != expected {
+            return Err(InvalidTrace::InvalidTrace {
+                expected,
+                actual,
+                index,
+            });
+        }
```
KtorZ: Given that this function is only really used for test code, I suggest relying on `assert_eq!` for _the entire vector_, so that we can get nice diffs. I don't think the assert lib will by default work well with json::value, but we can compare a vector of text-encoded traces and leverage the underlying tooling for diffs.

crates/tracing-json/src/lib.rs
```diff
@@ -1,30 +1,39 @@
...
     }
 
-    fn insert(&self, value: json::Value) {
-        self.lines.lock().unwrap().push(value);
+    pub fn get_traces(
+        &self,
+    ) -> Result<
+        RwLockReadGuard<'_, Vec<json::Value>>,
+        PoisonError<RwLockReadGuard<'_, Vec<json::Value>>>,
+    > {
```
KtorZ: We are leaking implementation details here; the guard and poison errors shouldn't make it to the consumer API.

PR #186, opened 2025-04-11
chore: make sure rules validation returns a structured result

jeluard:
Let `rules::validate_block` return validation results as structured data and not as an error. `Result::Err` should be used only for runtime errors and not validation failures.

**Before**

```rust
pub fn execute(...) ->anyhow::Result<InvalidBlock>;
```

**After**

```rust
pub enum BlockValidation {
    Valid,
    Invalid(InvalidBlock),
}

pub fn execute(...) -> BlockValidation;
```

A common signature has been introduced for `block` level rule validation. This allows to share common logic to deal with the `invalid` case. 
Note that introducing a trait to have stronger guarantees could be considered.

Execution of block validation fun is now abstracted in a trait:

```rust
pub trait RuleValidationExecutor<C: ValidationContext<FinalState = S>, S: From<C>> {
    fn execute(
        &mut self,
        context: &mut C,
        protocol_params: &ProtocolParameters,
        block: &MintedBlock<'_>,
        block_validation_fns: &[fn(
            &mut C,
            &MintedBlock<'_>,
            &ProtocolParameters,
        ) -> BlockValidation],
    ) -> BlockValidation;
}
```

This open the door down the line for more advanced way to deal with their execution than the simplistic linear execution (e.g. parallel execution, ordered execution, ...).

# Next steps

The same changes could be applied to `transaction` level rule validations.

## Summary by CodeRabbit

- **New Features**
  - Introduced explicit block validation results with detailed error variants for block validation.
  - Enhanced error reporting for block size, header size, and execution unit validation.

- **Refactor**
  - Updated validation functions to return structured validation results instead of generic errors.
  - Simplified and clarified validation control flow across block processing and tests.
  - Improved test assertions to match new validation result patterns.
  - Refactored ledger stage to handle validation results explicitly and streamline block processing.

- **Chores**
  - Updated Rust toolchain to a specific nightly version and revised minimum required Rust version.
  - Added ignore rule for coverage files and improved Makefile CI targets.
  - Enabled required Rust feature gates and suppressed specific Clippy lints.


yHSJ:
> Result::Err should be used only for runtime errors and not validation failures.

I'm not sure I agree with this statement. Perhaps it's not a system failure, but it's a failure to satisfy a predicate, which is a part of the happy-path. I have less rust experience than others working on Amaru so I may be wrong, but it feels very standard to use a `Result::Err` when there is a predicate failure in the `validate_block` function. I think returning a `BlockValidation` (which could be easily named `BlockValidationResult`) is communicating the exact same thing as a `Result`.

I'll need some convincing to agree that this PR is needed at all.

yHSJ:
> It looks sensible to distinguish both and in general we are moving in the direction of using Result::Err only for those cases.

I understand the idea here, it just feels unnatural to me. The `Ok` variant in every case *can* just be a `Result` since they're just enums with  `Valid` and `Invalid` variants. I always though the point of `anyhow` was to make error reporting simpler, i.e. I don't have to define a type for every error. But we already *are* AND, handling them. Why not just make the unrecoverable errors (or recoverable but unexpected errors) a variant of the `Error` type that we are returning?

Or, perhaps even better, instead of returning a `Result` at all (either `anyhow::Result` or the rust `Result`), return a custom `BlockValidationResult` that has a `Valid`, `Invalid`, and `Error` variant, so that we don't have to do extra maps and matches to get the same behavior.

I was just looking through some other rust implementations of blockchains (notably [reth](https://github.com/paradigmxyz/reth/blob/7f11a6f14839baca8d143a73a8b1a66ed0477c37/crates/engine/tree/src/tree/mod.rs#L1914) as we learned about it from an Ethereum dev in Paris), and they use a mix of basic `Result` values and the custom type with three variants. 

All this to say that I get the idea and will use it if it's what we want throughout the rest of Amaru, it just feels a bit strange to me.

crates/amaru-ledger/src/rules/block.rs
```diff
@@ -57,15 +57,26 @@ pub enum InvalidBlock {
...
     protocol_params: ProtocolParameters,
     block: MintedBlock<'_>,
-) -> Result<S, InvalidBlock> {
-    header_size::block_header_size_valid(block.header.raw_cbor(), &protocol_params)?;
+) -> BlockValidation<C, S> {
+    if let Err(err) =
+        header_size::block_header_size_valid(block.header.raw_cbor(), &protocol_params)
+    {
+        return BlockValidation::Invalid(err.into());
+    };
```
KtorZ: I'd eventually bite the bullet and do it for _all rules_.  The inconsistency (some rules returning errors, and some returning results) is going to drive us crazy in the long run 😰.

crates/amaru-ledger/src/rules/block.rs
```diff
@@ -57,15 +57,26 @@ pub enum InvalidBlock {
     UncategorizedError(String),
 }
 
+pub enum BlockValidation<C: ValidationContext<FinalState = S>, S: From<C>> {
```
KtorZ: Why the contraints here?

KtorZ:
@yHSJ this comes from a couple of discussions we had with @jeluard and @rkuhn about error handling. While we still don't generally agree about everything, I think the ledger rules is one particular location where the distinction between errors (as in, an application failure) and "invalid user inputs" is worth separating. 

One way to look at it is that from the perspective of Amaru (a block producer), an invalid block is part of the normal flow of execution and is just bound to a different logic flow. This is unlike a chain-indexer, who can reasonably expect all blocks it sees to be valid and if presented with an invalid one, it truly is an exception case. Amaru is different since we expect to sometimes run into invalid blocks if only because they are invalid _with respects to our local state_, but aren't "generally invalid". 

On the other hand, there are cases in the validation sequence that are truly invalid. For example, being unable to decode raw stake addresses bytes into a proper stake address. This error only occurs because of the partial de-serialization that happens in Pallas; and it is to be treated as an exception in this scenario.

yHSJ:
> I think the ledger rules is one particular location where the distinction between errors (as in, an application failure) and "invalid user inputs" is worth separating.

I totally get it and agree that this makes sense. I just think the way we're wrapping stuff and handling the results is strange now. If we want to have three different variants, we should create a new enum that has three variants, no?

crates/amaru-ledger/src/rules/block.rs
jeluard: Good catch!

rkuhn:
One very important about `anyhow`: its main utility is not to play it fast and loose regarding error types, its main utility is the ability to add context as an error is bubbling up the call chain. Without this, error reporting is almost guaranteed to be utterly useless, I’ve been frustrated by this when operating Rust code in production and getting panics or error logs in the past.

So, please never use `anyhow::anyhow!()`, ever (instead use `thiserror` to create good error enums). And use `anyhow::Context` generously. You will thank yourself later.

---

On the topic at hand: it would also be sensible to return `anyhow::Result<Result<(), ValidationError>>`, I’ve seen that in other places and probably used it myself somewhere. This is useful if you want to use the machinery that comes packaged with `Result<T, E>` in working further with the validation result — but not for reacting to runtime failures. The envisioned `BlockValidationResult` enum is isomorphic to `Result<(), ValidationError>`, so this is mostly a question of personal style. One argument for using a specific type is better IDE support for finding all usages.

crates/amaru-ledger/src/rules/block.rs
```diff
@@ -124,21 +137,22 @@ pub fn execute<C: ValidationContext<FinalState = S>, S: From<C>>(
             transaction_index: i as usize, // From u32
         };
 
-        transaction::execute(
-            &mut context,
+        if let Err(err) = transaction::execute(
```
jeluard, line 144: `transaction` rules haven't been changed; this could be done in a follow up PR.

crates/amaru-ledger/src/rules/block/body_size.rs
```diff
@@ -12,31 +12,34 @@
...
-) -> Result<(), InvalidBlockSize> {
+    _protocol_params: &ProtocolParameters,
+) -> BlockValidation {
+    let block_header = &block.header.header_body;
     let bh_size = block_header.block_body_size as usize;
     let actual_block_size = calculate_block_body_size(block);
 
     if bh_size != actual_block_size {
-        Err(InvalidBlockSize::SizeMismatch {
+        BlockValidation::Invalid(InvalidBlock::Size(InvalidBlockSize::SizeMismatch {
```
jeluard: Surely we can do better

crates/amaru-ledger/src/rules/block.rs
```diff
@@ -26,64 +26,74 @@ use crate::{
...
-        })
-        .flatten()
-        .collect::<Vec<_>>();
-
-    ex_units::block_ex_units_valid(ex_units, &protocol_params)?;
+) -> BlockValidation {
+    // Block level validations functions share the same signature.
+    // Currently apply them one by one
+    // TODO consider an abstract strategy pattern to apply them (e.g. in parallel, in a priority order, ...)
+    let block_validation_fns = vec![
```
jeluard: This is an nice side-effect of having unified signature for validation fns

rust-toolchain.toml
```diff
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.84" # ⚠️ Also change in top-level Cargo.toml
+channel = "nightly-2025-04-16" # ⚠️ Also change in top-level Cargo.toml
```
KtorZ, line 2: Doesn't match Cargo.toml's ?

crates/amaru-ledger/src/rules/block.rs
```diff
@@ -16,58 +16,101 @@ pub mod body_size;
...
 
+#[derive(Debug)]
+pub enum BlockValidation {
+    Valid,
+    Invalid(InvalidBlockDetails),
+}
+
+impl Try for BlockValidation {
+    type Output = ();
+    type Residual = BlockValidationResidual;
```
KtorZ: Why not: 

```rs
type Residual = InvalidBlockDetails
```

?

crates/amaru-ledger/src/rules/block.rs
KtorZ: Hmmm. Yes, although the downside being that they now all have the same signatures 😅 ... So we can no longer tell from the signature whether a rule: (1) need a context or (2) need a mutable context. It feels wrong down the line to have rule take a parameter that they simply ignore.

crates/amaru-ledger/src/rules/block.rs
jeluard: Indeed 👍

rust-toolchain.toml
jeluard: In `Cargo.toml` you can't use the nightly syntax so I used the closest stable version.

crates/amaru-ledger/src/rules/block.rs
jeluard: Maybe having a proper trait for that would alleviate this? I think it opens the door to nice optimizations down the line.

rust-toolchain.toml
KtorZ: 😬

examples/ledger-in-nodejs/rust-toolchain.toml
```diff
@@ -1 +1 @@
-toolchain = { channel = "nightly"}
\ No newline at end of file
+toolchain = { channel = "nightly-2025-04-16"}
```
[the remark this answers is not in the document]
jeluard: Looks like you are in the past, mate!

crates/amaru-consensus/src/consensus/chain_selection.rs
```diff
@@ -497,8 +497,10 @@ pub(crate) mod tests {
             chain_selector.select_roll_forward(&alice, *header);
         });
 
+        #[allow(clippy::double_ended_iterator_last)]
         let result = chain2
             .iter()
+            // TODO looks like this test relies on the fact that `select_roll_forward` is called on every element. `map` might not be correct then
```
KtorZ, line 503: I do not understand this comment about `map` ?

jeluard: `last` can be [replaced by next_back](https://rust-lang.github.io/rust-clippy/master/index.html#double_ended_iterator_last) in theory here but in practice it fails some test. This implies that this call expects `map` to traverse all element by calling a function with side-effects (`select_roll_forward`). `map` should be called with pure functions as a way to construct another iterator. `for_each` might help clarify the expectation here (but at the price of a more contrived usage as the last element should be tracked).

PR #187, opened 2025-04-11
sketch possible stage API and wiring

rkuhn:
no implementation at all, just API for now

## Summary by CodeRabbit

- **New Features**
  - Introduced a new framework that offers deterministic scheduling and robust asynchronous network processing for improved system performance.
  - Added comprehensive configuration files for blockchain node operations, covering various eras, governance models, and network topology.

- **Chores**
  - Upgraded the Rust toolchain support to version 1.85.
  - Incorporated an extra dependency to expand project capabilities.

- **Documentation**
  - Provided detailed guidelines on the new framework’s design and usage for clearer integration.


crates/pure-stage/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+Design goals:
+
+- no side effects, only explicit effects (which does include internal state changes)
+- easy to use, including the ability to hold references between invocations
+- executable on concurrent thread pool or using a deterministic simulator
+- fully back-pressured
+- ability for biased reading from inputs (which is necessary to avoid deadlocks with back-pressure)
+- wiring code should be nicely readable
+
```
stevana, line 11: Can we add "ability to (easily) dump the state before and after processing an input"? (So that we can write a nice debugger.)

crates/pure-stage/README.md
```diff
@@ -0,0 +1,130 @@
+# Pure Stage: Actor Model for deterministic scheduling
+
+Design goals:
+
+- no side effects, only explicit effects (which does include internal state changes)
+- easy to use, including the ability to hold references between invocations
+- executable on concurrent thread pool or using a deterministic simulator
```
stevana, line 7: This seems to suggest that the interleaving of threads doesn't matter in the "production deployment" (when using concurrent thread pool) as opposed to when in "simulation/testing deployment".

That's one strategy, but I think it would make it much harder to argue what we deploy is what we test this way. Alternatively we can try to stay deterministic while performant (in the production deployment) a la Disruptor.

cardano-node-config/alonzo-genesis.json
KtorZ, line 1: I am not fond of committing those. This is going to confuse the hell out of people if we start versioning configuration for other software at the root of the project. Few notes:

- If we explicitly need those configuration files for testing, let's put them in a tests folder, and make sure we add a relevant README.
- I would rather rely on a git submodule for those. The config "often" changes and are network-dependent. So it's best to rely on something like: https://github.com/cardano-foundation/cardano-configurations
- If they are only versioned for one's convenience because we sometimes need to run a Cardano node for our own development, then: don't 😅 ... Put them somewhere else on your machine, but not under Amaru.

crates/pure-stage/src/lib.rs
```diff
@@ -0,0 +1,273 @@
...
+
+pub struct Zero;
+pub struct Succ<T: Nat>(T);
+pub trait Nat {}
+impl Nat for Zero {}
+impl<T: Nat> Nat for Succ<T> {}
+pub const _0: Zero = Zero;
+pub const _1: Succ<Zero> = Succ(_0);
+pub const _2: Succ<Succ<Zero>> = Succ(_1);
+pub const _3: Succ<Succ<Succ<Zero>>> = Succ(_2);
```
abailly, line 12: Do we really want to go down the route of type-level naturals? although I understand the desire for type safety, I am not sure Rust is the best language to do that

crates/pure-stage/README.md
rkuhn: yes, indeed

crates/pure-stage/README.md
rkuhn: If we use discipline do ensure that the only effects within a stage consist of updates to local state updates, then it shouldn’t matter whether parallelism is introduced between different stages. One stage will never be executed concurrently with itself, so we get the same ability to parallelise as the actor model provides.

One assumption we need to make is that causality assumptions are transported between stages via proper messaging and not out-of-band (which is included in the above effect assumption). This allows a later state to depend for example on some information having been added to an append-only content-addressed storage facility.

My gut feeling is that adequate node performance — especially in the Leios era — will require parallel execution of processing steps, so we should design for this ability from the get go.

cardano-node-config/alonzo-genesis.json
rkuhn: Of course I didn’t intend to check these in. I have an update to `.gitignore` pending on another branch (IIRC), and I heroically managed to ignore these files in the beginning, but then they slipped through.

So basically: thanks for highlighting that `.gitignore` needs some love :-)

KtorZ: .git/info/exclude would probably be a better place for those as they don't _universally apply to the project_.

crates/pure-stage/README.md
```diff
@@ -0,0 +1,130 @@
...
+        }
+    }
+}
+```
+
+Findings:
+
+- quite complicated
+- was difficult to think up and probably still as bugs I didn’t realise
+- state progression is hard to see when reading the source code
```
abailly, line 89: As already discussed, I (we) have written code like that [in the past](https://github.com/cardano-scaling/hydra/blob/master/hydra-node/src/Hydra/HeadLogic.hs#L1284). By separating the interpretation of each transition into functions, things become quite readable and more importantly, the state machine structure becomes obvious.

abailly: In the above example, importing all the constructors' variants would already go a long way towards increasing legibility.

crates/pure-stage/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+## Sketch 2: effectful function from inputs to outputs
+
+```rust
+async fn chain_sync(state: ChainSync, input: ChainSyncMsg) -> (ChainSync, Vec<ChainSyncOutput>) {
+    match (state, input) {
+        (ChainSync::Initial(tip), ChainSyncMsg::BlockOp(op)) => 
+        // mostly like the above, but with less intermediate states e.g. for reading the store or for output
```
abailly, line 97: So basically effects would be just plain `async fn` functions on which the code `await`s?

crates/pure-stage/README.md
stevana: > One stage will never be executed concurrently with itself

I assumed we'd potentially want to shard slow stages (e.g. put two CPUs/threads on one stage and diverge even items to one CPU/thread and odd to the other). This is something that Disruptor can do without breaking determinism, while simply adding a thread pool of workers to each stage breaks determinism (which is what I thought you were suggesting when you mentioned a thread pool.)

crates/pure-stage/README.md
stevana: Perhaps something like this be useful: https://hoverbear.org/blog/rust-state-machine-pattern/ ?

But I suppose that the problem with any "raw state machine" (explicit function from `(i,s) -> (o, s)`) approach is the "intermediate states", that are not part of what's interesting. For example if we want to add async IO/RPC as part of a transition, then we need to introduce "intermediate states" that suspend the transition and resume it when the IO completes, these intermediate states are clutter if you are familiar with `await`s (they don't make things more obvious, but rather obscure the intent).

In the Maelstrom inspired DSL that I've been working with I've hidden these intermediate states in the interpreter/event loop, away from the programmer. What Roland is suggesting is to rely on the Rust compiler to fill in the intermediate states via `await`s.

I still don't see exactly how simulating `await`s would work though, because it's exactly at these intermediate states where you want to yield to the simulator. It would be good if we could sketch this a bit further. I've been trying to read the async from scratch series: https://natkr.com/tags/async-from-scratch-series/ but it's still not clear to me.

There's also be the option of doing some kind of DSL in Rust, closer to how I and [other's](https://crates.io/crates/maelstrom-node) have implemented the Maelstrom DSL in Rust. This DSL then gets interpreted into a "raw state machine", but this would obviously require reimplementing some of the `await` machinery. One potential advantage of this approach is that because we have the "raw state machine" (i.e. explicit function from `(i, s) -> (o, s)`) in the end, we can easily dump the state between steps in the simulator, which gives us a trace of state transitions which is useful for making assertions and writing a debugger. With the async approach we'd need to dump the state manually somehow, but should still be doable.

abailly: I agree with @stevana that the key here to have a convincing "framework" for writing state machines using async code is to demonstrate how this is amenable to simulating, deterministically, various execution paths generated from a model.
I did some archaeology and found that we discussed different approaches for Hydra in April 2021: https://github.com/cardano-scaling/hydra/wiki/Logbook-2021-H1#2021-04-26
I unfortunately don't have the code for the various other approaches (Monad stack, free monads) we tried.

rkuhn: Yes, I agree. I’m working on that (besides catching up to `main` with the downstream server), should have a demonstrator next week.

BTW: the Church numerals were only useful for the initial API sketch, they aren’t practical for an actual implementation. Will probably go with tuples and macros — we don’t expect more than a handful of inputs or outputs, right?

cardano-node-config/alonzo-genesis.json
rkuhn: Hmm, we may be talking past each other or this topic needs more communication between us.

My proposal is to include the instructions for ignoring files at a certain path within the version-controlled repository contents because within that contents there also are a README and a Makefile that populate that certain path, so everyone who follows the instructions will have to manually ignore that path. What downside would there be do have battery shieldings included upon cloning? (the batteries being files that are automatically downloaded for the reasons you state above)

I’ll include the concrete change in #149 so that we can review it there.

rkuhn:
Hi @stevana! I’m developing a better API with an actual implementation (same branch) but it isn’t in a state where anyone (or the rabbit) should look at it. Will open a new PR once I have something working.

PR #195, opened 2025-04-22
feat: further splitting of stages

abailly:
* Split old `header_validation` logic into smaller logical parts, matching 1-1 pipeline diagram
* Introduce both _modules_ containing "business logic" and "small" stages for wiring the various steps
* Rename some modules/stages towards more consistency with documentation and across the pipeline

## Summary by CodeRabbit

- **New Features**
  - Introduced a modular, multi-stage consensus pipeline for blockchain headers: receiving, validating, storing, selecting chain, and forwarding.
  - Added detailed event types and enriched tracing for better observability of consensus events.
- **Refactor**
  - Split monolithic consensus logic into focused stages, improving clarity and concurrency handling.
  - Updated event and enum variants to use struct-style named fields for better code readability.
  - Streamlined nonce management by unifying type usage and changing ownership to references.
  - Renamed structs and stages for consistency, e.g., ForwardStage to ForwardChainStage.
  - Reorganized imports and adjusted method signatures across consensus and simulation components.
  - Replaced single consensus component in the simulator with distinct validate, store, and select stages.
  - Updated simulator and sync components to align with new consensus event types and flow.
- **Documentation**
  - Added a comprehensive README with a pipeline flow diagram explaining consensus header processing stages.
- **Bug Fixes**
  - Enhanced error handling and propagation in consensus and simulation stages, preventing silent failures.


crates/amaru-consensus/README.md
```diff
@@ -0,0 +1,36 @@
+# Consensus pipeline
+
+The following graph represents a simplified sequence of _stages_ headers (either forwarded or rolled-back) undergo in the consensus module. Each square node is a processing step, and arrows are labelled with the type of messages flowing between the various steps.
+
```
stevana, line 4: Could we name the events in the diagram the same way as in the code? E.g. "PullEvent" rather than "pull event"?

This version of the diagram doesn't convey which effects happen at each stage, unlike your old diagram which had dotted up and down arrows to storage and/or upstream peers. Could we add e.g. "*effect: storage/networking*" or similar to the boxes that need it? (Alternatively we could add a node for storage and add dotted arrows to storage/upstream, not sure if that will be easier to read.)

simulation/amaru-sim/src/simulator/mod.rs
```diff
@@ -152,21 +164,32 @@ async fn run_simulator(
...
-                        PullEvent::Rollback(peer, rollback) => {
-                            consensus.handle_roll_back(&peer, &rollback).await
-                        }
+                        PullEvent::RollForward(peer, point, raw_header, _span) => consensus
+                            .handle_roll_forward(&peer, &point, &raw_header)
+                            .await
+                            .expect("unexpected error on roll forward"),
+                        PullEvent::Rollback(_, _) => event,
                     },
                     Err(_) => todo!(),
```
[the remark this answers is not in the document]
abailly: @coderabbitai this is incorrect, the `handle_roll_forward` now returns a `PullEvent`

crates/amaru-consensus/README.md
```diff
@@ -0,0 +1,36 @@
...
+    sto_hdr -- storage failure --> crash?
+    sto_hdr -- pull event --> select[select chain]
+    select -- invalid chain --> disc
+    select -- validate header events --> fetch[fetch block]
+    fetch -- fetch error --> disc
+    fetch -- validate block event --> sto_block[store block]
+    sto_block -- storage failure --> crash?
+    sto_block -- validate block event --> val_block[validate block]
+    val_block -- invalid block --> disc
+    val_block -- block validation result --> fwd[forward chain]
```
stevana: From my understanding of @rkuhn's PRs the "forward chain" stage has "BlockValidationResult + NewPeer" as input and needs to select between the two. Are there other stages that need to select? If so it would be good to make that explicit in the diagram as well?

crates/amaru-consensus/README.md
abailly: I wanted to have the diagram reflect the reality, so the effects are _in_ the boxes.

crates/amaru-consensus/README.md
```diff
@@ -0,0 +1,36 @@
...
+    fwd --> down([downstream])
+```
+
+Stages:
```
stevana, line 35: I'm not sure if this is the right time to have this discussion, but when I see that several stages have the same input type, e.g. "validate header", "store header" and "select chain" all have "PullEvent" as input, it makes me wonder if all these can be done in parallel? (We could still only proceed with "fetch block" if they all succeed.)

crates/amaru-consensus/README.md
stevana: I'm not sure I understand what you mean... You've left the effects implicit because in "reality" (= the Rust source code?) they are implicit?

abailly: Yes, but I am also tired so perhaps I can actually make the effects manifest :)

crates/amaru-consensus/README.md
abailly: I think `NewPeer` is an internal message?

crates/amaru-consensus/README.md
abailly: I think that would make sense indeed

stevana: "store block" and "validate block" both have "ValidateBlockEvent" as input, could they be done in parallel as well?

Can any problem arise from storing an invalid header/block? (Which could happen if validation and storage happens in parallel.)

crates/amaru-consensus/README.md
stevana: Internal to what?

(The reason I think this should be made explicit is because this choice of which message to receive will affect the DSL/[stage API](https://github.com/pragma-org/amaru/pull/187) for deterministic testing.)

rkuhn: Within the current `main` framework, the forward stage only accepts BlockValidationResult as input, new peers only occur within the actor-based internal implementation. Any future stage API design is still in the research phase and thus shouldn’t be taken too seriously for current PRs.

crates/amaru-consensus/README.md
rkuhn: I do agree if we make the following explicit: Rust types should not only denote structure but also meaning. So even if a freshly received header from the network and a validated header have the same structure, they should be carried by different Rust types to make the distinction clear.

While I might be missing context/knowledge, it does seem to me that storage and chain selection should logically be deferred until after validation (which would occur naturally in the regime of the sentence above).

crates/amaru-consensus/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+* [receive header](src/consensus/receive_header.rs): this stage is responsible for basic sanity check of _chain sync_ messages, deserialising raw headers, and potentially checking whether or not they should be further processed (eg. if a header is already known to be invalid, or known to be valid because it's part of our best chain, let's not waste time processing it!)
+* [validate header](src/consensus/validate_header.rs): protocol validation of the header, checks the correctness of the VRF leader election w.r.t relevant stake distribution, and epoch nonce
+* [store header](src/consensus/store_header.rs): store valid (and invalid?) headers indexed by hash
```
rkuhn, line 40: My gut feeling is that invalid headers should be stored separately, probably with a quota, so as to not degrade the processing facilities for valid headers.

crates/amaru-consensus/src/consensus/mod.rs
```diff
@@ -12,15 +12,32 @@
...
+
+#[derive(Clone, Debug)]
+pub enum ChainSyncEvent {
+    RollForward(Peer, Point, Vec<u8>, Span),
+    Rollback(Peer, Point),
+}
+
+#[derive(Clone, Debug)]
+#[allow(clippy::large_enum_variant)]
+pub enum PullEvent {
```
rkuhn: “pull” is a generic word that doesn’t much help the reader, how about `SanitizedChainSyncEvent`? The next stage should then probably be passed a `ValidatedChainSyncEvent` that contains a `ValidHeader`.

crates/amaru-consensus/src/consensus/mod.rs
```diff
@@ -12,15 +12,32 @@
...
+pub mod select_chain;
 pub mod store;
+pub mod store_header;
+pub mod validate_header;
+
+pub const EVENT_TARGET: &str = "amaru::consensus";
+
+#[derive(Clone, Debug)]
+pub enum ChainSyncEvent {
+    RollForward(Peer, Point, Vec<u8>, Span),
```
rkuhn: Tuple structs or variants are pervasive in this codebase and its dependencies, but they make it really hard to understand what the intent is. While `Peer` seems self-explanatory, there are multiple `Point`s that could be meant here, the `Vec<u8>` obviously needs a name, and the `Span` is quite unexpected and needs documentation.

crates/amaru-consensus/src/consensus/mod.rs
```diff
@@ -12,15 +12,32 @@
...
 pub mod chain_selection;
-
-pub mod header_validation;
+pub mod receive_header;
+pub mod select_chain;
 pub mod store;
+pub mod store_header;
+pub mod validate_header;
+
+pub const EVENT_TARGET: &str = "amaru::consensus";
```
rkuhn, line 27: this seems extraneous to me, why not just use `tracing` built-in module naming?

crates/amaru-consensus/src/consensus/select_chain.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2024 PRAGMA
+// Copyright 2025 PRAGMA
```
rkuhn, line 1: This should probably be updated automatically; my preference has so far been to include the year of inception as well, just for nostalgia ;-)

crates/amaru-consensus/src/consensus/store_header.rs
```diff
@@ -0,0 +1,58 @@
...
+
+    pub async fn store(&self, point: &Point, header: &Header) -> Result<(), ConsensusError> {
+        self.store
+            .lock()
+            .await
+            .store_header(&header.hash(), header)
+            .map_err(|e| ConsensusError::StoreHeaderFailed(point.clone(), e))
+    }
+
+    pub async fn handle_event(&self, event: &PullEvent) -> Result<PullEvent, ConsensusError> {
```
rkuhn: I’d pass the PullEvent by value to allow passing it through.

As a general rule: This type signature makes it clear that the body of the method will need to create an owned PullEvent, so we should request the caller to provide one.

With the current signature, the caller is forced to duplicate the PullEvent even if that is of no use.

crates/amaru-consensus/src/consensus/validate_header.rs
```diff
@@ -0,0 +1,117 @@
...
+)]
+pub fn header_is_valid(
+    point: &Point,
+    header: &Header,
+    raw_header_body: &[u8],
+    epoch_nonce: &Nonce,
+    ledger: &dyn HasStakeDistribution,
+) -> Result<(), ConsensusError> {
+    let active_slot_coeff: FixedDecimal =
+        FixedDecimal::from(1_u64) / FixedDecimal::from(ACTIVE_SLOT_COEFF_INVERSE as u64);
```
rkuhn, line 41: this does look like a constant — if it cannot be CONST then it would be a good use case for a OnceLock

crates/amaru-consensus/README.md
abailly: I agree the types should convey meaning, this PR is a first pass at splitting the stages into a coherent sequence. For the sake of simplicity and size, I will defer introducing specific types to a follow-up PR.

> it does seem to me that storage and chain selection should logically be deferred until after validation

Chain selection occurs after header validation, but before block validation which indeed is an issue. The output of `validate_header` should actually be _candidate_ chains, that can be turned into actual best chain should the body be valid _and_ no better chain exists. The `select_chain` stage should happen after `validate block` and the current `select_chain` should rather be `select_candidate`.
 
But I would like to have failing tests manifest this issue before implementing that.

About storage: there are 2 reasons we store the block body _before_ doing the validation (but indeed we could it in parallel, mostly). The first one is that we want to store both valid and invalid blocks, to be able to reject the latter early and penalise peers who propagate them. The second is because of a feature of the consensus diffusion called _pipelining_, whereby a node can propagate a `RollForward` before validating the block body iff the roll forward is the new best chain, and it's a child of the previous best chain.

crates/amaru-consensus/README.md
abailly: I tend to agree, but you want to be able to know whether or not a given header is already known as a bad or good header.

crates/amaru-consensus/src/consensus/select_chain.rs
abailly: Actually, this is because of renaming: Emacs automatically insert the current year in _new files_ and this was a new file before I copy-pasted previous code.

crates/amaru-consensus/src/consensus/validate_header.rs
abailly: It's currently a constant but `ACTIVE_SLOT_COEFF` should be a network parameter

crates/amaru-consensus/README.md
```diff
@@ -0,0 +1,36 @@
...
+    sto_block -- storage failure --> crash?
+    sto_block -- validate block event --> val_block[validate block]
+    val_block -- invalid block --> disc
+    val_block -- block validation result --> fwd[forward chain]
+    fwd --> down([downstream])
+```
+
+Stages:
+
+* [receive header](src/consensus/receive_header.rs): this stage is responsible for basic sanity check of _chain sync_ messages, deserialising raw headers, and potentially checking whether or not they should be further processed (eg. if a header is already known to be invalid, or known to be valid because it's part of our best chain, let's not waste time processing it!)
```
stevana, line 38: `crates/amaru/src/stages/pull.rs` seems to be missing from the list of stages.

crates/amaru-consensus/src/consensus/mod.rs
abailly: What do you mean @rkuhn ?

simulation/amaru-sim/src/simulator/mod.rs
```diff
@@ -152,21 +164,37 @@ async fn run_simulator(
...
 
-                match events {
+                // store header stage
+                let store_event = match store_header.handle_event(validation_event).await {
+                    Ok(stored) => stored,
+                    Err(_) => panic!("got error storing event"),
+                };
+
+                // chain selection stage
+                match select_chain.handle_chain_sync(store_event).await {
```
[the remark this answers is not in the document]
abailly: @coderabbitai This code is a simulator code, and until we implement a proper fault injection strategy that we can make the pipeline depends on, panicking is probably the right thing to do.

crates/amaru/src/stages/consensus/fetch_block.rs
```diff
@@ -52,19 +52,28 @@ impl BlockFetchStage {
...
     #[instrument(level = tracing::Level::TRACE, skip_all)]
-    async fn handle_event(&mut self, unit: &ValidateHeaderEvent) -> Result<(), WorkerError> {
+    async fn handle_event(&mut self, unit: ValidateHeaderEvent) -> Result<(), WorkerError> {
         match unit {
-            ValidateHeaderEvent::Validated(peer, point, span) => {
+            ValidateHeaderEvent::Validated { peer, point, span } => {
                 Span::current().set_parent(span.context());
-                let block = self.fetch_block(peer, point.clone()).await.or_panic()?;
+                let block = self.fetch_block(&peer, &point).await.or_panic()?;
                 self.downstream
```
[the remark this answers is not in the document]
abailly: @coderabbitai In this case we'll need to clone all the internals of the event anyway which makes the code more obscure. Taking ownership of the event by cloning makes sense

abailly:
Mergin in spite of low coverage, I feel a bit ashamed but the tests that come to my head are not really valuable if just for the sake of adding more coverage.

PR #197, opened 2025-04-28
feat: promote 'UncategorizedError' in ledger to a top-level anyhow::Error

KtorZ:
This makes the 'BlockValidation' a 3-variant enum, where `Exception` case reflects exceptional scenarii that aren't expected to occur as part of the normal operations.

## Summary by CodeRabbit

- **New Features**
  - Enhanced block validation with more flexible and expressive error handling, allowing for both domain-specific and generic errors.
  - Improved error context and propagation for block and header size validation.

- **Refactor**
  - Unified success and error result types across validation functions for consistency.
  - Updated validation logic to return detailed invalid block information using option types.
  - Streamlined test cases to match new validation result patterns.

- **Chores**
  - Updated dependencies for consistent workspace management.

### DIFF
--- a/crates/amaru-ledger/src/rules/block.rs
+++ b/crates/amaru-ledger/src/rules/block.rs
@@ -25,7 +25,10 @@ use amaru_kernel::{
     protocol_parameters::ProtocolParameters, AuxiliaryData, ExUnits, HasExUnits, Hash, MintedBlock,
     OriginalHash, StakeCredential, TransactionPointer,
 };
-use std::ops::{ControlFlow, Deref, FromResidual, Try};
+use std::{
+    ops::{ControlFlow, Deref, FromResidual, Try},
+    process::{ExitCode, Termination},
+};
 use tracing::{instrument, Level};
 
 #[derive(Debug)]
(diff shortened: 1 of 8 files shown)


yHSJ:
Thank you for making this change! I really do think this is much nicer to consume for the call site and clearly communicates the intention. Not sure why that test is failing in the coverage CI run though 😛

KtorZ:
> Not sure why that test is failing in the coverage CI run though

I have no idea either 😅 ... and it's slightly worrying that it fails on the coverage but not on the classic ones.

crates/amaru-ledger/src/rules/block.rs
```diff
@@ -47,42 +50,69 @@ pub enum InvalidBlockDetails {
...
         violation: InvalidTransaction,
     },
-    UncategorizedError(String),
 }
 
 #[derive(Debug)]
-pub enum BlockValidation {
-    Valid,
+pub enum BlockValidation<A, E> {
+    Valid(A),
```
jeluard, line 57: Is `A` used at all? Or that's for some future expected behavior?

crates/amaru-ledger/src/rules/block.rs
```diff
@@ -47,42 +50,69 @@ pub enum InvalidBlockDetails {
...
+        C: std::fmt::Display + Send + Sync + 'static,
+    {
+        match self {
+            BlockValidation::Err(err) => BlockValidation::Err(err.context(context)),
+            BlockValidation::Invalid { .. } | BlockValidation::Valid { .. } => self,
+        }
+    }
+}
+
+impl<A, E> Termination for BlockValidation<A, E> {
```
jeluard, line 85: This is only useful for the tests? [Termination](https://doc.rust-lang.org/std/process/trait.Termination.html) semantic is `A trait for implementing arbitrary return types in the main function.` so feels a bit unnatural to have it implemented at the `BlockValidation` level?

crates/amaru-ledger/src/rules/block/header_size.rs
```diff
@@ -16,28 +16,23 @@ use amaru_kernel::protocol_parameters::ProtocolParameters;
...
     header: &[u8],
     protocol_params: &ProtocolParameters,
-) -> BlockValidation {
+) -> BlockValidation<(), anyhow::Error> {
     let max_header_size = protocol_params
         .max_header_size
         .try_into()
-        .unwrap_or_else(|_| panic!("Failed to convert u32 to usize"));
+        .map(BlockValidation::Valid)
+        .unwrap_or_else(|e| BlockValidation::anyhow(e).context("failed to convert u32 to usize"))?;
```
jeluard, line 27: With a proper error type we wouldn't need to use `anyhow` here?

crates/amaru-ledger/src/rules/block.rs
```diff
@@ -113,10 +143,11 @@ pub fn execute<C: ValidationContext<FinalState = S>, S: From<C>>(
         let witness_set = match witness_sets.get(i as usize) {
             Some(witness_set) => witness_set,
             None => {
-                return BlockValidation::Invalid(InvalidBlockDetails::UncategorizedError(format!(
+                // TODO: Define a proper error for this.
+                return BlockValidation::bail(format!(
```
jeluard, line 147: If we had a proper error here there wouldn't be a need for `anyhow` at all?

crates/amaru-ledger/src/rules/block.rs
KtorZ: It's used in the header size rule!

crates/amaru-ledger/src/rules/block.rs
KtorZ: Indeed, this is used to satisfy some of the test code. 

Although, I don't think the semantic is particularly weird here. On the contrary.

crates/amaru-ledger/src/rules/block/header_size.rs
KtorZ: Yes, but those are exceptions; and I don't think we really care to turn them into user facing errors. 

So it seems to be a perfect use case for anyhow?

jeluard (approved):

jeluard:
I've seen this failure happen sporadically on windows too. Not sure why this test would be flaky.

PR #200, opened 2025-04-30
Refund pool deposits & re-implement DRep delegation bug

KtorZ:
- :round_pushpin: **fix: record and use drep previous de-registration**
   It is now used to implement a bug in the version 9 of the protocol, where it would be possible for one to delegate to a non-existing drep. Upon registering, the drep would immediately be granted with the associated stake.

  This commit also fix drep mandates when those are updating in the middle of a dormant epoch. We need some workaround for the beginning of the epoch because we are blind to past history.

- :round_pushpin: **chore: refund stake pool deposits when they retire.**
  

- :round_pushpin: **fix: amend test snapshots to cope with bloopers...**
    So there are two things at play here. One which I can explain, and one which I cannot.

  1. Regarding the divergence in treasury amount on snapshot 174:

     The test snapshots are powerful, but limited. We define the 'stake distribution' snapshot by looking at the ledger state right after the 'SNAP' rule.

     The 'rewards' snapshots contains data pertaining to the rewards of a particular epoch. Those rewards are calculated *later*, which makes some element of the snapshot a bit tricky to anchor in time.

     In particular, the treasury and reserves value used for rewards calculation are the values of the pots *at the moment the calculation begin*.

     So, for rewards corresponding to an epoch `e`, these calculation begin about `k` blocks within epoch `e + 3`; and are paid out in the transition from `e + 3` to `e + 4` (and thus, available in `e + 4`).

     BUT: for the 'rewards' snapshot at epoch `e`, we pull the treasury value from our snapshots at `e + 2`; which already includes rewards paid out to accounts as well as the leftovers from any unregistered account, but, does not include leftovers that may come from pool retirements since those are only process _at the beginning of epochs_ (and thus, after the snapshot has been taken).

     There's a pool retiring in the transition from 176 to 177, with a reward account that's already unregistered. And thus, the deposit is sent back to the treasury. So when the rewards calculation for 174 kicks in later in the epoch, the deposit was already added to the treasury... So it won't be present from our snapshot labeled 176 since it happened BEFORE the beginning of the epoch 177.

  2. Regarding the missing drep delegation in 178 and 179.

     This is a case of a delegation that occured BEFORE a drep registration. In fact, they are both done in the same transaction, but the certificates are swapped (delegation, then unregister). Because of a bug in the version 9 of the protocol (going until epoch 180), this was allowed by the protocol.

     However, since the DRep doesn't exist when the delegation is registered, its delegators list cannot be modified to hold the account. And thus, the account isn't reported as part of the DRep's delegators as it should (even though, its stake is being accounted for!).

## Summary by CodeRabbit

- **New Features**
  - Introduced detailed stake distribution snapshots with account, pool, and DRep states, supporting network-aware serialization.
  - Added new protocol version constants and protocol version-aware governance and mandate calculations.
  - Enabled retrieval of account information by stake credential in stores and APIs.
  - Added method to obtain the first epoch of an era in era history.
  - Enhanced demo script and Makefile to support configurable listen address.

- **Improvements**
  - Refined DRep registration/deregistration tracking and mandate calculations for protocol version compatibility.
  - Updated pool retirement logic to collect and process refunds after ticking pools.
  - Improved epoch and era boundary handling in ledger state imports.
  - Enhanced serialization of summary data with human-readable encodings.
  - Adjusted logging and tracing for span events and refund processing.

- **Bug Fixes**
  - Fixed filtering of stale or superseded DRep registrations in governance snapshots.
  - Corrected handling of DRep removals to record previous deregistration pointers.

- **Documentation**
  - Added and updated license headers in source files.

- **Style/Refactor**
  - Consolidated and reorganized imports across modules for clarity.
  - Simplified method signatures by removing unused parameters.
  - Re-enabled previously ignored tests and refined test imports.

- **Chores**
  - Updated CI workflow cache key version for ledger database.
  - Adjusted test coverage to include additional epochs and protocol versions.


crates/amaru-ledger/src/state.rs
```diff
@@ -341,8 +342,15 @@ fn recover_stake_distribution(
...
-    StakeDistribution::new(&snapshot, GovernanceSummary::new(&snapshot, era_history)?)
-        .map_err(StateError::Storage)
+    // FIXME: Obtain from current block
+    let protocol_version = PROTOCOL_VERSION_9;
+
+    StakeDistribution::new(
+        &snapshot,
+        protocol_version,
+        GovernanceSummary::new(&snapshot, protocol_version, era_history)?,
+    )
```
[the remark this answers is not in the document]
KtorZ: Thanks rabbit, I plan to pull the protocol version from block header in an upcoming PR. For now, I've just moved it from being hard-coded inside the summary functions to being hard-coded inside the ledger state. So it's one closer to the block header. But it's not the point of the PR to fix it, so will come in another.

crates/amaru-ledger/src/context.rs
```diff
@@ -211,7 +211,7 @@ pub trait DRepsSlice {
         anchor: Option<Anchor>,
     ) -> Result<(), UpdateError<StakeCredential>>;
 
-    fn unregister(&mut self, drep: StakeCredential, refund: Lovelace);
+    fn unregister(&mut self, drep: StakeCredential, refund: Lovelace, pointer: CertificatePointer);
```
KtorZ, line 214: This is required so that we can track the precise moment drep deregisters, in order to re-implement the v9 bug where one could delegate to a non-existing DRep. 

Without tracking the de-registration time (i.e. if we only look at re-registration times), we cannot properly prune delegation from before the de-registration (which ends when de-registering) from those that happened after the re-registration (which will be immediately effective on the re-registration).

crates/amaru-ledger/src/store/columns/pools.rs
```diff
@@ -102,14 +105,17 @@ impl Row {
...
                         last,
                         Some(&(None, current_epoch)),
                         "invariant violation: most recent retirement is not last certificate: {:?}",
                         last,
                     );
+
                     *pool = None;
-                    return;
+
+                    return Some(refund);
```
KtorZ, line 118: We now return the credential that needs refund when the pool is retired. This credential may or may not exist anymore. The case where it no longer exists is handled by the refund function.

crates/amaru-ledger/src/store/columns/dreps.rs
```diff
@@ -55,23 +61,29 @@ impl<C> cbor::encode::Encode<C> for Row {
...
         e.encode_with(self.last_interaction, ctx)?;
+        e.encode_with(self.previous_deregistration, ctx)?;
         Ok(())
     }
 }
 
 impl<'a, C> cbor::decode::Decode<'a, C> for Row {
     fn decode(d: &mut cbor::Decoder<'a>, ctx: &mut C) -> Result<Self, cbor::decode::Error> {
-        d.array()?;
+        let len = d.array()?;
```
jeluard, line 76: Is it to support migration from previous versions?

crates/amaru-stores/src/rocksdb/columns/dreps.rs
```diff
@@ -35,34 +34,60 @@ pub fn add<DB>(
...
-                last_interaction: epoch,
+                last_interaction: None,
+                previous_deregistration: None,
             };
             anchor.set_or_reset(&mut row.anchor);
-            db.put(key, as_value(row))
-                .map_err(|err| StoreError::Internal(err.into()))?;
+            Some(row)
+
+        // Technically impossible, sign of a logic error.
```
jeluard, line 73: Worth a `panic` ?

crates/amaru-stores/src/rocksdb/columns/dreps.rs
```diff
@@ -101,11 +126,26 @@ pub fn tick<DB>(
...
+    for (credential, pointer) in rows {
+        let key = as_key(&PREFIX, &credential);
+
+        if let Some(mut row) = db
+            .get(&key)
+            .map_err(|err| StoreError::Internal(err.into()))?
+            .map(Row::unsafe_decode)
+        {
+            row.previous_deregistration = Some(pointer);
+            db.put(key, as_value(row))
```
jeluard, line 140: So we never delete drep?

jeluard (approved):

crates/amaru-ledger/src/summary/rewards.rs
```diff
@@ -178,18 +179,32 @@ impl StakeDistribution {
...
+                                        None
+                                    }
                                 } else {
-                                    None
+                                    if &since >= registered_at {
+                                        Some(drep)
+                                    } else {
+                                        None
+                                    }
                                 }
```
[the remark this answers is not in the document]
jeluard: Even if it's not an issue for us, this suggestion makes the code more readable.

crates/amaru-ledger/src/state/volatile_db.rs
```diff
@@ -297,7 +298,17 @@ impl AnchoredVolatileState {
...
+                    .state
+                    .dreps
+                    .unregistered
+                    .into_iter()
+                    .filter_map(move |credential| {
+                        self.state
+                            .dreps_deregistrations
+                            .remove(&credential)
+                            .map(|pointer| (credential, pointer))
+                    }),
```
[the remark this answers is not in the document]
jeluard: I understand that it is expected that some `credential` might not be  in the `dreps_deregistrations`.
Maybe it would be worth some comment @KtorZ ?

crates/amaru-ledger/src/store/columns/dreps.rs
KtorZ: Ah yes; I did that at some point to avoid re-importing snapshots but it stayed 😅

crates/amaru-stores/src/rocksdb/columns/dreps.rs
KtorZ: At this point, no. Given the substantial deposit on DRep, it won't "blow up" anyway and remain significantly smaller than the UTxO set. Eventually, if we use a more recent snapshot (post v10), we can just go back to the previous behaviour 👍

crates/amaru-stores/src/rocksdb/columns/dreps.rs
KtorZ: It's already handled; this branch yields `None`, which will trigger the `error!` in the next expression. 

```rust
            None => {
                error!(
                    target: EVENT_TARGET,
                    ?credential,
                    "add.register_no_deposit",
                )
            }
```

crates/amaru-ledger/src/state/volatile_db.rs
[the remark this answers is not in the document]
KtorZ: They should actually all be there; I've made this chunk of code clearer:

https://github.com/pragma-org/amaru/pull/200/commits/596e5b5611c7fe6fdc8bde7ad480be7aa461ddcf

PR #204, opened 2025-05-02
feat: generalize parameters handling

jeluard:
Move all parameters from `amaru-kernel/src/lib.rs` into a `GlobalParameters` struct that is then threaded into all the relevant places. Down the line this struct will be filled at boot-time from a `store`. `GlobalParameters` are immutable and ultimately originates from `genesis` configs. In this PR they are still considered static and hard-coded.
`ProtocolParameters` are imported from `snapshots` and not hardcoded anymore. Usage has been adapted to prepare for `ProtocolParameters` changes via governance.

Also adapted `bootstrap` scripts so that `mainnet` and others network can be used for tests later on.

**Note** after those changes are merged it is required to delete the `snapshots` folder the first time or the proper structure won't be re-created.

## Summary by CodeRabbit

- **New Features**
  - Introduced a centralized Global Parameters configuration, allowing protocol constants and network settings to be managed dynamically instead of using hardcoded values.
  - Added new economic parameters for monetary and treasury expansion rates.

- **Refactor**
  - Updated validation, ledger, consensus, governance, and simulation logic to use Global Parameters, replacing fixed constants throughout the application.
  - Modified method and function signatures across the codebase to accept and utilize Global Parameters, enhancing flexibility and maintainability.
  - Enhanced block validation, state management, reward calculations, certificate processing, and nonce evolution to incorporate Global Parameters.
  - Integrated Global Parameters into initialization and testing workflows for consistent protocol parameter usage.
  - Refined protocol parameters encoding and decoding with CBOR for improved data handling.
  - Updated stake distribution, governance summaries, and rewards calculations to depend on protocol parameters instead of fixed constants.
  - Improved error handling and parameter fetching during ledger state import and validation stages.
  - Adjusted ledger state and stake distribution initialization to use epoch-aware protocol parameters.
  - Enhanced store interfaces to persist and retrieve protocol parameters per epoch.
  - Updated simulation and example code to propagate Global Parameters and protocol versions consistently.
  - Added property-based tests for protocol parameters encoding and decoding.
  - Suppressed unnecessary lint warnings related to protocol parameter conversions.
  - Reorganized import statements for clarity and maintainability.

- **Chores**
  - Removed obsolete protocol constants that are now provided via Global Parameters.
  - Removed default Makefile target to prevent automatic execution without arguments.
  - Updated snapshot download and import procedures to use network-specific directories for better organization.
  - Enhanced Makefile targets for dynamic snapshot, header, and nonce imports using JSON configuration files.
  - Cleaned up imports and lint warnings related to protocol parameters and conversions.
  - Improved temporary directory handling in tests to ensure proper cleanup.


crates/amaru-ledger/src/state/volatile_db.rs
```diff
@@ -278,7 +279,8 @@ impl AnchoredVolatileState {
                                     ProposalId::from(proposal_id),
                                     proposals::Value {
                                         proposed_in,
-                                        valid_until: epoch + GOV_ACTION_LIFETIME,
+                                        valid_until: epoch
+                                            + protocol_parameters.gov_action_lifetime as Epoch,
```
jeluard: TODO will be improved in a subsequent PR by introducing `Epoch` and `EpochInterval` struct.

crates/amaru/src/stages/ledger.rs
```diff
@@ -109,18 +127,39 @@ impl<S: Store + Send, HS: HistoricalStores + Send> ValidateBlockStage<S, HS> {
...
         let block = parse_block(&raw_block[..]).context("Failed to parse block")?;
+        let slot = point.slot_or_default();
+        let current_epoch = self
+            .state
+            .era_history
+            .slot_to_epoch(slot)
+            .map_err(|e| StateError::ErrorComputingEpoch(slot, e))?;
+
+        // Refresh from the store as those may change every epoch
+        // TODO make sure it's only refreshed when needed
```
jeluard: Will be modified once the governance based update is introduced.

Makefile
```diff
@@ -9,9 +13,7 @@ LISTEN_ADDRESS ?= 0.0.0.0:0
 LEDGER_DIR ?= ./ledger.db
 CHAIN_DIR ?= ./chain.db
 
-.PHONY: help bootstrap run import-snapshots import-headers import-nonces download-haskell-config  coverage-html coverage-lconv check-llvm-cov
-
-all: help
+.PHONY: help bootstrap run import-snapshots import-headers import-nonces download-haskell-config coverage-html coverage-lconv check-llvm-cov
```
KtorZ, line 17: Accident or done on purpose?

jeluard: `make` already uses the first target as [default](https://www.gnu.org/software/make/manual/html_node/How-Make-Works.html)

crates/amaru-consensus/src/consensus/validate_header.rs
```diff
@@ -36,9 +36,10 @@ pub fn header_is_valid(
     raw_header_body: &[u8],
     epoch_nonce: &Nonce,
     ledger: &dyn HasStakeDistribution,
+    global_parameters: &GlobalParameters,
 ) -> Result<(), ConsensusError> {
-    let active_slot_coeff: FixedDecimal =
-        FixedDecimal::from(1_u64) / FixedDecimal::from(ACTIVE_SLOT_COEFF_INVERSE as u64);
+    let active_slot_coeff: FixedDecimal = FixedDecimal::from(1_u64)
+        / FixedDecimal::from(global_parameters.active_slot_coeff_inverse as u64);
```
KtorZ, line 42: Feels like this could be a method on `GlobalParamters`

```rust
global_parameters.active_slot_coeff()
```

crates/amaru-kernel/src/protocol_parameters.rs
```diff
@@ -22,38 +31,393 @@ pub struct ProtocolParameters {
...
+}
+
+fn decode_u16(d: &mut Decoder<'_>) -> Result<u16, cbor::decode::Error> {
+    #[allow(clippy::wildcard_enum_match_arm)]
+    match d.datatype()? {
+        cbor::data::Type::U8 => Ok(d.u8()? as u16),
+        cbor::data::Type::U16 => Ok(d.u16()?),
+        _ => Err(cbor::decode::Error::message("Expected u8/u16 type")),
+    }
 }
```
KtorZ: Not so sure about this. I believe I understand the rationale, but we need to ensure that the decoding behavior is identical to the Haskell node. If the Haskell node fails to decode u8 as u16, we should to.

crates/amaru-kernel/src/protocol_parameters.rs
```diff
@@ -22,38 +31,393 @@ pub struct ProtocolParameters {
...
+    d.array()?;
+    let numerator = decode_u64(d)?;
+    let denominator = decode_u64(d)?;
+    Ok(RationalNumber {
+        numerator,
+        denominator,
+    })
+}
+
+impl<'b, C> cbor::decode::Decode<'b, C> for ProtocolParameters {
```
KtorZ, line 76: I feel like this structure is perhaps "mechanical enough" that we can leverage the macro annotation for writing the encoders/decoders (and reduce the boilerplate here)? Although, I think the downside is that we have to order fields in the Rust structure in the same way they are serialised (loosing the ability to nicely group them for readability). 

So unsure. I think I value more the Rust readability than reducing a bit of boilerplate (which is now already written anyway :D !)

crates/amaru-kernel/src/protocol_parameters.rs
```diff
@@ -276,9 +641,189 @@ impl Default for ProtocolParameters {
...
+        prop_cbor_roundtrip,
+        protocol_parameters::{
+            CostModels, DrepThresholds, PoolThresholds, Prices, ProtocolParameters,
+            ProtocolParametersThresholds,
+        },
+        Coin, Epoch, ExUnits, RationalNumber,
+    };
+    use proptest::prelude::*;
+
+    prop_cbor_roundtrip!(ProtocolParameters, any_protocol_paramater());
```
KtorZ, line 646: o/

crates/amaru-ledger/src/state.rs
```diff
@@ -60,15 +60,15 @@ const EVENT_TARGET: &str = "amaru::ledger::state";
...
+///   top of the _stable_ store. It contains at most 'GlobalParameters::consensus_security_param' entries; old entries
 ///   get persisted in the stable storage when they are popped out of the volatile state.
 pub struct State<S, HS>
 where
     S: Store,
     HS: HistoricalStores,
 {
     /// A handle to the stable store, shared across all ledger instances.
-    stable: Arc<Mutex<S>>,
+    pub stable: Arc<Mutex<S>>,
```
KtorZ: Now that's suspicious 👀

crates/amaru-ledger/src/state.rs
```diff
@@ -364,26 +369,53 @@ impl<S: Store, HS: HistoricalStores> State<S, HS> {
...
+// consumed for the rewards calculation.
+//
+// We always hold on two stake distributions:
+//
+// - The one from an epoch `e - 1` which is used for the ongoing leader schedule at epoch `e + 1`
+// - The one from an epoch `e - 2` which is used for the rewards calculations at epoch `e + 1`
+//
+// Note that the most recent snapshot we have is necessarily `e`, since `e + 1` designates
+// the ongoing epoch, not yet finished (and so, not available as snapshot).
+pub fn stake_distributions(
```
```suggestion
pub fn initial_stake_distributions(
```

This work is really only meant to happen once, on start-up.

crates/amaru-ledger/src/store.rs
```diff
@@ -67,6 +68,9 @@ pub enum StoreError {
 // ----------------------------------------------------------------------------
 
 pub trait ReadOnlyStore {
+    /// Get the current protocol parameters for a given epoch, or most recent one
+    fn get_protocol_parameters_for(&self, epoch: &Epoch) -> Result<ProtocolParameters, StoreError>;
```
KtorZ, line 72: Not sure about this API; time will tell but I don't foresee we'll ever need to lookup protocol parameters from previous epochs.

crates/amaru-ledger/src/summary/rewards.rs
```diff
@@ -166,10 +166,16 @@ impl PoolState {
...
+        &self,
+        available_rewards: Lovelace,
+        total_stake: Lovelace,
+        protocol_parameters: &ProtocolParameters,
+    ) -> Lovelace {
         let one = SafeRatio::one();
-        let a0 = &*PLEDGE_INFLUENCE;
-        let z0 = safe_ratio(1, OPTIMAL_STAKE_POOLS_COUNT as u64);
+        let a0 = protocol_parameters.pledge_influence.clone();
+        let a0 = safe_ratio(a0.numerator, a0.denominator);
```
KtorZ: That `.clone()` seems unnecessary. But the real problem is probably that don't use BigUInt ratio as types in the protocol parameter struct. So we have this conversion going on here for every pool. Now that we have a proper struct, I think we should simply parse and store them as BigUInt ratio when relevant (mostly all economic parameters needed in the rewards calculations).

crates/amaru-stores/src/rocksdb/mod.rs
```diff
@@ -160,44 +163,56 @@ impl RocksDB {
...
-                hex::encode(&key),
-                hex::encode(&value)
-            )
-        });
-        (decoded_key, decoded_value)
-    }))
+    let mut opts = ReadOptions::default();
+    opts.set_prefix_same_as_start(true);
+    Ok(db
+        .iterator_opt(IteratorMode::From(prefix.as_ref(), direction), opts)
```
KtorZ, line 189: Any reason for the change 🤔 ?

crates/amaru-stores/src/rocksdb/mod.rs
```diff
@@ -361,6 +376,21 @@ impl TransactionalContext<'_> for RocksDBTransactionalContext<'_> {
...
     }
 
+    fn set_protocol_parameters(
+        &self,
+        epoch: &Epoch,
+        protocol_parameters: &ProtocolParameters,
+    ) -> Result<(), StoreError> {
+        self.transaction
+            .put(
+                format!("{PROTOCOL_PARAMETERS_PREFIX}:{epoch}"),
```
KtorZ, line 386: Starts to feel like it should be in a separate _column_

crates/amaru-kernel/src/protocol_parameters.rs
jeluard: Apparently CBOR already support using smaller value natively and most decoder (including our own) handle the upcast at read time natively. So this code was useless :)

crates/amaru-ledger/src/store.rs
jeluard: We don't have a protocol params for each `Epoch` so this call must find the closest one (not implemented yet).

crates/amaru-stores/src/rocksdb/mod.rs
jeluard: This is in preparation for the implementation of `get_protocol_parameters_for `

PR #205, opened 2025-05-02
chore: move ledger state management logic out of the store

KtorZ:
This happened mostly out of convenience but was plain wrong in terms of dispatching responsibilities. We shall strive to keep the store as dumb as possible and have higher-level functions combining the stores primitives in the ledger.

This is now much cleaner.

## Summary by CodeRabbit

- **Refactor**
  - Streamlined and modularized epoch transition operations, improving clarity and maintainability of state updates at epoch boundaries.
  - Updated refund logic to process refunds individually instead of in batches.
  - Enhanced handling of missing data in the ledger, defaulting to safe values instead of errors.
  - Adjusted internal state reset and update mechanisms for improved reliability.
- **Bug Fixes**
  - Improved error handling to prevent panics when ledger data is missing.
- **Chores**
  - Simplified and cleaned up internal interfaces and removed unused methods.

### DIFF
--- a/crates/amaru-ledger/src/state.rs
+++ b/crates/amaru-ledger/src/state.rs
@@ -19,7 +19,10 @@ pub mod volatile_db;
 
 use crate::{
     state::volatile_db::{StoreUpdate, VolatileDB},
-    store::{EpochTransitionProgress, HistoricalStores, Store, StoreError, TransactionalContext},
+    store::{
+        columns::pools, EpochTransitionProgress, HistoricalStores, Store, StoreError,
+        TransactionalContext,
+    },
     summary::{
         governance::{self, GovernanceSummary},
         rewards::RewardsSummary,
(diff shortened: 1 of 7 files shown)


crates/amaru-ledger/src/state.rs
```diff
@@ -433,28 +440,42 @@ fn epoch_transition(
...
-        rewards_summary.delta_treasury(),
-        rewards_summary.delta_reserves(),
-        rewards_summary.unclaimed_rewards(),
-    )?;
+    // Adjust treasury and reserves accordingly.
+    db.with_pots(|mut row| {
+        let pots = row.borrow_mut();
+        pots.treasury += rewards_summary.delta_treasury() + rewards_summary.unclaimed_rewards();
+        pots.reserves -= rewards_summary.delta_reserves();
+    })?;
```
KtorZ, line 466: Note: I've inlined the `apply_rewards` function in here; otherwise, we would simply have something along the lines of:

```
fn end_epoch<'store>(db) {
   apply_rewards(db)
}
```

If we ever get to a point where we have more end of epoch operations, we could consider factoring them back into functions.

crates/amaru-stores/src/rocksdb/ledger/columns/pots.rs
```diff
@@ -21,11 +21,11 @@ pub const PREFIX: [u8; PREFIX_LEN] = [0x70, 0x6f, 0x74, 0x73];
 
 #[allow(clippy::panic)]
```
jeluard:
```suggestion
```

jeluard (approved):

crates/amaru-stores/src/rocksdb/mod.rs
```diff
@@ -351,71 +351,14 @@ impl TransactionalContext<'_> for RocksDBTransactionalContext<'_> {
...
-        self.with_pots(|mut row| {
-            let pots = row.borrow_mut();
-            pots.treasury += delta_treasury + unclaimed_rewards;
-            pots.reserves -= delta_reserves;
-        })
+        credential: &scolumns::accounts::Key,
+        deposit: Lovelace,
+    ) -> Result<Lovelace, StoreError> {
+        accounts::set(&self.transaction, credential, |balance| balance + deposit)
     }
```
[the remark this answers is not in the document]
KtorZ: Please take note that the maximum lovelace supply is capped at 45 billion, so there's no scenario whatsoever where the balance of an account can overflow a u64 (since it can't grow beyond 45B -- which is already quite unlikely and probably the end of the network anyway as it would mean someone has accumulated ALL the money and no economy is possible from that point...)

PR #209, opened 2025-05-09
More Output Validations

yHSJ:
This is a small PR, introducing a couple more validations on a transaction output:

- **`InvalidOutput::WrongNetwork`**
   - Validate that the network present in an address matches the network the node is running on
-   **`InvalidOutput::ValueTooLarge`**
    - Validate that the value of the output is less than `max_val_size`, a protocol parameter

## Summary by CodeRabbit

- **New Features**
	- Added network ID validation for transaction outputs to ensure they match the expected network.
	- Introduced error reporting for outputs that exceed the maximum allowed value size.
	- Enhanced error visibility by printing validation errors during block processing.
	- Provided unified interfaces for extracting addresses and network information from various address types.

- **Bug Fixes**
	- Improved consistency in address and network extraction across different address types.

- **Tests**
	- Expanded test coverage for outputs with incorrect network IDs and oversized values.


yHSJ:
There is an issue here with byron era network IDs. Do not merge yet

crates/amaru-kernel/src/protocol_parameters.rs
```diff
@@ -7,7 +7,7 @@ pub struct ProtocolParameters {
     pub max_header_size: u32,
     pub max_tx_ex_units: ExUnits,
     pub max_block_ex_units: ExUnits,
-    pub max_val_size: u32,
```
jeluard: `u32` looks like the correct value, why the change?

crates/amaru-ledger/src/rules/transaction.rs
```diff
@@ -85,6 +85,9 @@ pub fn execute(
     transaction_witness_set: &MintedWitnessSet<'_>,
     transaction_auxiliary_data: Option<&AuxiliaryData>,
 ) -> Result<(), InvalidTransaction> {
+    // FIXME: this is temporary, to be replaced when we have some state that determines the node's network
+    let network = Network::Testnet;
```
jeluard, line 89: `Network` is accessible from the [config](https://github.com/pragma-org/amaru/blob/main/crates/amaru/src/stages/mod.rs#L79). Make sure it is passed correctly from there.

yHSJ: Thank you!

crates/amaru-kernel/src/protocol_parameters.rs
yHSJ: it's a maximum size in bytes, I felt `usize` more appropriately conveyed that

jeluard: [usize](https://doc.rust-lang.org/std/primitive.usize.html) is platform dependent (32 bits on 32 bits system, 64 on 64, ..) so it doesn't look right here. Given this is something that we get from network and others node (via updates, snapshots, ..), it probably is best to stick with what is described in the [CDDL](https://github.com/IntersectMBO/cardano-ledger/blob/2d1e94cf96d00ba0da53883c388fa0aba6d74624/eras/conway/impl/cddl-files/conway.cddl#L373) (so `u32` in this case?)

I tried to align all values accordingly in this [PR](https://github.com/pragma-org/amaru/pull/204)

crates/amaru-ledger/src/rules/transaction.rs
yHSJ: I guess `network` has to be added to some state for this stage which feels like it should be a different PR.

crates/amaru-kernel/src/protocol_parameters.rs
yHSJ: Fair enough, I will change it back. I don't love having to do type conversions, but I think I agree it makes more sense for the `ProtocolParameters` type to match the CDDL

PR #212, opened 2025-05-12
chore: simplify readme

jeluard:
Simplify `README` instructions to bootstrap an`amaru` node.
Also move CI `demo` to `release` mode to improve stability and reduce processing time.

## Summary by CodeRabbit

- **Chores**
  - Updated all relevant build and test commands to use release mode for improved performance.
  - Simplified node bootstrapping in the documentation, replacing multiple manual steps with a single command.
  - Streamlined demo script to run in release mode for optimized execution.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -209,7 +209,7 @@ jobs:
 
       - name: Build Amaru
         run: |
-          cargo test --no-run -p amaru
+          cargo test --release --no-run -p amaru
 
       - name: Cache Amaru's ledger.db
         id: cache-ledger-db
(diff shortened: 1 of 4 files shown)


yHSJ (approved):
Thanks for simplifying the README!

I'm not sure if we want to use `--release` in the CI actions, as those are generally rather short lived tests and the extra build time may just increase the action duration without providing any benefit.

jeluard:
@yHSJ the main idea with `--release` is to make sure `make test-e2e` (which takes some minutes to run) uses the `release` binary so that it reflects real usage and can be used down the line as a scape goat to detect performance regression (something I plan to work on next).

KtorZ (approved):

PR #217, opened 2025-05-15
chore: keep dependencies uptodate

jeluard:
## Summary by CodeRabbit

- **Chores**
  - Updated several dependency versions to their latest releases.
  - Removed the `envpath` crate from development dependencies in multiple components.
  - Added a clarifying comment regarding `minicbor` version alignment.
  - Added new workspace dependencies `iter-borrow` and `rocksdb` for development.
- **Tests**
  - Updated test setup to use a different method for creating temporary directories, ensuring directories persist as needed.
  - Introduced new test modules to expand iteration and summary test coverage.
  - Improved test iterator handling for better integration with new dependencies.


crates/amaru-stores/src/rocksdb/consensus.rs
```diff
@@ -105,7 +105,7 @@ mod test {
     #[test]
     fn rocksdb_chain_store_can_get_what_it_puts() {
         let tempdir = tempfile::tempdir().unwrap();
-        let basedir = tempdir.path().join("rocksdb_chain_store");
+        let basedir = tempdir.keep().join("rocksdb_chain_store");
```
KtorZ: I don't think we want `keep` here; this used to be `into_path` (now deprecated), and I recently turned it to `path` (although the documentation wrongly suggests to use `keep` instead). Could it be that this change now comes from a merge-conflict resolution?

jeluard: Right I blindly followed the deprecation message and you are probably right that I screwed some merge.
Was it that we were incorrectly using `into_path` then? Naming is pretty confusing.

KtorZ: We were using `into_path` indeed; but taking ownership wasn't needed. So `path` is sufficient.

jeluard: [into_path](https://github.com/Stebalien/tempfile/blob/95540ed3fcb9ca74845c02aee058726b2dca58b7/src/dir/mod.rs#L410) prevents the dir to be deleted automatically (as `keep` now does) so I guess the original usage wasn't expected.

PR #218, opened 2025-05-16
Missing Required Datums Validation

yHSJ:
TODO:
- [x] Add Tests for new script validations

## Summary by CodeRabbit

- **New Features**
  - Enhanced validation and tracking of datums in transaction outputs and scripts, including support for supplemental datums.
  - Improved error reporting for missing, extraneous, or unspendable datums in scripts.
  - Added support for serializing and deserializing datum options in transaction outputs.
  - Introduced context-aware handling of supplemental datums during transaction validation.
  - Extended transaction output processing to integrate context-aware datum handling.
  - Added methods to track allowed supplemental datums within transaction witness sets.
  - Introduced new representations for script references paired with hashes to improve script management.

- **Bug Fixes**
  - Corrected and expanded test data for transaction contexts, including UTXO entries and addresses.

- **Tests**
  - Added and updated JSON test data to cover new datum handling scenarios and script validation cases.
  - Improved test coverage for unspendable inputs and datum validation.

- **Refactor**
  - Streamlined reference input handling and context management for transaction validation.
  - Optimized script and datum witness validation by resolving inputs upfront and refining set operations.
  - Updated transaction output and collateral return output processing to pass context explicitly.


crates/amaru-ledger/src/rules/transaction/scripts.rs
```diff
@@ -1,59 +1,115 @@
...
+            context
+                // We assume that the input exists as that's validated during the inputs validation
+                .lookup(input)
+                .map(|output| (input, output))
+        })
+        .collect::<Vec<_>>();
+
+    let empty_vec = vec![];
+    let resolved_reference_inputs = reference_inputs
+        .unwrap_or(&empty_vec)
```
jeluard, line 74: Always hurt my eyes now I don't see how to improve this :D

crates/amaru-ledger/src/rules/transaction/scripts.rs
```diff
@@ -1,59 +1,115 @@
...
+        .collect::<Vec<_>>();
+
     // provided reference scripts from inputs and reference inputs only include ScriptRefs that are required by an input
     let provided_reference_scripts = [
-        reference_inputs.unwrap_or(&vec![]).as_slice(),
-        inputs.as_slice(),
+        resolved_inputs.as_slice(),
+        resolved_reference_inputs.as_slice(),
     ]
     .concat()
```
jeluard, line 92: Looks like some intermediary `collect` and `concat` (allocation in general) could be avoided by keeping iterators alive longer?

yHSJ: Almost certainly. I'm working through a set of validations that will make up this `scripts.rs` in separate PRs, then I will go back through and cleanup/improve this. My first focus is just getting *something* working with some tests so that I can more effectively optimize.

crates/amaru-kernel/src/lib.rs
```diff
@@ -213,6 +214,30 @@ impl RedeemersExt for Redeemers {
     }
 }
 
+#[derive(Eq, PartialEq)]
+pub struct ScriptRefWithHash {
+    pub hash: ScriptHash,
+    pub script: ScriptRef,
+}
```
KtorZ, line 245: Debatable. I see room for something more holistic here. `WithHash<A>` or `Addressable<A>`. Theoretically, Pallas should also provides the API making this entirely superfluous.

crates/amaru-kernel/src/lib.rs
```diff
@@ -739,7 +767,10 @@ pub fn get_provided_scripts(witness_set: &MintedWitnessSet<'_>) -> BTreeSet<Scri
             .map(|plutus_v1_scripts| {
                 plutus_v1_scripts
                     .iter()
-                    .map(|script| script.script_hash())
+                    .map(|script| ScriptRefWithHash {
+                        hash: script.script_hash(),
+                        script: PseudoScript::PlutusV1Script(script.clone()),
```
KtorZ: I am not comfortable with the idea of cloning scripts here. Those are amongst the largest things we can clone; and this is on the critical path. Isn't a reference an option here?

crates/amaru-kernel/src/lib.rs
```diff
@@ -799,6 +836,41 @@ impl<'b> HasAddress for PseudoTransactionOutput<MintedPostAlonzoTransactionOutpu
...
+        match self {
+            PseudoTransactionOutput::Legacy(transaction_output) => {
+                transaction_output.datum_hash.map(DatumOption::Hash)
+            }
+            PseudoTransactionOutput::PostAlonzo(transaction_output) => transaction_output
+                .datum_option
+                .as_ref()
+                .and_then(|datum| match datum {
+                    PseudoDatumOption::Hash(hash) => Some(DatumOption::Hash(*hash)),
+                    PseudoDatumOption::Data(_) => None,
```
KtorZ: That's a confusing semantic here. The function returns an `Option<DatumOption>` and is called `has_datum`, but would yield `None` when the underlying datum is actually inline? If that's an oversight, let's fix it. If that's the intended behavior, then that function needs both another signature and a different name.

Seems like it always returns a hash; so we could fancy `-> Option<Hash<32>>` as a return annotation. However, the whole trait implementation is inconsistent with the one for TransactionOutput.

I am confused.

crates/amaru-kernel/src/serde_utils.rs
```diff
@@ -107,3 +110,28 @@ impl From<ScriptRefProxy> for ScriptRef {
...
+    fn from(value: DatumOptionProxy) -> Self {
+        match value {
+            DatumOptionProxy::Hash(bytes) => DatumOption::Hash(Hash::from(bytes.as_slice())),
+            // This code should only be run during tests, so a panic here is fine
+            DatumOptionProxy::Data(data) => {
+                DatumOption::Data(CborWrap(from_cbor(data.deref()).unwrap()))
+            }
+        }
+    }
+}
```
KtorZ, line 137: Perhaps, keep this function safe by defining `TryFrom` and move `unwrap` at call-site where the context makes it reasonable (or not) to be used (and same for `ScriptRefProxy` actually).

crates/amaru-ledger/src/rules/transaction/inputs.rs
```diff
@@ -52,16 +54,24 @@ where
...
-            .collect(),
-        None => Vec::new(),
-    };
+    let mut intersection = Vec::new();
 
+    if let Some(reference_inputs) = reference_inputs {
+        for reference_input in reference_inputs {
+            // Non-disjoint reference inputs
+            if inputs.contains(reference_input) {
+                intersection.push(reference_input.clone());
```
KtorZ, line 63: It should actually be acceptable for this function to take ownership of the reference inputs and consume them here to avoid cloning. Maybe I am wrong but I can't think of another place where we might need those reference inputs?

crates/amaru-ledger/src/rules/transaction/inputs.rs
```diff
@@ -52,16 +54,24 @@ where
...
+            }
+
+            if let Some(output) = context.lookup(reference_input) {
+                if let Some(DatumOption::Hash(hash)) = output.has_datum() {
+                    context.allow_supplemental_datum(hash);
```
KtorZ: 👍

crates/amaru-ledger/src/rules/transaction/outputs.rs
```diff
@@ -64,8 +71,12 @@ pub fn execute(
         validate_network(&output, network)
             .unwrap_or_else(|element| invalid_outputs.push(WithPosition { position, element }));
 
+        if let Some(DatumOption::Hash(hash)) = output.has_datum() {
+            context.allow_supplemental_datum(hash);
```
KtorZ: 👍

crates/amaru-ledger/src/rules/transaction/outputs.rs
```diff
@@ -50,12 +53,16 @@ pub enum InvalidOutput {
...
 
-pub fn execute(
+pub fn execute<C>(
+    context: &mut C,
     protocol_parameters: &ProtocolParameters,
     network: &Network,
     outputs: Vec<MintedTransactionOutput<'_>>,
-    yield_output: &mut impl FnMut(u64, TransactionOutput),
-) -> Result<(), InvalidOutputs> {
+    yield_output: &mut impl FnMut(&mut C, u64, TransactionOutput),
```
KtorZ: Doesn't sound right. Why are we yielding back the context to a caller that already has it?

crates/amaru-ledger/src/rules/transaction/scripts.rs
```diff
@@ -1,59 +1,115 @@
...
+        provided.iter().map(|hash| hash.to_string()).collect::<Vec<_>>().join(", "),
+    )]
+    MissingRequiredDatums {
+        missing: Vec<Hash<32>>,
+        provided: BTreeSet<Hash<32>>,
+    },
+    #[error(
+        "extraneous supplemental datums: allowed: [{}] provided [{}]",
+        allowed.iter().map(|hash| hash.to_string()).collect::<Vec<_>>().join(", "),
+        provided.iter().map(|hash| hash.to_string()).collect::<Vec<_>>().join(", "),
```
KtorZ:
```rust
.iter().map(|hash| hash.to_string()).collect::<Vec<_>>().join(", ")
```

is repeated 6 times in those 40 lines; perhaps consider moving it as a separate function?

crates/amaru-ledger/src/rules/transaction/scripts.rs
```diff
@@ -1,59 +1,115 @@
...
 {
     let required_scripts = context.required_scripts();
 
+    let resolved_inputs = inputs
+        .iter()
+        .filter_map(|input| {
+            context
+                // We assume that the input exists as that's validated during the inputs validation
+                .lookup(input)
+                .map(|output| (input, output))
```
KtorZ: Then, let's not use `filter_map` but, `map` + `unreachable!` or at the very least `debug_assert!`. Those kind of "assumptions" are important and shall be explicit rather than silently swallowed. Unfortunately, the comment won't make it loudly fail in case we breach that contract.

crates/amaru-ledger/src/rules/transaction/scripts.rs
```diff
@@ -1,59 +1,115 @@
...
+        })
+        .collect::<Vec<_>>();
+
+    let empty_vec = vec![];
+    let resolved_reference_inputs = reference_inputs
+        .unwrap_or(&empty_vec)
+        .iter()
+        .filter_map(|input| {
+            context
+                // We assume that the reference input exists as that's validated during the inputs validation
```
KtorZ: same

crates/amaru-ledger/src/rules/transaction/scripts.rs
KtorZ: Use https://go.dev/

crates/amaru-ledger/src/rules/transaction/scripts.rs
```diff
@@ -1,59 +1,115 @@
...
+                            Some(ScriptRefWithHash {
+                                hash: script_ref.script_hash(),
+                                script: script_ref.clone(),
+                            })
+                        } else {
+                            None
+                        }
+                    })
+            }
+            amaru_kernel::PseudoTransactionOutput::Legacy(_) => None,
```
KtorZ: I know it's just re-indenting, but it now occurs to me that this feels like a wrong abstraction level here. I would hope that the ledger rules can be free of serialisation details (i.e. we shall never have to pattern match on a `PostAlonzo` struct in here). 

So there's room for moving stuff out as helper function or trait.

crates/amaru-ledger/src/rules/transaction/scripts.rs
```diff
@@ -72,13 +133,110 @@ where
     }
 
     let extra_scripts: Vec<ScriptHash> = provided_scripts
+        .iter()
+        .map(ScriptHash::from)
+        .collect::<BTreeSet<_>>()
         .difference(&required_scripts)
         .cloned()
         .collect();
```
KtorZ: double clone + iteration; can be simplified:

```rs
    let extra_scripts: Vec<ScriptHash> =
        provided_scripts
            .iter()
            .fold(Vec::new(), |mut accum, script| {
                if !required_scripts.contains(&script.hash) {
                    accum.push(script.hash);
                }

                accum
            });
```

crates/amaru-ledger/src/rules/transaction/scripts.rs
```diff
@@ -72,13 +133,110 @@ where
...
+
+    let mut input_datum_hashes: BTreeSet<Hash<32>> = BTreeSet::new();
+    let mut inputs_missing_datum: Vec<&TransactionInput> = Vec::new();
+
+    required_script_inputs
+        .into_iter()
+        .for_each(|((input, output), script)| match output.has_datum() {
+            None => {
+                if !(matches!(script.script, PseudoScript::PlutusV3Script(..))
+                    || matches!(script.script, PseudoScript::NativeScript(..)))
```
KtorZ: Dangerous; we might forget to update it later with PlutusV4; I would suggest using a complete pattern match here.

crates/amaru-ledger/src/rules/transaction/scripts.rs
```diff
@@ -1,59 +1,115 @@
...
+        provided: BTreeSet<Hash<32>>,
+    },
 }
 
+// TODO: this can be made MUCH more efficient. Remove clones, don't iterate the same list several times, etc... Lots of low hanging fruit.
 pub fn execute<C>(
     context: &mut C,
     reference_inputs: Option<&Vec<TransactionInput>>,
-    inputs: &Vec<TransactionInput>,
+    inputs: &[TransactionInput],
```
KtorZ, line 51: I don't think these guys should be here as they've already been processed during the `inputs` rules. Rather; while processing inputs (and outputs), we shall register scripts and datums that are found. In the same way we do it for certificates and credentials.

So this particular rule could have immediate access to the scripts present in the transaction body. It may only need the script witness as an argument (and not the whole witness since we check vk witnesses elsewhere).

crates/amaru-ledger/tests/data/transactions/preprod/7a098c13f3fb0119bc1ea6a418af3b9b8fef18bb65147872bf5037d28dda7b7b/unknown-reference-input/expected.traces
```diff
@@ -1,6 +1 @@
-[ {"name":"require_witness_span","credential": {"hash":"5309fa786856c1262d095b89adf64fe8a5255ad19142c9c537359e41","type":"verification_key"}}
-, {"name":"require_witness_span","credential": {"hash":"4e3b75eb68829ca9e63b541f3645f78afe8d119932a7a2528171e749","type":"script"}}
-, {"name":"require_witness_span","credential": {"hash":"4e3b75eb68829ca9e63b541f3645f78afe8d119932a7a2528171e749","type":"script"}}
-, {"name":"require_witness_span","credential": {"hash":"4e3b75eb68829ca9e63b541f3645f78afe8d119932a7a2528171e749","type":"script"}}
-, {"name":"require_witness_span","credential": {"hash":"5309fa786856c1262d095b89adf64fe8a5255ad19142c9c537359e41","type":"verification_key"}}
-]
+[]
```
KtorZ, line 1: 🤔 ?

yHSJ: The ordering of the input validation changed, so now traces aren't happening when there is an err found in the reference inputs

crates/amaru-ledger/src/rules/transaction/scripts.rs
yHSJ: I agree -- but I think that this will be addressed in future PRs. I've coupled 3 different validation rules together for the `scripts.rs` file, and I'm working on finishing up the redeemers now. My plan was to go back and improve all of this once I have them all written up with tests, so it's easier to tell what I'm breaking.

crates/amaru-ledger/src/rules/transaction/scripts.rs
yHSJ: Good call, I had that at first but decided it was obnoxious. But "wen PlutusV4?" I guess :P

crates/amaru-ledger/src/rules/transaction/outputs.rs
yHSJ: Because of ownership issues -- the `outputs::execute` function needs to use the `context` to allow supplemental datums. But then the yield output function can't also call the context, so we pass it to the argument function.

I'd argue that's a code smell, but I just wanted to get it "working" for now

crates/amaru-ledger/src/rules/transaction/scripts.rs
yHSJ: I agree with this. I'll write up a trait

crates/amaru-ledger/src/rules/transaction/scripts.rs
KtorZ: I insist <3 

And it's a rule of thumb for pretty much everything; avoid wildcards and partial pattern match whenever possible. It makes the code harder to maintain in the long run and it's only usually one line shorter not to.

crates/amaru-ledger/src/rules/transaction/inputs.rs
yHSJ: I'm not sure I agree with this yet. Once we refactor the logic to move script references and datums to the context I'd agree, but we haven't done that yet and I don't want to overload this PR.

I'm also not particularly worried about a clone in this case, as it will immediately error after this, so we're not cloning a significant amount of data (maximum of some relatively small number of reference inputs)

crates/amaru-kernel/src/lib.rs
yHSJ: I agree that we shouldn't clone the scripts.

A reference isn't ideal, because in the witness set we have `PlutusScript<N>` and `NativeScript` types, so I have to construct the enum variant (which requires owned data). I can create a `BorrowedPseudoScript` enum which allows the scripts to be borrowed, but it's a pretty unpleasant solution in my opinion. And unfortunately there is no shared trait between all the script variants that I can rely on. I guess we could write that ourselves, but it will mostly be duplicate work. Feels like this is something we should be committing upstream to Pallas, once we decide on an API that makes sense.

crates/amaru-kernel/src/lib.rs
yHSJ: I think that we should define a Pallas API that removes the need for this. That being said, I think this is "fine"

crates/amaru-kernel/src/lib.rs
yHSJ: You're right, looking back on code I wrote a week ago I got confused 😅 

The idea is this returns the DatumHash, if present, and the `TransactionOutput` implementation is wrong. Will clean this up.

crates/amaru-kernel/src/serde_utils.rs
yHSJ: The `HasProxy` logic requires a `From`, which is why I went with this solution. I don't love the moving of the `unwrap` to the call-site, as it only moves it one level up to the `OutputProxy`, which doesn't handle a result. We could refactor the whole thing, which feels more involved, or we could leave it as is as this code is only used in test and it doesn't *really* matter if this function is safe. Happy to do either one, *more* happy to leave it as is though 😛

crates/amaru-ledger/src/rules/transaction/outputs.rs
KtorZ: I am not convinced. "Getting it working for now" isn't a valid excuse 😶 (I am exaggerating of course, no big deal here but still, not convinced!). There's no such thing as a "temporary work-around". So I am afraid this would stay. 

At the very least, please add a TODO.

yHSJ: Cleaned it up. Just gonna do any context updating in the `yield_output` function.

crates/amaru-kernel/src/lib.rs
jeluard: Would it make sense to create a pallas issue to track this?

crates/amaru-kernel/src/serde_utils.rs
jeluard: If it's only for tests we might want to hide it behind a `#[cfg(test)]` annotation.

crates/amaru-ledger/src/rules/transaction/scripts.rs
jeluard: Would be worth detailing in the PR summary what is planned to be done in subsequent PRs.

crates/amaru-ledger/src/rules/transaction/outputs.rs
```diff
@@ -64,8 +72,23 @@ pub fn execute(
...
+        // supplemental datums for outputs (regardless of whether transaction fails or not).
+        //
+        // In particular, any datum present in a collateral return does NOT count towards the
+        // allowed supplemental datums.
+        //
+        // However, I am not fixing this now, because we have no test covering the case whatsoever.
+        // At the moment, that line can actually be fully removed without making any test fail.
+        if let Some(BorrowedDatumOption::Hash(hash)) = output.datum() {
+            context.allow_supplemental_datum(*hash);
+        }
```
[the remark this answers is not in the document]
KtorZ: See this comment @yHSJ ☝️.

PR #221, opened 2025-05-18
feat: pure-stage add Clock/Wait/Call effects

rkuhn:
Also refactor the implementation to make it safer, more regular, and fully tested. Tokio now also uses the effects, but with its own interpreter for managing time, mailboxes, and task concurrency.

## Summary by CodeRabbit

- **New Features**
  - Introduced a comprehensive effect system for asynchronous stage-based simulations, including structured message passing, timing, call/response, and interruption primitives.
  - Added a new time abstraction with an `Instant` type and utilities for time manipulation and formatting.
  - Enabled effectful asynchronous operations within stage handlers via an `Effects` interface.
  - Added support for call/response patterns with unique call identifiers and response channels.

- **Refactor**
  - Refactored simulation and Tokio runtime to use explicit effect handling, improving control over asynchronous stage execution and timing.
  - Simplified and streamlined stage references, removing direct message sending from `StageRef`.

- **Bug Fixes**
  - Improved error handling and correctness checks in simulation runtime, including robust state transitions and effect validation.

- **Tests**
  - Enhanced and expanded the test suite to cover timing, call/response, and new effect features.

- **Chores**
  - Updated dependencies and workspace configuration.


crates/pure-stage/src/effect.rs
```diff
@@ -0,0 +1,264 @@
...
+/// upon whose resumption the stage waits (in which case T is `()`).
+#[derive(Debug)]
+pub(crate) enum StageEffect<T> {
+    Receive,
+    Send(
+        Name,
+        T,
+        // this is present in case the send is the first part of a call effect
+        Option<(Duration, oneshot::Receiver<Box<dyn Message>>, CallId)>,
+    ),
```
abailly, line 15: seems to me having named fields is clearer as it's probably the case we'll quickly forget exactly what those tuple values are

```suggestion
    Send {
        name: Name,
        msg: T,
        // this is present in case the send is the first part of a call effect
        call_param: Option<(Duration, oneshot::Receiver<Box<dyn Message>>, CallId)>,
    },

```

crates/pure-stage/src/effect.rs
```diff
@@ -0,0 +1,264 @@
...
+    Clock,
+    Wait(Duration),
+    Call(
```
abailly, line 18: Same issue here?

abailly (commented):
That's quite a lot of code to digest 😅 Given it's pretty much self-contained in single crate and we'll only figure out whether it really works when rewiring the pipeline stages, I suggest we just merge and this and proceed with a concrete example to both run simulation and actual tokio-based stages. Seems like @stevana's echo service would be useful here?

PR #222, opened 2025-05-19
feat: add pure-stage simulation of echo

stevana:
This PR implements a `NodeHandle` based on @rkuhn's pure-stage and uses this to communicate with the echo simulator. A test run looks something like this (note that there's a bug inserted on purpose which uppercases the string that gets echoed back every 5th message):

```
$ cargo test simulator::simulate::tests::simulate_pure_stage_echo -- --nocapture
[...]
Found minimal failing case:

  Envelope { src: "c1", dest: "n1", body: Echo { msg_id: 0, echo: "Please echo 0" } }
  Envelope { src: "c1", dest: "n1", body: Echo { msg_id: 0, echo: "Please echo 0" } }
  Envelope { src: "c1", dest: "n1", body: Echo { msg_id: 0, echo: "Please echo 0" } }
  Envelope { src: "c1", dest: "n1", body: Echo { msg_id: 0, echo: "Please echo 0" } }
  Envelope { src: "c1", dest: "n1", body: Echo { msg_id: 0, echo: "Please echo 1" } }

Error message:

  No matching response found for echo request:
    Envelope { src: "c1", dest: "n1", body: Echo { msg_id: 0, echo: "Please echo 1" } }

Trace:
  Envelope { src: "c1", dest: "n1", body: Echo { msg_id: 0, echo: "Please echo 0" } }
  Envelope { src: "n1", dest: "c1", body: EchoOk { msg_id: 1, in_reply_to: 0, echo: "Please echo 0" } }
  Envelope { src: "c1", dest: "n1", body: Echo { msg_id: 0, echo: "Please echo 0" } }
  Envelope { src: "n1", dest: "c1", body: EchoOk { msg_id: 2, in_reply_to: 0, echo: "Please echo 0" } }
  Envelope { src: "c1", dest: "n1", body: Echo { msg_id: 0, echo: "Please echo 0" } }
  Envelope { src: "n1", dest: "c1", body: EchoOk { msg_id: 3, in_reply_to: 0, echo: "Please echo 0" } }
  Envelope { src: "c1", dest: "n1", body: Echo { msg_id: 0, echo: "Please echo 0" } }
  Envelope { src: "n1", dest: "c1", body: EchoOk { msg_id: 4, in_reply_to: 0, echo: "Please echo 0" } }
  Envelope { src: "c1", dest: "n1", body: Echo { msg_id: 0, echo: "Please echo 1" } }
  Envelope { src: "n1", dest: "c1", body: EchoOk { msg_id: 5, in_reply_to: 0, echo: "PLEASE ECHO 1" } }
 at simulation/amaru-sim/src/simulator/simulate.rs:262
```

## Summary by CodeRabbit

- **New Features**
  - Added support for in-process simulation nodes using the pure-stage framework.
  - Introduced new property-based tests for verifying echo message correctness in both in-process and external node scenarios.
  - Extended public API by adding the `Void` entity to the stage module exports.

- **Bug Fixes**
  - Improved error reporting in simulation tests, providing more detailed diagnostics on failures.

- **Chores**
  - Updated dependencies to include the pure-stage crate.

### DIFF
--- a/simulation/amaru-sim/src/simulator/simulate.rs
+++ b/simulation/amaru-sim/src/simulator/simulate.rs
@@ -65,7 +65,6 @@ impl<Msg: PartialEq> Eq for Entry<Msg> {}
 
 type NodeId = String;
 
-// TODO: should be RK's handle to interact with a node
 pub struct NodeHandle {
     handle:
         Box<dyn FnMut(Envelope<EchoMessage>) -> Result<Vec<Envelope<EchoMessage>>, anyhow::Error>>,
@@ -79,7 +78,6 @@ pub fn pure_stage_node_handle(
     mut running: SimulationRunning,
 ) -> anyhow::Result<NodeHandle> {
     let handle = Box::new(move |msg: Envelope<EchoMessage>| {
-        println!("handle called on message:\n  {:?}", msg);
         running.enqueue_msg(&stage, [msg]);
         running.run_until_blocked().assert_idle();
         Ok(rx.drain().collect::<Vec<_>>())


simulation/amaru-sim/src/simulator/simulate.rs
```diff
@@ -262,11 +291,105 @@ mod tests {
         assert_eq!(world.run_world(), &Vec::new());
     }
 
-    // NOTE: This will be replaced with a test that uses an in-process NodeHandle later, which
-    // avoids the problem of locating and spawning the binary.
+    #[test]
+    #[should_panic]
+    fn simulate_pure_stage_echo() {
+        let config = Default::default();
```
rkuhn: Please don’t use `Default::default()`: the inferred type is unavailable here on github. There is no downside to using `WhateverType::default()` instead.

simulation/amaru-sim/src/simulator/simulate.rs
```diff
@@ -262,11 +291,105 @@ mod tests {
...
+        let spawn: fn() -> NodeHandle = || {
+            println!("*** Spawning node!");
+            let mut network = SimulationBuilder::default();
+            let stage = network.stage(
+                "echo",
+                async |(mut state, out), msg: Envelope<EchoMessage>, eff| {
+                    if let EchoMessage::Echo { msg_id, echo } = &msg.body {
+                        state += 1;
+                        // Insert a bug every 5 messages.
+                        let reply;
```
rkuhn: more idiomatic would be

```rust
let reply = if ... { ... } else { ... };
```

simulation/amaru-sim/src/simulator/simulate.rs
```diff
@@ -262,11 +291,105 @@ mod tests {
...
+                        if state % 5 == 0 {
+                            reply = Envelope {
+                                src: msg.dest,
+                                dest: msg.src,
+                                body: EchoMessage::EchoOk {
+                                    msg_id: state,
+                                    in_reply_to: *msg_id,
+                                    echo: echo.to_string().to_uppercase(),
```
rkuhn: I’d compute `echo` before constructing the Envelope, limiting the scope of the `if` (less code to check for differences)

simulation/amaru-sim/src/simulator/simulate.rs
```diff
@@ -69,6 +72,24 @@ pub struct NodeHandle {
     close: Box<dyn FnMut()>,
 }
 
+#[allow(dead_code)]
```
rkuhn: this should be `#[cfg(test)]` instead — this is the reason for the pattern
```rust
#[cfg(test)]
mod tests {
    ....
}
```
at the end of Rust modules: you can put the infrastructure helpers in there without needing further annotations.

simulation/amaru-sim/src/simulator/simulate.rs
```diff
@@ -262,11 +291,105 @@ mod tests {
...
+            pure_stage_node_handle(rx, stage, running).unwrap()
+        };
+        let generate_message = (0..128u8).prop_map(|i| EchoMessage::Echo {
+            msg_id: 0,
+            echo: format!("Please echo {}", i),
+        });
+        let property = |trace: Trace| {
+            for msg in trace.0.iter().filter(|msg| msg.src.starts_with("c")) {
+                if let EchoMessage::Echo { msg_id, echo } = &msg.body {
+                    let response = trace.0.iter().find(|resp| {
```
rkuhn: if we were serious about this test, we’d want to start looking only after the echo request and stop looking after a certain time has passed in the trace

simulation/amaru-sim/src/simulator/simulate.rs
```diff
@@ -262,11 +291,105 @@ mod tests {
...
+                        resp.dest == msg.src
+                            && if let EchoMessage::EchoOk {
+                                in_reply_to,
+                                echo: resp_echo,
+                                ..
+                            } = &resp.body
+                            {
+                                *in_reply_to == *msg_id && *resp_echo == *echo
```
rkuhn: this can also be writte
```rust
matches!(&resp.body, EchoMessage::EchoOk { in_reply_to, echo: resp_echo, .. }
    if in_reply_to == msg_id && resp_echo == echo
)
```

simulation/amaru-sim/src/simulator/simulate.rs
```diff
@@ -262,11 +291,105 @@ mod tests {
...
+                    if response.is_none() {
+                        let mut err = String::new();
+                        err += &format!(
+                            "No matching response found for echo request:\n    {:?}\n\nTrace:\n",
+                            msg
+                        );
+                        trace
+                            .0
+                            .clone()
+                            .into_iter()
```
rkuhn: better `trace.0.iter().cloned()`, maybe using a `for` loop for readability:
```rust
for envelope in trace.0.iter().cloned() {
    writeln!(&mut err, "  {envelope:?}");
}
```

simulation/amaru-sim/src/simulator/simulate.rs
stevana: I want the library to export this function, so that users of the library can use it. Would it be better to use `#[allow(unused)]` in this case?

simulation/amaru-sim/src/simulator/simulate.rs
stevana: Fixed the first part, added note about stop looking after a certain time passed (as arrival times are not part of the trace yet).

simulation/amaru-sim/src/simulator/simulate.rs
stevana: Fixed, thanks!

simulation/amaru-sim/src/simulator/simulate.rs
stevana: I couldn't use `writeln!` here, but change it to be a for loop.

simulation/amaru-sim/src/simulator/simulate.rs
stevana: Fixed, thanks!

simulation/amaru-sim/src/simulator/simulate.rs
```diff
@@ -69,6 +72,24 @@ pub struct NodeHandle {
...
+        println!("handle called on message:\n  {:?}", msg);
+        running.enqueue_msg(&stage, [msg]);
+        running.run_until_blocked().assert_idle();
+        Ok(rx.drain().collect::<Vec<_>>())
+    });
+
+    let close = Box::new(move || ());
+
+    Ok(NodeHandle { handle, close })
+}
```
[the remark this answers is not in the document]
abailly: @stevana What do you think about this suggestion? Seems sensible to me but I don't know enough of the details of the runtime to know whether or not it makes sense given the fact this call already exists in the `handle` function

simulation/amaru-sim/src/simulator/simulate.rs
```diff
@@ -69,7 +72,25 @@ pub struct NodeHandle {
...
 
-#[allow(dead_code)]
+#[allow(unused)]
+pub fn pure_stage_node_handle(
+    mut rx: Receiver<Envelope<EchoMessage>>,
+    stage: StageRef<Envelope<EchoMessage>, (u64, StageRef<Envelope<EchoMessage>, Void>)>,
+    mut running: SimulationRunning,
+) -> anyhow::Result<NodeHandle> {
+    let handle = Box::new(move |msg: Envelope<EchoMessage>| {
+        println!("handle called on message:\n  {:?}", msg);
```
abailly: I would rather use `tracing` here, it's quite useful to have proper logs even for simulation tests I think?

abailly (commented):
Seems the 🐰 has some interesting suggestions, wdyt @stevana ?

simulation/amaru-sim/src/simulator/simulate.rs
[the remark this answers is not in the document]
stevana: It's not clear to me that it's necessary, also it doesn't typecheck.

simulation/amaru-sim/src/simulator/simulate.rs
stevana: Agreed. Not sure if this particular log line should be there at all though, so removed it for now, thanks!

abailly (approved):

PR #223, opened 2025-05-19
Abailly/add block storage

abailly:


jeluard:
Somehow this PR breaks `wasm` compilation apparently because [getrandom use is introduced](https://github.com/pragma-org/amaru/actions/runs/15123149201/job/42509894403#step:6:96)

abailly:
> Somehow this PR breaks `wasm` compilation apparently because [getrandom use is introduced](https://github.com/pragma-org/amaru/actions/runs/15123149201/job/42509894403#step:6:96)

I think it comes from dependency to tracing-opentelemetry which I added to add `#[instrument]` macro to block store handler, I will move it to the corresponding stage.

PR #224, opened 2025-05-20
chore: make Epoch a type

jeluard:
Make sure `Epoch` is now a type as detailed in this [EDR](https://github.com/pragma-org/amaru/blob/main/engineering-decision-records/009-avoid-primitive-obsession.md)

## Summary by CodeRabbit

- **New Features**
  - Introduced a dedicated `Epoch` type for improved type safety, replacing raw `u64` values for epochs throughout the application.

- **Refactor**
  - Updated method and function signatures, struct fields, and test code to use the new `Epoch` and `Slot` types consistently.
  - Standardized conversions and arithmetic involving epochs and slots for greater clarity and consistency.
  - Replaced implicit conversions with explicit `Epoch::from` and `Slot::from` calls across the codebase.

- **Bug Fixes**
  - Improved type consistency in epoch and slot calculations to reduce potential errors.
  - Corrected epoch-related arithmetic and logging for clearer diagnostics.

- **Chores**
  - Added the new `slot-arithmetic` crate as a dependency and updated workspace configurations.
  - Removed unused dependencies and modularized testing utilities with feature flags.

### DIFF
--- a/crates/slot-arithmetic/src/lib.rs
+++ b/crates/slot-arithmetic/src/lib.rs
@@ -12,10 +12,21 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::fmt::Display;
+#![feature(step_trait)]
+
+use std::{
+    fmt::Display,
+    iter::Step,
+    ops::{Add, Mul, Sub},
+    str::FromStr,
+};
 
 use minicbor::{Decode, Decoder, Encode};
+
 use serde::{Deserialize, Serialize};
+
+#[cfg(test)]
+use proptest::prelude::{Arbitrary, BoxedStrategy, Strategy};
 
 #[derive(Clone, Debug, Copy, PartialEq, PartialOrd, Ord, Eq, Serialize, Deserialize, Default)]
 #[repr(transparent)]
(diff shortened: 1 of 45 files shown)


crates/slot-arithmetic/src/lib.rs
```diff
@@ -74,11 +85,113 @@ impl<'b, C> Decode<'b, C> for Slot {
     }
 }
 
+#[derive(Clone, Debug, Copy, PartialEq, PartialOrd, Ord, Eq, Serialize, Deserialize, Default)]
+#[repr(transparent)]
+pub struct Epoch(u64);
+
+#[cfg(any(test, feature = "test-utils"))]
```
abailly, line 92: Cool, I now know how to make features :)

crates/slot-arithmetic/src/lib.rs
```diff
@@ -74,11 +85,113 @@ impl<'b, C> Decode<'b, C> for Slot {
...
+    }
+}
+
+impl<'b, C> Decode<'b, C> for Epoch {
+    fn decode(d: &mut Decoder<'b>, _ctx: &mut C) -> Result<Self, minicbor::decode::Error> {
+        d.u64().map(Epoch)
+    }
+}
+
+impl Add<u64> for Epoch {
```
abailly, line 144: I think we discussed "domain types" support for arithmetic operations when I implemented this for `Slot` and I vaguely remember we decided to not do this, but can't remember the specifics. I think the argument was that it was better to provided dedicated method where it made sense.

simulation/amaru-sim/src/simulator/ledger.rs
```diff
@@ -100,8 +101,8 @@ impl HasStakeDistribution for FakeStakeDistribution {
             })
     }
 
-    fn slot_to_kes_period(&self, slot: u64) -> u64 {
-        slot / self.slots_per_kes_period
+    fn slot_to_kes_period(&self, slot: Slot) -> u64 {
```
abailly, line 104: Good catch!

abailly (approved):

abailly:
> Regarding @abailly's point; I recall the "dispute" to have been about having arithmetic traits for the type itself (e.g. allowing to add two slots to one another); but here the case is more about arithmetic with scalars which I think could be argue for 🤔

I tend to agree, although...

> (Although to some sense, why would we allow `2 * epoch` but not `epoch + epoch` 😶...)

and does it even make sense to multiply epochs?

jeluard:
> and does it even make sense to multiply epochs?

Clearly it doesn't, and actually it's not used so I removed it. Not sure why it's been added in the first place :)

Adding and substracting `u64` make sense to me, and substracting two `Epoch`s is actually the difference between thm. Could be renamed.

abailly (approved):

PR #225, opened 2025-05-20
chore: make GlobalParameters derived from Network

jeluard:
`GlobalParameters` are stil hardcoded but derived from `NetworkName`.

## Summary by CodeRabbit

- **New Features**
  - Added support for network-specific protocol parameters for Preprod, Preview, and Testnet Cardano networks, ensuring more accurate configuration across environments.

- **Refactor**
  - Updated how global protocol parameters are initialized and accessed throughout the application and tests, removing reliance on default values and ensuring explicit network context is used.
  - Simplified and generalized the structure of protocol parameters by removing legacy fields and focusing on current epoch-wide parameters.
  - Modified simulation and bootstrap processes to consistently use network-based global parameters, enhancing configuration coherence.

- **Bug Fixes**
  - Improved calculation of reward efficiency by using the correct epoch length parameter, leading to more accurate metrics in rewards summaries.

- **Tests**
  - Updated test cases to use network-specific protocol parameters for greater consistency and accuracy.

### DIFF
--- a/crates/amaru-kernel/src/network.rs
+++ b/crates/amaru-kernel/src/network.rs
@@ -215,14 +215,54 @@ static PREPROD_GLOBAL_PARAMETERS: LazyLock<GlobalParameters> = LazyLock::new(|| 
     }
 });
 
+static PREVIEW_GLOBAL_PARAMETERS: LazyLock<GlobalParameters> = LazyLock::new(|| {
+    let consensus_security_param = 432;
+    let active_slot_coeff_inverse = 20;
+    let epoch_length_scale_factor = 10;
+    let epoch_length =
+        active_slot_coeff_inverse * epoch_length_scale_factor * consensus_security_param;
+    GlobalParameters {
+        consensus_security_param,
+        epoch_length_scale_factor,
+        active_slot_coeff_inverse,
+        max_lovelace_supply: 45_000_000_000_000_000,
+        slots_per_kes_period: 129_600,
+        max_kes_evolution: 62,
+        epoch_length,
+        stability_window: active_slot_coeff_inverse * consensus_security_param * 2,
+        randomness_stabilization_window: (4 * consensus_security_param * active_slot_coeff_inverse)
+            as u64,
+    }
+});
+
+static TESTNET_GLOBAL_PARAMETERS: LazyLock<GlobalParameters> = LazyLock::new(|| {
+    let consensus_security_param = 432;
+    let active_slot_coeff_inverse = 20;
+    let epoch_length_scale_factor = 10;
+    let epoch_length =
+        active_slot_coeff_inverse * epoch_length_scale_factor * consensus_security_param;
+    GlobalParameters {
+        consensus_security_param,
+        epoch_length_scale_factor,
+        active_slot_coeff_inverse,
+        max_lovelace_supply: 45_000_000_000_000_000,
+        slots_per_kes_period: 129_600,
+        max_kes_evolution: 62,
+        epoch_length,
+        stability_window: active_slot_coeff_inverse * consensus_security_param * 2,
+        randomness_stabilization_window: (4 * consensus_security_param * active_slot_coeff_inverse)
+            as u64,
+    }
+});
+
 #[allow(clippy::todo)]
 impl From<NetworkName> for &GlobalParameters {
     fn from(value: NetworkName) -> Self {
         match value {
             NetworkName::Mainnet => todo!(),
             NetworkName::Preprod => &PREPROD_GLOBAL_PARAMETERS,
-            NetworkName::Preview => todo!(),
-            NetworkName::Testnet(_) => todo!(),
+            NetworkName::Preview => &PREVIEW_GLOBAL_PARAMETERS,
+            NetworkName::Testnet(_) => &TESTNET_GLOBAL_PARAMETERS,
         }
     }
 }
--- a/crates/amaru-kernel/src/network.rs
+++ b/crates/amaru-kernel/src/network.rs
@@ -198,25 +198,18 @@ static PREPROD_GLOBAL_PARAMETERS: LazyLock<GlobalParameters> = LazyLock::new(|| 
     // https://cips.cardano.org/cip/CIP-9
     let consensus_security_param = 2160;
     let active_slot_coeff_inverse = 20;
-    let shelley_epoch_length_scale_factor = 10;
-    let shelley_epoch_length =
-        active_slot_coeff_inverse * shelley_epoch_length_scale_factor * consensus_security_param;
-    let byron_epoch_length_scale_factor = 10;
-    let byron_epoch_length = byron_epoch_length_scale_factor * consensus_security_param;
-    let shelley_transition_epoch = 4;
+    let epoch_length_scale_factor = 10;
+    let epoch_length =
+        active_slot_coeff_inverse * epoch_length_scale_factor * consensus_security_param;
     GlobalParameters {
         consensus_security_param,
-        shelley_epoch_length_scale_factor,
+        epoch_length_scale_factor,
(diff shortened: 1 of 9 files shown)


jeluard:
@abailly I believe that this PR should break `amaru-sim` but I understand it's not part of the CI? Would it be worth adding it?
How can I manually check that it still runs fine?

abailly:
Yes, we should probably add it to the CI but to properly test it would mean running the generator that itself spawns the SUT, which currently requires building the `simulation-testing` project which is in Haskell. @stevana is in the process of porting the simulator to Rust but we are not there yet. 
That said, compiling the project should be straightforward  (few dependencies) so perhaps worth starting it.

abailly:
To manually check it, you need to:

1.  `git clone https://github.com/pragma-org/simulation-testing`
2. `cd simulation-testing/moskstraumen`
3. (assuming you have a Haskell toolchain installed) `cabal build all`
4. something like `RUST_BACKTRACE=1 cabal run blackbox-test -- /Users/arnaud/projects/amaru/amaru/./target/debug/simulator amaru 1 --peer-address=127.0.0.1:3000  --stake-distribution-file data/stake.json`

.github/workflows/CI.yml
```diff
@@ -285,3 +285,29 @@ jobs:
...
+      - name: Build moskstraumen
+        run: |
+          git clone https://github.com/pragma-org/simulation-testing
+          cd simulation-testing/moskstraumen
+          cabal build all
+
+      - name: Run simulation
+        run: |
+          RUST_BACKTRACE=1 cabal run blackbox-test -- ../../target/debug/simulator amaru 1 --peer-address=127.0.0.1:3001  --stake-distribution-file data/stake.json
+        working-directory: simulation-testing/moskstraumen
```
KtorZ: Seems unrelated to the PR?

crates/amaru-kernel/src/network.rs
```diff
@@ -192,6 +194,46 @@ impl From<NetworkName> for &EraHistory {
...
+        slots_per_kes_period: 129_600,
+        max_kes_evolution: 62,
+        shelley_epoch_length,
+        stability_window: active_slot_coeff_inverse * consensus_security_param * 2,
+        byron_epoch_length,
+        byron_total_slots: byron_epoch_length * shelley_transition_epoch,
+        randomness_stabilization_window: (4 * consensus_security_param * active_slot_coeff_inverse)
+            as u64,
+    }
+});
```
KtorZ:
```suggestion
static PREVIEW_GLOBAL_PARAMETERS: LazyLock<GlobalParameters> = LazyLock::new(|| {
    let consensus_security_param = 432;
    let active_slot_coeff_inverse = 20;
    let shelley_epoch_length_scale_factor = 10;
    let shelley_epoch_length =
        active_slot_coeff_inverse * shelley_epoch_length_scale_factor * consensus_security_param;
    let byron_epoch_length_scale_factor = 0;
    let byron_epoch_length = byron_epoch_length_scale_factor * consensus_security_param;
    let shelley_transition_epoch = 0;
    GlobalParameters {
        consensus_security_param,
        shelley_epoch_length_scale_factor,
        active_slot_coeff_inverse,
        byron_epoch_length_scale_factor,
        shelley_transition_epoch,
        max_lovelace_supply: 45_000_000_000_000_000,
        slots_per_kes_period: 129_600,
        max_kes_evolution: 62,
        shelley_epoch_length,
        stability_window: active_slot_coeff_inverse * consensus_security_param * 2,
        byron_epoch_length,
        byron_total_slots: byron_epoch_length * shelley_transition_epoch,
        randomness_stabilization_window: (4 * consensus_security_param * active_slot_coeff_inverse)
            as u64,
    }
});

static PREPROD_GLOBAL_PARAMETERS: LazyLock<GlobalParameters> = LazyLock::new(|| {
    let consensus_security_param = 2160;
    let active_slot_coeff_inverse = 20;
    let shelley_epoch_length_scale_factor = 10;
    let shelley_epoch_length =
        active_slot_coeff_inverse * shelley_epoch_length_scale_factor * consensus_security_param;
    let byron_epoch_length_scale_factor = 10;
    let byron_epoch_length = byron_epoch_length_scale_factor * consensus_security_param;
    let shelley_transition_epoch = 4;
    GlobalParameters {
        consensus_security_param,
        shelley_epoch_length_scale_factor,
        active_slot_coeff_inverse,
        byron_epoch_length_scale_factor,
        shelley_transition_epoch,
        max_lovelace_supply: 45_000_000_000_000_000,
        slots_per_kes_period: 129_600,
        max_kes_evolution: 62,
        shelley_epoch_length,
        stability_window: active_slot_coeff_inverse * consensus_security_param * 2,
        byron_epoch_length,
        byron_total_slots: byron_epoch_length * shelley_transition_epoch,
        randomness_stabilization_window: (4 * consensus_security_param * active_slot_coeff_inverse)
            as u64,
    }
});
```

crates/amaru-kernel/src/network.rs
```diff
@@ -192,6 +194,46 @@ impl From<NetworkName> for &EraHistory {
...
+        shelley_epoch_length_scale_factor,
+        active_slot_coeff_inverse,
+        byron_epoch_length_scale_factor,
+        shelley_transition_epoch,
+        max_lovelace_supply: 45_000_000_000_000_000,
+        slots_per_kes_period: 129_600,
+        max_kes_evolution: 62,
+        shelley_epoch_length,
+        stability_window: active_slot_coeff_inverse * consensus_security_param * 2,
+        byron_epoch_length,
```
KtorZ: Note that the "byron_epoch_length", "byron_epoch_length_scale_factor" and "shelley_transition_epoch" should no longer be needed now that we started using the Era history for slot arithmetic. They should probably be removed from the globals to avoid shooting ourselves in the foot later. 

Which would also mean we can then rename "shelley_epoch_length" to simply "epoch_length".

.github/workflows/CI.yml
jeluard: Right, I will extract it a separate PR.
This PR actually breaks the simulator hence the idea to include them in the CI.

KtorZ: I totally missed your comments above with Arnaud. My bad.

crates/amaru-kernel/src/network.rs
```diff
@@ -192,6 +194,79 @@ impl From<NetworkName> for &EraHistory {
...
+        max_kes_evolution: 62,
+        epoch_length,
+        stability_window: active_slot_coeff_inverse * consensus_security_param * 2,
+        randomness_stabilization_window: (4 * consensus_security_param * active_slot_coeff_inverse)
+            as u64,
+    }
+});
+
+#[allow(clippy::todo)]
+impl From<NetworkName> for &GlobalParameters {
```
abailly, line 259: I am not sure I like the fact this transformation is "hidden" behind a generic `.into()` call, but maybe it's only me? 

I would rather have an explicit method

```
impl GlobalParameters {
   fn from_network_name (network: NetworkName) -> Self {
   }
}
```

or the other way round:

```
impl NetworkName {
  fn to_global_parameters(&self) -> GlobalParameters {
  }
```

}.

jeluard: I was following the pattern introduced for `EraHistory`.
That said I am confident this will change again once we support custom testnet (which I am looking at now).

KtorZ: @abailly you don't have to use `.into()`; I (almost) never do because I also don't like it. Instead, you can do:

```rust
GlobalParameters::from(network)
```

It's still useful to be able to rely on (highly common) traits for this, so we can write generic functions if needed. Otherwise, this rapidly gets out of hands with libraries all having their own naming conventions.

PR #228, opened 2025-05-21
feat: refactor bootstrap to be testable with in-mem stores

abailly:
This change makes it possible to test the top-level bootstrap function that wires all stages together, using in-memory "dummy" stores to simplify the setup process for ledger and consensus. Those stores are mostly non-functional but the idea is to make it possible to use those and enhance their functionalities later in order to ease testing the whole process using "production" code rather than having to rewire everything as is currently done in simulator.


crates/amaru-stores/src/rocksdb/consensus.rs
```diff
@@ -86,6 +86,47 @@ impl<H: IsHeader + for<'d> cbor::Decode<'d, ()>> ChainStore<H> for RocksDBStore
...
+
+    fn put_nonces(&mut self, header: &Hash<32>, nonces: &Nonces) -> Result<(), StoreError> {
+        self.nonces.insert(*header, nonces.clone());
+        Ok(())
+    }
+
+    fn era_history(&self) -> &amaru_kernel::EraHistory {
+        NetworkName::Testnet(42).into()
+    }
+}
```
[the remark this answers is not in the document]
abailly: We are going to fully flesh-out `InMemoryStore` on a by-need basis, driven by the kind of tests we want to write.

crates/amaru/src/stages/mod.rs
```diff
@@ -50,16 +52,37 @@ pub mod pull;
...
+#[derive(Clone, Debug)]
+pub enum StorePath {
+    InMem,
+    OnDisk(PathBuf),
+}
+
 pub struct Config {
-    pub ledger_dir: PathBuf,
-    pub chain_dir: PathBuf,
+    pub ledger_store: StorePath,
```
jeluard, line 63: I'm wondering if this choice has to be exposed at the configuration level. In memory stores are probably something only relevant for testing scenarios where we manually wire things?

abailly: `Config` is the result of parsing `Args` and not exposed to the user: It can only be set programmatically. While I agree this is mostly useful for testing, it seems to me we shall be using this extensively in the near future and there might be use cases for in-memory only nodes, so I would like to keep this option open.

PR #229, opened 2025-05-22
feat: add stage to store block bodies

abailly:
This change "completes" the block diffusion pipeline handled by consensus logic and makes it possible for an Amaru node to use another Amaru node as an upstream peer, and therefore for an Amaru node to be used as a validating relay in the network.

## Summary by CodeRabbit

- **New Features**
  - Introduced a new pipeline stage for storing raw blockchain blocks before validation, improving data handling and supporting pipelined processing.
  - Added support for storing and retrieving raw blocks in the consensus store, including RocksDB-backed storage.
  - Added new event types and results for block validation and rollback, with enhanced data tracking.

- **Bug Fixes**
  - Clarified and corrected the order and description of consensus pipeline stages in the documentation.

- **Refactor**
  - Updated imports and reorganized code to centralize block event types and raw block definitions.
  - Integrated the new store block stage into the pipeline, updating internal connections and naming for consistency.

- **Documentation**
  - Improved documentation to clarify the consensus pipeline and storage process.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -126,6 +126,7 @@ dependencies = [
  "slot-arithmetic",
  "test-case",
  "thiserror 2.0.12",
+ "tracing",
 ]
 
 [[package]]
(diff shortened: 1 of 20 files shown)


abailly:
Note: I have tested locally connecting 2 Amaru nodes and it does not work yet, as the upstream node fails to provide a proper response to the intersection request from downstream node.

PR #233, opened 2025-05-28
feat: generic snapshots

jeluard:


crates/amaru/build.rs
```diff
@@ -12,7 +12,85 @@
...
+use std::{
+    env,
+    fs::{read_dir, File},
+    io::Write,
+    path::Path,
+};
+
+use amaru_kernel::network::NetworkName;
+
+fn network_name_to_string(network: &NetworkName) -> String {
```
abailly: Would this not be better as a `impl Display` ?

crates/amaru/tests/summary.rs
```diff
@@ -65,50 +65,42 @@ fn db(epoch: Epoch) -> Arc<impl Snapshot + Send + Sync> {
...
-#[test_case(172)]
-#[test_case(173)]
-#[test_case(174)]
-#[test_case(175)]
-#[test_case(176)]
-#[test_case(177)]
-#[test_case(178)]
-#[test_case(179)]
-#[ignore]
+include!(concat!(
```
abailly: I like having explicit list of test cases, could we keep it but add a `NetworkName` parameter? Also, not a great fan of relying on environment variable to run tests.

jeluard: Re. env variable: I see this as more a technical constraints and to my knowledge this is how it's done in rust to generate things that depends on filesystem at compilation time.

The idea is that we want to tests for all epoch, including the live ones. Then there's a parallel job (to be done) that will create the `snap` files by following the chain with `cardano-node`.

crates/amaru/build.rs
jeluard: There's already one with different case. Here it's more a way to get enum from a string representation.

crates/amaru/tests/summary.rs
```diff
@@ -123,10 +115,17 @@ fn compare_preprod_snapshot(epoch: u64) {
...
+    }, {
+        insta::assert_json_snapshot!(
+        format!("rewards_summary_{}", epoch),
+        rewards_summary
+        );
+    });
 }
 
-fn preprod_protocol_version(epoch: Epoch) -> ProtocolVersion {
+fn protocol_version(epoch: Epoch) -> ProtocolVersion {
```
KtorZ: The body of this function is only valid for PreProd, though. We might want to pass in the network name here too.

crates/amaru/tests/summary.rs
KtorZ: The `#[ignore]` annotation was dropped, is that intended? Since those snapshots test require to have already run the node, we need to exclude them from standard unit tests.

KtorZ: Ah no, I see, the test cases have been moved to be generated by the `build.rs`

crates/amaru/build.rs
```diff
@@ -12,7 +12,85 @@
...
+        compare_snapshot(network_name, epoch)
+    }";
+    writeln!(file, "{}", test_content).unwrap();
+}
+
 #[allow(clippy::expect_used)]
 fn main() {
     built::write_built_file().expect("Failed to acquire build-time information");
+
+    generated_compare_snapshot_test_cases();
```
KtorZ: I am not sure about doing this on every build? Seems like it's only needed once and if there are new snapshots added. I feel like this would be better in a side script / Makefile recipe?

jeluard: Yes I considered that option. The advantage of the current one is that tests definitions will always be in sync with the `snap` files and there's no way something can get out of sync. The downside is some extra processing at build time, although I expect this cost to be rather slim. I would have done it at macro expansion time if macro could access the file system.

crates/amaru/build.rs
abailly: it's also more code where `Display` could just work?

jeluard: There's already a `Display` implementation for this type, already used I assumed. But it's a different implementation (case is handled differently).

abailly: I understand that, my question is: why handle the case differently so that the existing function can be used as is?

jeluard: This function is used by the macro to generate enums name so the case has to respect that. 
As I am not entirely sure if the case of the other implementation can be changed, I'd rather play it safe.

jeluard: Changed it to a bash file

crates/amaru/build.rs
jeluard: Now must be done manually

PR #237, opened 2025-06-02
Has exact redeemer set

yHSJ:
This PR introduces the `HasExactRedeemerSet` validation, ensuring there are no missing/extra redeemers present in the witnesses.


crates/amaru-kernel/src/lib.rs
```diff
@@ -1189,6 +1200,21 @@ pub fn sum_ex_units(left: ExUnits, right: ExUnits) -> ExUnits {
     }
 }
 
+// FIXME: cleanup clones?
+pub fn to_redeemer_keys(redeemers: &Redeemers) -> Vec<RedeemersKey> {
```
jeluard: Worth implementing `From` ? Same for `script_purpose_to_string `

jeluard (approved):

crates/amaru-kernel/src/lib.rs
yHSJ: I can't, these are types defined in Pallas so compiler yells at me.

Quantumplation: We can submit PRs to pallas; can you drop a TODO? this will create great fodder for Sundae Labs to contribute during our mercenary era

yHSJ: Sure, I can. But, most things that exist in this `kernel/lib.rs` exist because they are things that don't really belong to Amaru. Generally, the goal is to move much of this logic to Pallas at some point, and @KtorZ and I discussed having the first stages of this process start soon, after I finish up the current phase 1 script validations.

So it existing here kind of implies a `TODO` already.

crates/amaru-ledger/src/rules/transaction/scripts.rs
```diff
@@ -188,6 +188,59 @@ where
...
+    }
+
+    if !redeemers_required.is_empty() {
+        return Err(InvalidScripts::MissingRedeemers(redeemers_required));
+    }
+
+    if !extra_redeemers.is_empty() {
+        return Err(InvalidScripts::ExtraneousRedeemers(extra_redeemers));
+    }
+
```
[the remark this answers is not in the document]
yHSJ: `RedeemersKey` doesn't implement `Hash` and `Eq`, so this isn't possible right now.

Issue #211, opened 2025-05-12
Failure while syncing


jeluard:
### What revision are you using?

4b8bd48c41018fbf3d36375ce407e6b54d00c0f3

### What operating system are you using?

Linux

### What architecture are you using?

x86_64/amd64

### What problem are you having? What should be the expected behaviour?

`amaru` crashes while syncing epoch `188`.

Also we should ensure that it makes sense to crash in this scenario.

```rust
2025-05-12T12:00:44.522621Z ERROR amaru::consensus::chain_forward: block_validation_failed slot=Slot(80153462) hash=f51bd6ea5ddcd06f710433e586104f681c484483856eca8dbbf064e553b0cc39

   amaru::fatal::error
   Whoops! The Amaru process panicked, rather than handling the error it encountered gracefully.
   
   This is almost certainly a bug, and we'd appreciate a report so we can improve Amaru.
   
   Please report this error at https://github.com/pragma-org/amaru/issues/new.
   
   In your bug report please provide the information below and if possible the code
   that produced it.
   
   Operating System: macos
   Architecture:     aarch64
   Version:          v0.1.0+4b8bd48
   

thread '<unnamed>' panicked at crates/amaru/src/stages/consensus/forward_chain.rs:270:21:
assertion `left == right` failed
  left: 3047776
 right: 3047775
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
2025-05-12T12:00:44.523131Z ERROR gasket::framework: x=error sending work unit through output port
2025-05-12T12:00:44.523141Z ERROR gasket::runtime: stage should stop retry=Retry(None)
2025-05-12T12:00:44.523178Z ERROR gasket::framework: x=error sending work unit through output port
2025-05-12T12:00:44.523189Z ERROR gasket::runtime: stage should stop retry=Retry(None)
2025-05-12T12:00:44.523214Z ERROR gasket::framework: x=error sending work unit through output port
2025-05-12T12:00:44.523222Z ERROR gasket::runtime: stage should stop retry=Retry(None)
2025-05-12T12:00:45.073658Z ERROR gasket::framework: x=error sending work unit through output port
2025-05-12T12:00:45.073681Z ERROR gasket::runtime: stage should stop retry=Retry(None)
2025-05-12T12:00:45.073789Z ERROR gasket::runtime: stage should stop retry=Retry(None)
2025-05-12T12:00:45.113424Z ERROR gasket::framework: x=error sending work unit through output port
[...]

Issue #161, opened 2025-03-24
Improve json traces format


jeluard:
### What revision are you using?

main

### What operating system are you using?

MacOS

### What architecture are you using?

aarch64/armv8

### What problem are you having? What should be the expected behaviour?

Fields are not escaped when formatted as json. This make json traces pretty easy to break.
Make sure json traces are always valid json.

See https://github.com/tokio-rs/tracing/pull/377#discussion_r335311790 for context.

Also scripts leveraging those traces (e.g. `demo`) should probably better handle incorrect traces and not swallow errors.

### Please confirm that you've already searched issues, CHANGELOG and available documentation.

- [x] I searched, but no luck.

abailly:
Or just trace _everything_ in JSON?

abailly:
Not sure how much this is related, but I would like to benefit from traces and instrumentation in the simulator. At least in that case, it does not make much sense to distinguish between logs and traces: everything should be logged to stderr (stdout is used to communicate with tester) and we don't want to connect jaeger. Should I move the logging module out of the binary and into the library?

KtorZ:
> everything should be logged to stderr

You can simply set: `AMARU_LOG=amaru=trace`.

abailly:
> it's not clear to me why moving the logging module out of the binary would help?
Because it's not exposed in the crate's library? The simulator lives in `amaru-sim` crate

abailly:
It does not need any internals specific to the binary, but we are emitting useful traces through `#[instrument]` macros which require some of this machinery. Perhaps there's a simple way to get those to be emitted, it's just that I am not aware of it.

KtorZ:
Most of the code in the logging module (if you refer to the `observability.rs` one) is really about setting up and connecting the open telemetry. 

For generating JSON traces, I don't think there's much more than:


```rust
tracing_subscriber::registry()
  .with_layer(tracing_subscriber::fmt::layer()
    .event_format(tracing_subscriber::fmt::format().json())
    .fmt_fields(JsonFields::new())
 )
 .init();
```

abailly:
Here is what I currently do, I will try your suggestion:
```
    tracing_subscriber::fmt()
        .with_writer(std::io::stderr)
        .json()
        .init();
```
I was under the impression the `instrument` macros were somehow tied to Otel but I may be wrong.

abailly:
`with_layer()` does not work out of the box though

jeluard:
[...]

PR #239, opened 2025-06-03
feat: run long tests when merging in main branch

jeluard:
Sync the chain up to latest revolut epoch when commiting on main branch.
This currently [fails](https://github.com/pragma-org/amaru/actions/runs/15450419719/job/43490865530#step:14:252) at epoch `188` due to https://github.com/pragma-org/amaru/issues/211

## Summary by CodeRabbit

- **Chores**
  - Updated the CI workflow to install Cardano CLI version 10.11.0.0 and refined environment variable setup.
  - Modified the Makefile to update the demo script invocation.
  - Improved the demo script's error handling by enabling detection of unset variables early.
- **Bug Fixes**
  - Corrected metadata hash error handling to properly detect conflicts.
- **Improvements**
  - Enhanced error messages for invalid blocks and transaction metadata for clearer readability.
  - Added error logging for invalid blocks to aid in troubleshooting.


crates/amaru/src/stages/ledger.rs
```diff
@@ -115,7 +115,7 @@ impl<S: Store + Send, HS: HistoricalStores + Send> ValidateBlockStage<S, HS> {
         match rules::validate_block(&mut context, self.state.protocol_parameters(), &block) {
             BlockValidation::Err(err) => return Err(err),
             BlockValidation::Invalid(err) => {
-                error!("Block invalid: {:?}", err);
+                error!("Block invalid: {}", err);
```
jeluard, line 118: `Debug`  breaks JSON traces generation as it introduces `"`.

abailly: 🤔 Is it not supposed to be properly escaped?

jeluard: [No](https://github.com/pragma-org/amaru/issues/161)

abailly (commented):
I am not sure I understand the idea: we want to extend the test run to the latest epoch of the network we are testing against?

jeluard:
@abailly Exactly, to test the node can cope with more blocks than just stopping at an arbitrary limit. Then there are others PR than will extend snapshot testing to recent epochs too.

.github/workflows/CI.yml
```diff
@@ -247,10 +249,32 @@ jobs:
...
+          VERSION="10.11.0.0"
+
+          curl -L -o cardano-cli.tar.gz \
+            "https://github.com/IntersectMBO/cardano-cli/releases/download/cardano-cli-${VERSION}/cardano-cli-${VERSION}-x86_64-linux.tar.gz"
+
+          mkdir -p cardano-cli-bin
+          tar -xzf cardano-cli.tar.gz -C cardano-cli-bin
+          mv cardano-cli-bin/cardano-cli* cardano-cli-bin/cardano-cli
+          ls $PWD/cardano-cli-bin/
+          chmod +x cardano-cli-bin/cardano-cli
```
abailly, line 266: I am not a big fan of long scripts in CI. At this stage, would it not make sense to have a docker-compose with all we need?

abailly (approved):

.github/workflows/CI.yml
jeluard: Agree. This one is simply to install `cardano-cli` so I would say it's ok? Or we could consider some shared GH action. Another option is to rely on ogmios now it introduces other deps like a websocket client. Or down the line have it part of `amaru` CLI itself?
So I would err on the side on keeping it simple for now, thus as is :D

abailly: Or add it to the `Makefile`?

jeluard: That's not something I would expect anyone to do via makefile. I think it's fine as is.

abailly: Then to a `scripts/install-cli.sh` file? My point Is I that having long-ish scripts in a YAML file makes our code more tightly coupled with a particular CI engine for no particular reason (eg. why not providing the capability to run long tests to anyone?), and is error-prone (code as text is not great for syntax highlighting or linters).
Note that this is a remark that's also relevant to the `run tests` step.

KtorZ: Inasmuch as those scripts are merely just "bash" scripts, I don't mind. It doesn't create any particular coupling with the underlying CI system (it's just bash!). I agree that for scripts that are somewhat related to testing or running amaru-specific tasks (e.g. demo) it makes sense to have them as standalone scripts. 

For installing the cardano-cli; besides that it's quite puzzling to me that it requires 10 lines of bash to do so and not 1 😬; I don't think it's a good idea to have an `install-cli.sh` script (and take responsibility for everyone down the line coming at us to tell us that the script 'doesn't work on [their] machine' ...). I would instead **consider moving this into a proper Github action `pragma-org/setup-cardano-cli` which installs the cardano-cli and optionally the cardano-node**; which we (and others) could rely on. 

Tagging @Quantumplation for no particular reason.

abailly: I have spent too much time fiddling with scripts inside YAML to be comfortable with having scripts inside YAML. 

> I don't think it's a good idea to have an install-cli.sh script (and take responsibility for everyone down the line coming at us to tell us that the script 'doesn't work on [their] machine' ...)

I don't buy this argument: It could very well live in a `scripts/ci/install-cli.sh` with explanation about the scope and usage. Furthermore, I would wait for the first user to tell us "your script does not work on my machine" to act on this feedback, rather than anticipating something which I doubt will ever happen. But I am 100% sure that someone, possibly me, you, or any other contributor, will need to modify this script at some point, will mess up or introduce a typo, and will lose time because the CI will fail possibly for a cryptic reason.

I am on the other hand not particularly excited with the prospect of creating and *maintaining* a GitHub action, just as I am not particularly excited with the prospect of tying our CI to GitHub. 

All that being said, I understand it's just me and I won't fight an uphill battle 🤷

KtorZ: > I am on the other hand not particularly excited with the prospect of creating and maintaining a GitHub action

Somehow, I am 😅...  Just because it's (installing and needing the cardano-cli in github actions) something I have needed in almost all of my Cardano-related projects so far. So it _should_ already exist.

PR #241, opened 2025-06-03
Pre-commit Hook

yHSJ:
This PR provides a pre-commit hook that can be used to run clippy, and provides a script to install said hook.

## Summary by CodeRabbit

- **Documentation**
  - Updated contributing guidelines to include instructions for setting up Git hooks to automate code linting before commits.

- **Chores**
  - Added a script to automate the installation of Git hooks in the repository.
  - Introduced a pre-commit hook to automatically run a code linter and block commits with linting issues.

### DIFF
--- a/scripts/hooks/pre-commit
+++ b/scripts/hooks/pre-commit
@@ -1,0 +1,18 @@
+#!/bin/bash
+
+set -e
+
+echo "Running clippy..."
+
+if cargo clippy-amaru 2>&1; then
+    echo ""
+else
+    echo "Discovered a clippy issue"
+    echo ""
+    echo "Please fix the clippy issues before committing."
+    echo "You can run 'cargo clippy-amaru' to see the full output."
+    echo ""
+    echo "To bypass this check (not recommended), use:"
+    echo "  git commit --no-verify"
+    exit 1
+fi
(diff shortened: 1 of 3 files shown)


scripts/hooks/pre-commit
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+set -e
+
+echo "Running clippy..."
+
+if cargo clippy-amaru 2>&1; then
```
abailly, line 7: I would add `cargo fmt-amaru`, although of course that should be part of the IDE setup.

KtorZ:
I don't like this 😶... especially on a 'pre-commit' (would be more acceptable on a 'pre-push' IMO). You should just configure your editor/IDE to do it for you.

yHSJ:
@KtorZ We can make it pre-push if you'd like. But importantly, it's optional. You don't have to install the hooks.

I can think of several reasons one wouldn't want to configure their editor/IDE to do it for them, notably they may have different tooling/configurations for different projects. I don't understand the pushback against providing optional hooks for contributors

KtorZ:
> I don't understand the pushback against providing optional hooks for contributors

It's actually the same reason I don't want to Nix configuration on the repository usually: if no one in the core contributors use it, it'll likely just stop working at some point and someone one day will have a bad experience with the repo. 

In this case, at least you are using it, and it's a fairly static set of config. So 🤷.

PR #243, opened 2025-06-05
fix: use test build profile throughout CI, to ensure meaningful cache hit.

KtorZ:
We only populate caches when merging on `main`, and those caches are prepared with a test profile. The simulation and e2e snapshots tests currently causes a full rebuild, adding 7-8 extra minutes on all CI checks. That's unfortunate and this commit shall fix this.

## Summary by CodeRabbit

- **Chores**
  - Updated CI workflow for improved build profile handling, caching, and formatting.
  - Modified Makefile and demo script to support dynamic build profiles.
  - Cleaned up and reorganized simulation project dependencies for better maintainability.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -42,16 +42,20 @@ jobs:
         environments:
           - runner: ubuntu-latest
             target: x86_64-unknown-linux-gnu
+
           - runner: macos-latest
             target: aarch64-apple-darwin
+
           - runner: windows-latest
             target: x86_64-pc-windows-msvc
             command: test --profile dev-debug
+
           - runner: ubuntu-latest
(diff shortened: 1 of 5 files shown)


.github/workflows/CI.yml
```diff
@@ -209,7 +215,7 @@ jobs:
 
       - name: Build Amaru
         run: |
-          cargo test --release --no-run -p amaru
+          cargo test --no-run -p amaru
```
jeluard, line 218: I made the change to use `release` profile some time ago as I expect to use this build to track performance regression during `snapshots` tests.
Now we can always revisite later once we figure out how to do this.

jeluard (approved):

PR #251, opened 2025-06-11
Allow using Amaru as an upstream Cardano node

abailly:
This PR adds a minimal ability to let other Cardano nodes connect to an Amaru upstream node and start syncing the chain from them:
* fix the block type which was wrongly set to 1
* actually implement minimal block-fetch server to serve blocks to downstream peers.

This PR has only been tested with 2 Amaru nodes and usage documents are not updated because this new feature can be used simply by running an Amaru node and setting its `--peer-address` argument to the `--listen-address` of the other Amaru node. We would want to check what happens if a _cardano-node_ connects to the Amaru node and sync from them.

Also, there's no automated tests which is annoying: I don't have a good mental model of  how to write tests for these low-level network features.

## Summary by CodeRabbit

- **New Features**
  - Implemented a minimal block fetch server that serves single-block requests and returns a specific error for unsupported range requests.

- **Bug Fixes**
  - Improved error handling and logging in header reception, validation, block fetching, and chain selection for better troubleshooting.

- **Documentation**
  - Removed outdated information about protocol implementation status.

- **Chores**
  - Updated CI workflow to use a newer runner environment and ensure required system packages are installed.

- **Refactor**
  - Introduced macros to simplify message sending and scheduling across multiple components, reducing boilerplate and enhancing error logging.
  - Enhanced logging for chain selection and protocol operations to aid in debugging and monitoring.
  - Added structured tracing instrumentation and improved span context propagation across stages.
  - Refactored event handling to adopt tracing spans from events, improving observability.
  - Added display formatting for peer entities to improve logging clarity.

### DIFF
--- a/crates/amaru/src/stages/common.rs
+++ b/crates/amaru/src/stages/common.rs
@@ -12,6 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use amaru_consensus::consensus::{ChainSyncEvent, DecodedChainSyncEvent, ValidateHeaderEvent};
+use amaru_kernel::block::{BlockValidationResult, ValidateBlockEvent};
+use opentelemetry::Context;
 use tracing::Span;
 use tracing_opentelemetry::OpenTelemetrySpanExt;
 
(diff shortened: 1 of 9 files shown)


abailly:
I can rebase it to main I think, let me try.

abailly:
Actually, I wanted to have some proper e2e test in place before undrafting it

crates/amaru/src/stages/consensus/forward_chain/client_protocol.rs
```diff
@@ -214,15 +216,34 @@ enum BlockFetchMsg {}
...
+        server.send_batch_done().await?;
     }
+}
 
-    Ok(())
+fn from_pallas_point(lb_point: &Point) -> amaru_kernel::Point {
+    match lb_point {
+        Point::Origin => amaru_kernel::Point::Origin,
+        Point::Specific(slot, hash) => amaru_kernel::Point::Specific(*slot, hash.clone()),
+    }
```
KtorZ: I am pretty sure this already exists somewhere else 🤔 ?

jeluard: [This one](https://github.com/pragma-org/amaru/blob/34bff53a61cd6fe0562d22a4f3ac3ba51f993610/crates/amaru/src/point.rs#L18) maybe?

KtorZ: Haha! I knew it looked familiar.

abailly: Fair enough! I looked for it but couldn't find it

PR #259, opened 2025-06-12
feat: do not panic when block validation fails

jeluard:
Let `amaru` continue validating after a block validation fails.

This currently does not work and still panics higher up in the stack due to a mishandling in the consensus layer. This will be addressed in a subsequent PR.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved error handling and logging during block validation to provide clearer feedback when issues occur.
  - Enhanced log output formatting for hash values to aid in debugging.

- **Refactor**
  - Streamlined control flow for block validation processes, making error handling more explicit and robust.

### DIFF
--- a/crates/amaru/src/stages/consensus/forward_chain.rs
+++ b/crates/amaru/src/stages/consensus/forward_chain.rs
@@ -318,7 +318,7 @@ impl gasket::framework::Worker<ForwardChainStage> for Worker {
                     target: EVENT_TARGET,
                     parent: span,
                     slot = ?point.slot_or_default(),
-                    hash = %Hash::<32>::from(point),
+                    hash = ?Hash::<32>::from(point),
                     "block_validation_failed"
                 );
 
--- a/crates/amaru/src/stages/ledger.rs
+++ b/crates/amaru/src/stages/ledger.rs
@@ -114,10 +114,7 @@ impl<S: Store + Send, HS: HistoricalStores + Send> ValidateBlockStage<S, HS> {
         let protocol_version = block.header.header_body.protocol_version;
         match rules::validate_block(&mut context, self.state.protocol_parameters(), &block) {
             BlockValidation::Err(err) => return Err(err),
-            BlockValidation::Invalid(err) => {
-                error!("Block invalid: {:?}", err);
-                return Ok(Some(err));
-            }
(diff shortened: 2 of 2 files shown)


yHSJ:
Great, thank you! For now, while we're receiving invalid blocks from upstream (trusted) peers, we *know* that something is wrong with our logic. But once we're connected to unknown peers we'll have to handle this appropriately. This is one piece of that.

yHSJ (approved):

abailly:
What we need to do in this case is to disconnect from the upstream peer which for all intents and purposes, in the current stage of the system, translates to crashing. What do you refer to 

> due to a mishandling in the consensus layer.

jeluard:
@abailly I refer to this panic: `thread '<unnamed>' panicked at crates/amaru/src/stages/consensus/forward_chain.rs:270:21:` that can be seen [here](https://github.com/pragma-org/amaru/issues/211). Might be more complex than that but `amaru` as a node should just continue processing new blocks even when invalid ones are received?

yHSJ:
> but `amaru` as a node should just continue processing new blocks even when invalid ones are received?

Yes, from the ledger perspective it should just continue processing blocks. There is consensus and networking things that have to occur, as Arnaud said, such as removing the peer. When we have a single peer, that is effectively stopping the process, so it's no different than the current situation. But I don't mind removing the panic and preparing for more robust logic. I don't know enough about the networking/consensus specs, so @abailly will be better for that side

abailly:
> Might be more complex than that but `amaru` as a node should just continue processing new blocks even when invalid ones are received?

Definitely! But as I said, the correct behaviour when a block is invalid is to assume the remote peer is adversarial and therefore to disconnect from it. In the current state of amaru this is exactly equivalent to crashing, because there's no more progress the node can make once disconnected from its one and only peer.

So do you suggest we just ignore block validation failures and assume we can proceed, at least until we add this disconnection behaviour?

jeluard:
In the absence of node rating and eventual subsequent disconnection I would assume we consider them honest. Thus amaru should keep accepting blocks from them?
If the node was disconnected amaru should just stay up, maybe with some special logs if there's no more peers?

The current error sounds a bit unrelated.

abailly:
As I said on discord, happy to discuss f2f if that helps

PR #265, opened 2025-06-16
Rework, reorganize and relocate import-ledger-state functionality

rrruko:
## Summary by CodeRabbit

* **New Features**
  * Introduced a new progress bar library with terminal support, providing consistent progress reporting across commands.
  * Implemented support for importing and bootstrapping initial ledger state from external snapshots, with improved progress reporting.
  * Modularized and exposed additional ledger state structures for broader use.

* **Improvements**
  * Refactored and simplified ledger state import commands to use external snapshot import logic and unified progress bars.
  * Enhanced protocol parameter handling to support optional Plutus cost models.

* **Chores**
  * Updated dependencies and workspace configuration to include the new progress bar crate.
  * Improved CI workflow to support submodule checkout.


crates/amaru/src/bin/amaru/cmd/import_ledger_state.rs
```diff
@@ -82,6 +88,14 @@ pub struct Args {
...
     )]
     network: NetworkName,
+
+    #[arg(
+        long,
+    )]
+    import_vector: bool,
+
+    #[arg(long, value_name = "DIR")]
+    pparams_dir: Option<PathBuf>,
```
KtorZ: These feel shoehorned as they trigger an action completely separate from the import-ledger-state command. It seems to me that this would be better done in a test context? A bit similar to how we do the snapshots tests: https://github.com/pragma-org/amaru/blob/main/crates/amaru/tests/summary.rs

crates/amaru/src/bin/amaru/cmd/import_ledger_state.rs
```diff
@@ -285,8 +405,22 @@ fn decode_new_epoch_state(
...
+        let pparams_hash: &minicbor::bytes::ByteSlice = d.decode()?;
+        let pparams_file_path = fs::read_dir(pparams_dir)?
+            .filter_map(|entry| entry.ok().map(|e| e.path()))
+            .find(|path| path.file_name().map(|filename| filename.to_str() == Some(&hex::encode(pparams_hash.as_ref()))).unwrap_or(false))
+            .ok_or(Error::MissingPparamsFile)?;
+        let pparams_file = fs::read(pparams_file_path)?;
+        cbor::Decoder::new(&pparams_file).decode()?
+    } else {
+        d.decode()?
+    };
```
KtorZ: I don't understand why this is necessary? If the goal is to test against a NES, why not use the current protocol params specified in that NES ?

.github/workflows/UnpackTestVectors.yml
```diff
@@ -0,0 +1,25 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          repository: rrruko/cardano-blueprint
+          path: cardano-blueprint
+          ref: rrruko/ledger-conformance-test-vectors
+      - name: Unpack vectors
+        run: |
+          tar -xzf cardano-blueprint/src/ledger/conformance-test-vectors/vectors.tar.gz
```
KtorZ: Feels like this should just be a git submodule.

crates/amaru/src/bin/amaru/cmd/import_ledger_state.rs
rrruko: the pparams field of the NES in a test vector is stored separately. the NES just gives the hash, so you have to go look up the preimage. this is an adjustment that i made to the NES serialization format on my branch, to optimize the size of the dumped vectors

crates/amaru/src/bin/amaru/cmd/import_ledger_state.rs
rrruko: makes sense

crates/amaru/src/bin/amaru/cmd/import_ledger_state.rs
KtorZ: The parameters are in the NES, although a bit hidden so maybe you missed them 😅. Unless you mean something else which I don't understand?

--> https://github.com/IntersectMBO/cardano-ledger/blob/3ae9159f5868a31d393a5029569ff28969d5329d/eras/conway/impl/src/Cardano/Ledger/Conway/Governance.hs#L251

They are in the governance state, which is itself in the UTxO state (don't ask why...).

Quantumplation: That's the haskell node definition; he's saying that when we saved those out to disk naively, the pparams were almost always just duplicated across every test case, massively ballooning the size, and so we modified the format on disk slightly; protocol parameters are stored separately, and the test vector refers to it by hash.

That being said, that was before we added gzip, so maybe that's not needed anymore.

rrruko:
- moved test vector decode logic into a test case at `crates/amaru/tests/evaluate_ledger_states.rs` that evaluates all vectors from a cardano-blueprint git module (vectors need to be unpacked first)
- moved decode_new_epoch_state into amaru-stores so that it can be invoked from test

crates/amaru-ledger/src/state/snapshot.rs
```diff
@@ -0,0 +1,720 @@
+// Copyright 2024 PRAGMA
```
jeluard: This looks largely pasted from amaru/cmd?
We should find a way to not have this huge duplication.

jeluard: Hmm that's what you did apparently? The name `state/snapshot` is confusing then.

KtorZ: Any better suggestion 😅 ?

jeluard: `snapshot` is pretty overloaded in amaru's context. @coderabbitai how should we call this file?

KtorZ: I kind of hate calling this `new_epoch_state`, because that's Haskell's ledger jargon and it's confusing in this particular instance. 

Perhaps "bootstrap" is not bad. Although to be fair @jeluard, this does refers to the things we call _snapshots_ everywhere else (the very one we use in the import-ledger-state command). Code has simply been moved around.

abailly:
@rrruko I am trying to rebase #317 on top of your PR. It seems relatively straightforward as I understand `decode_new_epoch_state` was mostly moved to another file and not significantly changed, is that correct? My goal is to be able to feed this function with data coming from other sources, notably the genesis files, in order to support running Amaru in testnets.

rrruko:
> @rrruko I am trying to rebase #317 on top of your PR. It seems relatively straightforward as I understand `decode_new_epoch_state` was mostly moved to another file and not significantly changed, is that correct? My goal is to be able to feed this function with data coming from other sources, notably the genesis files, in order to support running Amaru in testnets.

the only changes are that i added the option for protocol parameters to be loaded separately, and that i made the rewards part optional

abailly:
@rrruko Thanks. How far are you from merging this? My PR is ready and as I am leaving for 3 weeks of vacations I would rather have it merged than leave it stalled.

rrruko:
>     3. Each test case is currently one opaque blob of CBOR; with little information about the scenario being tested other than the folder's name. Plus, the blob seems to be a record with an old NES, new NES and a transaction. It would be better for consumption to have field in a separate file, and possibly some metadata in a machine-readable format (e.g. JSON, YAML or TOML) that documents the test scenario. At least the title is probably fine for now, although I could imagine adding the ledger protocol version and the rules it relates to.

The last field of the CBOR blob already contains the title of the test (the same one that's used for the folder name).

crates/amaru-kernel/src/protocol_parameters.rs
```diff
@@ -96,14 +97,25 @@ impl<'b, C> cbor::decode::Decode<'b, C> for ProtocolParameters {
...
+                0 => {
+                    plutus_v1 = Some(v);
+                }
+                1 => {
+                    plutus_v2 = Some(v);
+                }
+                2 => {
+                    plutus_v3 = Some(v);
+                }
+                _ => {}
```
KtorZ:
```suggestion
                _ => unreachable!("unexpected language version: {k}"),
```

Avoid silently swallowing unexpected behaviors.

PR #266, opened 2025-06-16
Migrate to Pallas v1-alpha

yHSJ:
This is a second attempt at completing a migration to pallas v1-alpha. This PR will not necessarily have production ready code, so it should not be merged.

But it's a way to have a conversation about the potential issues, difficulties, etc. we run into while upgrading Pallas.


crates/amaru-kernel/src/serde_utils.rs
```diff
@@ -14,15 +14,24 @@
...
 use std::{collections::BTreeMap, ops::Deref};
 
+// FIXME: Here we are constructing KeepRaw wrappers with an empty raw vec (KeepRaw::from)
+// This logically is only (at this point, and probably forever) for testing purposes to build data structures from JSON representations of them.
+// Generally, I'd say we don't care about the fact we aren't storing the raw bytes, as these values are used to mock data coming from out rocksdb store,
+// which involves reserializing the data, so we shouldn't rely on the raw bytes anyway.
+// However, it's notable that we still have to construct the KeepRaw wrappers with an empty vec, and that is not clearly communicated anywhere else in the codebase,
+// so it could lead to confusion and potential bugs in tests (or even in production code based on usage from the store)
+// Should there be a type that clearly communicates that there is no raw data and thus, cannot be trusted as a source of truth for the "on the wire" representations?
+// Is there a better way to clearly communicate that these `KeepRaw` are not actually, keeping raw?
```
SupernaviX, line 33: IMO, `raw_cbor()` shouldn't exist, and you should use `minicbor::encode()` (which uses the raw vec under the hood) whenever you want to read the raw bytes. That lets you dodge the footgun completely.

yHSJ: The problem with that is that when we're working with raw bytes, it's because we're intentionally avoiding re-serializing data (wen canonical serialization). If we don't know whether or not we're getting the original bytes or some re-serialized representation, it can cause other issues which are possibly more challenging to handle.

yHSJ:
Closing due to discussion in Discord. We are not quite ready to make this migration yet.

PR #270, opened 2025-06-18
feat: Implement MemoryStore with ReadOnlyStore support

whankinsiv:
This PR: 
- Defines Memory store using RefCell in-memory structures for all relevant store columns.
- Implements the ReadOnlyStore trait for MemoryStore, replacing previous placeholder code.
- Adds test coverage for all ReadOnlyStore methods except iter_proposals (pending Ord implementation for ProposalId).
- Added StoreError::NotFound, used when an entry is not missing from the memory store.
- Updates make_ledger to use the now implemented MemoryStore definition instead of the placeholder. 

Issues:
- This PR causes the bootstrap_all_stages test in crates/amaru/src/stages/mod.rs to fail when attempting use the make_ledger function which calls get_protocol_parameters_for on an empty store. Data input will be implemented at a later time.

I am now working on implementing the TransactionalContext trait for the memory store. A followup draft PR will be submitted in the coming days.

### DIFF
--- a/crates/amaru-ledger/src/store/in_memory.rs
+++ b/crates/amaru-ledger/src/store/in_memory.rs
@@ -1,15 +1,49 @@
-use crate::{
-    store::{
-        EpochTransitionProgress, HistoricalStores, ReadOnlyStore, Snapshot, Store, StoreError,
-        TransactionalContext,
-    },
-    summary::Pots,
+use crate::store::{
+    columns::{accounts, dreps, pools, pots, proposals, slots, utxo},
+    EpochTransitionProgress, HistoricalStores, ReadOnlyStore, Snapshot, Store, StoreError,
+    TransactionalContext,
 };
-use amaru_kernel::{protocol_parameters::ProtocolParameters, Lovelace, Point, StakeCredential};
+use amaru_kernel::{
+    protocol_parameters::ProtocolParameters, Lovelace, Point, PoolId, ProposalId, Slot,
+    StakeCredential, TransactionInput,
+};
 use slot_arithmetic::Epoch;
-use std::collections::BTreeSet;
+use std::{
+    cell::RefCell,
+    collections::{BTreeMap, BTreeSet},
+};
 
-pub struct MemoryStore {}
+pub struct MemoryStore {
+    utxos: RefCell<BTreeMap<TransactionInput, Option<utxo::Value>>>,
+    accounts: RefCell<BTreeMap<StakeCredential, Option<accounts::Row>>>,
+    pools: RefCell<BTreeMap<PoolId, Option<pools::Row>>>,
+    pots: RefCell<pots::Row>,
+    slots: RefCell<BTreeMap<Slot, Option<slots::Row>>>,
+    dreps: RefCell<BTreeMap<StakeCredential, Option<dreps::Row>>>,
+    proposals: RefCell<BTreeMap<ProposalId, Option<proposals::Row>>>,
+    p_params: RefCell<BTreeMap<Epoch, ProtocolParameters>>,
+}
+
+impl Default for MemoryStore {
+    fn default() -> Self {
+        MemoryStore {
+            utxos: RefCell::new(BTreeMap::new()),
+            accounts: RefCell::new(BTreeMap::new()),
+            pools: RefCell::new(BTreeMap::new()),
+            pots: RefCell::new(pots::Row::default()),
+            slots: RefCell::new(BTreeMap::new()),
+            dreps: RefCell::new(BTreeMap::new()),
+            proposals: RefCell::new(BTreeMap::new()),
+            p_params: RefCell::new(BTreeMap::new()),
+        }
+    }
+}
+
+impl MemoryStore {
+    pub fn new() -> Self {
+        Self::default()
+    }
+}
 
 impl Snapshot for MemoryStore {
     fn epoch(&self) -> Epoch {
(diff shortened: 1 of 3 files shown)


abailly:
This is really cool and will tremendously help with simulation testing.

jeluard:
Nice!

Some questions:

* how should the existing `rocksdb` implementation deal with `StoreError::NotFound` ?
* it would be great to make tests abstract over `Store` so that we can use them for others implementations
* what do you have in mind for `TransactionalContext` ?

crates/amaru-ledger/src/store/in_memory.rs
```diff
@@ -1,15 +1,49 @@
...
+
+pub struct MemoryStore {
+    utxos: RefCell<BTreeMap<TransactionInput, Option<utxo::Value>>>,
+    accounts: RefCell<BTreeMap<StakeCredential, Option<accounts::Row>>>,
+    pools: RefCell<BTreeMap<PoolId, Option<pools::Row>>>,
+    pots: RefCell<pots::Row>,
+    slots: RefCell<BTreeMap<Slot, Option<slots::Row>>>,
+    dreps: RefCell<BTreeMap<StakeCredential, Option<dreps::Row>>>,
+    proposals: RefCell<BTreeMap<ProposalId, Option<proposals::Row>>>,
+    p_params: RefCell<BTreeMap<Epoch, ProtocolParameters>>,
```
KtorZ, line 24: Why the `RefCell`s here? We should be able to have fully static mutability in this context. If we need to modify the store interface here and there (because I reckon that rocksdb's API doesn't always require mutable handles), then I'd prefer that over RefCells.

crates/amaru-ledger/src/store/in_memory.rs
```diff
@@ -18,118 +52,185 @@ impl Snapshot for MemoryStore {
...
+            .utxos
+            .borrow()
+            .iter()
+            .filter_map(|(tx_input, opt_tx_output)| {
+                opt_tx_output
+                    .as_ref()
+                    .map(|tx_output| (tx_input.clone(), tx_output.clone()))
+            })
+            .collect();
+        Ok(utxo_vec.into_iter())
```
KtorZ, line 111: Feels odd to collect and then iter? We should be able to return the first iterator to begin with.

crates/amaru-ledger/src/store/in_memory.rs
```diff
@@ -1,15 +1,49 @@
...
-use std::collections::BTreeSet;
+use std::{
+    cell::RefCell,
+    collections::{BTreeMap, BTreeSet},
+};
+
+pub struct MemoryStore {
+    utxos: RefCell<BTreeMap<TransactionInput, Option<utxo::Value>>>,
+    accounts: RefCell<BTreeMap<StakeCredential, Option<accounts::Row>>>,
+    pools: RefCell<BTreeMap<PoolId, Option<pools::Row>>>,
```
KtorZ, line 19: I see a lot of `filter_map` in the implementation; which made me wonder. Why is this storing `Option` instead of values directly? It's already in a map and I don't see a case where we would want to preserve a key without a value?

crates/amaru-ledger/src/store/in_memory.rs
```diff
@@ -292,6 +393,632 @@ impl Store for MemoryStore {
...
+            utxos: RefCell::new(BTreeMap::new()),
+            accounts: RefCell::new(BTreeMap::new()),
+            pools: RefCell::new(BTreeMap::new()),
+            pots: RefCell::new(pots::Row::default()),
+            slots: RefCell::new(BTreeMap::new()),
+            dreps: RefCell::new(BTreeMap::new()),
+            proposals: RefCell::new(BTreeMap::new()),
+            p_params: RefCell::new(BTreeMap::new()),
+        })
+    }
```
KtorZ, line 406: Seems redundant with the `Default` implementation (and `::new()` as well).

crates/amaru-ledger/src/store/in_memory.rs
```diff
@@ -292,6 +393,632 @@ impl Store for MemoryStore {
...
+                },
+            },
+            cc_min_size: 13,
+            cc_max_term_length: 30u32,
+            gov_action_lifetime: 20u32,
+            gov_action_deposit: 10_000_010u64,
+            drep_deposit: 2_000_010u64,
+            drep_expiry: 60u32,
+        }
+    }
```
KtorZ, line 797: We shall re-use arbitrary generators we've already defined in various _columns_ for all the above. Although, I imagine this can be quite foreign to you so we may want to do this in a pairing session or as a commit on top.

crates/amaru-ledger/src/store/in_memory.rs
```diff
@@ -292,6 +393,632 @@ impl Store for MemoryStore {
...
+        let store = MemoryStore::new();
+
+        let txin = TransactionInput {
+            transaction_id: Hash::new([0u8; 32]),
+            index: 0,
+        };
+
+        let output = PseudoTransactionOutput::PostAlonzo(dummy_post_alonzo_output());
+
+        store.utxos.borrow_mut().insert(txin.clone(), Some(output));
```
KtorZ, line 810: I don't think those tests should be calling internal stuff directly; but rather rely on the API (i.e. `save`). Since the `new()` function at the moment is rather redundant with the `Default::default` already; I would suggest to leverage `new()` for this and allow passing a set of initial maps to preserve a good separation between tests and implementation details. 

As @jeluard also pointed; the tests should probably be ultimately re-usable across all stores, and thus solely rely on the store's API.

crates/amaru-ledger/src/store/in_memory.rs
```diff
@@ -292,6 +393,632 @@ impl Store for MemoryStore {
...
+
+        let result = store.utxo(&txin).unwrap();
+
+        assert!(result.is_some(), "UTXO should be found");
```
KtorZ, line 814: The statement is a bit weak; why not assert that the UTxO is the expected one?

crates/amaru-ledger/src/store/in_memory.rs
whankinsiv: > Why the `RefCell`s here? We should be able to have fully static mutability in this context. If we need to modify the store interface here and there (because I reckon that rocksdb's API doesn't always require mutable handles), then I'd prefer that over RefCells.

RefCell's were used to allow mutation within methods that take `&self`, avoiding the need to change the TransactionalContext trait to use `&mut self`. I agree that modifying the interface would be a cleaner approach.

crates/amaru-ledger/src/store/in_memory.rs
whankinsiv: The reason I decided to store `Option` instead of the values directly was to maintain compatibility with the `Iter` types defined in each column module, which are shared with the RocksDB implementation.

crates/amaru-ledger/src/store/in_memory.rs
whankinsiv: Thanks for pointing out the arbitrary generators. I'm not too familiar with that testing approach yet, but will give it a try and reach out if I run into any issues.

crates/amaru-ledger/src/store/in_memory.rs
whankinsiv: Thanks for the feedback! I’ll refactor the tests to avoid directly accessing internal fields. The current approach was just a temporary setup while `TransactionalContext` is still a work in progress. I'll also work on generalizing the tests so they can be used across all store implementations.

crates/amaru-ledger/src/store/in_memory.rs
KtorZ: I don't get the rationale / justification here. One is an interface concern (the use of Option in mutable iterators for some of the store entrypoints) and the other is a choice of internal data structures. The interface can be satisfied even if the internal data-structures do not use `Option`; so I really don't see reasons to carry `Option` down there. It is confusing.

crates/amaru-ledger/src/store/in_memory.rs
KtorZ: That's indeed what I realized, hence the suggestion to rely on `new`. Once the transactional context is implemented, that could be potentially made a little less verbose.

whankinsiv:
Updated PR opened here: #307

PR #275, opened 2025-06-18
fix: wrong merge changed the musl ubuntu image

jeluard:
Also update dependencies

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,14 +15,14 @@ default-members = ["crates/*"]
 resolver = "2"
 
 [workspace.dependencies]
-acto = { version = "0.7.2", features = ["tokio"] }
+acto = { version = "0.7.4", features = ["tokio"] }
 anyhow = "1.0.95"
 async-trait = "0.1.83"
-async-compression = { version = "0.4.24",  features = ["tokio", "gzip"] }
+async-compression = { version = "0.4.25",  features = ["tokio", "gzip"] }
 bech32 = "0.11.0"
 cbor4ii = { version = "1.0.0", features = ["serde1", "use_std"] }
 clap = { version = "4.5.38", features = ["derive", "env"] }
-reqwest = { version = "0.11.24", default-features = false, features = ["json", "stream", "rustls-tls"] }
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -15,14 +15,14 @@ default-members = ["crates/*"]
...
+acto = { version = "0.7.4", features = ["tokio"] }
 anyhow = "1.0.95"
 async-trait = "0.1.83"
-async-compression = { version = "0.4.24",  features = ["tokio", "gzip"] }
+async-compression = { version = "0.4.25",  features = ["tokio", "gzip"] }
 bech32 = "0.11.0"
 cbor4ii = { version = "1.0.0", features = ["serde1", "use_std"] }
 clap = { version = "4.5.38", features = ["derive", "env"] }
-reqwest = { version = "0.11.24", default-features = false, features = ["json", "stream", "rustls-tls"] }
+reqwest = { version = "0.12.20", default-features = false, features = ["json", "stream", "rustls-tls"] }
```
KtorZ, line 25: Note that we seem to only depend on `reqwest` in the bootstrap command; perhaps there are better deps choices if we have issues with the license? I am not sure what "if data is bundled with the software" means in this context?

KtorZ (approved):

Cargo.toml
jeluard: I understand that the issue is if we download the data dynamically from a binary. Which we don't. We probably don't use the relevant 'data' at all but I didn't dig enough to be confident about that.

KtorZ: The repository says it's Apache-2.0 licensed though, where did you see a CDLA-Permissive? In one of the dependencies introduced by Reqwest perhaps?

jeluard: Yes exactly, detected by cargo deny

jeluard: See https://github.com/pragma-org/amaru/actions/runs/15741213710/job/44366729538#step:4:17

jeluard: Offender is https://github.com/rustls/webpki-roots, used by `hyper-rustls`

PR #279, opened 2025-06-19
feat: set service name at start of daemon

abailly:
Currently, the otel property 'service.name' is hardcoded to 'amaru' which is confusing when one wants to monitor several amaru services. This change provides a dead simple solution allowing the process name to be set at start time and defaulting to 'amaru'. A possibly more robust solution would use host name resolution but this is not needed atm.

### DIFF
--- a/crates/amaru/src/bin/amaru/main.rs
+++ b/crates/amaru/src/bin/amaru/main.rs
@@ -58,11 +58,18 @@ enum Command {
 struct Cli {
     #[command(subcommand)]
     command: Command,
+
     #[clap(long, action, env("AMARU_WITH_OPEN_TELEMETRY"))]
     with_open_telemetry: bool,
+
     #[clap(long, action, env("AMARU_WITH_JSON_TRACES"))]
     with_json_traces: bool,
+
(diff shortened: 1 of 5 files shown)


crates/amaru/src/bin/amaru/main.rs
```diff
@@ -58,12 +58,19 @@ enum Command {
...
     #[command(subcommand)]
     command: Command,
+
     #[clap(long, action, env("AMARU_WITH_OPEN_TELEMETRY"))]
     with_open_telemetry: bool,
+
     #[clap(long, action, env("AMARU_WITH_JSON_TRACES"))]
     with_json_traces: bool,
+
+    #[arg(long, value_name = "STRING", default_value_t = DEFAULT_SERVICE_NAME.to_string())]
```
jeluard: Would it make sense to add `env` support, as e.g. `with_json_traces` and `AMARU_WITH_JSON_TRACES` ?

abailly: I wanted to add it for consistency's sake and forgot 👍

PR #286, opened 2025-06-19
allow passing explicit seed to simulation tests

abailly:
this requires some work to allow passing arguments through `cargo test`


simulation/amaru-sim/tests/simulation.rs
```diff
@@ -9,6 +9,7 @@ fn run_simulator() {
         chain_dir: ".".into(),
         block_tree_file: "tests/data/chain.json".into(),
         start_header: Hash::from([0; 32]),
+        seed: Some(10244329322600784733)
```
stevana:
```suggestion
        seed: Some(10244329322600784733),
```

PR #287, opened 2025-06-19
wip: thread stage graph TraceBuffer through simulation test

abailly:
we want to
1. remove duplication
2. write the trace file in a temporary file
3. copy the file locally if test fails
3. (optional) allow to keep the trace buffer file based on configuration of the test

### DIFF
--- a/simulation/amaru-sim/src/simulator/simulate.rs
+++ b/simulation/amaru-sim/src/simulator/simulate.rs
@@ -25,6 +25,7 @@
 // Make assertions on the trace to ensure the execution was correct, if not, shrink and present minimal trace that breaks the assertion together with the seed that allows us to reproduce the execution.
 
 use crate::echo::{EchoMessage, Envelope};
+use pure_stage::trace_buffer::TraceBuffer;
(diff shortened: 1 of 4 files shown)


simulation/amaru-sim/src/simulator/simulate.rs
```diff
@@ -265,12 +270,38 @@ pub fn simulate<Msg, F>(
...
+                "success-{}.trace",
+                now.duration_since(SystemTime::UNIX_EPOCH)
+                    .unwrap()
+                    .as_secs()
+            ))
+            .expect("Failed to create temporary file");
+            for bytes in trace_buffer.lock().iter() {
+                file.write_all(bytes)
+                    .expect("Failed to write JSON data to file");
+            }
```
rkuhn: Since we want the success status to be determined by the test and not by possible file writing problems, I’d put this code into a separate function, pass in the name prefix, return a result and then print any errors to stderr. Then this function can also be used in the error case and file writing errors won’t prevent printing the actual test failure.

PR #292, opened 2025-06-22
Configure otlp endpoints

abailly:
This PR allows one to configure where telemetry data is sent, while keeping current default of `localhost:4317` and `localhost:4318` for spans and metrics respectively.

## Summary by CodeRabbit

- **New Features**
  - Added new CLI options and environment variables to configure OpenTelemetry span and metric endpoints, as well as the service name.
- **Documentation**
  - Updated documentation with instructions for configuring OpenTelemetry options, including new flags and environment variables for spans, metrics, and service naming.

### DIFF
--- a/monitoring/README.md
+++ b/monitoring/README.md
@@ -355,3 +355,11 @@ Coming soon.
 > The plan so far is to maximise compatibility with the existing Haskell node Prometheus metrics such that tools like [`gLiveView`](https://cardano-community.github.io/guild-operators/Scripts/gliveview/?h=gliveview) and [`nview`](https://github.com/blinklabs-io/nview) keep working out-of-the-box.
 >
 > We are planning, however, to add more metrics to Amaru.
+
+## Configuring OpenTelemetry
+
+Amaru provides a few options to configure OpenTelemetry:
+
+* `--amaru-service-name <STRING>` (environment variable `AMARU_SERVICE_NAME`): Sets the `service.name` key used to identify metrics and traces. This is useful when a single OTLP service stack collects telemetry from several Amaru instances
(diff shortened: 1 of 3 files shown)


monitoring/README.md
```diff
@@ -355,3 +355,11 @@ Coming soon.
 > The plan so far is to maximise compatibility with the existing Haskell node Prometheus metrics such that tools like [`gLiveView`](https://cardano-community.github.io/guild-operators/Scripts/gliveview/?h=gliveview) and [`nview`](https://github.com/blinklabs-io/nview) keep working out-of-the-box.
 >
 > We are planning, however, to add more metrics to Amaru.
+
+## Configuring OpenTelemetry
+
+Amaru provides a few options to configure OpenTelemetry:
+
+- `--amaru-service-name <STRING>` (environment variable `AMARU_SERVICE_NAME`): Sets the `service.name` key used to identify metrics and traces. This is useful when a single OTLP service stack collects telemetry from several Amaru instances
+- `--otlp_span_url <STRING>` (environment variable `AMARU_OTLP_SPAN_URL`): Sets the endpoint used to send spans, defaults to `http://localhost:4317`
```
jeluard, line 364: Are the `_` in the argument name a typo?

abailly: indeed, thanks a lot for spotting this

jeluard:
It looks like OpenTelemetry defines a number ([1](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/), [2](https://opentelemetry.io/docs/specs/otel/protocol/exporter/)) of standard variables to address. They might even be supported by default (or by reusing some bundled logic) by the libs we are using.

Worth looking at that?

PR #294, opened 2025-06-23
feat: generate more realistic arrival times

stevana:
## Summary by CodeRabbit

- **New Features**
  - Message arrival times in simulations are now randomized using an exponential distribution, providing more realistic and varied timing scenarios.
- **Tests**
  - Added tests to verify the accuracy of the exponential distribution sampling and the correctness of generated arrival times.

### DIFF
--- a/simulation/amaru-sim/src/simulator/generate.rs
+++ b/simulation/amaru-sim/src/simulator/generate.rs
@@ -17,6 +17,7 @@ use pure_stage::Instant;
 use rand::rngs::StdRng;
 use rand::Rng;
 use rand::SeedableRng;
+use rand_distr::{Distribution, Exp};
 use serde::Deserialize;
 use serde_json::Result;
 use std::collections::{BTreeMap, BTreeSet};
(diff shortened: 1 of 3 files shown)


simulation/amaru-sim/src/simulator/generate.rs
```diff
@@ -307,6 +309,28 @@ pub fn generate_inputs_strategy(
     })
 }
 
+fn exponential<R: Rng>(rng: &mut R, mean: f64) -> f64 {
```
abailly: I think the [rand_distr](https://docs.rs/rand_distr/latest/rand_distr/) crate provides standard distributions, perhaps we could use that?

abailly (approved):
LGTM, but worth to look at packaged distributions in `rand_distr` as I plan to use it in another PR

simulation/amaru-sim/src/simulator/generate.rs
stevana: Fixed.

PR #298, opened 2025-06-24
fix: reset SelectChain between test runs

stevana:
As it contains lingering state that causes chain selection to fail on
subsequent test runs.

## Summary by CodeRabbit

- **Refactor**
  - Modularized node initialization and stage wiring for improved simulation setup.
- **Style**
  - Added informational logging to enhance visibility into simulation execution and message handling.
- **Tests**
  - Updated simulation test to use an environment-driven seed for more flexible test configurations.

### DIFF
--- a/simulation/amaru-sim/src/simulator/mod.rs
+++ b/simulation/amaru-sim/src/simulator/mod.rs
@@ -41,11 +41,11 @@ use generate::{generate_inputs_strategy, parse_json, read_chain_json};
 use ledger::{populate_chain_store, FakeStakeDistribution};
 use proptest::test_runner::Config;
 use pure_stage::{simulation::SimulationBuilder, trace_buffer::TraceBuffer, StageRef};
-use pure_stage::{StageGraph, Void};
+use pure_stage::{Receiver, StageGraph, Void};
 use simulate::{pure_stage_node_handle, simulate, Trace};
 use std::{path::PathBuf, sync::Arc};
 use tokio::sync::Mutex;
-use tracing::Span;
+use tracing::{info, Span};
 
 pub use sync::*;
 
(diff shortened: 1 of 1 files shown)


simulation/amaru-sim/src/simulator/mod.rs
```diff
@@ -87,57 +87,36 @@ pub struct Args {
...
-    select_chain: SelectChain,
-    chain_data_path: &PathBuf,
-    seed: Option<u64>,
-) {
     let number_of_nodes = 1;
     let trace_buffer = Arc::new(parking_lot::Mutex::new(TraceBuffer::new(42, 1_000_000_000)));
     let buffer = trace_buffer.clone();
+
     let spawn = move || {
         println!("*** Spawning node!");
```
abailly: We should replace these `println!` with `info!` or the like

simulation/amaru-sim/src/simulator/mod.rs
```diff
@@ -87,57 +87,36 @@ pub struct Args {
...
-    validate_header: &mut ValidateHeader,
-    select_chain: SelectChain,
-    chain_data_path: &PathBuf,
-    seed: Option<u64>,
-) {
     let number_of_nodes = 1;
     let trace_buffer = Arc::new(parking_lot::Mutex::new(TraceBuffer::new(42, 1_000_000_000)));
     let buffer = trace_buffer.clone();
+
     let spawn = move || {
```
abailly: Can't we make `spawn` a function of its own now? Or a function that returns a `Fn` or something?

stevana: Fixed.

simulation/amaru-sim/src/simulator/mod.rs
stevana: Fixed.

PR #305, opened 2025-06-25
feat: don't persist empty schedules

### LINKED ISSUE #288: Simulation trace clutters tests output
When running simulations tests the full output of the simulator's logs are dumped as is to `stderr` which clutters the output of tests. We should probably write those logs to a temporary file and copy it in the working directory when the test fails. 

_(optional)_ We might want to provide a test execution flag to _always_ save the trace even when the test succeeds.

stevana:
There's a test that we expect to fail and which has a trace buffer of size 0, running this test used to cause failure schedule to be persisted.

Fixes #288

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented creation of empty schedule files when there is no trace data to save.  
- **New Features**
  - Schedule files are now saved with unique timestamped filenames in specified directories.  
  - Users receive clear feedback on the success or failure of schedule file saving operations.

### DIFF
--- a/simulation/amaru-sim/src/simulator/simulate.rs
+++ b/simulation/amaru-sim/src/simulator/simulate.rs
@@ -324,19 +324,14 @@ fn persist_schedule(
 
     let now = SystemTime::now();
 
-    let dir_str: &str = dir
-        .as_os_str()
-        .to_str()
-        .ok_or(anyhow::anyhow!("invalid bytes in dir"))?;
-
-    let path = PathBuf::from(format!(
-        "{}/{}-{}.schedule",
-        dir_str,
+    let filename = format!(
+        "{}-{}.schedule",
         prefix,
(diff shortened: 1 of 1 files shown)


simulation/amaru-sim/src/simulator/simulate.rs
```diff
@@ -310,6 +310,10 @@ fn persist_schedule(
     prefix: &str,
     trace_buffer: Arc<Mutex<TraceBuffer>>,
 ) -> Result<(), anyhow::Error> {
+    if trace_buffer.lock().used_size() == 0 {
```
abailly: Might be worthwhile to add  a couple "unit" tests for this function, perhaps?

stevana: Fixed

PR #306, opened 2025-06-25
chore: clarify IncompatibleVersion error message

pgrange:
When trying to connect to a cardano node using an incompatible protocol version the error message was not explicit:
```
Error: IncompatibleVersion
```

This commit changes that to:
```
failed to connect to peer <peer>: handshake version not accepted
```

It would be even better if we could share the versions supported by the remote node but we need to make pallas-network share this information with us so that's probably the best we can do right now.

Also note that, in the daemon, the first node with which we fail to connect will make the whole process to stop. Not sure this is intentional but that's what we had before so I keep it that way for now.

I'm not super happy with tracing errors and then returning them. I considered handling errors globally in the `main` function and I think this is what we should do but, at this point, doing so reduces the information that we can share and so make diagnosis a bit difficult. Such an approach would need to define amaru specific global errors that can be handled at a very high level but still carry enough information to be usefull... next time ;)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved error handling when connecting to peers by adding explicit error logging for connection failures.
- **New Features**
  - Enhanced peer connection process with a new asynchronous connection method for better reliability.

### DIFF
--- a/crates/amaru/src/bin/amaru/cmd/daemon.rs
+++ b/crates/amaru/src/bin/amaru/cmd/daemon.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crate::metrics::track_system_metrics;
+use crate::{cmd::connect_to_peer, metrics::track_system_metrics};
 use amaru::stages::{bootstrap, Config, StorePath};
 use amaru_kernel::{default_chain_dir, default_ledger_dir, network::NetworkName};
 use clap::{ArgAction, Parser};
@@ -70,9 +70,7 @@ pub async fn run(
 
     let mut clients: Vec<(String, Arc<Mutex<PeerClient>>)> = vec![];
     for peer in &config.upstream_peers {
-        let client =
-            PeerClient::connect(peer.clone(), config.network.to_network_magic() as u64).await
-            .inspect_err(|e| tracing::error!("failed to connect to peer {}: {}", peer, e))?;
+        let client = connect_to_peer(peer, &config.network).await?;
         clients.push((peer.clone(), Arc::new(Mutex::new(client))));
     }
 
--- a/crates/amaru/src/bin/amaru/cmd/import_headers.rs
+++ b/crates/amaru/src/bin/amaru/cmd/import_headers.rs
@@ -1,3 +1,4 @@
+use crate::cmd::connect_to_peer;
 use amaru::stages::{pull, PeerSession};
 use amaru_consensus::{consensus::store::ChainStore, peer::Peer, IsHeader};
 use amaru_kernel::{default_chain_dir, from_cbor, network::NetworkName, Header, Point};
@@ -5,10 +6,7 @@ use amaru_stores::rocksdb::consensus::RocksDBStore;
 use clap::Parser;
 use gasket::framework::*;
 use indicatif::{ProgressBar, ProgressStyle};
-use pallas_network::{
-    facades::PeerClient,
-    miniprotocols::chainsync::{self, HeaderContent, NextResponse},
-};
+use pallas_network::miniprotocols::chainsync::{self, HeaderContent, NextResponse};
 use std::{error::Error, path::PathBuf, sync::Arc, time::Duration};
 use tokio::{sync::Mutex, time::timeout};
 use tracing::info;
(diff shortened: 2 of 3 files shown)


Quantumplation:
If it's hard to include the version the *remote* supported versions... would it be possible to include *our* local supported version(s)?

crates/amaru/src/bin/amaru/cmd/daemon.rs
```diff
@@ -70,9 +70,15 @@ pub async fn run(
 
     let mut clients: Vec<(String, Arc<Mutex<PeerClient>>)> = vec![];
     for peer in &config.upstream_peers {
-        let client =
-            PeerClient::connect(peer.clone(), config.network.to_network_magic() as u64).await?;
-        clients.push((peer.clone(), Arc::new(Mutex::new(client))));
+        match PeerClient::connect(peer.clone(), config.network.to_network_magic() as u64).await {
```
yHSJ: Another option that is a bit less verbose would be to chain `.await.inspect_err(|e| tracing::error!("failed to connect to peer {}: {}", peer, e)?;`

jeluard:
>Such an approach would need to define amaru specific global errors that can be handled at a very high level but still carry enough information to be usefull... next time ;)

It's sounds like a good time to solve this properly and not push the ball further?

crates/amaru/src/bin/amaru/cmd/daemon.rs
```diff
@@ -70,9 +70,15 @@ pub async fn run(
...
+            Ok(client) => {
+                clients.push((peer.clone(), Arc::new(Mutex::new(client))));
+            }
+            Err(e) => {
+                tracing::error!("failed to connect to peer {}: {}", peer, e);
```
jeluard: Tracing is not logging, the philosophy is slightly different.
Besides we must ensure data logged will not [break JSON generation](https://github.com/pragma-org/amaru/issues/161) so tracing full errors is not advised.

crates/amaru/src/bin/amaru/cmd/daemon.rs
pgrange: Thanks for the feedback, greatly appreciated as I may often miss a rust idioms.

Done.

crates/amaru/src/bin/amaru/cmd/daemon.rs
pgrange: Hi @jeluard,

I just checked by injecting several _problematic_ characters in the error and they're all escaped correctly when I activate `--with-json-trace`. I did not observe any JSON generation breakage.

Seemingly, one can just inject whatever in this error message and it's escaped so seems to be safe in my tests. Do you have any specific exemple that could help me observe JSON generation breakage? That would help a lot.

Thanks.

jeluard: How is that escaped? Unless it changed in recent versions it's not. `"` breaks json.

Now the broader point is that traces are for oltp so it's better if what is traced is controlled. This is not regular logging.

pgrange: > How is that escaped? Unless it changed in recent versions it's not. " breaks json.

No idea. I just tested by injecting some problematic characters in the error message and I observe a valid JSON. For instance if I change my line to do this:

```
            .inspect_err(|e| tracing::error!("failed to connect \"to peer {}: {}", peer, "This is a \"problematic\" message with } and '"))?;
```

Running amaru outputs valid JSON:
```
#> cargo run --release -- --with-json-traces daemon  \
  --peer-address=$AMARU_PEER_ADDRESS \
  --network=preprod
{"timestamp":"2025-06-25T15:26:17.805240Z","level":"ERROR","fields":{"message":"failed to connect \"to peer 3.248.56.78:3001: This is a \"problematic\" message with } and '"},"target":"amaru::cmd::daemon"}
[...]
```

KtorZ:
@pgrange @jeluard 

I rewrote the error to be more structured and leverage fields:

```json
{"timestamp":"2025-07-08T11:42:02.621579Z","level":"ERROR","fields":{"message":"failed to connect to peer","peer":"127.0.0.1:3001","reason":"handshake version not accepted"},"target":"amaru::cmd"}
```

or

```
2025-07-08T11:44:28.022830Z ERROR amaru::cmd: failed to connect to peer 
  peer=127.0.0.1:3001 
  reason=handshake version not accepted
```

Which I believe makes everyone equally happy. At least it makes me happy.

KtorZ:
I have another mission for you @pgrange. Possibly for the next 2 months. Let's talk tomorrow.

PR #307, opened 2025-06-26
feat: Implement MemoryStore with ReadOnlyStore & TransactionContext support

whankinsiv:
This PR is a follow-up to a previous draft and introduces a MemoryStore implementation using RefCell-wrapped BTreeMaps. It now fully implements the ReadOnlyStore and TransactionalContext traits. It also introduces a suite of generic property based tests that work across both rocksDB and in-memory store implementations.

Main Additions:
- Defining a BTreeMap based MemoryStore wrapped in RefCells to allow mutability.
- Implementing TransactionalContext and ReadOnlyStore traits on MemoryStore.
- Implementing generic property based tests for ReadOnlyStore and TransactionalContext methods.

Organizational changes:
- Moved in_memory.rs definition from amaru-ledger to amaru-stores.
- Moved property-based generators from amaru-ledger to amaru-stores.

Feedback addressed from Draft PR:
- Removed Option wrappers from MemoryStore fields.
- Rewrote tests to be generic, provide deeper coverage, and use arbitrary generators. 

Remaining work:
- Support for storing and loading snapshots
- Integration into the bootstrap process for initial population. 

## Summary by CodeRabbit

* **New Features**
  * Introduced a new modular in-memory ledger store supporting accounts, pools, UTXOs, delegation representatives, proposals, committee members, and transactional operations.
  * Added public functions for adding, removing, and updating ledger entries across multiple columns in the in-memory store.
  * Enhanced property-based testing strategies and utilities for ledger columns, improving test coverage and robustness.
  * Enabled default instantiation for EraHistory and equality comparison for transaction outputs.
  * Added a public constructor for memoized Plutus data to facilitate CBOR serialization.

* **Bug Fixes**
  * Improved test utilities and property-based testing for ledger columns, increasing reliability.

* **Refactor**
  * Reorganized and expanded test code with broader conditional compilation for better visibility and reuse.
  * Updated in-memory store initialization to include era history context.
  * Removed the previous in-memory store implementation and replaced it with a new modular version.
  * Adjusted in-memory store construction and imports for consistency.
  * Refined pool update logic and test assertions for clarity and correctness.

* **Chores**
  * Updated dependency and feature configurations to improve modularity and test support.
  * Enhanced documentation and conditional compilation for test utilities and features.

* **Style**
  * Improved code organization and import structure for clarity and maintainability.


crates/amaru-stores/src/in_memory/mod.rs
```diff
@@ -0,0 +1,882 @@
...
+            .borrow()
+            .iter()
+            .map(|(stake_credential, row)| (stake_credential.clone(), row.clone()))
+            .collect();
+
+        Ok(accounts_vec.into_iter())
+    }
+
+    #[allow(refining_impl_trait)]
+    #[allow(refining_impl_trait)]
```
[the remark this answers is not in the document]
whankinsiv: Removed.

crates/amaru-stores/src/in_memory/mod.rs
```diff
@@ -0,0 +1,881 @@
...
+                        drep: None,
+                        rewards: 0,
+                        deposit,
+                    });
+
+            match delegatee {
+                Resettable::Set(val) => row.delegatee = Some(val),
+                Resettable::Reset => row.delegatee = None,
+                Resettable::Unchanged => {}
+            }
```
KtorZ: This is the definition of `Resettable::set_or_reset`

crates/amaru-stores/src/in_memory/mod.rs
```diff
@@ -0,0 +1,881 @@
...
+
+            match drep {
+                Resettable::Set(val) => row.drep = Some(val),
+                Resettable::Reset => row.drep = None,
+                Resettable::Unchanged => {}
+            }
```
KtorZ: Same here.

crates/amaru-stores/src/in_memory/mod.rs
```diff
@@ -0,0 +1,881 @@
...
+                    amaru_ledger::store::columns::dreps::Key,
+                    amaru_kernel::CertificatePointer,
+                ),
+            >,
+            impl Iterator<Item = amaru_ledger::store::columns::cc_members::Key>,
+            impl Iterator<Item = amaru_ledger::store::columns::proposals::Key>,
+        >,
+        withdrawals: impl Iterator<Item = amaru_ledger::store::columns::accounts::Key>,
+        voting_dreps: BTreeSet<StakeCredential>,
+    ) -> Result<(), amaru_ledger::store::StoreError> {
```
KtorZ, line 426: The whole function is almost 300 lines long, which should hint to the fact that sub-parts of it should be split into smaller function. 

Try to avoid inlining each operations related to each sub-store, and instead move them as separate functions.

KtorZ: Note that one of the goals of implementing the memory store is to identify and fix parts of the store which actually contains some business logic. So any duplication between two stores implementation is an indicator that some logic probably lives in the wrong place. 

Pure state management logic should live in the ledger, whereas specific manipulation of a particular column type should live with that column definition. The rest can be in the store.

crates/amaru-stores/src/in_memory/mod.rs
```diff
@@ -0,0 +1,881 @@
...
+        with(Box::new(iter));
+
+        let mut reconstructed: BTreeMap<TransactionInput, TransactionOutput> = values
+            .into_iter()
+            .filter_map(|(k, v)| Some((k, v?)))
+            .collect();
+
+        std::mem::swap(utxos.deref_mut(), &mut reconstructed);
+        Ok(())
+    }
```
KtorZ: Nice, this interface isn't particularly easy to wield; And somewhat a bit tailored to the needs of interacting with a real key:value store; So it feels a bit more artificial with the in-memory store but that's a rather nice solution. 

However, this could be slightly simplified by using functions already available on RefCell:

```rs
    fn with_utxo(
        &self,
        mut with: impl FnMut(amaru_ledger::store::columns::utxo::Iter<'_, '_>),
    ) -> Result<(), amaru_ledger::store::StoreError> {
        let utxos = self.store.utxos.take();

        let mut values: Vec<_> = utxos.into_iter().map(|(k, v)| (k, Some(v))).collect();

        let iter = values.iter_mut().map(|(k, v)| {
            let key = k.clone();
            let boxed: Box<dyn BorrowMut<Option<utxo::Value>>> = Box::new(RefMutAdapterMut::new(v));
            (key, boxed)
        });

        with(Box::new(iter));

        let reconstructed: BTreeMap<TransactionInput, TransactionOutput> = values
            .into_iter()
            .filter_map(|(k, v)| Some((k, v?)))
            .collect();

        self.store.utxos.replace(reconstructed);

        Ok(())
    }
```

Also, there's an apparent duplication between all the `with_...` implementations. The surrounding logic should be generalized and factored out:

```rs
fn with_column<K: Ord + Clone, V>(
    column: &RefCell<BTreeMap<K, V>>,
    mut with: impl FnMut(IterBorrow<'_, '_, K, Option<V>>),
) -> Result<(), amaru_ledger::store::StoreError> {
    let key_values = column.take();

    let mut items: Vec<_> = key_values.into_iter().map(|(k, v)| (k, Some(v))).collect();

    let iter = items.iter_mut().map(|(k, v)| {
        let key = k.clone();
        let boxed: Box<dyn BorrowMut<Option<V>>> = Box::new(RefMutAdapterMut::new(v));
        (key, boxed)
    });

    with(Box::new(iter));

    let reconstructed: BTreeMap<K, V> = items
        .into_iter()
        .filter_map(|(k, v)| Some((k, v?)))
        .collect();

    column.replace(reconstructed);

    Ok(())
}
```

So that we can simply write: 

```rs
    fn with_utxo(
        &self,
        with: impl FnMut(amaru_ledger::store::columns::utxo::Iter<'_, '_>),
    ) -> Result<(), amaru_ledger::store::StoreError> {
        with_column(&self.store.utxos, with)
    }
```

crates/amaru-stores/src/test_utils/pots.rs
```diff
@@ -0,0 +1,36 @@
...
+#[cfg(test)]
+pub(crate) mod test {
+    use amaru_kernel::{prop_cbor_roundtrip, Lovelace};
+    use amaru_ledger::store::columns::pots::Row;
+    use proptest::prelude::*;
+
+    prop_cbor_roundtrip!(Row, any_row());
+
+    prop_compose! {
+        pub fn any_row()(
```
KtorZ: I am not sure to understand the split logic here. The `prop_cbor_roundtrip` is a test, not a test_util; so I believe it should still live with the `pots`. And similarly for all columns.

Also, the generators (`any_...`) aren't really specific to the stores; so it is odd to find them now in this crate. They should rather belong to the `amaru-ledger` crate.

crates/amaru-stores/src/in_memory/mod.rs
```diff
@@ -0,0 +1,881 @@
...
+        test_epoch_transition(&store)?;
+        test_slot_updated(&store, &seeded)?;
+
+        // Validate removal tests
+        test_remove_utxo(&store, &seeded)?;
+        test_remove_account(&store, &seeded)?;
+        test_remove_pool(&store, &seeded)?;
+        test_remove_drep(&store, &seeded)?;
+        test_remove_proposal(&store, &seeded)?;
+        // TODO: Add cc_members iterator to validate removal works as intended
```
KtorZ: All these should be separate `test`; tests should be as much as possible, atomic and about one thing.

crates/amaru-stores/src/lib.rs
```diff
@@ -12,4 +12,720 @@
...
+        slots::tests::any_slot,
+        utxo::test::{any_pseudo_transaction_output, any_txin},
+    };
+
+    fn generate_txin() -> TransactionInput {
+        any_txin()
+            .new_tree(&mut TestRunner::default())
+            .unwrap()
+            .current()
+    }
```
KtorZ: This doesn't seem quite right since it probably always yields the same value? The name at least is misleading; if the intent is to provide some dummy value, then a static constant is probably better. If the intent is to have a strategy for generating arbitrary values, then it is already `any_txin`.

crates/amaru-stores/src/lib.rs
```diff
@@ -12,4 +12,720 @@
...
+                .unwrap()
+                .current();
+            row.hot_credential = Some(hot);
+        }
+
+        row
+    }
+
+    #[derive(Debug, Clone)]
+    pub struct SeededData {
```
KtorZ: The name is a bit weird; in test jargon, this would probably be called a _Fixture_.

crates/amaru-stores/src/test_utils/pots.rs
SupernaviX: Items defined in `#[cfg(test)]` are apparently not shareable between crates. So if generators are defined in `amaru-ledger` but needed in `amaru-stores`, they have to be outside of `#[cfg(test)]` blocks (and technically part of the crate's public API)

jeluard: They can, see what's been done [here](https://github.com/pragma-org/amaru/blob/6ba3c70c2b9c286464ee8cf095cf84eadc740951/crates/amaru-ledger/Cargo.toml#L31) (and search for `test-utils`)

crates/amaru-stores/src/in_memory/mod.rs
```diff
@@ -0,0 +1,881 @@
...
+
+        // Reset rewards data for accounts on withdrawal
+        for key in withdrawals {
+            if let Some(account) = self.store.accounts.borrow_mut().get_mut(&key) {
+                account.rewards = 0;
+            }
+        }
+
+        Ok(())
+    }
```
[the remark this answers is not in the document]
whankinsiv: Resolved. The save method has been reduced to 93 lines.

crates/amaru-stores/src/in_memory/mod.rs
```diff
@@ -0,0 +1,881 @@
...
+            .ok_or_else(|| StoreError::Internal("tip not yet set".into()))
+    }
+}
+
+// TODO: Implement HistoricalStores on MemoryStore (currently returns a new empty MemoryStore)
+impl HistoricalStores for MemoryStore {
+    fn for_epoch(&self, _epoch: Epoch) -> Result<impl Snapshot, amaru_ledger::store::StoreError> {
+        let era_history: &EraHistory = NetworkName::Preprod.into();
+        Ok(MemoryStore::new(era_history.clone()))
+    }
```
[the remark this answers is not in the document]
whankinsiv: Resolved.

crates/amaru-stores/src/in_memory/mod.rs
whankinsiv: Updated to use set_or_reset [here](https://github.com/pragma-org/amaru/blob/8442d766bee6a4d036da88b6763647a7fa3c9e6b/crates/amaru-stores/src/in_memory/ledger/columns/accounts.rs#L23-L24).

crates/amaru-stores/src/in_memory/mod.rs
whankinsiv: Updated to use set_or_reset [here](https://github.com/pragma-org/amaru/blob/8442d766bee6a4d036da88b6763647a7fa3c9e6b/crates/amaru-stores/src/in_memory/ledger/columns/accounts.rs#L23-L24).

crates/amaru-stores/src/in_memory/mod.rs
whankinsiv: Updated [here](https://github.com/pragma-org/amaru/blob/8442d766bee6a4d036da88b6763647a7fa3c9e6b/crates/amaru-stores/src/in_memory/mod.rs#L290-L320) to use RefCell functions inside a generalized with_columns function.

crates/amaru-stores/src/in_memory/mod.rs
whankinsiv: Fixed [here](https://github.com/pragma-org/amaru/blob/8442d766bee6a4d036da88b6763647a7fa3c9e6b/crates/amaru-stores/src/in_memory/mod.rs#L567-L650).

crates/amaru-stores/src/lib.rs
whankinsiv: Fixed [here](https://github.com/pragma-org/amaru/blob/8442d766bee6a4d036da88b6763647a7fa3c9e6b/crates/amaru-stores/src/lib.rs#L159).

crates/amaru-stores/src/in_memory/mod.rs
whankinsiv: I have reduced the function to 93 lines by using a generalized with_columns function and moving the add and remove operations to their own submodules as is done in the rocksdb implementation.

whankinsiv:
Hi @KtorZ `ledger-in-nodejs` now fails to build in wasm due to moving `in_memory.rs` to `amaru-stores`, as `examples/shared` now depends on `amaru-stores`. Would you prefer moving the in-memory implementation back to `amaru-ledger` or should this go in its own crate? The generalized tests will also need to be moved out of `amaru-stores` so that the in-memory tests are not dependent on the crate. Please advise.

jeluard:
For context `ledger-in-nodejs` is merely here to ensure amaru can still be used via wasm from nodejs. Longer term for real use-cases I would expect that implementations of `Stores` trait are done using `IndexedDB` or similar.

I can see a couple options:

* move back the store to it's previous location
* keep a copy of the previous store in the original location
* use `cfg` to allow to selectively enable `rocksdb` backends when using `amaru-stores`

My personal choice would be the 3rd option.

whankinsiv:
> For context `ledger-in-nodejs` is merely here to ensure amaru can still be used via wasm from nodejs. Longer term for real use-cases I would expect that implementations of `Stores` trait are done using `IndexedDB` or similar.
> 
> I can see a couple options:
> 
> * move back the store to it's previous location
> * keep a copy of the previous store in the original location
> * use `cfg` to allow to selectively enable `rocksdb` backends when using `amaru-stores`
> 
> My personal choice would be the 3rd option.

Thank you for your input @jeluard. I've added the cfg flag to `rocksdb` [here](https://github.com/pragma-org/amaru/pull/307/commits/b905ec758fc293dbe52c973233caa28f947fa542). `ledger-in-nodejs` now builds successfully.

whankinsiv:
Hi @jeluard. It appears likely that the reason this PR is failing to build with `x86_64-pc-windows-msvc` is due to changes you made [here](https://github.com/pragma-org/amaru/commit/0d91c74b94df0e90333da2661f947cc582529d5e). When using the `--profile dev-debug` flag the tests consistently fail but previously passed. Is it okay to remove this line from the CI?

jeluard:
Is this test specific when it comes to tack requirement? The changes you mention have been introduced in march and that's the only failing test.

crates/amaru-ledger/src/store/columns/accounts.rs
```diff
@@ -80,54 +80,55 @@ impl<'a, C> cbor::decode::Decode<'a, C> for Row {
         })
     }
 }
+#[cfg(any(test, feature = "test-utils"))]
```
yHSJ, line 83: Nitpick: Instead of using an attribute for each file, why not just define a public test module and apply the attribute to the whole module?

Makes it easier to work with later, less likely to accidentally change/remove an attribute.

yHSJ: This seems to apply to almost every file where you added the feature conditional compilation

crates/amaru-stores/src/in_memory/mod.rs
```diff
@@ -0,0 +1,698 @@
...
+            dreps: RefCell::new(BTreeMap::new()),
+            proposals: RefCell::new(BTreeMap::new()),
+            cc_members: RefCell::new(BTreeMap::new()),
+            p_params: RefCell::new(BTreeMap::new()),
+            era_history,
+        }
+    }
+}
+
+// TODO: Implement Snapshot on MemoryStore (Currently returns hard-coded epoch 10)
```
yHSJ, line 106: Since this isn't actually being used anywhere, this is fine. But, would like to see this get implemented at somepoint soon.

crates/amaru-stores/src/in_memory/mod.rs
```diff
@@ -0,0 +1,698 @@
...
+
+    #[test]
+    fn test_in_mem_slot_updated() -> Result<(), StoreError> {
+        let mut runner = TestRunner::default();
+        let (store, fixture) =
+            setup_memory_store(&mut runner).expect("Failed to add test data to store");
+        test_slot_updated(&store, &fixture)
+    }
+
+    /* Disabled until MemoizedTransactionOutput is created
```
yHSJ: `MemoizedTransactionOutput` exists now. Please uncomment and fix the tests if necessary, we shouldn't merge commented code.

crates/amaru-stores/src/lib.rs
```diff
@@ -12,4 +12,683 @@
...
+    pub fn add_test_data_to_store(
+        store: &impl Store,
+        era_history: &EraHistory,
+        runner: &mut TestRunner,
+    ) -> Result<Fixture, StoreError> {
+        use diff_bind::Resettable;
+
+        // utxos
+        /*
+        let txin = any_txin().new_tree(runner).unwrap().current();
```
yHSJ, line 92: Same here, why is there commented out code?

crates/amaru-stores/src/lib.rs
```diff
@@ -12,4 +12,683 @@
...
+            proposal_key,
+            #[cfg(not(target_os = "windows"))]
+            proposal_row,
+            slot,
+            slot_leader,
+        })
+    }
+
+    /*
+    pub fn test_read_utxo(store: &impl ReadStore, fixture: &Fixture) {
```
yHSJ, line 268: Commented out code -- decide whether it needs to exist or not

crates/amaru-stores/src/lib.rs
```diff
@@ -12,4 +12,683 @@
...
+            stored_proposal.valid_until, fixture.proposal_row.valid_until,
+            "proposal valid_until mismatch"
+        );
+        assert_eq!(
+            stored_proposal.proposal, fixture.proposal_row.proposal,
+            "proposal data mismatch"
+        );
+    }
+
+    /* Disabled until MemoizedTransactionOutput generator is created
```
yHSJ: Either remove it (we can add it later) or create a generator for `MemoizedTransactionOutput` (which would be better)

crates/amaru-stores/src/rocksdb/mod.rs
```diff
@@ -743,6 +766,111 @@ mod tests {
...
+    }
+
+    #[test]
+    fn test_rocksdb_slot_updated() -> Result<(), StoreError> {
+        let mut runner = TestRunner::default();
+        let (store, fixture) = setup_rocksdb_store(&mut runner)?;
+        test_slot_updated(&store, &fixture)
+    }
+
+    /* Disabled until MemoizedTransactionOutput is created
```
yHSJ: Same as above

examples/shared/src/lib.rs
```diff
@@ -21,13 +21,14 @@ pub fn forward_ledger(raw_block: &str) {
 
     let (_hash, block): BlockWrapper = cbor::decode(&bytes).unwrap();
     let network = NetworkName::Preprod;
-    let era_history: &EraHistory = network.into();
+    let era_history: &EraHistory = NetworkName::Preprod.into();
```
yHSJ: Not sure why this changed

jeluard:
@whankinsiv Is this PR done? I can see there are still some `Remaining work` items.

crates/amaru-stores/Cargo.toml
```diff
@@ -14,9 +14,9 @@ rust-version.workspace = true
 
 [dependencies]
 hex.workspace = true
-rocksdb.workspace = true
+rocksdb = { workspace = true, optional = true }
```
jeluard, line 17: I'm curious about the scenario that would require this?

whankinsiv: This was added so that `examples/ledger-in-nodejs` can build properly. Although this isn't complete yet, I would like to get it merged into main and deliver the remaining work as smaller incremental PRs. I have been assigned to full-time Acropolis work but will continue spending a couple hours a day working to get this into a fully usable alternative to RocksDB.

crates/amaru-stores/src/lib.rs
whankinsiv: I have implemented `any_memoized_transaction_output` and enabled the `test_read_utxo` and `test_remove_utxo` tests for both in-memory and rockdb stores.

crates/amaru-stores/src/rocksdb/mod.rs
whankinsiv: I have implemented `any_memoized_transaction_output` and enabled the `test_read_utxo` and `test_remove_utxo` tests for both in-memory and rocksdb stores.

examples/shared/src/lib.rs
whankinsiv: I have reverted this change.

crates/amaru-stores/src/lib.rs
whankinsiv: I have implemented `any_memoized_transaction_output` and enabled the `test_read_utxo` and `test_remove_utxo` tests for both in-memory and rocksdb stores.

crates/amaru-stores/src/lib.rs
whankinsiv: fixed.

crates/amaru-stores/src/in_memory/mod.rs
whankinsiv: fixed.

crates/amaru-stores/src/in_memory/mod.rs
whankinsiv: I will implement Snapshot in a follow up PR.

crates/amaru-ledger/src/store/columns/accounts.rs
whankinsiv: fixed for all `/columns` files.

PR #308, opened 2025-06-26
feat: add consensus metrics

abailly:
This PR introduces tracking a bunch of metrics related to the consensus pipeline stages, following our [EDR](https://github.com/pragma-org/amaru/blob/main/engineering-decision-records/007-observability.md).

## Summary by CodeRabbit

* **New Features**
  * Introduced comprehensive consensus metrics collection, including tracking of chain tip slot, chain selection size, block and header validation, storage, and peer message counts.
  * Metrics are now integrated across consensus, block fetch, header validation, chain selection, and ledger validation stages.
  * Added OpenTelemetry support for consensus metrics, enabling enhanced observability.

* **Documentation**
  * Expanded monitoring documentation to describe new consensus metrics and updated the CLI option for the OpenTelemetry metrics endpoint.

### DIFF
--- a/monitoring/README.md
+++ b/monitoring/README.md
@@ -347,9 +347,29 @@ Filters can be provided as a sequence of `,`-separated values. Right-most filter
 
 ø
 
+
+
 ## Metrics
 
-Coming soon.
+### target: `consensus`
+
+These metrics are defined in the [amaru-consensus](https://github.com/pragma-org/amaru/blob/74652de42695ba47572d5bebb7383ec24d75bbdc/crates/amaru-consensus/src/lib.rs#L102) module.
+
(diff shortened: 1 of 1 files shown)


jeluard:
Still a draft?

abailly:
Yes, for 2 reasons:
* It breaks Wasm compilation because of dependency on opentelemetry so need to rethink the approach
* I want to revisit the metrics as most of them are redundant with what you can get from spans

jeluard:
Have you considered using [global::meter](https://github.com/open-telemetry/opentelemetry-rust/blob/main/examples/metrics-basic/src/main.rs#L30)? That and `LazyCell` might allow to define globally accessible meters.

PR #311, opened 2025-06-26
Collateral Validations

yHSJ:
This PR introduces collateral validation, which is now separate from `inputs` where it was (incorrectly, and partially) implemented previously.

In the Haskell node, `feesOk` validates the collateral inputs while a separate function, `validateTooManyCollateralInputs`, validates the max collateral inputs.

TODO:
- [x] Implement validation logic
- [x] Write unit tests for collateral code covearge 

## Summary by CodeRabbit

* **New Features**
  * Collateral validation added and applied only for transactions with redeemers.
  * Improved transaction input display and broader availability of on-chain types.

* **Bug Fixes**
  * Value deserialization now handles coin-only and full multi-asset formats.

* **Tests**
  * Numerous new ledger test scenarios added; UTXO value representations standardized across fixtures.


crates/amaru-ledger/src/rules/transaction/collateral.rs
```diff
@@ -0,0 +1,221 @@
...
+#[derive(Debug, Error)]
+pub enum InvalidCollateral {
+    #[error("Unknown input: {0}")]
+    UnknownInput(DisplayableTransactionInput),
+    #[error("too many collateral inputs: provided: {provided} allowed: {allowed}")]
+    TooManyInputs { provided: usize, allowed: usize },
+    #[error("a collateral input is locked at a script address: {0}")]
+    LockedAtScriptAddress(DisplayableTransactionInput),
+    #[error("a collateral input contains non ADA value: {0}")]
+    ContainsNonAda(DisplayableTransactionInput),
```
Quantumplation: This doesn't look used anymore

crates/amaru-kernel/src/lib.rs
```diff
@@ -214,6 +214,36 @@ impl<'b> Decode<'b, ()> for Point {
     }
 }
 
+#[derive(Debug, Clone)]
+pub struct TransactionInputAdapter(TransactionInput);
+pub type DisplayableTransactionInput = TransactionInputAdapter;
```
Quantumplation: This feels like a strange construct; I think I understand why the adapter is needed, but the type alias seems pointless, and it's strange to have "Displayable__" leak into the name;

Instead, I would just call this transaction input, and have it wrap a fully qualified `pallas::TransactionInput`; it's purpose won't be *just* to make it displayable; it's to isolate Pallas details from the rest of the code so that if we need to replace it, it's less annoying.

It just so happens that we also provide additional / alternative display implementations for this struct. We may add other implementations to it as well over time.

yHSJ: I agree it's strange to have `Displayable` leak into the name. I added that for clarity at the call site, what the adapter is being used for.

> Instead, I would just call this transaction input, and have it wrap a fully qualified pallas::TransactionInput; it's purpose won't be just to make it displayable; it's to isolate Pallas details from the rest of the code so that if we need to replace it, it's less annoying.


This is what I initially tried, but it would mean we're (albeit very slightly) increasing the size of our on disk storage without significant changes to the state that felt out of scope for this PR. It also means that we're doing additional iterations during validation just to convert `TransactionInput`s to `TransactionInputAdapter`s, which I figured we'd rather avoid.

So, I think I'll attempt to replace `TransactionInput` with `TransactionInputAdapter` everywhere else in the codebase in a different PR.

crates/amaru-ledger/src/rules/transaction/collateral.rs
```diff
@@ -0,0 +1,336 @@
...
+        match balance.multiasset.get_mut(&key) {
+            Some(v) => {
+                *v -= value;
+                if *v == 0 {
+                    balance.multiasset.remove(&key);
+                }
+            }
+            None => return Err(InvalidCollateral::ValueNotConserved),
+        };
+    }
```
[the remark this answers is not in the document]
jeluard: @yHSJ Is this of concern to us here?

crates/amaru-kernel/src/memoized/transaction_output.rs
```diff
@@ -137,6 +139,50 @@ impl<'a> TryFrom<MintedTransactionOutput<'a>> for MemoizedTransactionOutput {
     }
 }
 
+// TODO: avoid clones (just getting the rebase to continue, will fix after rebase)
```
jeluard: @Jsh I understand you plan to address this as part of this PR?

crates/amaru-kernel/src/memoized/transaction_output.rs
```diff
@@ -206,11 +252,57 @@ fn serialize_value<S: serde::ser::Serializer>(
...
+        ValueHelper::Multiasset(coin, multiasset_data) => {
+            let mut converted_multiasset = Vec::new();
+
+            for (policy_id, assets) in multiasset_data {
+                let policy_id = hex::decode(&policy_id).map_err(|_| {
+                    serde::de::Error::custom(format!("invalid hex string: {policy_id}"))
+                })?;
+
+                let mut converted_assets = Vec::new();
+                for (asset_name, quantity) in assets {
```
jeluard, line 365: Looks like a map could do the job?

crates/amaru-ledger/src/rules/transaction/collateral.rs
```diff
@@ -0,0 +1,312 @@
...
+    pub coin: u64,
+    pub multiasset: BTreeMap<Vec<u8>, u64>,
+}
+
+impl From<&Value> for CollateralBalance {
+    fn from(value: &Value) -> Self {
+        match value {
+            Value::Multiasset(coin, multiasset) => {
+                let mut map = BTreeMap::new();
+                multiasset.iter().for_each(|(policy, assets)| {
```
jeluard: Also smells like could be do without `for_each`

crates/amaru-kernel/src/memoized/transaction_output.rs
yHSJ: it's much easier to handle the errors with an imperative for loop here than with a `.iter().map()`. It's possible of course, but the code is less readable imo

crates/amaru-ledger/src/rules/transaction/collateral.rs
yHSJ: Sure, could be a map, save a few lines!

crates/amaru-kernel/src/memoized/transaction_output.rs
yHSJ: Yes, I totally forgot about this. Thanks!

crates/amaru-kernel/src/lib.rs
```diff
@@ -268,6 +271,35 @@ impl From<&[u8]> for RawBlock {
...
+    fn from(value: TransactionInput) -> Self {
+        Self(value)
+    }
+}
+
+impl From<&TransactionInput> for TransactionInputAdapter {
+    fn from(value: &TransactionInput) -> Self {
+        value.clone().into()
+    }
+}
```
KtorZ: I'd rather not have that. That's a hidden clone, and we already have `From<TransactionInput> for TransactionInputAdapter`. So if you have a reference which needs transformation, it's better to let the caller clone.

crates/amaru-ledger/src/rules/transaction.rs
```diff
@@ -124,12 +130,22 @@ where
             .reference_inputs
             .as_deref()
             .map(|vec| vec.as_slice()),
-        transaction_body
-            .collateral
-            .as_deref()
-            .map(|vec| vec.as_slice()),
     )?;
 
+    if transaction_witness_set.redeemer.is_some() {
```
KtorZ, line 135: This condition seems like it could backfire. What if a transaction has no redeemer but provides (potentially invalid) collateral? We should validate collateral always, and if they are not necessary but present, then it's an error. A test should be illustrating this.

crates/amaru-ledger/src/rules/transaction/collateral.rs
```diff
@@ -0,0 +1,363 @@
...
+ * i64 has a lower maximum than u64 due to the data structure, but the value for coins must always be in the range:
+ *  [-9223372036854775808, 18446744073709551615]
+ * Which is within the bounds of an i64, so conversions are safe i64<->u64
+ */
+pub struct CollateralBalance {
+    pub coin: i64,
+    pub multiasset: BTreeMap<Vec<u8>, i64>,
+}
+
+impl fmt::Debug for CollateralBalance {
```
KtorZ:
```suggestion
impl fmt::Display for CollateralBalance {
```

Do not write `Debug` instances by hand, unless you have a really good reason to.

crates/amaru-ledger/src/rules/transaction/collateral.rs
```diff
@@ -0,0 +1,363 @@
...
+                Some(v) => {
+                    *v -= value;
+                    if *v == 0 {
+                        self.multiasset.remove(&key);
+                    }
+                }
+                None => {
+                    self.multiasset.insert(key, -value);
+                }
+            };
```
KtorZ: This would be better written using `.entry`

crates/amaru-ledger/src/rules/transaction/collateral.rs
```diff
@@ -0,0 +1,363 @@
...
+        return Err(InvalidCollateral::IncorrectTotalCollateral {
+            provided: balance.coin as u64,
+            expected: expected_balance,
+        });
+    }
+
+    Ok(())
+}
+
+fn add_value_to_balance(balance: &mut CollateralBalance, output: &MemoizedTransactionOutput) {
```
KtorZ: Why not simply a method on CollateralBalance?

crates/amaru-ledger/src/rules/transaction/collateral.rs
```diff
@@ -0,0 +1,363 @@
...
+        },
+        None => CollateralBalance {
+            coin: 0,
+            multiasset: BTreeMap::new(),
+        },
+    };
+
+    balance.sub(collateral_return_balance);
+
+    if !balance.multiasset.is_empty() || balance.coin < 0 {
```
KtorZ: This sounds like there should be a `is_empty()` method on `CollateralBalance`.

crates/amaru-ledger/src/rules/transaction/collateral.rs
```diff
@@ -0,0 +1,363 @@
...
+    C: UtxoSlice,
+{
+    let collaterals = collaterals
+        .filter(|c| !c.is_empty())
+        .ok_or(InvalidCollateral::NoCollateral)?;
+
+    let mut balance = CollateralBalance {
+        coin: 0,
+        multiasset: BTreeMap::new(),
+    };
```
KtorZ: Implementation details are leaked here, please define a `::new()` or `::default()` or `::empty()` method for `CollateralBalance`.

crates/amaru-ledger/src/rules/transaction/collateral.rs
```diff
@@ -0,0 +1,363 @@
...
+
+            amaru_kernel::PseudoTransactionOutput::PostAlonzo(output) => {
+                CollateralBalance::from(&output.value)
+            }
+        },
+        None => CollateralBalance {
+            coin: 0,
+            multiasset: BTreeMap::new(),
+        },
+    };
```
KtorZ: That code is too low-level for rules; please refactor so that the pattern-match lives inside kernel's helper functions or CollateralBalance's helper function.

crates/amaru-ledger/src/rules/transaction/collateral.rs
```diff
@@ -0,0 +1,363 @@
...
+    if !balance.multiasset.is_empty() || balance.coin < 0 {
+        return Err(InvalidCollateral::ValueNotConserved(balance));
+    }
+
+    // We're avoiding floating points and truncating values (exactly what the Haskell node does)
+    // so we check that balance * 100 = fee * collPercentage
+    // When we display to the user, we want to display minimum_collateral in lovelace, not in 100ths of a lovelace,
+    // so we divide by 100 and round up
+    let minimum_collateral = fee * protocol_parameters.collateral_percentage as u64;
+    if balance.coin * 100 < minimum_collateral as i64 {
```
KtorZ: Can't this overflow?

crates/amaru-ledger/src/rules/transaction.rs
yHSJ: https://github.com/IntersectMBO/cardano-ledger/blob/071530a66ac9ef8474b87f68765573c9edd2a5b4/eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxo.hs#L195-L222

The Haskell node provides some additional context. The logic that *always* runs, is the `minFee` check. That is done by a different submodule in the case of Amaru. The rest of it depends on the presence of redeemers.

You're able to provide otherwise invalid collateral if there is no provided redeemer (why would a node waste time validating the collateral if it's never being spent?)

KtorZ: Interesting. I'd love a test that captures that.

> why would a node waste time validating the collateral if it's never being spent?
Because a node cares that only _(phase-1) valid_ transactions make it to the ledger. In the same way that a node checks that dreps or pools you're trying to delegate to exist; there's no damage to the system if they don't really. But it's just the right thing to do. 

Users have no reasons to set a collateral if that's not needed. But I guess, the Haskell ledger doesn't actually care here.

PR #313, opened 2025-06-27
chore: remove unused pub fns

jeluard:
Remove unused pub functions as detected by [cargo-workspace-unused-pub](https://github.com/cpg314/cargo-workspace-unused-pub)

## Summary by CodeRabbit

* **Refactor**
  * Removed several public methods related to execution unit calculation and slot/time conversions, resulting in a simplified interface for users.

### DIFF
--- a/crates/amaru-kernel/src/lib.rs
+++ b/crates/amaru-kernel/src/lib.rs
@@ -825,25 +825,6 @@ impl HasExUnits for MintedBlock<'_> {
     }
 }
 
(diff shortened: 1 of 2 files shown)


abailly (changes requested):
I would defer to @rkuhn for `pure-stage` functions, I think those are unused _for now_ but they will be useful as we extend the scope of simulation testing.

abailly (approved):

PR #314, opened 2025-06-27
Memoize transaction outputs.

KtorZ:
After running into yet-another-reserialisation issue, I decided to byte the bullet. The whole thing turns out quite okay in the end; although we have quite some boilerplate to bridge things with Pallas. My hope is that we can eventually converge towards a similar solution in Pallas too. 

Although, we're also removing boilerplate in other places thanks to this. So all-in-all, it's not a bad trade-off.

## Summary by CodeRabbit

* **New Features**
  * Introduced memoized data types for transaction outputs, scripts, and datums, enhancing performance and consistency.
  * Added comprehensive serialization and deserialization support for memoized types, compatible with legacy and modern transaction formats.

* **Refactor**
  * Replaced legacy transaction output and script types with memoized variants across the codebase.
  * Simplified script type representation and unified script hash computation.
  * Updated datum and script field names and structures for improved clarity and consistency.
  * Refactored context and validation layers to utilize memoized types.

* **Bug Fixes**
  * Enhanced error handling and validation during transaction output conversions and address parsing.

* **Chores**
  * Standardized and enriched test data files with explicit value, datum, and script fields.
  * Removed deprecated proxy types and streamlined serialization logic.

### DIFF
--- a/crates/amaru-kernel/src/memoized/native_script.rs
+++ b/crates/amaru-kernel/src/memoized/native_script.rs
@@ -30,6 +30,12 @@ impl MemoizedNativeScript {
     }
 }
 
+impl AsRef<NativeScript> for MemoizedNativeScript {
+    fn as_ref(&self) -> &NativeScript {
+        &self.expr
+    }
+}
+
 impl TryFrom<Bytes> for MemoizedNativeScript {
     type Error = String;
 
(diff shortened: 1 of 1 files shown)


crates/amaru-kernel/src/lib.rs
```diff
@@ -267,29 +267,27 @@ impl RedeemersExt for Redeemers {
...
-// This allows us to avoid cloning, but it's a pretty awful API.
-// Ideally, this is something that Pallas would own and cleanup.
-#[derive(Debug, PartialEq, Eq)]
-pub enum BorrowedScript<'a> {
-    NativeScript(&'a NativeScript),
-    PlutusV1Script(&'a PlutusScript<1>),
-    PlutusV2Script(&'a PlutusScript<2>),
-    PlutusV3Script(&'a PlutusScript<3>),
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+pub enum ScriptKind {
```
yHSJ, line 272: Does the type that already exists in Pallas [here](https://github.com/txpipe/pallas/blob/v0.32.1/pallas-txbuilder/src/transaction/model.rs#L515) not meet our needs?

crates/amaru-kernel/src/memoized.rs
```diff
@@ -0,0 +1,265 @@
...
+pub struct MemoizedTransactionOutput {
+    is_legacy: bool,
+    pub address: Address,
+    pub value: Value,
+    pub datum: MemoizedDatum,
+    pub script: Option<MemoizedScript>,
+}
+
+impl<'b, C> cbor::Decode<'b, C> for MemoizedTransactionOutput {
+    // TODO: This implementation still copies the bytes when we convert from the minted version to
```
yHSJ: Yes, this is a great first step and makes things easier to work with, but the true benefit will be once we update the decoding.

Are we concerned that with Pallas v1 this "issue" will be solved, thus rendering this work obsolete? Or is the plan that this becomes the solution in Pallas v1

crates/amaru-kernel/src/memoized.rs
```diff
@@ -0,0 +1,265 @@
...
+    // the owned version. To avoid that, we should actually rewrite the decoders for all the
+    // memoized structures, so that they can retain their original bytes.
+    fn decode(d: &mut cbor::Decoder<'b>, ctx: &mut C) -> Result<Self, cbor::decode::Error> {
+        let minted: MintedTransactionOutput<'_> = d.decode_with(ctx)?;
+        Self::try_from(minted).map_err(cbor::decode::Error::message)
+    }
+}
+
+impl<C> cbor::Encode<C> for MemoizedTransactionOutput {
+    fn encode<W: cbor::encode::Write>(
```
yHSJ: I understand what's going on here since I've spent so much time staring at the CDDL recently, but it is probably helpful to provide some comments here providing context to the next reader (or ourselves in 6 months at 3am trying to fix a bug...).

Maybe even just a "the cbor encoding for a transaction output is defined [here](https://github.com/IntersectMBO/cardano-ledger/blob/master/eras/conway/impl/cddl-files/conway.cddl#L162)"

crates/amaru-kernel/src/memoized.rs
```diff
@@ -0,0 +1,265 @@
...
+        } else {
+            e.begin_map()?;
+
+            e.u8(0)?;
+            e.bytes(&self.address.to_vec())?;
+
+            e.u8(1)?;
+            e.encode_with(&self.value, ctx)?;
+
+            match &self.datum {
```
yHSJ: Perhaps it makes more sense to have the datum encoding belong in it's own implementation of `minicbor::Encode`?

crates/amaru-kernel/src/memoized.rs
```diff
@@ -0,0 +1,265 @@
...
+                MemoizedDatum::Inline(data) => {
+                    e.u8(2)?;
+                    e.array(2)?;
+                    e.u8(1)?;
+                    e.tag(IanaTag::Cbor)?;
+                    e.bytes(data.original_bytes())?;
+                }
+            }
+
+            match &self.script {
```
yHSJ: Same as datum, can exist as it's own implementation of `minicbor::Encode`

crates/amaru-kernel/src/memoized.rs
KtorZ: Definitely. This is just preliminary / drafty. I stashed that away in a PR, with the goal to first to the whole vertical and see if it *actually solves the problem* as expect 😅. 

Loads of refactoring and commenting to come.

crates/amaru-kernel/src/lib.rs
KtorZ: Yes, but it's in `txbuilder` which don't depend on and I don't want to start depending on it :/ ... I'll add a TODO, as it should likely just live in pallas-primitives to begin with.

KtorZ:
> I'm a fan of these changes -- they make a lot of sense. While it _does_ feel like we're just moving the problem around, at least it feels more isolated and easier to "fix" later.

Well, moving the problem around is what we've kept doing here and there; but this PR actually finally addresses the problem of carrying the decoded bytes all-the-way to the database, and preserving them upon reading from the database. 

Funny enough, I had to do this not because of a non-canonical serialisation issue, but because of a bug in pallas-primitive (for which I'll issue a fix ASAP) which generates invalid encoding for native scripts 😬. So I thought I might as well bite the bullet and see where it leads us. 

It's not too bad in the end.

crates/amaru-kernel/src/lib.rs
Quantumplation: That's something we can have one of our new hires tackle :)

PR #316, opened 2025-06-30
feat: allow ledgerdb to be concurrently opened for reading

geofflittle:
This change introduces `open_for_read_only` to both `RocksDB` and `RocksDBHistoricalStores`, enabling concurrent read access without requiring a write lock.

This is particularly useful for tools like amaru-doctor, which need to inspect the ledger database while amaru is actively running and modifying it.


crates/amaru-stores/src/rocksdb/mod.rs
```diff
@@ -125,82 +143,119 @@ impl RocksDB {
...
-        Ok(RocksDB {
-            dir: dir.to_path_buf(),
-            incremental_save: false,
-            db: OptimisticTransactionDB::open(&opts, dir.join("live"))
-                .map_err(|err| StoreError::Internal(err.into()))?,
-            era_history: era_history.clone(), // TODO: remove clone?
-            ongoing_transaction: OngoingTransaction::new(),
-        })
+    #[allow(dead_code)]
+    fn open_for_read_only(dir: &Path) -> Result<ROLedgerDB, StoreError> {
```
jeluard: Don't we want those to be pub?

crates/amaru-stores/src/rocksdb/mod.rs
```diff
@@ -125,82 +143,119 @@ impl RocksDB {
...
-        Ok(RocksDB {
-            dir: dir.to_path_buf(),
-            incremental_save: true,
-            db: OptimisticTransactionDB::open(&opts, dir.join("live"))
-                .map_err(|err| StoreError::Internal(err.into()))?,
-            era_history: era_history.clone(),
-            ongoing_transaction: OngoingTransaction::new(),
-        })
+        OptimisticTransactionDB::open(&opts, dir.join("live"))
+            .map(|db| Self {
```
jeluard, line 151: That's a pattern I wasn't expecting! Not sure how I feel about it :)

crates/amaru-stores/src/rocksdb/mod.rs
```diff
@@ -125,82 +143,119 @@ impl RocksDB {
...
+                "unable to decode value ({}) for key ({}): {e:?}",
+                hex::encode(&value),
+                hex::encode(&key)
+            )
+        });
+        (k, v)
+    })
+}
+
+pub fn iter_db<'a, K, V>(
```
jeluard: The iter / iter_db (and same for get) duplication could be removed by introducing some traits that would be implemented by twice. Not yet clear if it's worth the hassle.

crates/amaru-stores/src/rocksdb/ledger/columns/accounts.rs
```diff
@@ -101,16 +101,28 @@ pub fn reset_many<DB>(
...
-        .map(Row::unsafe_decode))
+    let bytes = db.get(&key);
+    get_row_from_bytes(bytes)
+}
+
+pub fn get_from_db(db: &DB, credential: &Key) -> Result<Option<Row>, StoreError> {
+    let key = as_key(&PREFIX, credential);
+    let bytes = db.get(&key);
+    get_row_from_bytes(bytes)
 }
```
KtorZ: Note: I don't think this buys us much. We could reduce duplication across all columns definitions by defining a more generic function which relies on `Decode`, since, it is true that we end up doing the same operation for all rows.

KtorZ: My rule of thumb when it comes to refactoring is: don't. Unless it's needed to introduce something new which is made easier by first refactoring the code. But we should avoid refactoring for the sake of simply rewriting code to something more aesthetically pleasing. The main reason being that this never ends for people won't agree about what "clean code" is.

crates/amaru-stores/src/rocksdb/ledger/columns/utxo.rs
```diff
@@ -18,18 +18,16 @@ use amaru_ledger::store::{
...
-pub fn get<T: ThreadMode>(
-    db: &OptimisticTransactionDB<T>,
-    key: &Key,
+fn get_value_from_bytes(
+    bytes: Result<Option<Vec<u8>>, rocksdb::Error>,
 ) -> Result<Option<Value>, StoreError> {
-    Ok(db
-        .get(as_key(&PREFIX, key))
+    Ok(bytes
         .map_err(|err| StoreError::Internal(err.into()))?
```
KtorZ: This function definition is weird. It should take in a `Result` since it does nothing with it. And more specifically, it doesn't `get` anything but merely _decode_ something 😅

crates/amaru-stores/src/rocksdb/mod.rs
```diff
@@ -93,6 +93,25 @@ pub struct RocksDB {
...
+    Ok(())
+}
+
+fn default_opts_with_prefix() -> Options {
+    let mut opts = Options::default();
+    opts.set_prefix_extractor(SliceTransform::create_fixed_prefix(PREFIX_LEN));
+    opts
+}
+
+pub struct ROLedgerDB {
```
KtorZ:
```suggestion
pub struct ReadOnlyLedgerDB {
```

KtorZ: Words are cheap!

crates/amaru-stores/src/rocksdb/mod.rs
```diff
@@ -125,82 +143,119 @@ impl RocksDB {
...
 }
 
+fn decode_cbor<T: for<'d> cbor::decode::Decode<'d, ()>>(
+    bytes: Option<impl AsRef<[u8]>>,
+) -> Result<Option<T>, StoreError> {
+    bytes
+        .map(|b| cbor::decode(b.as_ref()))
+        .transpose()
+        .map_err(StoreError::Undecodable)
+}
```
KtorZ: This function is also weird-looking (decoding CBOR, but from an option and returning a `StoreError`).

crates/amaru-stores/src/rocksdb/ledger/columns/accounts.rs
KtorZ: I see that this is introduced because of the need to reuse the code between the read-write and the read-only stores now. @jeluard already hinted at a possible trait, which could indeed removes the duplication but I also agree it's perhaps overkill here. 

So instead, I'd suggest factoring out the problematic part: i.e. the db. Which would lead to a simpler implementation in the ledger to replace `get`, `get_row_from_bytes` and `get_from_db` by a single:

```rs
pub fn get(
    with_db: impl Fn(&[u8]) -> Result<Option<Vec<u8>>, rocksdb::Error>,
    credential: &Key,
) -> Result<Option<Row>, StoreError> {
    let key = as_key(&PREFIX, credential);
    let bytes = with_db(&key).map_err(|err| StoreError::Internal(err.into()))?;
    Ok(bytes.map(Row::unsafe_decode))
}
```

while then only changing the call-sites to:

```rs
    accounts::get(|key| self.db.get(key), credential)
```
Which funny enough, reads exactly the same in both cases. Which means that it might be possible to avoid the double-macro altogether and simply define all 4 implementations through a single macro parameterized by the type of store `impl_StoreForDB`.

Also, with some minor generalization of the error, we might even be able to completely decouple that definition from rocksdb.

crates/amaru-stores/src/rocksdb/mod.rs
```diff
@@ -210,62 +238,62 @@ macro_rules! impl_ReadOnlyStore {
...
                 &self,
             ) -> Result<impl Iterator<Item = (scolumns::utxo::Key, scolumns::utxo::Value)>, StoreError>
             {
-                iter::<scolumns::utxo::Key, scolumns::utxo::Value>(&self.db, utxo::PREFIX, Direction::Forward)
+                iter(|mode, opts| self.db.iterator_opt(mode, opts), utxo::PREFIX, Direction::Forward)
             }
 
             fn pots(&self) -> Result<Pots, StoreError> {
-                pots::get(&self.db.transaction()).map(|row| Pots::from(&row))
+                pots::get(|key| self.db.get(key)).map(|row| Pots::from(&row))
```
geofflittle, line 265: I'm wondering if this is a valid change. Does the pots get need to occur within a transaction? Nothing else happens in this transaction, so I'm curious what's its purpose.

If it's needed, I'll need to find how to square this with the read only db, since it doesn't have a `transaction` method.

crates/amaru-stores/src/rocksdb/mod.rs
geofflittle: Removed!

crates/amaru-stores/src/rocksdb/mod.rs
geofflittle: Changed!

crates/amaru-stores/src/rocksdb/ledger/columns/utxo.rs
geofflittle: Removed!

crates/amaru-stores/src/rocksdb/ledger/columns/accounts.rs
geofflittle: Made this refactor in terms of a closure over `db.get` or `db.iterator_opt`.

Note: Since `transaction()` is not a method on the read only db, I removed the call from the pots get, since it stopped the unification of the read/write + read only impls of `ReadOnlyStore`. Lmk if pots requires the get under a transaction.

crates/amaru-stores/src/rocksdb/mod.rs
geofflittle: Removed!

crates/amaru-stores/src/rocksdb/mod.rs
geofflittle: I do like it this way but I don't mind reverting it--lmk

crates/amaru-stores/src/rocksdb/mod.rs
geofflittle: Changed!

crates/amaru-stores/src/rocksdb/ledger/columns/accounts.rs
KtorZ: > Lmk if pots requires the get under a transaction.

That... requires some careful thinking. The pots are only updated at epoch boundaries and there's only one writer; so it's **probably fine** if we only read them out-of-transaction.

crates/amaru-stores/src/rocksdb/mod.rs
KtorZ: That in particular looks completely okay.

crates/amaru-stores/src/rocksdb/ledger/columns/pools.rs
```diff
@@ -17,21 +17,22 @@ use amaru_ledger::store::{
...
 ) -> Result<Option<Row>, StoreError> {
-    Ok(db
-        .get(as_key(&PREFIX, pool))
-        .map_err(|err| StoreError::Internal(err.into()))?
-        .map(Row::unsafe_decode))
+    let key = as_key(&PREFIX, pool);
+    let bytes = (db_get)(&key);
+    bytes
+        .map_err(|err| StoreError::Internal(err.into()))
+        .map(|opt| opt.map(Row::unsafe_decode))
```
KtorZ:
```suggestion
        .map_err(|err| StoreError::Internal(err.into()))?
        .map(Row::unsafe_decode)
```

crates/amaru-stores/src/rocksdb/mod.rs
```diff
@@ -93,6 +95,25 @@ pub struct RocksDB {
     ongoing_transaction: OngoingTransaction,
 }
 
+fn validate_snapshots(dir: &Path) -> Result<(), StoreError> {
```
KtorZ:
```suggestion
fn assert_non_empty(dir: &Path) -> Result<(), StoreError> {
```

crates/amaru-stores/src/rocksdb/mod.rs
```diff
@@ -125,40 +145,41 @@ impl RocksDB {
...
-        let snapshots = RocksDB::snapshots(dir)?;
-
-        info!(target: EVENT_TARGET, snapshots = ?snapshots, "new.known_snapshots");
-
-        if snapshots.is_empty() {
-            return Err(StoreError::Open(OpenErrorKind::NoStableSnapshot));
-        }
-
-        let mut opts = Options::default();
+        validate_snapshots(dir)?;
```
KtorZ:
```suggestion
        assert_non_empty(dir)?;
```

crates/amaru-stores/src/rocksdb/mod.rs
```diff
@@ -125,40 +145,41 @@ impl RocksDB {
...
-        Ok(RocksDB {
-            dir: dir.to_path_buf(),
-            incremental_save: false,
-            db: OptimisticTransactionDB::open(&opts, dir.join("live"))
-                .map_err(|err| StoreError::Internal(err.into()))?,
-            era_history: era_history.clone(), // TODO: remove clone?
-            ongoing_transaction: OngoingTransaction::new(),
-        })
+    pub fn open_for_read_only(dir: &Path) -> Result<ReadOnlyLedgerDB, StoreError> {
+        validate_snapshots(dir)?;
```
KtorZ:
```suggestion
        assert_non_empty(dir)?;
```

crates/amaru-stores/src/rocksdb/ledger/columns/accounts.rs
```diff
@@ -102,15 +102,15 @@ pub fn reset_many<DB>(
...
+pub fn get(
+    db_get: impl Fn(&[u8]) -> Result<Option<Vec<u8>>, rocksdb::Error>,
     credential: &Key,
 ) -> Result<Option<Row>, StoreError> {
     let key = as_key(&PREFIX, credential);
-    Ok(db
-        .get(&key)
-        .map_err(|err| StoreError::Internal(err.into()))?
-        .map(Row::unsafe_decode))
+    let bytes = (db_get)(&key);
```
KtorZ:
```suggestion
    let bytes = db_get(&key);
```

crates/amaru-stores/src/rocksdb/ledger/columns/pools.rs
```diff
@@ -17,21 +17,22 @@ use amaru_ledger::store::{
...
+pub fn get(
+    db_get: impl Fn(&[u8]) -> Result<Option<Vec<u8>>, rocksdb::Error>,
     pool: &Key,
 ) -> Result<Option<Row>, StoreError> {
-    Ok(db
-        .get(as_key(&PREFIX, pool))
-        .map_err(|err| StoreError::Internal(err.into()))?
-        .map(Row::unsafe_decode))
+    let key = as_key(&PREFIX, pool);
+    let bytes = (db_get)(&key);
```
KtorZ:
```suggestion
    let bytes = db_get(&key);
```

crates/amaru-stores/src/rocksdb/ledger/columns/pots.rs
```diff
@@ -19,9 +19,11 @@ use rocksdb::Transaction;
 /// Name prefixed used for storing protocol pots. UTF-8 encoding for "pots"
 pub const PREFIX: [u8; PREFIX_LEN] = [0x70, 0x6f, 0x74, 0x73];
 
-pub fn get<DB>(db: &Transaction<'_, DB>) -> Result<Row, StoreError> {
-    Ok(db
-        .get(PREFIX)
+pub fn get(
+    db_get: impl Fn(&[u8]) -> Result<Option<Vec<u8>>, rocksdb::Error>,
+) -> Result<Row, StoreError> {
+    let bytes = (db_get)(&PREFIX);
```
KtorZ:
```suggestion
    let bytes = db_get(&PREFIX);
```

crates/amaru-stores/src/rocksdb/ledger/columns/utxo.rs
```diff
@@ -18,18 +18,19 @@ use amaru_ledger::store::{
...
-pub fn get<T: ThreadMode>(
-    db: &OptimisticTransactionDB<T>,
+pub fn get(
+    db_get: impl Fn(&[u8]) -> Result<Option<Vec<u8>>, rocksdb::Error>,
     key: &Key,
 ) -> Result<Option<Value>, StoreError> {
-    Ok(db
-        .get(as_key(&PREFIX, key))
+    let key = as_key(&PREFIX, key);
+    let bytes = (db_get)(&key);
```
KtorZ:
```suggestion
    let bytes = db_get(&key);
```

PR #317, opened 2025-07-01
WIP: Abstract epoch state

abailly:
This PR is a stab at abstracting how we populate ledger data, the design of which I would appreciate feedback on before sinking time. The current way is very imperative and tied to the specifics of the CBOR encoding, but ultimately it just needs various parts of the state. To provide a safe refactoring path, I introduce a trait implementing `Iterator<StateFragment>` upon which the [decode_new_epoch_state](https://github.com/pragma-org/amaru/blob/9a9e0f345a87697ccf7477fe230ca70680e27f2f/crates/amaru/src/bin/amaru/cmd/import_ledger_state.rs#L294) depend: it iterates over all `StateFragment`, injecting data in the db as needed. This `Iterator` trait can then easily be implemented from genesis files, or any other kind of data.


crates/amaru/src/bin/amaru/cmd/import_ledger_state.rs
```diff
@@ -148,12 +149,110 @@ async fn import_all(
...
+    EpochNo(Epoch),
+    CurrentBlockIssuers(BTreeMap<PoolId, u64>),
+    Treasury { treasury: i64, reserve: i64 },
+}
+
+/// Simple wrapper over an iterator to enumerate `StateFragment` for
+/// a given epoch
+trait EpochState<'a, E>: Iterator<Item = Result<StateFragment, E>> {
+    /// FIXME: temporary kludge while refactoring code
+    fn decoder(&self) -> minicbor::Decoder<'a>;
```
jeluard: Use `cbor` ?

crates/amaru/src/bin/amaru/cmd/import_ledger_state.rs
```diff
@@ -190,37 +291,36 @@ async fn import_one(
...
-    import_block_issuers(db, d.decode()?)?;
-
-    // Epoch State
-    d.array()?;
+) -> Result<Epoch, Box<dyn Error>> {
+    let mut epoch = Epoch::from(0);
+    let mut treasury = 0;
+    let mut reserves = 0;
+
+    for state_fragment in &mut *epoch_state {
```
jeluard, line 544: What does the `Iterator` abstraction give us compared to a simple function returning `(Epoch, BTreeMap<PoolId, u64>, ..)`?
This looks counter intuitive to me plus introduces mut states.

abailly: That was my first attempt indeed, but:

1.  I wanted a way to refactor the function incrementally and the fact we need to go through the various steps incrementally as we are parsing the CBOR at the same time we are populating the DB
2. it seems to me to actually make a lot of sense to do this parsing lazily or incrementally, if you think that, down the road, the snapshot we'll need to parse could weigh several gigabytes (the size of the ledger state DB on maintain right now is ~2.6 GB).

Hence the `Iterator` idea: It allows _pulling_ the data on-demand instead of having to parse the whole available data. In the current implementation the code still loads the whole bytes in an array but I am pretty sure this can be turned into a lazily read stream of bytes.

crates/amaru/src/bin/amaru/cmd/import_ledger_state.rs
abailly: I did not change anything here, just used what was there already.

crates/amaru/src/bin/amaru/cmd/import_ledger_state.rs
abailly: The `mut` variables are mostly temporary I thikn

jeluard: Sure we can always iterate

KtorZ:
Haven't looked in depth into the PR, but it seems to me that it'll heavily clash with: https://github.com/pragma-org/amaru/pull/265 

So it's probably better to base it on top of #265 if possible.

PR #318, opened 2025-07-01
feat: allow chaindb to be concurrently opened for reading

geofflittle:
This change introduces `open_for_read_only` to `RocksDBStore`, enabling concurrent read access without requiring a write lock.

This is particularly useful for tools like amaru-doctor, which need to inspect the chain database while amaru is actively running and modifying it.

## Summary by CodeRabbit

* **New Features**
  * Added support for opening consensus stores in read-only mode, including a dedicated read-only store type and method for opening databases without write access.

* **Refactor**
  * Separated store interfaces into distinct read-only and read-write traits, clarifying which operations are available for each mode.
  * Updated mock and test store implementations to align with the new trait hierarchy, improving modularity and clarity in test infrastructure.
  * Simplified test mocks by replacing custom mocks with standardized fake stores.

* **Tests**
  * Enhanced tests to verify concurrent access for both read-only and read-write stores and updated test imports to use the new read-only trait.

### DIFF
--- a/crates/amaru-consensus/src/consensus/store_block.rs
+++ b/crates/amaru-consensus/src/consensus/store_block.rs
@@ -54,7 +54,7 @@ impl StoreBlock {
 
 #[cfg(test)]
 mod tests {
-    use crate::consensus::store::StoreError;
+    use crate::consensus::store::{ReadOnlyChainStore, StoreError};
 
     use super::*;
     use amaru_kernel::{Hash, Point, RawBlock};
@@ -75,25 +75,25 @@ mod tests {
         }
     }
 
+    impl ReadOnlyChainStore<Header> for MockChainStore {
+        fn load_header(&self, _hash: &Hash<32>) -> Option<Header> {
+            unimplemented!()
+        }
+        fn load_block(&self, _hash: &Hash<32>) -> Result<RawBlock, StoreError> {
+            unimplemented!()
+        }
+        fn get_nonces(&self, _header: &Hash<32>) -> Option<amaru_ouroboros::Nonces> {
+            unimplemented!()
+        }
+    }
+
     impl ChainStore<Header> for MockChainStore {
         fn store_block(&mut self, point: &Hash<32>, block: &RawBlock) -> Result<(), StoreError> {
             self.stored_blocks.insert(*point, block.clone());
             Ok(())
         }
 
-        fn load_header(&self, _hash: &Hash<32>) -> Option<Header> {
-            unimplemented!()
-        }
-
         fn store_header(&mut self, _hash: &Hash<32>, _header: &Header) -> Result<(), StoreError> {
-            unimplemented!()
-        }
-
-        fn load_block(&self, _hash: &Hash<32>) -> Result<RawBlock, StoreError> {
-            unimplemented!()
-        }
-
-        fn get_nonces(&self, _header: &Hash<32>) -> Option<amaru_ouroboros::Nonces> {
             unimplemented!()
         }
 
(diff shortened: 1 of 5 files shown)


crates/amaru-consensus/src/consensus/store_block.rs
```diff
@@ -75,25 +75,25 @@ mod tests {
         }
     }
 
-    impl ChainStore<Header> for MockChainStore {
-        fn store_block(&mut self, point: &Hash<32>, block: &RawBlock) -> Result<(), StoreError> {
-            self.stored_blocks.insert(*point, block.clone());
-            Ok(())
-        }
-
+    impl ReadOnlyChainStore<Header> for MockChainStore {
```
abailly: this should be replaced by `FakeStore` perhaps?

crates/amaru-stores/src/rocksdb/consensus.rs
```diff
@@ -43,20 +47,58 @@ impl RocksDBStore {
             era_history: era_history.clone(),
         })
     }
+
+    pub fn open_for_readonly(basedir: &PathBuf) -> Result<ReadOnlyChainDB, StoreError> {
+        let mut opts = Options::default();
+        opts.create_if_missing(true);
```
abailly: Does it make sense to set `create_if_missing` to true in read-only mode?

crates/amaru-stores/src/rocksdb/consensus.rs
```diff
@@ -43,20 +47,58 @@ impl RocksDBStore {
...
+                    })?
+                    .ok_or(StoreError::NotFound { hash: *hash })
+                    .map(|bytes| bytes.as_ref().into())
+            }
+
+        })*
     }
+}
 
+impl_ReadOnlyChainStore!(for ReadOnlyChainDB, RocksDBStore);
```
abailly, line 99: What's the point of a macro with a single use?

crates/amaru-stores/src/rocksdb/consensus.rs
```diff
@@ -233,4 +274,11 @@ mod test {
             result
         );
     }
+
+    #[test]
+    fn both_rw_and_ro_can_be_open_on_same_dir() {
+        let tempdir = tempfile::tempdir().unwrap();
+        let _rw_store = initialise_test_rw_store(&tempdir);
+        let _ro_store = initialise_test_ro_store(&tempdir);
```
abailly: I would make the test more explicit, something like:
```
if let Err(e) = initialise_test_ro_store(&tempdir) {
   panic!("failed to re-open DB in read-only mode: {}", e);
}
```

crates/amaru-stores/src/rocksdb/consensus.rs
KtorZ: In this particular case, it reduces the boilerplate of implementing the same trait (with many methods) twice (for the ReadOnlyChainDB, and the RocksDBStore)

abailly: But the macro is invoked only once, unless I missed something?

jeluard: It implements all methods for both `ReadOnlyChainDB` and `RocksDBStore`

crates/amaru-stores/src/rocksdb/consensus.rs
geofflittle: Added!

crates/amaru-stores/src/rocksdb/consensus.rs
geofflittle: I changed it to false

crates/amaru-consensus/src/consensus/store_block.rs
geofflittle: Replaced!

PR #320, opened 2025-07-02
Can test the CLI of Amaru

pgrange:
During a conversation with @KtorZ we were contemplating the idea of introducing tests of Amaru's CLI.

This PR implements several CLI tests. As these tests mainly consist of running a shell command which starts Amaru in various conditions and checking the outcome, we use [bash_unit](https://github.com/bash-unit/bash_unit) that is tailored to run tests written in bash (and which I maintain so I may be a bit biased here).

The `cli-tests` directory is introduced to contain this kind of tests:
*  `test_explains_snapshot_file_is_missing()` in `cli-tests/test_crash_with_helpful_information.sh` checks the output of amaru when bootstraped without snapshots is helpfull;
* `test_no_color_if_no_terminal` and `test_color_if_in_terminal` in `cli-tests/test_manage_colors_in_output.sh` check that Amaru is correctly managing colors depending on wether it's started in a terminal or not.

## How to run the test

Ensure you have bash_unit installed. See [how to install bash_unit](https://github.com/bash-unit/bash_unit/blob/main/README.adoc#how-to-install-bash_unit).

Run the following command:

```
#> bash_unit cli-tests/test_*
```

## How are the tests structured

We often need some sort of setup specific to the kind of checks we want to make on the CLI. The `fixtures` subdirectory contains such setups and the tests `cd` into them before starting Amaru.

## Integration with the CI

I've added a job to run this tests. I did my best to try to use rust compilation cache from the build stage but I'm not 100% sure I'm not missing something there.

This is what the test looks like when it fails: https://github.com/pgrange/amaru/actions/runs/16028793730/job/45223640352

This is what the test looks like when it succeeds:
https://github.com/pgrange/amaru/actions/runs/16028360576/job/45222108688

## Summary by CodeRabbit

- Tests
  - Added CLI integration tests for missing snapshots error messaging and terminal-aware color output, with supporting fixtures.
- Chores
  - Extended continuous integration to run the CLI test suite on Linux with required tooling.
  - Ignored test snapshot directories to keep the repo clean.
- Documentation
  - Added instructions for running the CLI tests locally, including prerequisites and test structure.

### DIFF
--- a/cli-tests/fixtures/missing_snapshots_file_setup/data/preprod/rewards-provenance
+++ b/cli-tests/fixtures/missing_snapshots_file_setup/data/preprod/rewards-provenance
@@ -1,0 +1,1 @@
+../../../../../data/preprod/rewards-provenance
--- a/cli-tests/test_crash_with_helpful_information.sh
+++ b/cli-tests/test_crash_with_helpful_information.sh
@@ -1,0 +1,28 @@
+test_explains_snapshot_file_is_missing() {
+	given_snapshots_file_is_missing
+
+	assert_matches "data/preprod/snapshots.json.*NotFound" "$(bootstrap_amaru)"
+}
+
+setup() {
+	export NETWORK=preprod
+	export CONFIG_FOLDER=data
+	export LEDGER_DIR=./ledger.${NETWORK}.db
+	export CHAIN_DIR=./chain.${NETWORK}.db
+	export BUILD_PROFILE=dev
+	export AMARU_PEER_ADDRESS=127.0.0.1:3001
+}
+
+given_snapshots_file_is_missing() {
+	cd fixtures/missing_snapshots_file_setup
+}
+
+bootstrap_amaru() {
+	cargo run --profile ${BUILD_PROFILE} -- bootstrap \
+		--peer-address ${AMARU_PEER_ADDRESS} \
+		--config-dir ${CONFIG_FOLDER} \
+		--ledger-dir ${LEDGER_DIR} \
+		--chain-dir ${CHAIN_DIR} \
+		--network ${NETWORK} \
+		2>&1
+}
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -319,3 +319,23 @@ jobs:
             cargo-x86_64-unknown-linux-gnu
       - name: Run amaru-consensus benches
         run: cargo bench -p amaru-consensus --features="test-utils profiling"
+
+  cli:
+    name: Test the CLI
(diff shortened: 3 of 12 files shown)


.github/workflows/CI.yml
```diff
@@ -320,3 +320,19 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/cache/restore@v4
+        with:
+          path: ${{ env.RUST_CACHE_PATH }}
+          key: cargo-x86_64-unknown-linux-gnu
+          restore-keys: |
+            cargo-x86_64-unknown-linux-gnu
+      - name: Install bash_unit
+        run: curl -s https://raw.githubusercontent.com/pgrange/bash_unit/master/install.sh | bash
```
jeluard: Not `https://raw.githubusercontent.com/bash-unit/bash_unit/master/install.sh` ?

pgrange: Good catch! Looks like someone forgot to tidy things up in bash-unit itself and I just copy/pasted that here.

Fixed.

cli-tests/fixtures/missing_snapshots_file_setup/data/preprod/rewards-provenance
```diff
@@ -0,0 +1 @@
+../../../../../data/preprod/rewards-provenance
```
jeluard, line 1: What are those?

pgrange: Those are links to the data directory. cli-tests/fixtures/missing_snapshots_file_setup is a directory simulating an Amaru setup form which the snapshot files are missing. This allows to test for the output of the cli in that condition.

cli-tests/test_crash_with_helpful_information.sh
```diff
@@ -0,0 +1,28 @@
...
+	export CONFIG_FOLDER=data
+	export LEDGER_DIR=./ledger.${NETWORK}.db
+	export CHAIN_DIR=./chain.${NETWORK}.db
+	export BUILD_PROFILE=dev
+	export AMARU_PEER_ADDRESS=127.0.0.1:3001
+}
+
+given_snapshots_file_is_missing() {
+	cd fixtures/missing_snapshots_file_setup
+}
```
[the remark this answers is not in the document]
pgrange: That's ok, the job of the `given_snapshots_file_is_missing` function is, indeed, to simulate a bad configuration by changing the current working directory to `fixtures/missing_snapshots_file_setup`.

Let's keep it that way.

PR #326, opened 2025-07-04
feat: era history boundary

jeluard:
Deal with current EraHistory which doesn't define an end.


crates/amaru-ledger/src/state.rs
```diff
@@ -102,7 +102,7 @@ where
     /// The era history for the network this store is related to.
     era_history: Arc<EraHistory>,
 
-    global_parameters: Arc<GlobalParameters>,
+    pub global_parameters: Arc<GlobalParameters>,
```
KtorZ: What about providing an accessor instead? So that we keep the boundary clean. 

```rust
    pub fn global_parameters(&self) -> &GlobalParameters {
        &self.global_parameters
    }
```

crates/slot-arithmetic/src/lib.rs
```diff
@@ -362,12 +362,33 @@ pub struct EpochBounds {
 // horizon is the end of the epoch containing the end of the current era's safe zone relative to
 // the current tip. Returns number of milliseconds elapsed since the system start time.
 impl EraHistory {
-    pub fn slot_to_relative_time(&self, slot: Slot) -> Result<u64, TimeHorizonError> {
+    pub fn new(eras: &[Summary]) -> EraHistory {
```
KtorZ: We could also take the stability window here, when constructing the era history. It's a global constant that is network-specific; so it shouldn't change. That would avoid having to pass it with every checked calls. We can just pull it from `self`.

crates/slot-arithmetic/src/lib.rs
```diff
@@ -362,12 +362,33 @@ pub struct EpochBounds {
...
+        #[allow(clippy::panic)]
+        if eras.is_empty() {
+            panic!("EraHistory cannot be empty");
+        }
+        // TODO ensures only last era ends with Option
```
KtorZ, line 454: Alternatively, we could define:

```rust
pub struct Summary<E> {
    pub start: Bound,
    pub end: Bound,
    pub end: E,
    pub params: EraParams,
}
```

and 

```
pub struct EraHistory {
    past_eras: Vec<Summary<Bound>>,
    current_era: Vec<Summary<()>>,
}
```

Which would _by construction_, ensures that the last era has no bound. It might make the implementation of the function below a bit more annoying though because less uniform; yet perhaps more "safe".

crates/slot-arithmetic/src/lib.rs
```diff
@@ -362,12 +362,33 @@ pub struct EpochBounds {
...
                 return Err(TimeHorizonError::InvalidEraHistory);
             }
-            if era.end.slot >= slot {
+            #[allow(clippy::expect_used)]
+            if era
+                .clone()
+                .end
+                .expect("last EraHistory; shouldn't be tested")
+                .slot
+                >= slot
```
KtorZ:
```suggestion
            if era.end.as_ref().unwrap_or_else(|| panic!("last EraHistory; shouldn't be tested")) >= &slot {
```

To avoid clone?

KtorZ: Actually, I don't think the behavior is _correct_ here? We shouldn't fail when calling the `_unchecked_horizon` version, but simply proceed as if the end bound was +infinity.

KtorZ: So it should rather be:

```rust
if era.end.as_ref().map(|end| end >= &slot).unwrap_or(true) {

}
```

crates/amaru-stores/src/rocksdb/mod.rs
```diff
@@ -436,9 +429,8 @@ impl TransactionalContext<'_> for RocksDBTransactionalContext<'_> {
 
                 accounts::reset_many(&self.transaction, withdrawals)?;
                 dreps::tick(&self.transaction, voting_dreps, {
-                    let slot = point.slot_or_default();
                     era_history
-                        .slot_to_epoch(slot)
+                        .slot_to_epoch(tip, tip)
```
KtorZ, line 433: Could simply be _unchecked_ here.

crates/amaru/src/stages/ledger.rs
```diff
@@ -136,8 +136,11 @@ impl<S: Store + Send, HS: HistoricalStores + Send> ValidateBlockStage<S, HS> {
             BlockValidation::Valid(()) => {
                 let state: VolatileState = context.into();
                 let issuer = Hasher::<224>::hash(&block.header.header_body.issuer_vkey[..]);
-                self.state
-                    .forward(protocol_version, state.anchor(&point, issuer))?;
+                self.state.forward(
+                    protocol_version,
+                    state.anchor(&point, issuer),
+                    point.slot_or_default(),
```
KtorZ: Feels like duplication here; the tip is already part of the anchor, we shouldn't need to pass it as argument

crates/amaru/src/stages/pull.rs
```diff
@@ -45,25 +47,42 @@ pub enum WorkUnit {
...
+        }
+    }
+
+    #[allow(clippy::unwrap_used)]
+    fn catching_up(is_catching_up: &RwLock<bool>) {
+        // Do not acquire the lock unless necessary.
+        if is_catching_up.read().map(|lock| *lock).unwrap_or(false) {
+            *is_catching_up.write().unwrap() = true;
+        }
     }
```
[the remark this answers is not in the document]
KtorZ: Nice catch.

Issue #290, opened 2025-06-19
Provide detailed design rationale and documentation about "stage graph"


abailly:
A lot of work has been done to provide needed infrastructure for model-based testing of Amaru core logic:

* The `pure-stage` crate provides an actor-like and `Effect` system to structure the various "effectual" parts of Amaru, and most notably the networking and consensus layers in such a way that they it run in production leveraging `tokio`'s parallelism, while still being able to run deterministically in simulations
* It includes a `TraceBuffer` mechanism to collect, store, and potentially replay any particular execution, with the goal of easing troubleshooting on tests failures but also capture/playback of production runs
* `simulation-testing` has been ported to Rust so that model-based tests of (part of) the consensus logic can be executed efficiently, with the simulation code leveraging the deterministic `pure-stage` execution engine

It's important we document and motivate this design, and share it across the whole team to ensure everyone's comfortable with the approach taken:

* [x] provide detailed explanation of simulation tests in `Amaru-sim`, how they are executed, and how to investigate failures to showcase the benefits of the approach
* [x] write an _Engineering Decision Record_ to validate use of stage-graph architecture w/in Amaru
* [x] walk the whole team of core contributors to amaru over the code and design

PR #328, opened 2025-07-07
docs: add EDR for deterministic simulation testing

stevana:
Part of #290.

## Summary by CodeRabbit

* **Documentation**
  * Added a new architecture decision record outlining the design and rationale for a deterministic simulation testing framework for distributed systems. This document covers motivation, design considerations, fault models, test generation, replay capabilities, and integration with production environments.

### DIFF
--- a/engineering-decision-records/011-deterministic-simulation-testing.md
+++ b/engineering-decision-records/011-deterministic-simulation-testing.md
@@ -169,7 +169,7 @@ sketch of how to implement them:
   running on. I/O pauses can be implemented at the pure-stage level by delaying
   when a future becomes available;
 
-* Time skews (the clock of nodes drifts a part). The simulator controls the
+* Time skews (the clock of nodes drifts apart). The simulator controls the
   clock of the nodes, via pure-stage's API, so skews can be implemented by
   advancing some node's clocks more than others;
 


engineering-decision-records/011-deterministic-simulation-testing.md
```diff
@@ -0,0 +1,229 @@
...
+
+## Decision
+
+In order to tackle the problem of testing and debugging we've introduced two
+changes:
+
+  1. Introduced a library, called pure-stage, for building parallel processing
+     networks that can be run in a completely deterministic fashion. The idea
+     being that the consensus pipeline will be ported from using gasket (which
+     wasn't designed with determinism in mind) to pure-stage;
```
KtorZ: Not only the consensus but all Amaru stages? Or do you foresee this being only useful for the consensus-related stages?

engineering-decision-records/011-deterministic-simulation-testing.md
```diff
@@ -0,0 +1,229 @@
...
+
+#### Byzantine faults
+
+The space of possibilities here is infinite. Any combination of changing any
+message or any piece of state on a faulty node is allowed. Although the
+protocol, with its use of cryptographically secure hashes, should render all of
+these attacks useless, nevertheless this needs to be tested to some extent.
+
+As a first approximation one could have the simulator mutate messages between
+or the stable storage of a node to ensure that hashes are checked.
```
KtorZ: 🤔 ?

engineering-decision-records/011-deterministic-simulation-testing.md
```diff
@@ -0,0 +1,229 @@
...
+
+Coordinated attacks that explore the extremes of protocol parameters and timings
+of messages is out of scope of this decision. Let us just note there's the
+possibility to add Byzantine (malicious) nodes to the simulation test and
+otherwise run the tests as before, and ensure that the non-faulty nodes still
+work as intended.
```
KtorZ, line 201: An interesting thing to do along those lines could also be to validate the system behaves 'as expected' under a successful attack. Ouroboros being fundamentally stochastic, we should be able to observe probability of success of attacks matching the theoretical ones from the paper.

engineering-decision-records/011-deterministic-simulation-testing.md
```diff
@@ -0,0 +1,229 @@
...
+     networks that can be run in a completely deterministic fashion. The idea
+     being that the consensus pipeline will be ported from using gasket (which
+     wasn't designed with determinism in mind) to pure-stage;
+
+  2. A discrete-event simulator which spawns a network of nodes and simulates
+     the network connections between them. The simulator decides when, or if,
+     messages get delivered to the nodes and can step through the execution
+     deterministically using the pure-stage API.
+
+## Discussion points
```
KtorZ, line 45: From reading those discussion points, I am lacking one important element: how is _the truth_ determined? There's a mention near the end of using the Haskell lib to generate some of the initial test tree; but generally speaking, how are we approaching doing conformance with a system that is itself a moving target?

stevana: I thought I covered this under "Properties to test for", or do you mean something else?

engineering-decision-records/011-deterministic-simulation-testing.md
stevana: I'm refering to these stages: https://github.com/pragma-org/amaru/tree/main/crates/amaru-consensus , pardon my ignorance, but are there other stages that are not in that diagram?

engineering-decision-records/011-deterministic-simulation-testing.md
stevana: Should say "mutate messages between *being sent between nodes* or *mutate data* in the stable storage".

Let me know if it still doesn't make sense.

engineering-decision-records/011-deterministic-simulation-testing.md
rkuhn: I’d reformulate that the whole node-internal processing network will be expressed using pure-stage. Whether that covers the so-called consensus (which is a bit of a misnomer if you ask me, since eventual consensus is only one of the desirable outcomes of running these modules) or something else is not so important.

engineering-decision-records/011-deterministic-simulation-testing.md
```diff
@@ -0,0 +1,279 @@
...
+it is the fact that the processing of incoming messages happens in a parallel
+pipeline. So the trick of implementing the system as a state machine of type:
+
+  (input, state) -> (output, state)
+
+doesn't quite work. One rather needs to implement it as several state machines
+(that can run in parallel) connected with queues. Further complications arise
+due to the fact that the different stages in the pipeline share state (or one
+stage needs to be able to send messages to another to update the other copy of
+the state).
```
rkuhn:
```suggestion
due to the fact that the different stages in the pipeline share external state (e.g.
the contents of a persistent storage for blocks, headers, tips). Here, the design
goal for the concrete stages should be that updates to some entity be made by
only one stage and messages to other stages carry the knowledge about what
has been written and can thus be read. This is aided by the append-only nature
of the content-addressed storage items used within a blockchain node.
```

engineering-decision-records/011-deterministic-simulation-testing.md
```diff
@@ -0,0 +1,279 @@
...
+
+ 2. Pop world heap to figure out which message to deliever next;
+
+ 3. If there are no interesting times from step 1 or they all happen after the
+    next message is supposed to be delivered, then advance the time to the
+    arrival time of the message on all nodes and deliver the popped message.
+
+    Otherwise put all the interesting times into the world heap and recurse.
+    (We need to make it so that if we ask for the next interesting points in
+    time in step 1, the nodes returns an empty list?)
```
rkuhn: Hmm, maybe the coffee isn’t yet working: this description seems more complicated than it needs to be. How about

1. the simulation heap starts out with external input messages, each paired with an arrival time and a target node
2. each node is polled for the next time at which something would occur internally; the node is also inserted into the simulation heap with that time
3. the simulator pops the task with the smallest time from the heap (message delivery or internal node action) and performs it; this may enqueue messages for other nodes that cause them to be woken up sooner than dictated by their current position in the heap — if so, remove and re-insert
4. that node is then queried for the next interesting time and inserted back into the heap with that time
5. the simulator starts over from 3 (unless some stop condition is satisfied)

engineering-decision-records/011-deterministic-simulation-testing.md
```diff
@@ -0,0 +1,279 @@
...
+
+### Maximising the overlap of what's being tested vs deployed
+
+We want the simulator to be as true to the "real world" as possible. Ideally
+any failure that could happen in a real world deployment should be reproducible
+within the simulator (and any failure to do so should be considered a bug in
+the simulator), but of course one has to draw a line somewhere.
+
+To achieve this the pure-stage library has a runtime interface that can be
+toggled between being deterministic or to use tokio.
```
rkuhn: I think this doesn’t belong here: we’re discussing fault injection in the simulator, which is independent of the Tokio implementation.

engineering-decision-records/011-deterministic-simulation-testing.md
rkuhn: Makes sense to me, we should test that corruption is indeed recognised and rejected in all parts of the system.

engineering-decision-records/011-deterministic-simulation-testing.md
rkuhn: One interesting angle could be to challenge the community to contribute node implementations that deviate in interesting ways, with some bounty for catching interesting edge cases. We aren’t there yet, though.

engineering-decision-records/011-deterministic-simulation-testing.md
```diff
@@ -0,0 +1,279 @@
...
+While we've not put much thought into this yet, the simulator has been designed
+in such a way that it can communicate with the nodes via stdin/stdout a la
+Jepsen's Maelstrom, which means we could technically run different versions of
+the node binaries (this would be slower than the default in-process tests).
+
+### Test case generation
+
+Due to the consensus protocol's requirement that all messages need to be hashed
+and refer to the previous message's hash (i.e. hash-chained), it's not possible
+to generate "random" traffic as one normally would with property-based testing.
```
rkuhn, line 222: It sure isn’t straight-forward and the libraries aren’t made for this, meaning it will be inconvenient, but the “not possible” part of the statement is incorrect. It is a matter of generating test inputs that then are used to compute a resulting block tree with the correct hashes, e.g. using an integer to point to a previous block (using modulo arithmetic to adequately “shorten” it).

engineering-decision-records/011-deterministic-simulation-testing.md
```diff
@@ -0,0 +1,279 @@
...
+   possible to use the tests during development as a sanity check;
+
+3. Antithesis doesn't support for example disk faults currently;
+
+4. Good system architecture and being easy to test locally (i.e. with `cargo
+   test`) goes hand in hand. The efforts described above leads to for example
+   being able to write a unit test for a regression test involving a whole network
+   of nodes and faults, which is something that would not been possible if the
+   system had not been made deterministic (and solely relied on Antithesis for
+   determinism).
```
rkuhn, line 287:
```suggestion
   determinism).

## Relation to Production Mode

The above description focuses on a special mode of execution for the pure-stage infrastructure that
allows us to deterministically run a whole network while precisely following what each node is doing.
This mode is single-threaded by definition due to the ordering guarantees it needs to make.
When running an Amaru node in production, the very same pure-stage components are used with a
non-deterministic execution engine that allows for parallelism, namely the Tokio runtime. The idea
here is that simulation testing should eventually exhaust all possible interleavings of messages and
other effects, thus covering the range of possible behaviours of the Tokio implementation.

We still want to be able to diagnose, reproduce, and correct any runtime failures observed during
production mode. This is why all relevant effects (which include message reception but also the
results of external effects like reading the clock or accessing disk storage) are recorded in a trace
that can then be taken to an offline system and be replayed. This requires that the offline system has
been primed with a suitable starting state (disk storage as well as stage state) because the trace
will usually be truncated (e.g. by recording to an in-memory ring buffer that overwrites old entries).
Care has been taken to ensure that all relevant state can be snapshotted and stored periodically,
making it possible to recreate the exact runtime state of an Amaru node from the youngest snapshot
and the trace written since.

Due to the true parallelism available in production mode it is impossible to recreate the exact
temporal evolution of a node’s state when replaying a trace. When focusing on a single stage
the result is exact, though, because a stage only processes inputs sequentially due to its state
machine signature. We can thus envision a time travelling debugger for a single stage, with a
bit of fuzz regarding the execution order between different stages.
```

rkuhn (commented):
Thanks @stevana, great write-up! What I think is still missing is the discussion of what exactly we need to record from a live (i.e. non-deterministic) system to enable an exact replay from a trace. But that may not be our current focus, so I added a paragraph in the end hinting at the reasoning.

engineering-decision-records/011-deterministic-simulation-testing.md
stevana: What I was trying to say is that the code using in testing and production is largely the same, except for the "runtime part" (maybe you call it something else?) which can be deterministic (during testing) or non-deterministic and fast using tokio (in production), or did I understand that wrong?

engineering-decision-records/011-deterministic-simulation-testing.md
stevana: Fixed, thanks!

engineering-decision-records/011-deterministic-simulation-testing.md
stevana: Fixed, thanks!

engineering-decision-records/011-deterministic-simulation-testing.md
stevana: Fixed.

engineering-decision-records/011-deterministic-simulation-testing.md
stevana: Fixed.

engineering-decision-records/011-deterministic-simulation-testing.md
rkuhn: Yes, we agree on this; my suggestion was to move this aspect from this isolated single sentence into its own section for a deeper discussion (which I added at the end of the text).

stevana: Ok, I added a forward reference to that secion.

PR #332, opened 2025-07-10
Rework metrics to show more relevant (process) metrics

KtorZ:
Synchronizing through PreProd:

Screenshot 2025-07-10 at 14 38 41

Screenshot 2025-07-10 at 14 39 08


crates/amaru/src/bin/amaru/metrics.rs
```diff
@@ -12,86 +12,180 @@
...
+            let disk_total_write_bytes = meter
+                .u64_gauge("process_disk_total_write")
+                .with_description("Total number of written bytes (in bytes).")
+                .with_unit("bytes")
+                .build();
+
+            let disk_live_read_bytes = meter
+                .u64_gauge("process_disk_live_read")
+                .with_description("Number of written bytes since the last refresh (in bytes)")
+                .with_unit("bytes")
```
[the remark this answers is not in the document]
KtorZ: Thanks mate. Here's a treat 🥕

crates/amaru/src/bin/amaru/metrics.rs
```diff
@@ -12,86 +12,180 @@
...
+    pub struct ProcessMetrics {
+        /// Number of available CPUs on the machine.
+        number_of_cpus: u64,
 
-    pub struct SystemCounters {
-        total_memory: Gauge<u64>,
-        free_memory: Gauge<u64>,
-        cpu_usage: Gauge<f64>,
+        /// How much time the process has been running (in seconds).
+        runtime_seconds: Gauge<u64>,
```
Quantumplation, line 75: should `runtime_seconds`, `dis_total_read_bytes`, and `disk_total_write_bytes` be Counters instead of Gauges?

KtorZ: I initially thought the same and had them as counters; but `sysinfo` actually returns us already aggregated numbers; so we would just be adding quantities to themselves over and over (I did that, and realized that Amaru ingesting multiple terrabytes of data in only a couple of minutes wasn't probably right...).

So no, gauges. in theory, we could use `Counter` if we added the live disk usages on each tick, but since sysinfo already provides the total data, there's no need.

PR #333, opened 2025-07-11
Preview conformance fixes

KtorZ:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved serialization of IPv6 addresses to output a standardized string format instead of raw bytes, enhancing readability and compatibility.

* **Refactor**
  * Updated protocol version logic in tests to explicitly handle multiple network types and epochs, providing clearer and more robust network-specific behavior.

### DIFF
--- a/crates/amaru-kernel/src/lib.rs
+++ b/crates/amaru-kernel/src/lib.rs
@@ -443,7 +443,16 @@ impl serde::Serialize for PoolParams {
                             )?;
                         }
                         if let Nullable::Some(ipv6) = ipv6 {
-                            s.serialize_field("ipv6", ipv6)?;
+                            let bytes: [u8; 16] = [
+                                ipv6[3], ipv6[2], ipv6[1], ipv6[0], // 1st fragment
+                                ipv6[7], ipv6[6], ipv6[5], ipv6[4], // 2nd fragment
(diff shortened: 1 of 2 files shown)


crates/amaru-kernel/src/lib.rs
```diff
@@ -443,7 +443,16 @@ impl serde::Serialize for PoolParams {
...
+                            let bytes: [u8; 16] = [
+                                ipv6[3], ipv6[2], ipv6[1], ipv6[0], // 1st fragment
+                                ipv6[7], ipv6[6], ipv6[5], ipv6[4], // 2nd fragment
+                                ipv6[11], ipv6[10], ipv6[9], ipv6[8], // 3rd fragment
+                                ipv6[15], ipv6[14], ipv6[13], ipv6[12], // 4th fragment
+                            ];
+                            s.serialize_field(
+                                "ipv6",
+                                &format!("{}", std::net::Ipv6Addr::from(bytes)),
+                            )?;
```
KtorZ, line 455: Not sure if it's the best choice long-term for sharing those snapshots but... IPv6 are currently encoded according to RFC 5952: http://tools.ietf.org/html/rfc5952 😅 ... we should probably just stick with hex-encoded bytes for the tests even if they're really cute.

jeluard (approved):

PR #341, opened 2025-07-17
feat/concurrent requests

stevana:
This PR adds support for simulating multiple upstream peers. Initially it wasn't working and I couldn't immediately see why, so I improved the failure reporting to make debugging easier.

## Summary by CodeRabbit

* **New Features**
  * Added configurable parameters for number of tests, nodes, and upstream peers in the simulation, allowing more flexible and dynamic simulation setups.
  * Introduced new helper functions for generating randomized test data and message entries.

* **Improvements**
  * Enhanced error handling and reporting in simulation runs, providing clearer feedback on failures.
  * Replaced property-based testing integration with a deterministic simulation loop for improved control and reproducibility.
  * Updated test structure to use new configuration options and improved message generation.

* **Bug Fixes**
  * Improved error propagation to prevent abrupt panics and provide detailed failure information.

### DIFF
--- a/simulation/amaru-sim/src/simulator/simulate.rs
+++ b/simulation/amaru-sim/src/simulator/simulate.rs
@@ -263,7 +263,7 @@ pub fn simulate<Msg, F>(
     let mut rng = StdRng::seed_from_u64(config.seed);
 
     for test_number in 0..config.number_of_tests {
-        let entries = generator(&mut rng);
+        let entries: Vec<Reverse<Entry<Msg>>> = generator(&mut rng);
 
         let node_handles: Vec<_> = (1..=config.number_of_nodes)
             .map(|i| (format!("n{}", i), spawn()))
@@ -272,24 +272,30 @@ pub fn simulate<Msg, F>(
         let mut world = World::new(entries.clone(), node_handles);
 
         match world.run_world() {
-            Err((reason, history)) => display_failure(
-                test_number,
-                config.seed,
-                entries,
-                History(history.to_vec()),
-                trace_buffer.clone(),
-                reason,
-            ),
-            Ok(history) => match property(History(history.to_vec())) {
-                Ok(()) => continue,
-                Err(reason) => display_failure(
+            Err((reason, history)) => {
+                display_failure(
                     test_number,
                     config.seed,
                     entries,
                     History(history.to_vec()),
                     trace_buffer.clone(),
                     reason,
-                ),
+                );
+                break;
+            }
+            Ok(history) => match property(History(history.to_vec())) {
+                Ok(()) => continue,
+                Err(reason) => {
+                    display_failure(
+                        test_number,
+                        config.seed,
+                        entries,
+                        History(history.to_vec()),
+                        trace_buffer.clone(),
+                        reason,
+                    );
+                    break;
+                }
             },
         }
     }


simulation/amaru-sim/src/simulator/generate.rs
```diff
@@ -329,12 +319,89 @@ pub fn generate_arrival_times<R: Rng>(
     arrival_times
 }
 
+pub fn generate_entries<R: Rng>(
+    file_path: &PathBuf,
+    start_time: Instant,
+    mean_millis: f64,
+    number_of_upstream_peers: u8,
+) -> impl Fn(&mut R) -> Vec<Reverse<Entry<ChainSyncMessage>>> {
```
rkuhn: @stevana you had added `+ use<'_, R>` here and so I went and investigated that new feature; it seems that `+ '_` would have done the trick, but then I thought why not update to edition 2024 which does this by default, so we'll need to fix this anyway. Feel free to remove my commit from your branch again if you don’t like it.

crates/amaru/src/bin/amaru/cmd/import_ledger_state.rs
```diff
@@ -653,7 +655,7 @@ fn import_stake_pools(
     updates: BTreeMap<PoolId, PoolParams>,
     retirements: BTreeMap<PoolId, Epoch>,
     era_history: &EraHistory,
-) -> Result<(), impl std::error::Error> {
+) -> Result<(), impl std::error::Error + 'static> {
```
rkuhn: @stevana This, BTW, is the concise way of saying “give me the edition 2021 behaviour back”.

simulation/amaru-sim/src/simulator/generate.rs
```diff
@@ -329,12 +319,89 @@ pub fn generate_arrival_times<R: Rng>(
...
+        let mut entries: Vec<Reverse<Entry<ChainSyncMessage>>> = vec![];
+        for client in 1..=number_of_upstream_peers {
+            let messages =
+                generate_inputs(rng, file_path).expect("Failed to generate inputs from chain file");
+            let arrival_times =
+                generate_arrival_times(rng, start_time, mean_millis, messages.len());
+            entries.extend(
+                messages
+                    .into_iter()
+                    .enumerate()
```
rkuhn, line 338: could also zip it with the arrival_times

simulation/amaru-sim/src/simulator/simulate.rs
```diff
@@ -242,69 +250,105 @@ impl<Msg> Drop for World<Msg> {
...
-                })
-            })
-            .collect()
-    });
-    let result = runner.run(&generate_messages, |initial_messages| {
-        let node_handles: Vec<_> = (1..=number_of_nodes)
+    let mut rng = StdRng::seed_from_u64(config.seed);
+
+    for test_number in 0..config.number_of_tests {
+        let entries = generator(&mut rng);
```
rkuhn: adding the type annotation here would make reading the diff on github easier

simulation/amaru-sim/src/simulator/simulate.rs
```diff
@@ -242,69 +250,105 @@ impl<Msg> Drop for World<Msg> {
...
+            Err((reason, history)) => display_failure(
+                test_number,
+                config.seed,
+                entries,
+                History(history.to_vec()),
+                trace_buffer.clone(),
+                reason,
+            ),
+            Ok(history) => match property(History(history.to_vec())) {
+                Ok(()) => continue,
```
rkuhn: this `continue` seems extraneous, but does it hint that the `Err(reason)` case should actually `break`?

simulation/amaru-sim/src/simulator/generate.rs
```diff
@@ -329,12 +319,89 @@ pub fn generate_arrival_times<R: Rng>(
...
+    mean_millis: f64,
+    number_of_upstream_peers: u8,
+) -> impl Fn(&mut R) -> Vec<Reverse<Entry<ChainSyncMessage>>> {
+    move |rng| {
+        let mut entries: Vec<Reverse<Entry<ChainSyncMessage>>> = vec![];
+        for client in 1..=number_of_upstream_peers {
+            let messages =
+                generate_inputs(rng, file_path).expect("Failed to generate inputs from chain file");
+            let arrival_times =
+                generate_arrival_times(rng, start_time, mean_millis, messages.len());
```
rkuhn, line 334: is this the main change in the PR, allowing arrival times to be close to each other?

rkuhn (commented):
Looks reasonable, I hope I got the point of the PR, though.

simulation/amaru-sim/src/simulator/generate.rs
stevana: Edition of what? The language "standard"? I don't mind, seems reasonable to use the latest.

simulation/amaru-sim/src/simulator/generate.rs
stevana: It's not just a zip, right? Because Im building a `Reverse Entry`. What I'd really need is a `zipWith :: (a -> b -> c) -> [a] -> [b] -> [c]`, but I couldn't find that without depending on some third-party library?

simulation/amaru-sim/src/simulator/generate.rs
stevana: My fault, I should have explained better: the main change is the `for client in 1..=number_of_upstream_peers` part, which simulates several upstream peers walking the pre-generated block tree, potentially creating situations where the SUT has to choose the longest chain from two possible candidates. Given that the pre-generated block tree only contains one branch which leads to an dead end, it's prerhaps not so interesting, but that what we got so far.

(I'm not 100% sure that I've done it correctly, but at least this how I understood Arnaud explaining it.)

simulation/amaru-sim/src/simulator/simulate.rs
stevana: I think the `Err` case always results in a `panic!`, but yeah maybe an explicit break would make things more clear?

stevana: Added breaks.

simulation/amaru-sim/src/simulator/simulate.rs
stevana: Fixed, thanks

simulation/amaru-sim/src/simulator/generate.rs
stevana: Actually, I reverted the edition change as it was causing conflicts with main. Let's change this in a separate PR.

PR #345, opened 2025-07-22
Add shrinking a la delta debugging

stevana:
Add shrinking to the simulator, essentially bisecting the input for as long as the same error is returned by the test.

## Summary by CodeRabbit

* **New Features**
  * Added automatic shrinking of failing simulation inputs, minimizing them to the smallest case that still triggers errors.
  * Introduced delta debugging to help identify minimal failing scenarios in simulations.
  * Added a new function to combine generated vectors element-wise for enhanced test input creation.

* **Refactor**
  * Improved simulation test loop structure for better maintainability and error handling.
  * Updated error reporting to provide concise failure messages and display minimized failing inputs.

* **Chores**
  * Enhanced message generation in tests to include randomized arrival times for more robust simulation scenarios.

### DIFF
--- a/simulation/amaru-sim/src/simulator/generate.rs
+++ b/simulation/amaru-sim/src/simulator/generate.rs
@@ -386,12 +386,7 @@ pub fn generate_zip_with<A: Copy, B: Copy, C>(
         let xs = generator1(rng);
         let ys = generator2(rng);
         assert_eq!(xs.len(), ys.len());
-        let mut zs = Vec::with_capacity(xs.len());
-
-        for i in 0..xs.len() {
-            zs.push(f(xs[i], ys[i]));
-        }
-        zs
+        xs.into_iter().zip(ys).map(|(x, y)| f(x, y)).collect()
     }
 }
 
--- a/simulation/amaru-sim/src/simulator/shrink.rs
+++ b/simulation/amaru-sim/src/simulator/shrink.rs
@@ -54,7 +54,7 @@ pub fn shrink<A: Debug + Clone, B: Debug>(
                 number_of_shrinks += 1;
                 last_error = result;
                 input = complement;
-                n = n.saturating_sub(1).max(2);
+                n = n.max(2) - 1;
                 some_complement_is_failing = true;
                 break;
             }
--- a/simulation/amaru-sim/src/simulator/simulate.rs
+++ b/simulation/amaru-sim/src/simulator/simulate.rs
@@ -250,6 +250,29 @@ impl<Msg> Drop for World<Msg> {
     }
 }
 
+fn run_test<Msg: Debug + PartialEq + Clone, F: Fn() -> NodeHandle<Msg>>(
+    number_of_nodes: u8,
+    spawn: F,
+    property: impl Fn(History<Msg>) -> Result<(), String>,
+) -> impl Fn(&[Reverse<Entry<Msg>>]) -> (History<Msg>, Result<(), String>) {
+    move |entries| {
+        let node_handles: Vec<_> = (1..=number_of_nodes)
+            .map(|i| (format!("n{}", i), spawn()))
+            .collect();
+
+        let mut world = World::new(entries.to_vec(), node_handles);
+
+        match world.run_world() {
+            Ok(history) => {
+                let history = History(history.to_vec());
+                let result = property(history.clone());
+                (history, result)
+            }
+            Err((reason, history)) => (History(history.to_vec()), Err(reason)),
+        }
+    }
+}
+
 pub fn simulate<Msg, F>(
     config: SimulateConfig,
     spawn: F,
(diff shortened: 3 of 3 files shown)


simulation/amaru-sim/src/simulator/simulate.rs
```diff
@@ -262,41 +286,29 @@ pub fn simulate<Msg, F>(
...
 
-        let node_handles: Vec<_> = (1..=config.number_of_nodes)
-            .map(|i| (format!("n{}", i), spawn()))
-            .collect();
-
-        let mut world = World::new(entries.clone(), node_handles);
-
-        match world.run_world() {
-            Err((reason, history)) => {
+        match run_test(config.number_of_nodes, &spawn, &property)(&entries) {
```
abailly, line 292: can you reuse `run_test(..)` here, eg. something like 
```
let test = run_test(config.number_of_nodes, &spawn, &property)
match test(&entries) {
...
}
```

Did not try it so take with a grain of salt

abailly (approved):
nice job! Just minor suggestion which might or might not work

simulation/amaru-sim/src/simulator/simulate.rs
stevana: Seems to work, will make it part of next PR, thanks!

simulation/amaru-sim/src/simulator/generate.rs
```diff
@@ -377,6 +377,19 @@ pub fn generate_u8(low: u8, high: u8) -> impl Fn(&mut StdRng) -> u8 {
...
 
+pub fn generate_zip_with<A: Copy, B: Copy, C>(
+    generator1: impl Fn(&mut StdRng) -> Vec<A>,
+    generator2: impl Fn(&mut StdRng) -> Vec<B>,
+    f: impl Fn(A, B) -> C,
+) -> impl Fn(&mut StdRng) -> Vec<C> {
+    move |rng| {
+        let xs = generator1(rng);
+        let ys = generator2(rng);
+        assert_eq!(xs.len(), ys.len());
```
rkuhn, line 388: this seems like a pitfall: how about generating the length first and passing it to the generators as a parameter?

simulation/amaru-sim/src/simulator/shrink.rs
```diff
@@ -0,0 +1,161 @@
...
+                complement.extend_from_slice(&input[start + subset_length..]);
+            }
+            // NOTE: that if we get a different error than the expected one, we treat it as a
+            // passing test.
+            let result = test(&complement);
+            if error_predicate(&result) {
+                number_of_shrinks += 1;
+                last_error = result;
+                input = complement;
+                n = n.max(2) - 1;
```
rkuhn, line 57: when we get here within the first iteration of the outer loop, we set `n=1` while `some_ completement_is_failing=true`, which means we try the empty input; if the empty input is accepted by the error_predicate, then this leads to an endless loop.

Instead of doing an extra round with the empty input, how about leaving `n` alone when it was 2 to begin with? So basically

```rust
n = (n - 1).max(2);
```

simulation/amaru-sim/src/simulator/shrink.rs
```diff
@@ -0,0 +1,161 @@
...
+            "shrink, error predicate doesn't hold for initial input: '{:?}'",
+            input
+        )
+    }
+    let mut n = 2;
+    while input.len() >= 2 {
+        let mut start = 0;
+        let subset_length = input.len() / n;
+        let mut some_complement_is_failing = false;
+        while start < input.len() {
```
rkuhn, line 44: when `input.len()` is odd, this will run a third time with a single input element removed at the end — is this required by the algorithm, or shouldn’t we rather round up and change the second `extend_from_slice` to just cap the copy?

simulation/amaru-sim/src/simulator/simulate.rs
```diff
@@ -249,6 +250,29 @@ impl<Msg> Drop for World<Msg> {
     }
 }
 
+fn run_test<Msg: Debug + PartialEq + Clone, F: Fn() -> NodeHandle<Msg>>(
+    number_of_nodes: u8,
+    spawn: F,
+    property: impl Fn(History<Msg>) -> Result<(), String>,
```
rkuhn, line 256: shouldn’t the property just get `&History`?

simulation/amaru-sim/src/simulator/generate.rs
stevana: Fixed in new PR: https://github.com/pragma-org/amaru/pull/353, thanks!

simulation/amaru-sim/src/simulator/shrink.rs
stevana: Good catch!

simulation/amaru-sim/src/simulator/shrink.rs
stevana: I think removing a single input element from the end is correct...

simulation/amaru-sim/src/simulator/simulate.rs
stevana: Fixed, thanks!

PR #347, opened 2025-07-23
Support arm

pgrange:
This P.R. adapts the C.I. to build and run tests on armv6 and armv7 architecture so that we ensure we support both of them.

## Summary by CodeRabbit

* **Chores**
  * Added cross-compilation testing for ARM architectures in the CI workflow.
  * Updated build job naming for improved clarity in CI.
  * Enhanced target-specific configuration for ARM builds.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -78,6 +78,11 @@ jobs:
             command: test
             cross-compile: true
 
+          - runner: ubuntu-latest
+            target: armv7-unknown-linux-gnueabihf
+            command: test
+            cross-compile: true
+
     timeout-minutes: 30
     runs-on: ${{ matrix.environments.runner }}
     steps:
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -73,6 +73,11 @@ jobs:
             setup: rustup target add aarch64-unknown-linux-musl
             cross-compile: true
 
+          - runner: ubuntu-latest
+            target: arm-unknown-linux-gnueabi #armv6 architecture
+            command: test
+            cross-compile: true
+
     timeout-minutes: 30
     runs-on: ${{ matrix.environments.runner }}
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/CI.yml
```diff
@@ -73,6 +73,16 @@ jobs:
             setup: rustup target add aarch64-unknown-linux-musl
             cross-compile: true
 
+          - runner: ubuntu-latest
+            target: arm-unknown-linux-gnueabi #armv6 architecture
+            command: test
+            cross-compile: true
+
+          - runner: ubuntu-latest
+            target: armv7-unknown-linux-gnueabihf
```
jeluard, line 82: I believe this is redundant with the `aarch64-unknown-linux-musl` we already have for recent RPI.

.github/workflows/CI.yml
```diff
@@ -73,6 +73,16 @@ jobs:
             setup: rustup target add aarch64-unknown-linux-musl
             cross-compile: true
 
+          - runner: ubuntu-latest
+            target: arm-unknown-linux-gnueabi #armv6 architecture
```
jeluard, line 77: Which RPI generation does this target?

.github/workflows/CI.yml
pgrange: It's my understanding that arch64 is for 64 bits architecture with armv8 when armv7 is for, well armv7 and 32 bits.
I could be mistaken here.

.github/workflows/CI.yml
pgrange: armv6 targets Pi Zero and Pi 1

jeluard: We're looking for recent PI support like zero2 and 5.

.github/workflows/CI.yml
jeluard: I understand for 4/5 `aarch64` should be good if we target PiOS 64 bits (preferred one)?

jeluard:
After more experimentations with RPIs I believe target `aarch64-linux` so we probably don't need those changes.

PR #351, opened 2025-07-29
Rewrote the encoding for MemoizedTransactionOutput

jonathanlim222:
This PR rewrites the encoding for MemoizedTransactionOutput to avoid encoding to MintedTransactionOutput and uses the original bytes to decode.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced transaction output decoding to support both current and legacy formats, improving compatibility and reliability.

* **Tests**
  * Added tests to verify encoding and decoding of transaction outputs with datum hashes in legacy and current formats.

### DIFF
--- a/crates/amaru-kernel/src/memoized/transaction_output.rs
+++ b/crates/amaru-kernel/src/memoized/transaction_output.rs
@@ -82,6 +82,7 @@ impl<'b, C> cbor::Decode<'b, C> for MemoizedTransactionOutput {
                                 d.tag()?;
                                 // Not sure why, but there is 1 extra byte when decoding
                                 // the PlutusData if this is not moved ahead one position
+                                // TODO: Determine why this is necessary
                                 d.set_position(d.position() + 1);
                                 let plutus_data: KeepRaw<'_, PlutusData> = d.decode_with(ctx)?;
                                 let memoized_data = MemoizedPlutusData::from(plutus_data);
@@ -116,8 +117,10 @@ impl<'b, C> cbor::Decode<'b, C> for MemoizedTransactionOutput {
                                 }
                             },
                             Err(e) => {
-                                println!("Failed to decode script: {:?}", e);
-                                None
+                                return Err(cbor::decode::Error::message(format!(
+                                    "failed to decode script: {:?}",
+                                    e
+                                )));
                             }
                         };
                     }
@@ -140,7 +143,10 @@ impl<'b, C> cbor::Decode<'b, C> for MemoizedTransactionOutput {
                 datum = MemoizedDatum::Hash(d.bytes()?.into());
             }
         } else {
-            unreachable!("output not in array or map ?!")
+            return Err(cbor::decode::Error::message(format!(
+                "unexpected CBOR type for output: {:?}",
+                data_type
+            )));
         }
 
         let address = address
--- a/crates/amaru-kernel/src/memoized/transaction_output.rs
+++ b/crates/amaru-kernel/src/memoized/transaction_output.rs
@@ -365,13 +365,11 @@ mod tests {
             script: None,
         };
 
-        // Encode
         let mut encoder = cbor::Encoder::new(Vec::new());
         let mut ctx = ();
         original.encode(&mut encoder, &mut ctx).unwrap();
         let encoded_bytes = encoder.writer().clone();
 
-        // Decode
         let mut decoder = cbor::Decoder::new(&encoded_bytes);
         let decoded: MemoizedTransactionOutput = decoder.decode_with(&mut ctx).unwrap();
 
@@ -396,13 +394,11 @@ mod tests {
             script: None,
         };
 
-        // Encode
         let mut encoder = cbor::Encoder::new(Vec::new());
         let mut ctx = ();
         original.encode(&mut encoder, &mut ctx).unwrap();
         let encoded_bytes = encoder.writer().clone();
 
-        // Decode
         let mut decoder = cbor::Decoder::new(&encoded_bytes);
         let decoded: MemoizedTransactionOutput = decoder.decode_with(&mut ctx).unwrap();
 
--- a/crates/amaru-kernel/src/memoized/transaction_output.rs
+++ b/crates/amaru-kernel/src/memoized/transaction_output.rs
@@ -80,11 +80,8 @@ impl<'b, C> cbor::Decode<'b, C> for MemoizedTransactionOutput {
                             }
                             1 => {
                                 d.tag()?;
-                                // Not sure why, but there is 1 extra byte when decoding
-                                // the PlutusData if this is not moved ahead one position
-                                // TODO: Determine why this is necessary
-                                d.set_position(d.position() + 1);
-                                let plutus_data: KeepRaw<'_, PlutusData> = d.decode_with(ctx)?;
+                                let plutus_data: KeepRaw<'_, PlutusData> =
+                                    cbor::decode(d.bytes()?)?;
                                 let memoized_data = MemoizedPlutusData::from(plutus_data);
                                 datum = MemoizedDatum::Inline(memoized_data);
                             }
@@ -367,7 +364,7 @@ mod tests {
             )
             .unwrap(),
             value: Value::Coin(1500000),
-            datum: datum,
+            datum,
             script: None,
         };
 
@@ -396,7 +393,7 @@ mod tests {
             )
             .unwrap(),
             value: Value::Coin(1500000),
-            datum: datum,
+            datum,
             script: None,
         };
 
--- a/crates/amaru-kernel/src/memoized/transaction_output.rs
+++ b/crates/amaru-kernel/src/memoized/transaction_output.rs
@@ -16,9 +16,11 @@ use crate::{
     cbor,
     script::{encode_script, serialize_memoized_script, PlaceholderScript},
     Address, AlonzoValue, AssetName, KeyValuePairs, Lovelace, MemoizedDatum, MemoizedNativeScript,
-    MemoizedScript, MintedTransactionOutput, NonEmptyKeyValuePairs, PseudoScript, ScriptHash,
-    Value,
+    MemoizedPlutusData, MemoizedScript, MintedTransactionOutput, NonEmptyKeyValuePairs,
+    PseudoScript, ScriptHash, Value,
 };
+
+use pallas_primitives::{conway::NativeScript, KeepRaw, PlutusData};
 
 #[derive(Debug, Clone, serde::Deserialize, serde::Serialize, PartialEq)]
 pub struct MemoizedTransactionOutput {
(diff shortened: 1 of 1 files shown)


crates/amaru-kernel/src/memoized/transaction_output.rs
```diff
@@ -41,12 +43,119 @@ pub struct MemoizedTransactionOutput {
...
+                            }
+                            1 => {
+                                d.tag()?;
+                                let plutus_data: KeepRaw<'_, PlutusData> =
+                                    cbor::decode(d.bytes()?)?;
+                                let memoized_data = MemoizedPlutusData::from(plutus_data);
+                                datum = MemoizedDatum::Inline(memoized_data);
+                            }
+                            _ => {
+                                d.skip()?;
```
yHSJ: We should not be using `d.skip()` in the case of an invalid value here. If there is something that does not follow the CDDL definition, it shouldn't be considered a `TransactionOutput` and thus we shouldn't create a `MemoizedTransactionOutput`. Instead, we should produce a deserialization error in this case

crates/amaru-kernel/src/memoized/transaction_output.rs
```diff
@@ -41,12 +43,119 @@ pub struct MemoizedTransactionOutput {
...
+                    }
+                    2 => {
+                        d.array()?;
+                        let datum_option = d.u8()?;
+                        match datum_option {
+                            0 => {
+                                datum = MemoizedDatum::Hash(d.bytes()?.into());
+                            }
+                            1 => {
+                                d.tag()?;
```
yHSJ: Similar to the `skip`, we should also be validating the tag matches the tag we expect here.

crates/amaru-kernel/src/memoized/transaction_output.rs
```diff
@@ -41,12 +43,119 @@ pub struct MemoizedTransactionOutput {
...
+                            Err(e) => {
+                                return Err(cbor::decode::Error::message(format!(
+                                    "failed to decode script: {:?}",
+                                    e
+                                )));
+                            }
+                        };
+                    }
+                    _ => {
+                        d.skip()?;
```
yHSJ: Same as above

yHSJ (changes requested):
The happy path logic looks good, just missing errors. We shouldn't decode something that doesn't fit our CDDL definition.

yHSJ:
Please also make sure your commits are signed, for the DCO validation. You can sign old commits with `git rebase HEAD~5 --signoff`

Also, With this logic in place it's worth writing generators and property tests for this logic in the future. Not necessary for this PR, but something to keep in mind (@KtorZ)

crates/amaru-kernel/src/memoized/transaction_output.rs
jonathanlim222: I added tag validation for the script tag, but do we need to add a tag validation here if there is a validation in the PlutusData decoding? 

https://github.com/txpipe/pallas/blob/5a32392d15a41e8311a1f5b562bd2a83848a0a43/pallas-primitives/src/plutus_data.rs#L58

```match type_ {
            minicbor::data::Type::Tag => {
                let mut probe = d.probe();
                let tag = probe.tag()?;

                if tag == IanaTag::PosBignum.tag() || tag == IanaTag::NegBignum.tag() {
                    Ok(Self::BigInt(d.decode_with(ctx)?))
                } else {
                    match tag.as_u64() {
                        (121..=127) | (1280..=1400) | 102 => Ok(Self::Constr(d.decode_with(ctx)?)),
                        _ => Err(minicbor::decode::Error::message(
                            "unknown tag for plutus data tag",
                        )),
                    }
                }
            }...
```

crates/amaru-kernel/src/memoized/transaction_output.rs
jonathanlim222: Added error:
```
return Err(cbor::decode::Error::message(format!(
                                    "unknown datum option: {}",
                                    datum_option
                                )));
```

crates/amaru-kernel/src/memoized/transaction_output.rs
jonathanlim222: Added error:
```
return Err(cbor::decode::Error::message(format!(
                                        "failed to decode script: {:?}",
                                        e
                                    )));
```

crates/amaru-kernel/src/memoized/transaction_output.rs
yHSJ: Yes, it is still needed. According to the [CDDL](https://github.com/IntersectMBO/cardano-ledger/blob/master/eras/conway/impl/cddl-files/conway.cddl#L243), there should be a tag `24` wrapped around the `PlutusData` bytes. The tags _inside_ the `PlutusData` are validated by the logic you shared.

jonathanlim222: Got it, added tag validation checks.

crates/amaru-kernel/src/memoized/transaction_output.rs
```diff
@@ -41,12 +43,145 @@ pub struct MemoizedTransactionOutput {
...
 
 impl<'b, C> cbor::Decode<'b, C> for MemoizedTransactionOutput {
-    // TODO: This implementation still copies the bytes when we convert from the minted version to
-    // the owned version. To avoid that, we should actually rewrite the decoders for all the
-    // memoized structures, so that they can retain their original bytes.
     fn decode(d: &mut cbor::Decoder<'b>, ctx: &mut C) -> Result<Self, cbor::decode::Error> {
-        let minted: MintedTransactionOutput<'_> = d.decode_with(ctx)?;
-        Self::try_from(minted).map_err(cbor::decode::Error::message)
+        let data_type = d.datatype()?;
+        let mut is_legacy = false;
```
jeluard, line 48: Functional style is favored in the codebase

crates/amaru-kernel/src/memoized/transaction_output.rs
```diff
@@ -41,12 +43,145 @@ pub struct MemoizedTransactionOutput {
...
+        let mut address: Option<Address> = None;
+        let mut value: Value = Value::Coin(0);
+        let mut datum: MemoizedDatum = MemoizedDatum::None;
+        let mut script: Option<MemoizedScript> = None;
+
+        if data_type == cbor::data::Type::MapIndef || data_type == cbor::data::Type::Map {
```
jeluard, line 54: Pretty hard to follow, use smaller self explanatory functions

crates/amaru-kernel/src/memoized/transaction_output.rs
```diff
@@ -41,12 +43,145 @@ pub struct MemoizedTransactionOutput {
...
+                                        PseudoScript::PlutusV1Script(script_bytes) => {
+                                            Some(PseudoScript::PlutusV1Script(script_bytes))
+                                        }
+                                        PseudoScript::PlutusV2Script(script_bytes) => {
+                                            Some(PseudoScript::PlutusV2Script(script_bytes))
+                                        }
+                                        PseudoScript::PlutusV3Script(script_byte) => {
+                                            Some(PseudoScript::PlutusV3Script(script_byte))
+                                        }
+                                    },
```
KtorZ, line 134: This is now duplicated with the `TryFrom` implementation. It's a significant-enough portion to live in a separate function.

crates/amaru-kernel/src/memoized/transaction_output.rs
```diff
@@ -41,12 +43,145 @@ pub struct MemoizedTransactionOutput {
...
+                                    datum_option
+                                )));
+                            }
+                        }
+                    }
+                    3 => {
+                        let tag = d.tag()?;
+
+                        match tag.as_u64() {
+                            24 => {
```
KtorZ, line 110: Avoid magic numbers here; this is a well known IANA tag for CBOR-in-CBOR structure, and minicbor comes with a nice module API for well known tags: https://docs.rs/minicbor/latest/minicbor/data/enum.IanaTag.html

crates/amaru-kernel/src/memoized/transaction_output.rs
```diff
@@ -41,12 +43,145 @@ pub struct MemoizedTransactionOutput {
...
+                match key {
+                    0 => {
+                        address = Some(Address::from_bytes(d.bytes()?).map_err(|e| {
+                            cbor::decode::Error::message(format!("invalid address: {e:?}"))
+                        })?);
+                    }
+                    1 => {
+                        value = d.decode_with(ctx)?;
+                    }
+                    2 => {
```
KtorZ, line 74: The code spread enough here to deserves moving the code for decoding fields into separate functions; thus improving the readability of the match here:

```rs
match d.u8()? {
   0 => decode_address(&mut address, d)?,
   1 => decode_value(&mut value, d)?,
   2 => decode_datum(&mut datum, d)?,
   3 => decode_reference_script(&mut script, d)?, 
}
```

crates/amaru-kernel/src/memoized/transaction_output.rs
```diff
@@ -41,12 +43,145 @@ pub struct MemoizedTransactionOutput {
...
+                        d.array()?;
```
KtorZ, line 75: This presents two issues (which ideally should be identified with tests):

1. The length of definite arrays is discarded; although it should in theory be checked to be 2. 
2. When using an indefinite array, this never decodes the end break; leaving some trailing bytes that can cause issues when decoding further down.

crates/amaru-kernel/src/memoized/transaction_output.rs
```diff
@@ -41,12 +43,145 @@ pub struct MemoizedTransactionOutput {
...
+                        let datum_option = d.u8()?;
+                        match datum_option {
+                            0 => {
+                                datum = MemoizedDatum::Hash(d.bytes()?.into());
```
KtorZ, line 79: Please avoid opaque `.into`; favor `.from` instead using the target type as a namespace.

crates/amaru-kernel/src/memoized/transaction_output.rs
```diff
@@ -41,12 +43,145 @@ pub struct MemoizedTransactionOutput {
...
+                }
+            }
+        } else if data_type == cbor::data::Type::ArrayIndef || data_type == cbor::data::Type::Array
+        {
+            is_legacy = true;
+            d.array()?;
+            address =
+                Some(Address::from_bytes(d.bytes()?).map_err(|e| {
+                    cbor::decode::Error::message(format!("invalid address: {e:?}"))
+                })?);
```
KtorZ, line 162: This code is duplicated from just above.

crates/amaru-kernel/src/memoized/transaction_output.rs
```diff
@@ -41,12 +43,145 @@ pub struct MemoizedTransactionOutput {
...
+            value = d.decode_with(ctx)?;
+
+            if d.datatype()? == cbor::data::Type::Bytes {
+                datum = MemoizedDatum::Hash(d.bytes()?.into());
```
KtorZ, line 166: Avoid opaque `.into` here ☝️, especially when decoding things where it's easy to sneak needless conversions that can be quite costly, without realizing it.

crates/amaru-kernel/src/memoized/transaction_output.rs
```diff
@@ -41,12 +43,145 @@ pub struct MemoizedTransactionOutput {
...
+        {
+            is_legacy = true;
+            d.array()?;
+            address =
+                Some(Address::from_bytes(d.bytes()?).map_err(|e| {
+                    cbor::decode::Error::message(format!("invalid address: {e:?}"))
+                })?);
+            value = d.decode_with(ctx)?;
+
+            if d.datatype()? == cbor::data::Type::Bytes {
```
KtorZ, line 165: That seems incorrect; what should dictate the decoding of the next item is the size of the array; not whether there are bytes.

crates/amaru-kernel/src/memoized/transaction_output.rs
```diff
@@ -41,12 +43,145 @@ pub struct MemoizedTransactionOutput {
...
+                    }
+                    _ => {
+                        return Err(cbor::decode::Error::message("unknown key option"));
+                    }
+                }
+            }
+        } else if data_type == cbor::data::Type::ArrayIndef || data_type == cbor::data::Type::Array
+        {
+            is_legacy = true;
+            d.array()?;
```
KtorZ, line 158: Same issue as above here ☝️

crates/amaru-kernel/src/memoized/transaction_output.rs
```diff
@@ -41,12 +43,145 @@ pub struct MemoizedTransactionOutput {
...
+            }
+        } else {
+            return Err(cbor::decode::Error::message(format!(
+                "unexpected CBOR type for output: {:?}",
+                data_type
+            )));
+        }
+
+        let address = address
+            .ok_or_else(|| cbor::decode::Error::message("missing required address field"))?;
```
KtorZ, line 176: Alternatively, to avoid a needless comparison and logic branching here; you could have both sub-decoders (legacy and modern) return a decoded address. 

This way, you can still have them split into separate independent functions, but you removes by construction the case where there's no address -- which is also not a cbor user error, but an unreachable statement.

crates/amaru-kernel/src/memoized/transaction_output.rs
jonathanlim222: Followup [PR](https://github.com/pragma-org/amaru/pull/361)

crates/amaru-kernel/src/memoized/transaction_output.rs
jonathanlim222: Followup [PR](https://github.com/pragma-org/amaru/pull/361) with smaller self explanatory functions

crates/amaru-kernel/src/memoized/transaction_output.rs
jonathanlim222: Followup [PR](https://github.com/pragma-org/amaru/pull/361)

Separated out this section to helper function.

crates/amaru-kernel/src/memoized/transaction_output.rs
jonathanlim222: Followup [PR](https://github.com/pragma-org/amaru/pull/361)

Rewrote to check tag against IanaTag::Cbor

crates/amaru-kernel/src/memoized/transaction_output.rs
jonathanlim222: Followup [PR](https://github.com/pragma-org/amaru/pull/361)

Rewrote to improve readability

crates/amaru-kernel/src/memoized/transaction_output.rs
jonathanlim222: Followup [PR](https://github.com/pragma-org/amaru/pull/361)

This array should be encoded as size 2, and not indefinite array.

https://github.com/pragma-org/amaru/blob/29930eedd2c2c5f055eebf88522a6a85a7ff30e7/crates/amaru-kernel/src/memoized/datum.rs#L73

I updated to check the size in the followup.

crates/amaru-kernel/src/memoized/transaction_output.rs
jonathanlim222: Followup [PR](https://github.com/pragma-org/amaru/pull/361)

Rewrote to use `.from`

crates/amaru-kernel/src/memoized/transaction_output.rs
jonathanlim222: Followup [PR](https://github.com/pragma-org/amaru/pull/361)

Extracted address decoding to a helper function.

crates/amaru-kernel/src/memoized/transaction_output.rs
jonathanlim222: Followup [PR](https://github.com/pragma-org/amaru/pull/361)

Rewrote to use `.into`

crates/amaru-kernel/src/memoized/transaction_output.rs
jonathanlim222: Followup [PR](https://github.com/pragma-org/amaru/pull/361)

This encoder encodes a indefinite array:

https://github.com/pragma-org/amaru/blob/29930eedd2c2c5f055eebf88522a6a85a7ff30e7/crates/amaru-kernel/src/memoized/transaction_output.rs#L195-L203

I rewrote this to check for Break or Byte to determine legacy datum.

crates/amaru-kernel/src/memoized/transaction_output.rs
jonathanlim222: Followup [PR](https://github.com/pragma-org/amaru/pull/361)

The encoder encodes a indefinite array. I rewrote the legacy decoding to assume the order of the array, the datum is the only optional field, which has a check for break or byte.

crates/amaru-kernel/src/memoized/transaction_output.rs
jonathanlim222: Followup [PR](https://github.com/pragma-org/amaru/pull/361)

I extracted the address decoding to a helper function, it returns the decoded address to avoid the construction.

crates/amaru-kernel/src/memoized/transaction_output.rs
jonathanlim222: Disregard the previous comment, working on adding definite array check and support.

crates/amaru-kernel/src/memoized/transaction_output.rs
jonathanlim222: Disregard the previous comment, working on adding definite array check and support.

jonathanlim222: Updated the followup PR to have definite array check and support

crates/amaru-kernel/src/memoized/transaction_output.rs
jonathanlim222: Updated the followup PR to have definite array check and support

PR #356, opened 2025-07-31
chore: add a make command to update the license header in source files

etorreborre:
I noticed that some source files had no license headers and I suspect that some others might not be up to date.

There are a few options for maintaining source file headers in the Rust ecosystem but not necessarily full-featured or well-maintained ([here](https://github.com/grtlr/cargo-license-template) is an example of an archived attempt), and I eventually chose [licensesnip](https://github.com/notken12/licensesnip).

I integrated `licensesnip` both as a `make` task: `make update-source-headers` (there's also `make check-source-headers`) and a commit hook.

_One caveat_: there's no way to leave an existing license header as it is with `licensesnip` if its year doesn't match with today's year. This is why I updated all existing headers even if some of them correspond to files that haven't been touched since 2024. Please tell me if you think that is an issue.

A follow-up question: do we really need license headers in every file? There is already a license file at the top of the project and the license is also mentioned in `Cargo.toml`.

## Summary by CodeRabbit

* **Chores**
  * Added Apache 2.0 license headers to all relevant source and test files across the project. No changes were made to code logic, functionality, or public interfaces.

### DIFF
--- a/crates/amaru-kernel/src/macros.rs
+++ b/crates/amaru-kernel/src/macros.rs
@@ -1,3 +1,17 @@
+// Copyright 2025 PRAGMA
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 /// Generate a roundtrip property to assert that Cbor encoder and decoder for a given type can
 /// safely be called in sequence and yield the original input.
 ///
--- a/crates/amaru-kernel/src/protocol_parameters.rs
+++ b/crates/amaru-kernel/src/protocol_parameters.rs
@@ -1,3 +1,17 @@
+// Copyright 2025 PRAGMA
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 use crate::{cbor, Coin, EpochInterval, ExUnits, Lovelace, RationalNumber};
 use pallas_codec::minicbor::{data::Tag, Decoder};
 use pallas_primitives::{conway::CostModels, CostModel};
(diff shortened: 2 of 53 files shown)


jeluard:
The approach we took so far is to use the file creation year in the license. So all the ones with `2024` shouldn't be touched. Similarly the missing ones should probably reflect this logic (hence use `2024` when it's relevant).

One thing that maybe would make sense here is to ensure that new files are created with the proper license with current year. Not sure it's worth the extra complexity.

BTW I don't think you intended to update `examples/Cargo.lock`

abailly:
@etorreborre Thanks for the contribution. I second your conclusion: I find the requirement to add a per-file header cumbersome and I am still wondering why we are doing that. I don't think anyone ever provided clear evidence and rationale this is needed so I appreciate your proposal gives us an opportunity to clarify that. Any thoughts @KtorZ @Dam-CZ @Quantumplation ?

etorreborre:
> FWIW, I have so far used editor-based configuration to ensure those headers are properly generated, where the year is dynamically defined.

@abailly yes that works too. 

I've just done a bit of spelunking on licenses and copyrights:

 - The Apache 2 license recommends that a copyright mention is added as a header on top of each file. That helps with the **copyright, patent, trademark** part below:
 ```
You must retain, in the Source form of any Derivative Works that You distribute, all copyright, patent, trademark, and attribution notices from the Source form of the Work, excluding those that do not pertain to any part of the Derivative Works; and
```
 - That copyright mention can be in abbreviated form. 
 - It is acceptable to put a date range on all files starting at the creation of the project and ending on the current year.
 - **However**, [it is not fundamentally necessary to add a copyright notice](https://daniel.haxx.se/blog/2023/01/08/copyright-without-years). Theses days  a simple `NOTICE` file with a copyright mention at the root of the project should be enough to assert copyrights. The practice of having a having a header on each file dates from a time where version control systems were not as prevalent as today and people were exchanging single files over email.

I'm happy to either close the PR or implement a solution that satisfies everyone, please tell me what you think :-).

abailly:
> that's what Apache-2.0 requires for applying to an Apache license

I am not a lawyer but I don't think this is true according to https://www.apache.org/licenses/LICENSE-2.0.html#apply. It says:

> To apply the Apache License to specific files in your work, attach the following boilerplate declaration, replacing the fields enclosed by brackets "[]" with your own identifying information.

It seems rather clear to me this means: "Do this when you want to apply Apache license to _some_ files but not all".

But IANAL and have no idea about the implications of having/not having a full/partial/reduced header added to each file with/without indication of years. Could we get a real lawyer's advice on this? I personally find wasting 14 lines of file real estate at the beginning of each file annoying but 🤷 

On top of that, may I suggest we add an EDR entry to this PR and/or update `CONTRIBUTING.md` to reflect whatever decision is taken?

KtorZ:
> Could we get a real lawyer's advice on this?

I actually got this advice from a lawyer (back from my IO{G,HK} days) 😅, and from Christian (our ex board member, who worked 15+ years with the Apache foundation). 

I have seen some people use the NOTICE trick that @etorreborre mentioned; but I have yet seen any justification as for whether it makes for valid licensing. Or if people just roll with it without further concerns.

abailly:
> I actually got this advice from a lawyer (back from my IO{G,HK} days) 😅, and from Christian (our ex board member, who worked 15+ years with the Apache foundation).

Thanks @KtorZ. I was about to file a request to CF's lawyers but obviously that's not needed.

KtorZ:
@etorreborre thanks! You'll need to sign-off on your commit (`git commit --signoff`) to comply with our [developer certificate of origin](https://developercertificate.org/) (you can make that an alias for Amaru if you want; there's no configuration for it).

PR #361, opened 2025-08-04
Followup PR for encoding rewrite to address readability and comments

jonathanlim222:
Followup for https://github.com/pragma-org/amaru/pull/351


crates/amaru-kernel/src/memoized/transaction_output.rs
```diff
@@ -43,148 +45,180 @@ pub struct MemoizedTransactionOutput {
...
+}
 
-                let key = d.u8()?;
-                match key {
-                    0 => {
-                        address = Some(Address::from_bytes(d.bytes()?).map_err(|e| {
-                            cbor::decode::Error::message(format!("invalid address: {e:?}"))
-                        })?);
-                    }
+fn decode_datum(d: &mut cbor::Decoder<'_>) -> Result<MemoizedDatum, cbor::decode::Error> {
```
yHSJ: This should probably be a `cbor::Decode` implementation on `MemoizedDatum` instead

crates/amaru-kernel/src/memoized/transaction_output.rs
```diff
@@ -253,20 +287,9 @@ impl<'a> TryFrom<MintedTransactionOutput<'a>> for MemoizedTransactionOutput {
...
-                    PseudoScript::PlutusV2Script(plutus_script) => {
-                        PseudoScript::PlutusV2Script(plutus_script)
-                    }
-                    PseudoScript::PlutusV3Script(plutus_script) => {
-                        PseudoScript::PlutusV3Script(plutus_script)
-                    }
-                }),
+                script: output
+                    .script_ref
+                    .map(|script| from_memoized_script(script.unwrap())),
```
yHSJ: what are we `unwrap`ing here? Why is it safe to do so?

crates/amaru-kernel/src/memoized/transaction_output.rs
```diff
@@ -43,148 +45,180 @@ pub struct MemoizedTransactionOutput {
...
-        })
+    // Process legacy datum
+    } else if d.datatype()? == cbor::data::Type::Break {
+        Ok(MemoizedDatum::None)
+    } else {
+        Ok(MemoizedDatum::from(Some(Hash::<32>::from(d.bytes()?))))
     }
 }
 
+fn decode_address(address_bytes: &[u8]) -> Result<Address, cbor::decode::Error> {
```
yHSJ, line 132: Not sure this should be a cbor error, since the error is in parsing the address (bytes) and not decoding cbor.

crates/amaru-kernel/src/memoized/transaction_output.rs
```diff
@@ -369,6 +392,17 @@ pub fn deserialize_script<'de, D: serde::de::Deserializer<'de>>(
     }
 }
 
+fn from_memoized_script(pseudo_script: PseudoScript<KeepRaw<'_, NativeScript>>) -> MemoizedScript {
```
yHSJ: I would write a `From<PseudoScript<KeepRaw<'_, NativeScript>` implementation on `MemoizedScript` instead of this.

crates/amaru-kernel/src/memoized/transaction_output.rs
```diff
@@ -43,148 +45,180 @@ pub struct MemoizedTransactionOutput {
...
+    d: &mut cbor::Decoder<'_>,
+) -> Result<MemoizedTransactionOutput, cbor::decode::Error> {
+    let mut address_bytes: &[u8] = &[];
+    let mut value: Value = Value::Coin(0);
+    let mut datum: MemoizedDatum = MemoizedDatum::None;
+    let mut script: Option<MemoizedScript> = None;
+
+    match d.map()? {
+        Some(size) => {
+            for _ in 0..size {
```
yHSJ: I don't love the loop here. I would prefer if you:

1) validate that the size is the expected size (4). We should error if the array is a different size than expected, no reason to continue decoding
2) initialize value here, instead of in `decode_map_value`. By that I mean directly access key `0` and decode the address, etc.

I guess point 2 is mostly stylistic, just that `decode_map_value` isn't *really* what you're doing. You're decoding a specific map's value. Doesn't make sense to me to make that it's own function.

crates/amaru-kernel/src/memoized/transaction_output.rs
```diff
@@ -43,148 +45,180 @@ pub struct MemoizedTransactionOutput {
...
+                decode_map_value(d, &mut address_bytes, &mut value, &mut datum, &mut script)?;
+            }
+        }
+        None => loop {
+            if d.datatype()? == cbor::data::Type::Break {
```
yHSJ: I'd probably so the same as above, minus the size check since we don't have one here?

yHSJ (changes requested):
This is looking better, thanks for the changes. Had to do a quick review as my flight is beginning it's descent, but I think things are looking better.

In general, try to rely on traits instead of functions when you can (`From`, `Decode`, etc.).

crates/amaru-kernel/src/memoized/transaction_output.rs
```diff
@@ -43,148 +45,180 @@ pub struct MemoizedTransactionOutput {
 }
 
 impl<'b, C> cbor::Decode<'b, C> for MemoizedTransactionOutput {
-    fn decode(d: &mut cbor::Decoder<'b>, ctx: &mut C) -> Result<Self, cbor::decode::Error> {
+    fn decode(d: &mut cbor::Decoder<'b>, _ctx: &mut C) -> Result<Self, cbor::decode::Error> {
```
KtorZ: This looks suspicious 🤔; the context should probably not be ignored. Even though we don't have yet decoders depending on context; we have plans to leverage them very soon. So that would cause hard-to-debug issues down the line.

crates/amaru-kernel/src/memoized/transaction_output.rs
```diff
@@ -43,148 +45,180 @@ pub struct MemoizedTransactionOutput {
...
-        let mut script: Option<MemoizedScript> = None;
 
         if data_type == cbor::data::Type::MapIndef || data_type == cbor::data::Type::Map {
-            let map_size = d.map()?.unwrap_or(4);
+            decode_modern_output(d)
+        } else if data_type == cbor::data::Type::ArrayIndef || data_type == cbor::data::Type::Array
+        {
+            decode_legacy_output(d)
+        } else {
+            Err(cbor::decode::Error::message(format!(
```
KtorZ: What about using: https://docs.rs/minicbor/latest/minicbor/decode/struct.Error.html#method.type_mismatch

crates/amaru-kernel/src/memoized/transaction_output.rs
```diff
@@ -43,148 +45,180 @@ pub struct MemoizedTransactionOutput {
...
+            address: decode_address(d.bytes()?)?,
+            value: d.decode()?,
+            datum: decode_datum(d)?,
+            script: None,
+        }),
+        _ => Err(cbor::decode::Error::message(format!(
+            "expected legacy transaction output array length of 2 or 3, got {len:?}",
+        ))),
+    }
+}
```
KtorZ:
```suggestion
fn decode_legacy_output(
    d: &mut cbor::Decoder<'_>,
) -> Result<MemoizedTransactionOutput, cbor::decode::Error> {
    let len = d.array()?;
    Ok(MemoizedTransactionOutput {
        is_legacy: true,
        address: decode_address(d.bytes()?)?,
        value: d.decode()?,
        datum: match len {
            Some(2) => MemoizedDatum::None,
            Some(3) => decode_datum(d)?,
            // TODO: Validate whether this assumption is correct with the Haskell implementation.
            // It could be that the decoders could still accept CBOR with invalid array length;
            Some(_) => Err(cbor::decode::Error::message(format!(
                "expected legacy transaction output array length of 2 or 3, got {len:?}",
            ))),
            // TODO: Also check whether this is consistent with the Haskell implementation. Are indefinite CBOR array allowed here?
            None => {
                decode_datum(d)?;
                decode_break(d)?;
            }
        },
        script: None,
    })
}

fn decode_break(d: &mut cbor::Decoder<'_>) -> Result<(), cbor::decode::Error> {
    let is_break = d.datatype()? == cbor::data::Type::Break;
    if !is_break {
        return Err(cbor::decode::Error::type_mismatch(cbor::data::Type::Break));
    }
    d.skip()
}
```

Most of the decoder is otherwise being duplicated here. Also, There's likely no reason to not allow indefinite CBOR arrays, although that should be checked with the Haskell impl.

crates/amaru-kernel/src/memoized/transaction_output.rs
```diff
@@ -43,148 +45,180 @@ pub struct MemoizedTransactionOutput {
...
+        2 => *datum = decode_datum(d)?,
+        3 => *script = decode_reference_script(d)?,
+        _ => {
+            return Err(cbor::decode::Error::message(
+                "unexpected key in transaction output map",
+            ));
+        }
+    }
+    Ok(())
+}
```
KtorZ: The logic split isn't quite right here. The switch between map field index would better live in the `decode_modern_output` -- which is visible here because this function ends up taking multiple mutable references as arguments while only updating one at a time. There's a clear connection between which argument is updated and the value of `d.u8()`; so the logic feels backward and is hard to read. 

See my other suggestion regarding `decode_modern_output`.

crates/amaru-kernel/src/memoized/transaction_output.rs
```diff
@@ -43,148 +45,180 @@ pub struct MemoizedTransactionOutput {
...
+    }
+}
+
+fn decode_modern_output(
+    d: &mut cbor::Decoder<'_>,
+) -> Result<MemoizedTransactionOutput, cbor::decode::Error> {
+    let mut address_bytes: &[u8] = &[];
+    let mut value: Value = Value::Coin(0);
+    let mut datum: MemoizedDatum = MemoizedDatum::None;
+    let mut script: Option<MemoizedScript> = None;
```
KtorZ: This is actually really wrong and dangerous. You should really avoid writing decoders like this (i.e. by plucking temporary default values). 

What's the problem exactly? Imagine we are given a map that contains no value (i.e. no field `1`). This implementation will succeed in decoding the output, and assign it a value of `0`. Whereas the given output is in fact totally invalid. 

I understand _why_ this is written this way, and also the motivations for `decode_map_value`. Yet, this also creates another logic inversion problem; while also being very much tailored to this particular decoder. 

I recognize, however, that it isn't an easy problem here. So here's a concrete suggestion and approach that we can then apply more holistically to other map decoders:

First, we need a more flexible heterogeneous map decoder (could be eventually upstream to `minicbor`); because the one that minicbor provides in the form of `map_iter` is not usable in this context. Since we don't know in which order the fields are coming, we have to give back the control to the caller when it comes to what is decoded when; but we can define a function that takes care of the surrounding:

```rs
fn decode_map<K, S>(
    d: &mut cbor::Decoder<'_>,
    mut state: S,
    decode_key: impl Fn(&mut cbor::Decoder<'_>) -> Result<K, cbor::decode::Error>,
    mut decode_value: impl FnMut(&mut cbor::Decoder<'_>, &mut S, K) -> Result<(), cbor::decode::Error>,
) -> Result<S, cbor::decode::Error> {
    let len = d.map()?;

    let mut n = 0;
    while len.is_none() || Some(n) < len {
        if d.datatype()? == cbor::data::Type::Break {
            // NOTE: If we encounter a rogue Break while decoding a definite map, that's an error.
            if len.is_some() {
                return Err(cbor::decode::Error::type_mismatch(cbor::data::Type::Break));
            }
            d.skip()?;
            break;
        }

        let k = decode_key(d)?;
        decode_value(d, &mut state, k)?;

        n += 1;
    }

    Ok(state)
}
```

Note that this is generic in the key type as well, although we _could_ specialize that to `u8` since I believe most (if not all?) maps we'll be dealing with use u8 as fields. But anyway; now we have ways of decoding maps into some generic state `S`. 

For `S`, we'll use a bunch of 4 "partial decoders", which in their most simplest form can be functions that yield the thing they are decoding (possibly with some error): 

```rs
type PartialDecoder<A> = Box<dyn FnOnce() -> Result<A, cbor::decode::Error>>;
```

So now we can rewrite `decode_modern_output` as: 

```rs
fn decode_modern_output(
    d: &mut cbor::Decoder<'_>,
) -> Result<MemoizedTransactionOutput, cbor::decode::Error> {
    let (address, value, datum, script) = decode_map(
        d,
        (
           Box::new(|| 
             Err(cbor::decode::Error::message("missing address (field tag '0') in output map"))
           ) as PartialDecoder<_>,
           Box::new(|| 
             Err(cbor::decode::Error::message("missing value (field tag '1') in output map"))
           ) as PartialDecoder<_>,
           Box::new(|| 
             Ok(MemoizedDatum::None)
           ) as PartialDecoder<_>,
           Box::new(|| 
             Ok(None)
           ) as PartialDecoder<_>,
        ),
        |d| d.u8(),
        |d, state, field| todo!(),
    )?;

    Ok(MemoizedTransactionOutput {
        is_legacy: false,
        address: address()?,
        value: value()?,
        datum: datum()?,
        script: script()?,
    })
}
```

Which now gives us a nice way to write the output decoder having default values when it makes sense (datum and script) while yielding errors where it doesn't. This starts to get a little too verbose though due to Rust syntax, so we can define a couple of helpers to turn this into a cute eDSL:

```rs
fn missing_field<A>(field_tag: u8, field_title: &str, container: &str) -> PartialDecoder<A> {
    let msg = format!("missing {field_title} (field tag {field_tag}) in {container} map");
    Box::new(move || Err(cbor::decode::Error::message(msg)))
}

fn with_default_value<A: 'static>(default: A) -> PartialDecoder<A> {
    Box::new(move || Ok(default))
}
```

which gives us:

```rs
fn decode_modern_output(
    d: &mut cbor::Decoder<'_>,
) -> Result<MemoizedTransactionOutput, cbor::decode::Error> {
    let (address, value, datum, script) = decode_map(
        d,
        (
            missing_field(0, "address", "output"),
            missing_field(1, "value", "output"),
            with_default_value(MemoizedDatum::None),
            with_default_value(None),
        ),
        |d| d.u8(),
        |d, state, field| todo!(),
    )?;

    Ok(MemoizedTransactionOutput {
        is_legacy: false,
        address: address()?,
        value: value()?,
        datum: datum()?,
        script: script()?,
    })
}
```

The implementation for the value decoder can now be written as:

```rs
        |d, state, field| {
            match field {
                0 => {
                    let address = d.bytes().and_then(decode_address);
                    state.0 = Box::new(|| address)
                }
                1 => {
                    let value = d.decode();
                    state.1 = Box::new(|| value)
                }
                2 => {
                    let datum = decode_datum(d);
                    state.2 = Box::new(|| datum)
                }
                3 => {
                    let script = decode_reference_script(d);
                    state.3 = Box::new(|| script)
                }
                _ => {
                    return Err(cbor::decode::Error::message(
                        "unexpected field in transaction output map",
                    ))
                }
            }
            Ok(())
```

Where we can see a recurring pattern again, which is a good indicator that we can segregate noise into another eDSL "primitive": 

```rs
fn decode_chunk<A: 'static>(
    d: &mut cbor::Decoder<'_>,
    decode: impl FnOnce(&mut cbor::Decoder<'_>) -> Result<A, cbor::decode::Error>,
) -> PartialDecoder<A> {
    // NOTE: It is crucial that this happens *outside* of the boxed closure, to ensure bytes are consumed 
    // when the closure is created; not when it is invoked!
    let a = decode(d);
    Box::new(|| a)
}
```

Which finally leaves us with:

```rs
fn decode_modern_output(
    d: &mut cbor::Decoder<'_>,
) -> Result<MemoizedTransactionOutput, cbor::decode::Error> {
    let (address, value, datum, script) = decode_map(
        d,
        (
            missing_field(0, "address", "output"),
            missing_field(1, "value", "output"),
            with_default_value(MemoizedDatum::None),
            with_default_value(None),
        ),
        |d| d.u8(),
        |d, state, field| {
            match field {
                0 => state.0 = decode_chunk(d, |d| decode_address(d.bytes()?)),
                1 => state.1 = decode_chunk(d, |d| d.decode()),
                2 => state.2 = decode_chunk(d, decode_datum),
                3 => state.3 = decode_chunk(d, decode_reference_script),
                _ => {
                    return Err(cbor::decode::Error::message(
                        "unexpected key in transaction output map",
                    ))
                }
            }
            Ok(())
        },
    )?;

    Ok(MemoizedTransactionOutput {
        is_legacy: false,
        address: address()?,
        value: value()?,
        datum: datum()?,
        script: script()?,
    })
}
```

Which is is (likely) correct, fairly readable, and at the right level of abstraction. Plus, it leaves us with: `decode_map`, `PartialDecoder`, `missing_field`, `with_default_value` and `decode_chunk` which are all generic functions that are completely decoupled from the transaction output decoding and can be reused in many other similar scenarii.

crates/amaru-kernel/src/memoized/transaction_output.rs
```diff
@@ -43,148 +45,180 @@ pub struct MemoizedTransactionOutput {
...
+                    return Err(cbor::decode::Error::message(format!(
+                        "failed to decode script: {e}"
+                    )));
                 }
+            }
+        }
+        false => {
+            return Err(cbor::decode::Error::message("unknown tag for script tag"));
+        }
+    })
```
KtorZ:
```suggestion
  if d.tag()? !=  IanaTag::Cbor.tag() {
     return Err(cbor::decode::Error::message("unexpected tag as script tag"));
  }
   
  let mut script_decoder: cbor::Decoder<'_> = cbor::Decoder::new(d.bytes()?);
  let failed_to_decode = |e| cbor::decode::Error::message(format!("failed to decode script: {e}"));
  
  from_memoized_script(
    script_decoder
      .decode()
      .map_err(failed_to_decode)?
  )   
```

crates/amaru-kernel/src/memoized/transaction_output.rs
KtorZ: `unwrap` is how Pallas decided to call the method that extract the cbor serialized bytes from within a `KeepRaw` structure 😬 ... so there's nothing unsafe about this one, but it is utterly confusing I concede.

Quantumplation: Lol maybe we can convince Santiago to change that

crates/amaru-kernel/src/memoized/transaction_output.rs
jonathanlim222: Moved this to a `Decode` implementation for MemoizedDatum in `datum.rs`.

crates/amaru-kernel/src/memoized/transaction_output.rs
jonathanlim222: The cbor error here is what gets bubbled up to the Decode, which has a signature for a `Result<Self, cbor::decode::Error> `.

crates/amaru-kernel/src/memoized/transaction_output.rs
jonathanlim222: Moved this to a From implementation of `LocalPseudoScript` wrapper in `script.rs` for `MemoizedScript`

crates/amaru-kernel/src/memoized/transaction_output.rs
jonathanlim222: This section has been rewritten per @KtorZ suggestions below.

crates/amaru-kernel/src/memoized/transaction_output.rs
jonathanlim222: This section has been rewritten per @KtorZ suggestions below.

crates/amaru-kernel/src/memoized/transaction_output.rs
jonathanlim222: I added a `#[allow(unused_variables)] // ctx will be used in the future` to ignore the compiler warning, and have a reminder about future usage.

crates/amaru-kernel/src/memoized/transaction_output.rs
jonathanlim222: Revised using the `Error::type_mismatch`

crates/amaru-kernel/src/memoized/transaction_output.rs
jonathanlim222: I double checked the Haskell implementation. It appears that transaction outputs can have infinite CBOR arrays when decoding.

https://github.com/IntersectMBO/cardano-ledger/blob/640fb66d27ac202764de0dda76621c6d57852ba9/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/TxOut.hs#L391-L414

https://github.com/IntersectMBO/cardano-ledger/blob/640fb66d27ac202764de0dda76621c6d57852ba9/eras/babbage/impl/src/Cardano/Ledger/Babbage/TxOut.hs#L560-L588

I have updated the code as per your suggestion.

crates/amaru-kernel/src/memoized/transaction_output.rs
jonathanlim222: Thank you for the in depth explanation and breakdown, this was definitely an area that I did have issues with. Your suggestions organize the map decoding in a clean manner. I have updated the PR as per your suggestions.

crates/amaru-kernel/src/memoized/transaction_output.rs
jonathanlim222: I have updated the PR as per your suggestions.

crates/amaru-kernel/src/memoized/transaction_output.rs
KtorZ: Note: I've now packed this as well as other cbor utils in --> crates/minicbor-extra.

For example:

https://github.com/pragma-org/amaru/blob/main/crates/minicbor-extra/src/decode.rs#L98

jonathanlim222: Updated to utilize the cbor utils.

crates/amaru-kernel/src/memoized/transaction_output.rs
KtorZ: I am afraid that this won't do. This variable should simply NOT be ignored; the context shall be passed down to the different sub-decoders. It isn't an issue for later.

crates/amaru-kernel/src/memoized/datum.rs
```diff
@@ -13,8 +13,12 @@
 // limitations under the License.
 
 use crate::{cbor, cbor::data::IanaTag, DatumHash, MemoizedPlutusData, MintedDatumOption};
+use pallas_crypto::hash::Hash;
+use pallas_primitives::{KeepRaw, PlutusData};
 use serde::ser::SerializeStruct;
```
KtorZ:
```suggestion
```

crates/amaru-kernel/src/memoized/datum.rs
```diff
@@ -61,6 +63,65 @@
...
+                None => Err(cbor::decode::Error::message(
+                    "expected datum array length of 2, got indefinite array",
+                )),
+            }
+        // Process legacy datum
+        } else if d.datatype()? == cbor::data::Type::Break {
+            Ok(MemoizedDatum::None)
+        } else {
+            Ok(MemoizedDatum::from(Some(Hash::<32>::from(d.bytes()?))))
+        }
```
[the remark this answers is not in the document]
KtorZ: The rabbit is right here; the `from` will blow up at runtime if provided with an incorrect number of bytes. This shouldn't panic, but fails the decoder with a proper decoding error.

crates/amaru-kernel/src/memoized/datum.rs
```diff
@@ -61,6 +65,46 @@ impl<'de> serde::Deserialize<'de> for MemoizedDatum {
     }
 }
 
+impl<'b, C> cbor::Decode<'b, C> for MemoizedDatum {
+    #[allow(unused_variables)] // ctx will be used in the future
+    fn decode(d: &mut cbor::Decoder<'b>, ctx: &mut C) -> Result<Self, cbor::decode::Error> {
+        // Process modern datum
+        if d.datatype()? == cbor::data::Type::Array {
```
KtorZ: I am afraid that this isn't quite correct; although it's subtle. 

The choice of whether the decoding needs to be of a datum or a datum_hash should isn't the responsibility of this particular decoder. The choice must be within the output decoder:

- If it is a legacy output, then the datum CAN ONLY be a hash (or be missing)
- If it is a modern output, then it MUST be a structured map. 

In particular, it isn't allowed to pass a full datum in a legacy output, or a naked datum hash in a modern output. So the decision must be pushed up. And to facilitate that, I would suggest adding a new generic type to `memoized/mod.rs`:

```rs
#[repr(transparent)]
pub(crate) struct Legacy<A>(A);
```

Which can then be used to define and chose between multiple decoders:

```rs
/// Modern decoder expecting a map
impl<'b, C> cbor::Decode<'b, C> for MemoizedDatum {
      todo!()
   }
}

/// Legacy decoder only allowing 32-bytes hashes.
impl<'b, C> cbor::Decode<'b, C> for Legacy<MemoizedDatum> {
      todo!()
    }
}
```

Not that the optional nature of the datum hash is also an extrinsic concern that belongs to the parent/caller that has context over the datum; the legacy decoder itself only allows decoding datum hashes.

crates/amaru-kernel/src/memoized/transaction_output.rs
```diff
@@ -253,20 +216,9 @@ impl<'a> TryFrom<MintedTransactionOutput<'a>> for MemoizedTransactionOutput {
...
-                    PseudoScript::PlutusV2Script(plutus_script) => {
-                        PseudoScript::PlutusV2Script(plutus_script)
-                    }
-                    PseudoScript::PlutusV3Script(plutus_script) => {
-                        PseudoScript::PlutusV3Script(plutus_script)
-                    }
-                }),
+                script: output
+                    .script_ref
+                    .map(|script| MemoizedScript::from(LocalPseudoScript(script.unwrap()))),
```
KtorZ:
```suggestion
                    .map(from_minted_script),
```

I like the idea; but the `LocalPseudoScript` adds an unnecessary overhead here. A function in the `memoized/script.rs` module shall simply do:

```rs
pub fn from_minted_script(
    wrapper: CborWrap<PseudoScript<KeepRaw<'_, NativeScript>>>,
) -> MemoizedScript {
    match wrapper.0 {
        PseudoScript::NativeScript(script) => {
            MemoizedScript::NativeScript(MemoizedNativeScript::from(script))
        }
        PseudoScript::PlutusV1Script(script) => MemoizedScript::PlutusV1Script(script),
        PseudoScript::PlutusV2Script(script) => MemoizedScript::PlutusV2Script(script),
        PseudoScript::PlutusV3Script(script) => MemoizedScript::PlutusV3Script(script),
    }
}
```

crates/amaru-kernel/src/memoized/transaction_output.rs
```diff
@@ -43,146 +47,105 @@ pub struct MemoizedTransactionOutput {
...
+    Address::from_bytes(address_bytes)
+        .map_err(|e| cbor::decode::Error::message(format!("invalid address: {e:?}")))
+}
+
+fn decode_break(d: &mut cbor::Decoder<'_>) -> Result<(), cbor::decode::Error> {
+    if d.datatype()? != Type::Break {
+        return Err(cbor::decode::Error::type_mismatch(Type::Break));
     }
+    d.skip()
 }
```
KtorZ: This is already defined in the `minicbor-extra`; although likely not exposed. I would suggest exposing it and using it from there instead of re-defining this.

crates/amaru-kernel/src/memoized/transaction_output.rs
```diff
@@ -43,146 +47,105 @@ pub struct MemoizedTransactionOutput {
...
+    if d.tag()? != IanaTag::Cbor.tag() {
+        return Err(cbor::decode::Error::message("unexpected tag as script tag"));
+    }
+    let mut script_decoder: cbor::Decoder<'_> = cbor::Decoder::new(d.bytes()?);
+    let failed_to_decode =
+        |e| cbor::decode::Error::message(format!("failed to decode script: {e}"));
+
+    Ok(Some(MemoizedScript::from(LocalPseudoScript(
+        script_decoder.decode().map_err(failed_to_decode)?,
+    ))))
```
KtorZ:
```suggestion
    Ok(Some(script_decoder.decode().map_err(failed_to_decode)?))
```

This will, however, demand to define a `Decode` instance for `MemoizedNativeScript`; which is the right way to go here.

KtorZ (changes requested):
Nice progress; but there are a few remaining issues which needs to be properly addressed.

crates/amaru-kernel/src/memoized/transaction_output.rs
jonathanlim222: I made the change and added a `Decode` for `MemoizedNativeScript` in `native_script.rs`.

crates/amaru-kernel/src/memoized/transaction_output.rs
jonathanlim222: I exposed the `decode_break` in `minicbor-extra` and updated the PR to use the exposed `decode_break`.

crates/amaru-kernel/src/memoized/transaction_output.rs
jonathanlim222: I removed the `LocalPseudoScript` code and added the `from_minted_script`. It's only used on encoding the script, the script decoding now uses the `Decode` for `MemoizedNativeScript`.

crates/amaru-kernel/src/memoized/datum.rs
jonathanlim222: Thank you for the explanation, I have made the suggestions and split up the legacy and modern datum decoding.

crates/amaru-kernel/src/memoized/datum.rs
jonathanlim222: I removed the imports.

crates/amaru-kernel/src/memoized/transaction_output.rs
jonathanlim222: I added the `ctx` to decode where applicable, this is no longer ignored.

crates/amaru-kernel/src/memoized/transaction_output.rs
```diff
@@ -45,144 +50,102 @@ pub struct MemoizedTransactionOutput {
...
+            Some(_) => {
+                return Err(cbor::decode::Error::message(format!(
+                    "expected legacy transaction output array length of 2 or 3, got {len:?}",
+                )))
+            }
+            None => {
+                let datum: Legacy<MemoizedDatum> = d.decode_with(ctx)?;
+                decode_break(d, len)?;
+                datum.0
+            }
```
SupernaviX, line 92: Will this work with an indefinite-length-encoded array that has two elements? it looks like this branch expects to find a datum.

jonathanlim222: In the case of a indefinite array with only two elements, the address is decoded, the value is decoded, the datum decoding here returns `MemoizedDatum::None` and then the break is decoded.
```
if d.datatype()? == cbor::data::Type::Break {
            Ok(memoized::Legacy(MemoizedDatum::None))
        } else {
```

SupernaviX: On closer inspection, the decode implementation for `Legacy<MemoizedDatum>` is looking for either a break (end of indefinite array) or a hash (though not consuming the break if it sees it). I think that's kind of confusing: it breaks an assumption I'd been making that `decode` only parses bytes belonging to the item getting decoded, and makes it look like we expect to consume bytes in this block when we actually don't.

More importantly, this is error-prone. `decode_break` won't error if the next element isn't a `break`, it'll just return `Ok(false)`. A 4-element indefinitely-encoded list will decode successfully here when it should've failed.

I'd recommend removing that break-checking logic from `Legacy<MemoizedDatum>` and moving it in here instead.
```suggestion
            None => {
                if decode_break(d, len)? {
                    MemoizedDatum::None
                } else {
                    let datum: Legacy<MemoizedDatum> = d.decode_with(ctx)?;
                    if !decode_break(d, len)? {
                        // return an error
                    }
                    datum.0
                }
            }
```

jonathanlim222: The suggestions have been implemented, the legacy `MemoizedDatum::None` check has been moved out of the datum decoding.

jeluard:
@jonathanlim222 Is this PR ready for a new round of reviews? You will need to rebase still.

jonathanlim222:
@jeluard Yes, I have rebased, and this is ready for reviews.

jeluard:
Looks like you need to rebase again @jonathanlim222

PR #365, opened 2025-08-07
move validate_header to pure_stage

rkuhn:
This is just the trivial move using the pure_stage implementation that already exists for the simulation tests. It turns out that only small adapters are needed between gasket and pure_stage to get local tests to pass, submitting PR to ask CI for an advanced opinion.


crates/amaru/src/stages/mod.rs
```diff
@@ -191,13 +192,29 @@ pub fn bootstrap(
...
     let (to_store_header, from_receive_header) = gasket::messaging::tokio::mpsc_channel(50);
-    let (to_validate_header, from_store_header) = gasket::messaging::tokio::mpsc_channel(50);
-    let (to_select_chain, from_validate_header) = gasket::messaging::tokio::mpsc_channel(50);
     let (to_fetch_block, from_select_chain) = gasket::messaging::tokio::mpsc_channel(50);
     let (to_store_block, from_fetch_block) = gasket::messaging::tokio::mpsc_channel(50);
     let (to_ledger, from_store_block) = gasket::messaging::tokio::mpsc_channel(50);
     let (to_block_forward, from_ledger) = gasket::messaging::tokio::mpsc_channel(50);
 
+    // start pure-stage parts, whose lifecycle is managed by a single gasket stage
+    let mut network = TokioBuilder::default();
```
abailly, line 201: This function is already quite long, can we encapsulate the "pure stage" creation logic in function(s)?

abailly (commented):
I am confused: The goal of this work was to remove the redundancy introduced by the coexistence of pure-stage and gasket with the former used in simulator and the latter in production code, redundancy which is manifest through the duplication of `handle_roll_forward` and `handle_roll_forward_new`. 
The 2 functions still exist, though.

rkuhn:
@abailly The duplication between the two functions you name is caused by the access to the store, which in the current store design is asynchronous. Without changing this aspect, it is impossible to remove the duplication because pure_stage only permits asynchrony via reified external effects while outside pure_stage these reified effects cannot be evaluated (for want of the interpreter). So if we really want to retain the gasket stage in parallel to the pure_stage one (which I didn’t know was a goal) then we will have to change the store access to synchronous code, i.e. using a standard Mutex instead of a Tokio one. Is that the change you wanted me to do? In that case I’m very sorry for the misunderstanding.

rkuhn:
Ah, now I understand: the gasket implementation of validate_header can now be removed, I just haven’t done it in this PR because it adds noise. In that sense the duplication will be resolved and the transition can be done piece by piece.

crates/amaru/src/stages/mod.rs
rkuhn: yes, good point

abailly (approved):

PR #366, opened 2025-08-07
persistent state vs. pure_stage

rkuhn:
This PR shall eventually provide a good answer to accessing and mutating persistent state from pure stages. Let’s start with a little design discussion.


crates/pure-stage/README.md
```diff
@@ -40,21 +40,34 @@ The wiring function ensures that message types do match, but it won't prevent co
...
+One important caveat is that stage states are deserialized independently for each stage, meaning that for example a database connection is maintained once per process in shared global mutable state or that each stage has a separate database connection.
+While the latter may work for some database drivers, it is not an option for resources that are intended to be shared between stages.
+A possible solution to this problem would be to add a per-StageGraph resource registry that deserialization code accesses using some kind of typed key objects.
+This means that deserialization needs to be parameterized, which is not foreseen in serde and thus implies using ThreadLocal or similar hacks.
+
+This reasoning leads me to the conclusion that moving resources out of the stages and manipulating them through external effects only is the more robust and more flexible approach:
+we control the API for evaluating external effects and can thus include access to a resource registry, which can then contain values that are either per-StageGraph or even shared between a selection of StageGraphs.
```
abailly, line 63: That makes sense, and it seems to me this has always been a key assumption of this approach. I think there are cases, eg. very localised effects like writing/reading temporary files or accessing "static" well-known resources where effect modelling through `ExternalEffect` would be overkill and can be avoided.

It seems to me an unfortunate but necessary side-effect of the heavy reliance on async/await to handle concurrency (or is it parallelism?) in Rust, and the opacity of the transformations done by the compiler. As mentioned earlier, the alternative would be to force programmers to write everything as explicit state-machines which might be quite annoying as code size grows.

abailly: An example of `ExternalEffect` implementation can be found [here](https://github.com/pragma-org/amaru/blob/c3fa143ee7a84c6316853569730a949d6fcb7c4a/crates/amaru-consensus/src/consensus/validate_header.rs#L149)

crates/pure-stage/README.md
```diff
@@ -40,21 +40,34 @@ The wiring function ensures that message types do match, but it won't prevent co
...
+
+The alternative is to move the resources into the stage state, which means that serialization records a brief declaration and deserialization recreates the resource based on that declaration (e.g. a database path and snapshot name).
+One important caveat is that stage states are deserialized independently for each stage, meaning that for example a database connection is maintained once per process in shared global mutable state or that each stage has a separate database connection.
+While the latter may work for some database drivers, it is not an option for resources that are intended to be shared between stages.
+A possible solution to this problem would be to add a per-StageGraph resource registry that deserialization code accesses using some kind of typed key objects.
+This means that deserialization needs to be parameterized, which is not foreseen in serde and thus implies using ThreadLocal or similar hacks.
+
+This reasoning leads me to the conclusion that moving resources out of the stages and manipulating them through external effects only is the more robust and more flexible approach:
```
KtorZ, line 62: Now I am a little scared by what you mean here and what the implication are from an implementation perspective. I do not like the thought of abstracting every database call behind "an effect". Effect systems have *never* brought anything but trouble to me :)

crates/pure-stage/README.md
rkuhn: That effect has now been adapted to the new resource handling scheme.

rkuhn (commented):
@abailly @stevana This completes the move of `validate_header` to pure_stage in the sense that resource handling is done as it should be done across the whole consensus stage graph.

abailly:
@rkuhn we should be good to go with this one, shan't we? Just need conflicts resolution

crates/amaru/src/stages/mod.rs
```diff
@@ -192,6 +192,13 @@ pub fn bootstrap(
     let (network_output, validate_header_input) =
         build_stage_graph(global_parameters, consensus, &mut network);
 
+    network
+        .resources()
+        .put::<ValidateHeaderResourceStore>(chain_store_ref);
+    network
+        .resources()
+        .put::<ValidateHeaderResourceParameters>(global_parameters.clone());
+
```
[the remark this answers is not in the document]
rkuhn: Please leave the check of whether something compiles to the actual compiler: CI proves that you’re wrong on this one.

PR #368, opened 2025-08-07
feat: nicer debug representation for Point

abailly:
We use the same representation for Debug and Display as there's no sense in displaying an array of bytes in Debug mode.

## Summary by CodeRabbit

* **New Features**
  * Introduced constants for header hash size and a default origin hash value.

* **Bug Fixes**
  * Improved debug output formatting for points to enhance readability.

* **Tests**
  * Expanded and restructured tests with property-based and parameterized tests for points.

### DIFF
--- a/crates/amaru-kernel/src/lib.rs
+++ b/crates/amaru-kernel/src/lib.rs
@@ -132,8 +132,9 @@ pub mod serde_utils;
 #[cfg(any(test, feature = "test-utils"))]
 pub mod tests {
     pub use crate::{
-        anchor::tests::*, ballot::tests::*, certificate_pointer::tests::*, pool_params::tests::*,
-        proposal_id::tests::*, proposal_pointer::tests::*, transaction_pointer::tests::*,
+        anchor::tests::*, ballot::tests::*, certificate_pointer::tests::*, point::tests::*,
+        pool_params::tests::*, proposal_id::tests::*, proposal_pointer::tests::*,
+        transaction_pointer::tests::*,
     };
 }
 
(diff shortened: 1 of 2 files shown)


crates/amaru-kernel/src/lib.rs
```diff
@@ -243,7 +243,16 @@ impl Display for Point {
...
 
-impl From<&Point> for Hash<32> {
+impl fmt::Debug for Point {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        Display::fmt(self, f)
+    }
+}
+
+pub const HEADER_HASH_SIZE: usize = 32;
+pub const ORIGIN_HASH: Hash<HEADER_HASH_SIZE> = Hash::new([0; HEADER_HASH_SIZE]);
```
KtorZ: Meh.

abailly: Could you elaborate on? My mind reading skills are somewhat lacking 😅. This should actually be used in `Display` implementation though

abailly: Anyhow, it was unused so I removed it

KtorZ: I don't think it makes much sense to debug-display the origin point as a all-zeroes hash... I also don't think Debug should equals Display generally speaking 😅 ... 

So, since I moved a few things from the gigantic amaru-kernel/lib.rs yesterday into their own modules, including points; I felt obliged to rebase your PR. And doing so, I've adjusted the Debug implementation to, I hope, find a good middleground.

abailly:
I honestly fail to understand why `Debug` should not be identical to `Display` esp. for such a well-known and simple structure as a `Point`. I agree representing `Origin` with all zeros is debatable though 🤷

abailly:
That said, I think you are definitely right in emphasising that `Debug` representations should be parseable valid Rust code, we'll need to revisit some implementation in the light of this good heuristic  cc @etorreborre

PR #369, opened 2025-08-08
Create EDR 012-interaction-between-consensus-and-ledger.md

rkuhn:
(will fix the commit later, was created using the github UI)


engineering-decision-records/012-interaction-between-consensus-and-ledger.md
```diff
@@ -0,0 +1,42 @@
...
+
+On the other hand, the main complexity of the ledger lies in its state management so that it can efficiently support the various required calculations for performing the business function of the Cardano blockchain.
+This is implemented using a combination of in-memory data structures and sophisticated database usage, none of which play nicely with the notion of encapsulating each and every interaction with persistent mutable state as a [`pure_stage::ExternalEffect`](https://docs.rs/pure-stage/latest/pure_stage/trait.ExternalEffect.html).
+
+The interaction between consensus and the ledger fundamentally has only two parts, with a third one added merely incidentally (which should be fixed):
+
+- whenever the chain selection finds a new longest chain candidate, it must pass it to the ledger for validation, which may reject the result for a variety of reasons (some indicating malice of a peer, some not)
+- at each epoch boundary, the ledger emits a new set of per-stakepool information (including stake, addresses, VRFs) that will need to be used by the consensus for header validation; this may include protocol version updates that may change consensus rules
```
KtorZ: So, this is maybe worth splitting in two separate points here; because there are really two pieces of information:

1. The stake distribution, which is computed DURING the epoch and could emitted anywhere between the `3*k/f`th slot and last slot of the epoch; Right now, this computation still happens synchronously so the stake distribution is available precisely after the first block that is applied beyond this boundary.

2. At the epoch boundary, pools may retire. 

Either way, the consensus looks at these pieces with a delay of 2 epochs.

engineering-decision-records/012-interaction-between-consensus-and-ledger.md
```diff
@@ -0,0 +1,42 @@
...
+- incidentally, the ledger also manages the generation of signing keys of the node — these should actually be managed by the consensus
```
KtorZ:
```suggestion
- incidentally, the ledger also manages the evolutions of operational certificates of the pool — these should actually be managed by the consensus
```

engineering-decision-records/012-interaction-between-consensus-and-ledger.md
rkuhn: I’ll make sure to add it, but I’d like to keep it as one bullet point because this information is communicated together — the purpose here is an interface description.

PR #372, opened 2025-08-09
refactor: store headers in a tree for chain selection

abailly:
This PR refactors the chain selection code by changing the data structure used to store chain fragments coming from peers.

Instead of a map of peers and their corresponding chain `Peer -> [Header]`, we use a tree of headers (backed-up by the [indextree](https://crates.io/crates/indextree) crate). In addition to the tree, we use a map `Peer -> NodeId` to register the position of each peer inside the tree (`NodeId` is a unique node identifier for each node generated by `indextree`).

The new code implements all the previous nominal and error cases. Notably:

 - Best chain selection.
 - Truncating the best chain when it becomes too long.
 - Cleaning up the memory when some headers or peers don't need to be tracked anymore.

The `amaru-consensus` README has also been extended to illustrate a bit better the chain selection algorithm.

This works puts us on par with the previous implementation and the future steps are:

 1. Load the existing headers from the `ChainDb` store in order to initialize the `HeadersTree`.
 3. Adapt the random tree generation to `proptest` in order to write some properties simulating the activity of several peers.
 4. Use block trees to benchmark chain selection stage
 5. Solve some specific cases listed in the `README` (at the end).
 6. Run more integration / system tests.
 7. Decide and implement a logging strategy to debug issues in production (for example by dumping the tree to disk).

## Summary by CodeRabbit

* **New Features**
  * Added a generic Tip type, new consensus error variants for unknown points and invalid header-parent details, and a default fragment-length constant.

* **Refactor**
  * Replaced internal chain-selection with an arena-backed, peer-aware HeadersTree improving fork handling, pruning, peer tracking, and chain-sync orchestration.

* **Documentation**
  * Added a comprehensive chain-selection guide covering forward/rollback/fork semantics.

* **Chores**
  * Added an arena/tree dependency and formatting cleanup.


crates/amaru-kernel/src/lib.rs
```diff
@@ -229,6 +229,14 @@ impl Point {
             Point::Specific(slot, _) => Slot::from(*slot),
         }
     }
+
+    pub fn hash(&self) -> Hash<HEADER_HASH_SIZE> {
```
KtorZ: Why remove the `From` instance in favor of a direct method?

crates/test-macros/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "test-macros"
```
KtorZ: Note that we already have: https://github.com/pragma-org/amaru/blob/main/crates/amaru-kernel/src/macros.rs

crates/amaru-kernel/src/lib.rs
abailly: I don't think it's removed, it's just that using `From` instance can be cumbersome requiring type annotation whereby using `.hash()` directly gives us what we want. Then `From` instance is defined in terms of `.hash()` method.

KtorZ: Ok my bad, the diff reads a little weird.

crates/test-macros/Cargo.toml
abailly: ping @etorreborre

crates/amaru-consensus/src/consensus/headers_tree.rs
```diff
@@ -0,0 +1,1206 @@
...
+            .collect();
+        chain.reverse();
+        chain
+    }
+
+    /// Return the headers tree size in terms of how many headers are being tracked.
+    /// This is used to check the garbage collection aspect of this data structure.
+    fn size(&self) -> usize {
+        self.arena.count()
+    }
```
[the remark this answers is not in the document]
abailly: Seems like you're referring to either a different library or an outdated version. As of version 4.7.4, indextree.Arena does have a [count()](https://docs.rs/indextree/latest/indextree/struct.Arena.html#method.count) method.

crates/test-macros/Cargo.toml
etorreborre: @KtorZ as far as I understand it is not possible to mix `proc-macros` with `macro-export` functions, so it has to be a separate crate.

Cargo.toml
```diff
@@ -85,6 +86,7 @@ progress-bar = { path = "crates/progress-bar" }
 pure-stage = { path = "crates/pure-stage" }
 slot-arithmetic = { path = "crates/slot-arithmetic" }
 tracing-json = { path = "crates/tracing-json" }
+test-macros = { path = "crates/test-macros" }
```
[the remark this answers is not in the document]
etorreborre: This is the pattern used for other similar dependencies like `rand` or `tempfile`.

crates/amaru-consensus/src/consensus/headers_tree.rs
```diff
@@ -0,0 +1,1206 @@
...
+        ancestors1.reverse();
+        ancestors2.reverse();
+
+        ancestors1
+            .into_iter()
+            .zip(ancestors2)
+            .take_while(|(n1, n2)| n1 == n2)
+            .last()
+            .map(|ns| ns.0).unwrap_or_else(|| panic!("by construction a tree must always have the same root for all chains. Found none for {node_id1} and {node_id2}"))
+    }
```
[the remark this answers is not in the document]
etorreborre: `node_id1` and `node_id2` are supplied as parameters of the `find_intersection_node_id` function.

crates/test-macros/src/lib.rs
```diff
@@ -0,0 +1,112 @@
...
+            use std::panic;
+
+            let original_hook = panic::take_hook();
+            panic::set_hook(Box::new(|_| {})); // disable panic output
+
+            let result = panic::catch_unwind(panic::AssertUnwindSafe(|| #fn_block ));
+
+            panic::set_hook(original_hook);
+
+            match result {
```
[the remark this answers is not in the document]
etorreborre: I removed the macro altogether.

crates/test-macros/Cargo.toml
etorreborre: I eventually the macro altogether because it is not possible to implement it reliably. The issue it was solving, spurious panic output in tests when running in an IDE, will have to be fixed at the IDE level.

crates/amaru-consensus/src/consensus/headers_tree.rs
```diff
@@ -0,0 +1,1206 @@
...
+                if let Some(node_id) = self.arena.get_node_id(node) {
+                    self.peers.insert(peer.clone(), node_id);
+                    if self.best_peer.is_none() {
+                        self.best_peer = Some(peer.clone())
+                    }
+                    return Ok(());
+                }
+            }
+        }
+        Err(ConsensusError::UnknownPoint(point.clone()))
```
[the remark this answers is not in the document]
etorreborre: The `Arena::get_node_id` function definitely exists.

crates/amaru-consensus/src/consensus/headers_tree.rs
```diff
@@ -0,0 +1,1206 @@
...
+    /// Return true if the header has already been added to the arena (and not rolled-back)
+    fn header_exists(&self, header: &H) -> bool {
+        let all_hashes: Vec<_> = self
+            .arena
+            .iter()
+            .filter(|n| !n.is_removed())
+            .map(|n| n.get().hash())
+            .collect();
+        all_hashes.contains(&header.hash())
+    }
```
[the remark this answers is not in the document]
etorreborre: I've modified this code to a short-circuiting function that should be fast enough for our purpose.

etorreborre:
@CodeRabbit review

etorreborre:
Let's try a full review: @coderabbitai full review.

simulation/amaru-sim/src/simulator/mod.rs
```diff
@@ -409,31 +404,20 @@ fn make_chain_selector(
...
-) -> &'a mut ChainSelectorBuilder<Header> {
-    match tip {
-        Origin => builder,
+) -> Arc<Mutex<HeadersTree<Header>>> {
+    let root = match tip {
+        Origin => None,
         Specific(..) => match chain_store.load_header(&Hash::from(&tip)) {
             None => panic!("Tip {:?} not found in chain store", tip),
-            Some(header) => builder.set_tip(&header),
+            Some(header) => Some(header),
```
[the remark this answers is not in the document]
etorreborre: This code is only used for simulation tests. It's ok.

crates/amaru-consensus/src/consensus/headers_tree.rs
```diff
@@ -0,0 +1,1469 @@
...
+                    for nid in to_remove {
+                        nid.remove_subtree(&mut self.arena);
+                    }
+                    root.remove(&mut self.arena);
+                    self.remove_dangling_peers();
+                }
+                _ => panic!("This chain selection tree is configured with a maximum chain length lower than 2 ({}), which does not make sense", self.max_length),
+            }
+        }
+    }
```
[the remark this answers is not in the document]
etorreborre: done.

abailly (commented):
Do we still need the `chain_selection` module?

etorreborre:
@abailly the data types in there can certainly go to `select_chain.rs` now. I suggest to do it in the next PR now that everything is green, unless there's a more pressing comment and then I'll do both.

crates/amaru-consensus/src/consensus/headers_tree.rs
```diff
@@ -0,0 +1,1475 @@
...
+/// This data type stores chains as a tree of headers.
+/// It also keeps track of what is the latest tip for each peer.
+///
+/// The main function of this data type is to be able to always return the best chain for the current
+/// tree of headers.
+///
+#[allow(dead_code)]
+pub struct HeadersTree<H> {
+    /// The arena maintains a list of headers and their parent/child relationship.
+    /// TODO: rename to `tree`, `arena` is very technical
```
KtorZ: I feel like `arena` is actually better; because it hints at the allocation strategy underneath. Even more so; the indextree is NOT implemented as a tree; but as a vector. So I would find calling this "tree" more confusing than "arena" -- which is relatively common in Rust lands anyway?

crates/amaru-consensus/src/consensus/headers_tree.rs
```diff
@@ -0,0 +1,1475 @@
...
+        // Create a new arena
+
+        // We estimate that a good starting point for the arena size is equivalent to having 2
+        // disjoint (except at the root) chains.
+        let capacity = max_length * 2;
+        let mut arena: Arena<Tip<H>> = Arena::with_capacity(capacity);
+        let root_node_id = if let Some(header) = root {
+            arena.new_node(Tip::Hdr(header.clone()))
+        } else {
+            arena.new_node(Tip::Genesis)
```
KtorZ, line 104: Just a general thought triggered by this clone, but, given everything happening around tips, points and headers in the consensus; it may be worth considering storing those as `Rc` or `Arc` to reduce re-allocations needed across all the different algorithms and data-structures manipulating headers?

crates/amaru-consensus/src/consensus/headers_tree.rs
```diff
@@ -0,0 +1,1475 @@
...
+        let peer = Peer::new("alice");
+        let header = generate_headers_chain(1)[0];
+        tree.initialize_peer(&peer, &Point::Origin.hash()).unwrap();
+        tree.select_roll_forward(&peer, header).unwrap();
+        assert_eq!(
+            tree.best_chain_tip(),
+            Some(&header),
+            "there is not best chain for an empty tree yet"
+        );
+    }
```
KtorZ: Then why are we asserting it returns `Some` 😅 ?

crates/amaru-consensus/src/consensus/headers_tree.rs
etorreborre: I agree, I removed the TODO.

crates/amaru-consensus/src/consensus/headers_tree.rs
etorreborre: Yes the message is confusing, I fixed it.

crates/amaru-consensus/src/consensus/headers_tree.rs
etorreborre: Yes, that could be a good thing. One nice side to such a change would be that `HeaderTree<T>` wouldn't have to change a bit :-).

crates/amaru-consensus/src/consensus/mod.rs
```diff
@@ -19,12 +19,13 @@ use amaru_kernel::{peer::Peer, Header, Point};
...
 use tracing::Span;
 
-pub mod chain_selection;
+pub mod headers_tree;
 pub mod receive_header;
 pub mod select_chain;
 pub mod store;
 pub mod store_block;
 pub mod store_header;
+pub mod tip;
```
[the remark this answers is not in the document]
etorreborre: fixed.

crates/amaru-consensus/src/consensus/select_chain.rs
```diff
@@ -12,37 +12,32 @@
...
-        max_fragment_length: DEFAULT_MAXIMUM_FRAGMENT_LENGTH,
-    }))
+/// FIXME: a default chain selector will not work as it is since a headers tree needs to be
+/// initialized before being used.
+fn default_chain_selector() -> Arc<Mutex<HeadersTree<Header>>> {
+    Arc::new(Mutex::new(HeadersTree::new(
+        DEFAULT_MAXIMUM_FRAGMENT_LENGTH,
+        &None,
+    )))
 }
```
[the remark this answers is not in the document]
etorreborre: We'll tackle the FIXME in a subsequent PR.

abailly:
Thanks a lot for the review @KtorZ. Although it seems you did not approve formally the PR I feel that given the amount of work done and the fact amaru still works as before it's a good first step and we should merge. Just revert tomorrow morning if you think more work should be done :)

PR #377, opened 2025-08-14
doc: adjust EDR-03 (git etiquette)

abailly:
There were decisions that obviously were not applied:
* We don't practice TBD as all work first go through PRs which have to be reviewed in parallel with other PRs and possibly live for several days
* We don't maintain a linear history on `main` branch
* We don't use squash merges
* We require _both_ GPG signing and `Signed-off-by`


abailly:
About TBD:

wolf31o2 (approved):

PR #383, opened 2025-08-19
fix: do not filter out pure_stage traces

jeluard:
Make sure `pure_stage` traces are not filtered out. Fixes regression with `demo` script and e2e tests.

## Summary by CodeRabbit

- Chores
  - Updated default logging configuration to surface info-level messages for the Pure Stage component, improving out-of-the-box observability.
  - Users may notice more informative logs from this component during normal operation, aiding monitoring and troubleshooting.
  - Existing environment variable overrides for logging continue to work as before; no other behavior or control-flow changes.

### DIFF
--- a/crates/amaru/src/bin/amaru/observability.rs
+++ b/crates/amaru/src/bin/amaru/observability.rs
@@ -290,7 +290,7 @@ fn default_filter(var: &str, default: &str) -> EnvFilter {
     // which is a not so nice side-effects of the tracing library.
     EnvFilter::builder()
         .parse(format!(
-            "none,gasket=error,{}",
+            "none,pure_stage=info,gasket=error,{}",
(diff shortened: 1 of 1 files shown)


KtorZ (approved):

crates/amaru/src/bin/amaru/observability.rs
```diff
@@ -290,7 +290,7 @@ fn default_filter(var: &str, default: &str) -> EnvFilter {
     // which is a not so nice side-effects of the tracing library.
     EnvFilter::builder()
         .parse(format!(
-            "none,gasket=error,{}",
+            "none,pure_stage=info,gasket=error,{}",
             env::var(var).ok().as_deref().unwrap_or(default)
```
rkuhn, line 294: This whole block of logic should be removed: use RUST_LOG if present and a specific fallback filter when not.

PR #384, opened 2025-08-19
pure-stage error handling

rkuhn:
This PR improves several small things:

- pure_stage no longer handles errors from stages, only termination (which is an explicit effect)
- as soon as one stage terminates, the whole graph terminates
- graph termination can be observed/awaited from the outside
- the gasket adapter now uses this termination signal to trigger the exit hook CancellationToken (without that Amaru wouldn’t terminate in case of a header validation failure, just sit there and do nothing)
- any panic in Amaru will now actually terminate the process by calling `exit(1)` (previously it just sat there doing nothing)
- AMARU_LOG is no longer embezzled with a leading `none,gasket=error,` because that doesn’t really do much in my experiments
- default logging has been changed to `warn,amaru=debug`, which yields the same terminal output as before (in my testing) but allows external overrides with less surprises

Concerning the business logic, the `validate_header` stage now no longer handles validation errors by failing, instead it sends those errors to a new `errors_stage` that will tear down the system after logging the error. Eventually, this stage will instruct the upstream network handling to drop the peer.

This PR still has some compilation errors and clippy remarks to resolve and it needs to be rebased, hence the draft status. Please do start the review already because the actual changes are there (and you’ll probably have comments).

## Summary by CodeRabbit

- New Features
  - Graceful shutdown: stages and runtime expose termination signals; bootstrap accepts a cancellation token for clean exits.
  - Validation pipeline: improved header validation that logs invalid headers and emits upstream actions (kick/ban/log).

- Refactor
  - Stage and call APIs simplified to return plain values (no Result); termination flows replaced error-paths.

- Tests
  - Updated tests for new return semantics and termination-aware call patterns.

- Chores
  - Added equality derives and small public API exports.

### DIFF
--- a/crates/amaru-consensus/src/consensus/mod.rs
+++ b/crates/amaru-consensus/src/consensus/mod.rs
@@ -41,11 +41,24 @@ pub fn build_stage_graph(
 ) {
     let validate_header_stage = network.stage("validate_header", validate_header::stage);
 
+    let errors_stage = network.stage("errors", async |_, (peer, error), eff| {
+        tracing::error!(%peer, %error, "invalid header");
+        // termination here will tear down the entire stage graph
+        eff.terminate().await
+    });
+
     let (network_output_ref, network_output) = network.output("network_output", 50);
+
+    let errors_stage = network.wire_up(errors_stage, ());
 
     let validate_header_stage = network.wire_up(
         validate_header_stage,
-        (consensus, global_parameters.clone(), network_output_ref),
+        (
+            consensus,
+            global_parameters.clone(),
+            network_output_ref,
+            errors_stage.without_state(),
+        ),
     );
 
     let validate_header_input = network.input(&validate_header_stage);
@@ -113,6 +126,15 @@ pub enum DecodedChainSyncEvent {
         #[serde(skip, default = "Span::none")]
         span: Span,
     },
+}
+
+impl DecodedChainSyncEvent {
+    pub fn peer(&self) -> Peer {
+        match self {
+            DecodedChainSyncEvent::RollForward { peer, .. } => peer.clone(),
+            DecodedChainSyncEvent::Rollback { peer, .. } => peer.clone(),
+        }
+    }
 }
 
 impl fmt::Debug for DecodedChainSyncEvent {
--- a/crates/amaru-consensus/src/consensus/mod.rs
+++ b/crates/amaru-consensus/src/consensus/mod.rs
@@ -14,9 +14,9 @@
 
 use std::fmt;
 
-use crate::is_header::IsHeader;
+use crate::{consensus::validate_header::ValidationFailed, is_header::IsHeader};
 use amaru_kernel::{peer::Peer, protocol_parameters::GlobalParameters, Header, Point};
-use pure_stage::StageGraph;
+use pure_stage::{StageGraph, StageRef, Void};
 use serde::{Deserialize, Serialize};
 use tracing::Span;
 
@@ -35,19 +35,23 @@ pub fn build_stage_graph(
     global_parameters: &GlobalParameters,
     consensus: validate_header::ValidateHeader,
     network: &mut impl StageGraph,
-) -> (
-    pure_stage::Receiver<DecodedChainSyncEvent>,
-    pure_stage::Sender<DecodedChainSyncEvent>,
-) {
+    validation_outputs: StageRef<DecodedChainSyncEvent, Void>,
+) -> StageRef<DecodedChainSyncEvent, Void> {
     let validate_header_stage = network.stage("validate_header", validate_header::stage);
 
-    let errors_stage = network.stage("errors", async |_, (peer, error), eff| {
-        tracing::error!(%peer, %error, "invalid header");
+    let errors_stage = network.stage("errors", async |_, msg, eff| {
+        let ValidationFailed {
+            peer,
+            error,
+            action,
+        } = msg;
+        tracing::error!(%peer, %error, ?action, "invalid header");
+
+        // TODO: implement specific actions once we have an upstream network
+
         // termination here will tear down the entire stage graph
         eff.terminate().await
     });
-
-    let (network_output_ref, network_output) = network.output("network_output", 50);
 
     let errors_stage = network.wire_up(errors_stage, ());
 
@@ -56,13 +60,12 @@ pub fn build_stage_graph(
         (
             consensus,
             global_parameters.clone(),
-            network_output_ref,
+            validation_outputs,
             errors_stage.without_state(),
         ),
     );
 
-    let validate_header_input = network.input(&validate_header_stage);
-    (network_output, validate_header_input)
+    validate_header_stage.without_state()
 }
 
 #[derive(Clone)]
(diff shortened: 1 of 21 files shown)


crates/pure-stage/src/effect.rs
```diff
@@ -198,6 +200,26 @@ impl<M: SendData, S> Effects<M, S> {
...
+    ///         return eff.terminate().await;
+    ///     }
+    ///     // ...
+    ///     state
+    /// }
+    /// ```
+    pub fn terminate<T>(&self) -> BoxFuture<'static, T> {
+        airlock_effect(&self.effect, StageEffect::Terminate, |_eff| never())
+    }
 }
```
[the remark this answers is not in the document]
rkuhn: No, if this effect returns it means that there is a (serious) bug in the framework, so we want to panic.

crates/pure-stage/tests/simulation.rs
```diff
@@ -415,3 +419,52 @@
...
+        Poll::Pending
+    );
+
+    sim.run_until_blocked(); // drive effects
+
+    assert!(sim.is_terminated(), "simulation should report terminated");
+    assert_eq!(
+        term.as_mut().poll(&mut Context::from_waker(Waker::noop())),
+        Poll::Ready(())
+    );
```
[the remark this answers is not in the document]
rkuhn: nope, Rust 1.85 stabilized the `Waker::noop()` API

crates/pure-stage/tests/simulation.rs
```diff
@@ -17,13 +17,14 @@ use pure_stage::{
...
-    StageGraph, StageRef, Void,
+    StageGraph, StageGraphRunning, StageRef, Void,
 };
 use std::{
     sync::{
         atomic::{AtomicUsize, Ordering},
         Arc,
     },
+    task::{Context, Poll, Waker},
     time::Duration,
```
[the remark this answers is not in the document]
rkuhn: nope, Rust 1.85 stabilized the `Waker::noop()` API

crates/amaru-consensus/src/consensus/mod.rs
```diff
@@ -30,6 +31,43 @@ pub mod validate_header;
...
+) -> StageRef<DecodedChainSyncEvent, Void> {
+    let validate_header_stage = network.stage("validate_header", validate_header::stage);
+
+    let errors_stage = network.stage("errors", async |_, msg, eff| {
+        let ValidationFailed {
+            peer,
+            error,
+            action,
+        } = msg;
+        tracing::error!(%peer, %error, ?action, "invalid header");
```
jeluard, line 48: There could be other kind of errors than `invalid header` ?

crates/amaru-consensus/src/consensus/validate_header.rs
```diff
@@ -220,19 +206,72 @@ impl ValidateHeader {
...
-) -> Result<State, anyhow::Error> {
-    let (mut state, global, downstream) = state;
-    let result = state.validate_header(&eff, msg, &global).await?;
+) -> State {
+    let (mut state, global, downstream, errors) = state;
+    let peer = msg.peer();
+    let result = match state.validate_header(&eff, msg, &global).await {
+        Ok(result) => result,
+        Err(error) => {
+            tracing::warn!(%peer, %error, "invalid header");
```
jeluard, line 266: Is this the same error than the one logged higher up with `error!` ?

crates/amaru/src/bin/amaru/panic.rs
```diff
@@ -41,6 +43,10 @@ pub fn panic_handler() {
         };
         eprintln!("\n{}", indent(&error_message, 3));
         prev(info);
+        // Exit with a non-zero code to indicate that the process crashed
+        // otherwise it will just sit there and wait for ctrl-c without saying so.
+        std::io::stderr().flush().ok();
```
jeluard, line 48: Are you saying the current behavior is that the process gets stuck forever?

crates/amaru-consensus/src/consensus/mod.rs
abailly: I think not right now because the only stage that propagates error is the `validate_header` stage.

crates/amaru-consensus/src/consensus/validate_header.rs
```diff
@@ -220,19 +206,72 @@ impl ValidateHeader {
...
+            eff.send(
+                &errors,
+                ValidationFailed::kick_peer(peer, error.to_string()),
```
abailly, line 269: I don't think it's the job of `validate_header` stage to decide to `kick_peer`. The whole point of the `error_handling` stage is to take this decision and channel it to the right place.

crates/amaru-consensus/src/consensus/mod.rs
rkuhn: right; my preliminary reasoning was that if header validation fails, that basically implies that the header must have been invalid :-)

crates/amaru-consensus/src/consensus/validate_header.rs
rkuhn: yes, this is not set in stone and I already downgraded it to `info` level in the next PR; I think it does make sense to log something at the source, i.e. where the error was detected, as well as at the point where the error is acted upon

crates/amaru/src/bin/amaru/panic.rs
rkuhn: Exactly, that is what was happening in my tests. Not sure exactly what caused it or whether it was doing that for everyone, but this now makes termination obviously reliable.

crates/amaru-consensus/src/consensus/validate_header.rs
rkuhn: Yes, I did it this way because at that time I thought I could get away without making all those error types `SendData` (which requires PartialEq, Serialize, Deserialize). Now that I’ve added those, I’ll move the decision to the error stage (which should probably be renamed to upstream_errors).

crates/amaru-consensus/src/consensus/mod.rs
jeluard: Isn't the whole point of the error to be generic and handle different errors?

crates/amaru/src/bin/amaru/panic.rs
jeluard: You could validate that in CI by modifying `scripts/demo` to simply kill and see if it helps then. Also I'm wondering if it is a workaround to gasket issues or expected behavior to call `exit` in a `panic_handler`

rkuhn: `panic!` has a very specific meaning in well-behaved Rust projects: the program cannot continue because an invariant it depends upon has been violated. It is therefore good practice to exit immediately, any further code could be tripped up worse than the state we are in already. This is also the reason for calling the `exit` syscall and thus NOT running any Drop glue.

A graceful exit should be performed in response to a user request like SIGINT or SIGTERM. I fixed that, too, by wiring the cancelation signal to gasket.

crates/amaru-consensus/src/consensus/mod.rs
rkuhn: The error is not generic, it is an enum. With my further changes in the next PR the precise structured ConsensusError is sent to the upstream_errors stage so that the reaction can use the full information. Once we have more error sources or cases, the input message type of that stage should become an  enum wrapping several error types.

PR #393, opened 2025-08-22
fix: remove optional failover; now failure are reported again

jeluard:


.github/workflows/continuous-integration.yml
```diff
@@ -112,15 +112,6 @@ jobs:
             # Run doc tests separately as --doc can't be used with --all-targets
             cargo test --doc
           fi
-          exitcode="$?"
-          if [[ "${{ matrix.environments.optional }}" == "true" && "$exitcode" != "0" ]] ; then
```
rkuhn, line 116: Hmm, it is weird that this should change something, because `optional` isn’t set in any of the environments, so how was the first test ever true?

jeluard: True so that's not the cause of failure not being caught. Still this bit is now useless

PR #396, opened 2025-08-23
Governance ratification tweaks & refactors

KtorZ:
- :round_pushpin: **chore: internalize in forest checks on withdrawals w.r.t treasury**
    This alleviates a bit the ratification plumbing, and moves some of the pre-conditions to ratification closer to the other similar conditions. Doing so also allow to include this particular behaviour as part of the proposals forest testing, by also covering withdrawals.

- :round_pushpin: **chore: internalize cc validity check in proposals forest's compass**
    For the same reason as the withdrawals; those checks are better all done in one place. It also allows testing them better and makes the encapsulation clearer.

- :round_pushpin: **feat: lazily compute governance approvals**
    CC acceptance is by far the least expensive, and there's no need to compute the other if this one is false. Same for SPOs; which should be in general smaller than dreps.

## Summary by CodeRabbit

* **New Features**
  * Treasury-aware proposal processing prevents overdrawn withdrawals and tracks remaining balance during ratification.
  * Protocol-parameter checks (e.g., maximum committee term length) enforced when navigating proposals.
  * Proposal orphan detection exposed for richer proposal handling.
  * Streamlined acceptance flow yields clearer outcomes from committee, pool, and DRep checks.

* **Bug Fixes**
  * Large treasury withdrawal totals display safely using saturating arithmetic.
  * Treasury updates now avoid underflow and remain consistent across steps.

* **Refactor**
  * Removed per-withdrawal cumulative state and adjusted proposal APIs.

* **Tests**
  * Property tests made deterministic by seeding RNGs and updated expectations.

### DIFF
--- a/crates/amaru-ledger/src/governance/ratification.rs
+++ b/crates/amaru-ledger/src/governance/ratification.rs
@@ -366,22 +366,40 @@ impl<'distr> RatificationContext<'distr> {
         let span = tracing::Span::current();
 
         let empty = BTreeMap::new();
-
         let (dreps_votes, cc_votes, pool_votes) =
             partition_votes(self.votes.get(id).unwrap_or(&empty));
 
+        // NOTE: because ratification is an expensive operation, and something that we *may* have
+        // to replay due to rollbacks, it's important to do the least amount of work possible.
+        //
+        // If the CC doesn't approve, then the majority of actions can be considered invalid and
+        // there's no need to check for DReps nor SPO votes. The following code is thus written in
+        // a way that the next governance body is only consulted should the previous one be "no".
+
         let cc_approved = self.is_accepted_by_constitutional_committee(proposal, cc_votes);
+
         span.record("approved.committee", cc_approved);
 
-        let spos_approved =
-            self.is_accepted_by_stake_pool_operators(proposal, pool_votes, stake_distribution);
-        span.record("approved.pools", spos_approved);
+        if cc_approved {
+            let spos_approved =
+                self.is_accepted_by_stake_pool_operators(proposal, pool_votes, stake_distribution);
 
-        let dreps_approved =
-            self.is_accepted_by_delegate_representatives(proposal, dreps_votes, stake_distribution);
-        span.record("approved.dreps", dreps_approved);
+            span.record("approved.pools", spos_approved);
 
-        cc_approved && spos_approved && dreps_approved
+            if spos_approved {
+                let dreps_approved = self.is_accepted_by_delegate_representatives(
+                    proposal,
+                    dreps_votes,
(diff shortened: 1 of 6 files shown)


crates/amaru-ledger/src/governance/ratification.rs
```diff
@@ -407,22 +366,40 @@ impl<'distr> RatificationContext<'distr> {
...
+        if cc_approved {
+            let spos_approved =
+                self.is_accepted_by_stake_pool_operators(proposal, pool_votes, stake_distribution);
+
+            span.record("approved.pools", spos_approved);
 
-        let dreps_approved =
-            self.is_accepted_by_delegate_representatives(proposal, dreps_votes, stake_distribution);
-        span.record("approved.dreps", dreps_approved);
+            if spos_approved {
```
abailly, line 389: Would it not be useful to have some unit tests for this logic? Not sure how complicated this would be to set up, but having a manifestation of this "shortcut" logic would help our future selves.

crates/amaru-ledger/src/governance/ratification/proposals_forest.rs
```diff
@@ -954,13 +1104,15 @@ mod tests {
...
 
                 // Control that the enacted proposal is one of the new root.
-                prop_assert!(
-                    forest.roots().protocol_parameters.as_deref() == Some(id.as_ref()) ||
-                    forest.roots().hard_fork.as_deref() == Some(id.as_ref()) ||
-                    forest.roots().constitutional_committee.as_deref() == Some(id.as_ref()) ||
-                    forest.roots().constitution.as_deref() == Some(id.as_ref()),
-                    "none of the roots match the just enacted proposal",
-                );
+                if !matches!(proposal, ProposalEnum::Orphan(..)) {
```
abailly: this `!matches!` looks weird at first sight. Although they are not needed, perhaps parens would help here?

abailly (approved):
This stuff is really insane, can't wait for people to start exploiting loopholes :(

crates/amaru-ledger/src/governance/ratification.rs
KtorZ: I've been trying to see how to best test the "plumbing" at the top, but I've been struggling so far. So my strategy has been to make it *as thin as possible* and have the lower parts well tested.

crates/amaru-ledger/src/governance/ratification/proposals_forest.rs
KtorZ: cargo-fmt gets rid of them. I've been using helpers like `is_orphan()` in other places, I could do the same here 👍

PR #397, opened 2025-08-24
Fix: dedup withdrawals and process them in stake credential order

KtorZ:
Importantly, script credentials come *before* key credentials.


crates/amaru-ledger/tests/data/transactions/preprod/6913ffb3588cad067c518fa1020c0f1f86adcc58abd7851bc380db058941c43b/tx.cbor
KtorZ, line 1: Note: this transaction is constructed such that the key is serialized before the script credential, but as the traces show, it is processed before. 

This test fails without the fix from this PR.

yHSJ (approved):

PR #401, opened 2025-08-25
fix: introduce 'consensus' stake distribution for leader election.

KtorZ:
Not quite the active stake. Not quite the voting stake. It's a new specimen.

TODO: 
- [ ] Open an issue in intersectMBO/cardano-ledger to describe the numerator/denominator discrepancy there.

## Summary by CodeRabbit

- New Features
  - Introduced a separate "consensus stake" metric alongside rewards stake.

- Refactor
  - Replaced pool "stake" with "rewards stake" for rewards, weighting and performance calculations.
  - Epoch calculation now uses the current horizon directly.
  - Refined deposit handling to separate expiring deposits for voting vs. rewards/consensus.

- Documentation
  - Clarified voting stake to exclude pools unregistering next epoch and deposits in their last valid epoch.

- Tests
  - Updated tests and models to reflect new stake metrics and calculations.

- Chores
  - Preserved external "stake" field in serialized output, now sourced from rewards stake.

### DIFF
--- a/crates/amaru-ledger/src/summary/stake_distribution.rs
+++ b/crates/amaru-ledger/src/summary/stake_distribution.rs
@@ -49,8 +49,9 @@ pub struct StakeDistribution {
     /// Total stake, in Lovelace, delegated to registered pools
     pub active_stake: Lovelace,
 
-    /// Active stake plus deposits of ongoing proposals whose reward accounts are delegated to
-    /// active stake pools.
+    /// Active stake plus deposits of ongoing proposals whose refund accounts are delegated to
+    /// active stake pools. This stake also omits the stake of pools that are unregistering in the
+    /// following epoch, and deposits of proposals in their last epoch of validity.
     pub pools_voting_stake: Lovelace,
 
     /// Total voting stake, in Lovelace, corresponding to the total stake assigned to registered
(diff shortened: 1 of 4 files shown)


crates/amaru-ledger/src/summary/stake_distribution.rs
```diff
@@ -228,20 +231,34 @@ impl StakeDistribution {
...
+                            // NOTE: Governance deposits do not count towards the pools' active
+                            // stake used in rewards calculation. They are only counted as part of
+                            // the voting power & for consensus leader election.
                             active_stake += &account.lovelace;
-                            pool.stake += &account.lovelace;
+                            pool.rewards_stake += &account.lovelace;
+
+                            // NOTE: An astute reader might wonder why there's no total
+                            // 'pools_consensus_stake' which would correspond to the sum of all
+                            // pool's consensus stake. *Why* is hard to answer here, but *how* we
```
abailly:
```suggestion
                            // pools' consensus stake. *Why* is hard to answer here, but *how* we
```

abailly: This comment feels bit coming out of the blue as there's no indication of any `total` in sight, perhaps it requires more context?

abailly (approved):

crates/amaru-ledger/src/summary/stake_distribution.rs
KtorZ: There are two other totals being computed in the same pass; the `active_stake` and the `pools_voting_stake`. I didn't rename the former to `pools_rewards_stake`, but perhaps I should and it would make this question more apparent: why is there no `pools_consensus_stake` despite each pool having the three individual stakes `rewards_stake`, `consensus_stake` and `voting_stake`.

KtorZ: Slightly reworded the comment, to make the connections more apparent.

PR #405, opened 2025-08-26
fix: incorrect length check

jeluard:


crates/amaru-ledger/src/bootstrap.rs
```diff
@@ -1054,7 +1054,7 @@ impl<'d, C> cbor::decode::Decode<'d, C> for ConstitutionalCommitteeAuthorization
     fn decode(d: &mut cbor::Decoder<'d>, ctx: &mut C) -> Result<Self, cbor::decode::Error> {
         heterogeneous_array(d, |d, assert_len| match d.u8()? {
             0 => {
-                assert_len(1)?;
+                assert_len(2)?;
```
KtorZ, line 1057: This doesn't seem right 🤨... the next step is only decoding a single CBOR object

jeluard: 1 for the `u8` variant read, 1 for the actual object?

KtorZ: Ho, right. 🤦

PR #406, opened 2025-08-26
feat: added amaru-ledger bin

jeluard:
A dedicated `amaru-ledger` binary that can ingest blocks from a folder. Allows to process blocks `offline`.
Also provides a command to download snapshots from mithril and convert them into `tar.gz` files of `CBOR` blocks that can then be ingested via `amaru-ledger`.
This effectively allows to bootstrap `amaru` from a mithril snapshot by applying all blocks on top of a bootstrapped amaru.

Complete flow:

```shell
export AMARU_NETWORK=preprod

# compile both `amaru` and `amaru-ledger` binary
cargo build --release

# bootstrap amaru from some amaru snapshot
./target/release/amaru bootstrap

# downloads and creates local `tar.gz` of blocks
./target/release/amaru-ledger mithril
# 25min

# sync the chain from those local `tar.gz`
./target/release/amaru-ledger sync
# 2h 12min

# run amaru like usual
./target/release/amaru run --peer-address localhost:3001
```

On mainnet, `amaru-ledger mithril` takes 25min and `amaru-ledger bootstrap` takes 2h (around 200 blocks/s).

Useful for:
* testing some specific scenarios
* performing controlled profiling
* triggering scenarios e.g. epoch transition
* create state for specific blocks

Note: I am not convinced of the utility to have a separate binary. Another approach would be to introduce 2 extra commands to the `amaru` binary: `download-blocks-from-mithril` and `sync-from-local` or similar.

Blocked by:
* [x] `mithril-client` must be updated to not rely on `rug`
* [x] use a complete `consensus` pipeline: after discussing with Eric we came to the conclusion that current approach is the right one for now

## Summary by CodeRabbit

* **New Features**
  * Added Mithril snapshot synchronization with verification and progress tracking
  * Introduced amaru-ledger tool for blockchain block ingestion and synchronization

* **Chores**
  * Updated dependencies including flate2, tar, pallas-hardano, and mithril-client
  * Enhanced CI workflows with artifact uploads on test failures
  * Expanded license allowlist to include Zlib and LGPL-3.0

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -157,17 +157,32 @@ jobs:
           token: ${{ secrets.CODECOV_TOKEN }}
           files: lcov.info
 
+
+  snapshots-instance-choice:
+    name: End-to-end snapshot preparation
+    runs-on: ubuntu-latest
+    outputs:
+      runner: ${{ steps.alt-runner.outputs.runner || 'ubuntu-latest' }}
+    steps:
+      - id: alt-runner
+        if: (github.event_name == 'push' || github.event_name == 'workflow_dispatch') && github.ref == 'refs/heads/main'
+        run: echo "runner=buildjet-4vcpu-ubuntu-2204" >> $GITHUB_OUTPUT
+
   snapshots:
     name: End-to-end snapshot tests
-    runs-on: ubuntu-latest
-    if: ${{ !github.event.pull_request.draft }}
+    needs: snapshots-instance-choice
+    runs-on: ${{ needs.snapshots-instance-choice.outputs.runner }}
     strategy:
       matrix:
         network:
           - name: preprod
             magic: 1
         cardano_node_version: [10.1.4]
+
+    if: ${{ !github.event.pull_request.draft }}
+
     continue-on-error: true
+
     steps:
       - uses: actions/checkout@v4
 
@@ -185,7 +200,7 @@ jobs:
 
       - name: Restore cardano-node DB
         id: cache-cardano-node-db
-        uses: actions/cache/restore@v4
+        uses: buildjet/cache/restore@v4
         with:
           # The path should match the one used for the 'Nightly Sync' workflow.
           path: ${{ runner.temp }}/db-${{ matrix.network.name }}
--- a/.github/workflows/nightly-synchronization.yml
+++ b/.github/workflows/nightly-synchronization.yml
@@ -30,7 +30,7 @@ jobs:
         echo "value=$(/bin/date -u '+%Y%m%d-%H%M%S')" >> $GITHUB_OUTPUT
 
     - id: cache
-      uses: actions/cache@v4
+      uses: buildjet/cache@v4
       with:
         path: ${{ runner.temp }}/db-${{ matrix.network }}
         key: cardano-node-ogmios-${{ matrix.network }}-${{ steps.timestamp.outputs.value }}
@@ -46,8 +46,8 @@ jobs:
 
     # Remove old immutable chunks from the database, we don't need them and they make cache & on-disk system larger.
     - name: prune node db
-      if: steps.cache.outputs.cache-hit == 'false'
       shell: bash
       working-directory: ${{ runner.temp }}/db-${{ matrix.network }}
       run: |
+        sudo chown -R $(whoami):$(whoami) immutable
         ls immutable/*.chunk | sort | head -n -2500 | xargs -r rm -f --
--- a/simulation/amaru-sim/src/simulator/run.rs
+++ b/simulation/amaru-sim/src/simulator/run.rs
@@ -75,7 +75,7 @@ pub fn run(args: Args) {
     let spawn = |node_id: String, rng: RandStdRng| {
         let mut network = SimulationBuilder::default()
             .with_trace_buffer(trace_buffer.clone())
-            .with_mailbox_size(1000)
+            .with_mailbox_size(10000)
             .with_eval_strategy(rng);
         let (input, init_messages, output) =
             spawn_node(node_id, node_config.clone(), &mut network, &rt);


crates/amaru/src/bin/ledger/cmd/bootstrap.rs
```diff
@@ -0,0 +1,184 @@
...
+        entries_with_keys.sort_by_key(|(num, _)| num.clone());
+
+        for (point, raw_block) in entries_with_keys.iter_mut() {
+            // Do not process points already in the ledger
+            if point.slot_or_default() <= tip.slot_or_default() {
+                continue;
+            }
+            processed += 1;
+
+            let block = parse_block(raw_block)?;
```
jeluard: @abailly I'm interested in your expertise here. This part should essentially mimic what the consensus part is doing assuming all blocks are fine (coming from a trusted source). After those steps, it should be ok to run amaru as expected. Anything wrong/missing?

abailly: Not sure I understand your question @jeluard

crates/amaru/src/bin/ledger/cmd/mithril.rs
```diff
@@ -0,0 +1,392 @@
...
+    let progress_bar = indicatif::MultiProgress::new();
+    let AggregatorDetails {
+        endpoint,
+        verification_key,
+        ancillary_verification_key,
+    } = aggregator_details(network);
+    let client = ClientBuilder::aggregator(endpoint, verification_key)
+        .set_ancillary_verification_key(ancillary_verification_key.to_string())
+        .add_feedback_receiver(Arc::new(IndicatifFeedbackReceiver::new(&progress_bar)))
+        .build()?;
```
jpraynaud, line 279: This will help track the origin of the snapshot download on the aggregator:
```suggestion
    let client = ClientBuilder::aggregator(endpoint, verification_key)
        .set_ancillary_verification_key(ancillary_verification_key.to_string())
        .with_origin_tag(Some("AMARU".to_string()))
        .add_feedback_receiver(Arc::new(IndicatifFeedbackReceiver::new(&progress_bar)))
        .build()?;
```

crates/amaru/src/bin/ledger/cmd/mithril.rs
```diff
@@ -0,0 +1,392 @@
...
+        .ok_or(format!("snapshot not found {}", snapshot_list_item.hash))?;
+    let certificate = client
+        .certificate()
+        .verify_chain(&snapshot.certificate_hash)
+        .await?;
+
+    let immutable_file_range = ImmutableFileRange::From(from_chunk);
+    let download_unpack_options = DownloadUnpackOptions {
+        allow_override: true,
+        include_ancillary: false,
```
jpraynaud, line 301: So you don't need the ledger state for the bootstrap of Amaru? 

- If you need it, you would need to include ancillaries:
```suggestion
        include_ancillary: true,
```

- Otherwise, you can probably get rid of the `ancillary_verification_key` in the `AggregatorDetails` which is needed only if `include_ancillary` is set to `true`.

crates/amaru/src/bin/ledger/cmd/mithril.rs
jeluard: Makes sense!

crates/amaru/src/bin/ledger/cmd/mithril.rs
jeluard: Right, we don't need it for this first approach as we re-execute all blocks. Thanks!

crates/amaru/src/bin/ledger/cmd/mithril.rs
```diff
@@ -0,0 +1,389 @@
...
+            .iter()
+            .filter_map(|cbor| {
+                // TODO only access the necessary info and do not decode the whole block
+                let block = MultiEraBlock::decode(cbor).ok()?;
+                let header = block.header();
+                let name = format!("{}.{}.cbor", header.slot(), header.hash());
+                Some((name, cbor))
+            })
+            .collect();
+        package_blocks(&network, &map).await?;
```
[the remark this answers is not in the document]
jeluard: Good catch!

crates/amaru/src/bin/ledger/cmd/sync.rs
```diff
@@ -0,0 +1,250 @@
...
+async fn process_block(
+    chain_store: &Arc<dyn ChainStore<BlockHeader>>,
+    praos_chain_store: &PraosChainStore<BlockHeader>,
+    block_validator: &BlockValidator<RocksDB, RocksDBHistoricalStores>,
+    point: &Point,
+    raw_block: &RawBlock,
+) -> Result<(), Box<dyn std::error::Error>> {
+    let block = parse_block(raw_block)?;
+    let header: Header = block.header.unwrap().into();
+    let block_header: BlockHeader = BlockHeader::from(header);
```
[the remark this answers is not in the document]
jeluard: Not Result#unwrap, a different one.

crates/amaru-ledger/src/state.rs
```diff
@@ -93,7 +93,7 @@ where
     stable: Arc<Mutex<S>>,
 
     /// A handle to the stable store, shared across all ledger instances.
-    snapshots: HS,
+    pub snapshots: HS,
```
KtorZ: I rather have an accessor here giving a ref, rather than making any of the ledger's internal state public. Been trying hard not to expose any :p

crates/amaru/src/bin/ledger/main.rs
```diff
@@ -0,0 +1,116 @@
...
+        (observability::OpenTelemetryHandle::default(), None)
+    };
+
+    let warning_json = if args.with_json_traces {
+        observability::setup_json_traces(&mut subscriber)
+    } else {
+        None
+    };
+
+    subscriber.init();
```
KtorZ: That's perhaps worth factoring out and sharing across both bins?

crates/amaru/src/bin/ledger/main.rs
```diff
@@ -0,0 +1,116 @@
...
+    #[arg(long, value_name = "URL", env("AMARU_OTLP_SPAN_URL"), default_value_t = DEFAULT_OTLP_SPAN_URL.to_string()
+    )]
+    otlp_span_url: String,
+
+    #[arg(long, value_name = "URL", env("AMARU_OTLP_METRIC_URL"), default_value_t = DEFAULT_OTLP_METRIC_URL.to_string()
+    )]
+    otlp_metric_url: String,
+}
+
+const DEFAULT_SERVICE_NAME: &str = "amaru";
```
KtorZ: Probably wanna use a different service name here, so that traces don't get conflated in the exporter (even if both bins aren't executed at the same time).

crates/amaru/src/bin/ledger/cmd/sync.rs
```diff
@@ -0,0 +1,257 @@
...
+
+    /// Ingest blocks until (and including) the given slot.
+    /// If not provided, will ingest all available blocks.
+    #[arg(long, value_name = "SLOT", env = "AMARU_INGEST_UNTIL_SLOT")]
+    ingest_until_slot: Option<u64>,
+
+    /// Ingest at most the given number of blocks.
+    /// If not provided, will ingest all available blocks.
+    #[arg(long, value_name = "INT", env = "AMARU_INGEST_MAXIMUM_BLOCKS")]
+    ingest_maximum_blocks: Option<usize>,
```
KtorZ, line 72: Are these two options mutually exclusive? (iirc, there's a way to specify that with clap)

crates/amaru/src/bin/ledger/cmd/sync.rs
```diff
@@ -0,0 +1,257 @@
...
+            } else {
+                None
+            }
+        })
+        .collect();
+    archives.sort_by(|a, b| {
+        let a_num = a
+            .split_once(".")
+            .and_then(|(prefix, _)| prefix.parse::<u32>().ok())
+            .unwrap_or_else(|| panic!("Invalid archive name format: {}", a));
```
KtorZ: Turn into a helper function?

crates/amaru/src/bin/ledger/cmd/sync.rs
```diff
@@ -0,0 +1,257 @@
...
+        }
+    }
+
+    let duration = Instant::now().saturating_duration_since(before);
+    info!(
+        "Processed {} blocks in {} seconds ({} blocks/s)",
+        processed,
+        duration.as_secs(),
+        processed as u64 / duration.as_secs().max(1)
+    );
```
KtorZ, line 259: Use fields!

crates/amaru/src/bin/ledger/cmd/sync.rs
```diff
@@ -0,0 +1,257 @@
...
+            process_block(
+                &chain_store,
+                &praos_chain_store,
+                &block_validator,
+                point,
+                raw_block,
+            )
+            .await?;
+
+            processed += 1;
```
KtorZ, line 242: Perhaps, consider using: https://github.com/pragma-org/amaru/blob/main/crates/amaru-progress-bar/src/lib.rs#L20-L23

crates/amaru/src/bin/ledger/cmd/mod.rs
```diff
@@ -0,0 +1,39 @@
...
+
+use amaru_kernel::{EraHistory, network::NetworkName, protocol_parameters::GlobalParameters};
+use amaru_ledger::block_validator::BlockValidator;
+use amaru_stores::rocksdb::{RocksDB, RocksDBHistoricalStores, RocksDbConfig};
+use std::{error::Error, path::PathBuf};
+
+pub(crate) mod mithril;
+pub(crate) mod sync;
+
+pub fn new_block_validator(
```
KtorZ, line 23: Not sure why this lives here?

crates/amaru/src/bin/ledger/cmd/mithril.rs
```diff
@@ -0,0 +1,428 @@
...
+        .add_feedback_receiver(Arc::new(IndicatifFeedbackReceiver::new(&progress_bar)))
+        .build()?;
+    let database_client = client.cardano_database_v2();
+    let snapshots = database_client.list().await?;
+    let snapshot_list_item = snapshots.first().ok_or("no snapshot found")?;
+
+    info!(
+        "Downloading and verifying mithril snapshot {}",
+        snapshot_list_item.hash
+    );
```
KtorZ, line 287: Use fields!

crates/amaru/src/bin/ledger/cmd/mithril.rs
```diff
@@ -0,0 +1,428 @@
...
+            ancillary_verification_key: "5b3138392c3139322c3231362c3135302c3131342c3231362c3233372c3231302c34352c31382c32312c3139362c3230382c3234362c3134362c322c3235322c3234332c3235312c3139372c32382c3135372c3230342c3134352c33302c31342c3232382c3136382c3132392c38332c3133362c33365d",
+        },
+        NetworkName::Preview => AggregatorDetails {
+            endpoint: "https://aggregator.testing-preview.api.mithril.network/aggregator",
+            verification_key: "5b3132372c37332c3132342c3136312c362c3133372c3133312c3231332c3230372c3131372c3139382c38352c3137362c3139392c3136322c3234312c36382c3132332c3131392c3134352c31332c3233322c3234332c34392c3232392c322c3234392c3230352c3230352c33392c3233352c34345d",
+            ancillary_verification_key: "5b3138392c3139322c3231362c3135302c3131342c3231362c3233372c3231302c34352c31382c32312c3139362c3230382c3234362c3134362c322c3235322c3234332c3235312c3139372c32382c3135372c3230342c3134352c33302c31342c3232382c3136382c3132392c38332c3133362c33365d",
+        },
+        NetworkName::Testnet(_) => panic!("Mithril not supported on testnets"),
+    }
+}
```
KtorZ, line 262: Shouldn't that be command-line options with _sensible defaults_ rather than being hard-coded here?

crates/amaru/src/bin/ledger/cmd/mithril.rs
```diff
@@ -0,0 +1,428 @@
...
+
+    // Determine the chunk to start from
+    let latest_chunk = get_latest_chunk(&immutable_dir)?;
+    let from_chunk =
+        latest_chunk.unwrap_or(infer_chunk_from_slot(tip.slot_or_default().into()) - 1);
+
+    info!(
+        "Downloading mithril immutable starting chunk {} (tip {})",
+        from_chunk, tip
+    );
```
KtorZ, line 397: 😢

crates/amaru/src/bin/ledger/cmd/mithril.rs
jeluard: Unclear as I don't think that it can meaningfully be changed?

crates/amaru/src/bin/ledger/cmd/sync.rs
jeluard: I tried, it was not trivial. Will revisit.

crates/amaru/src/bin/ledger/cmd/sync.rs
jeluard: I guess you can combine them? Unclear what we would get from making them exclusive, there might be some use for both at the same time.

crates/amaru/src/bin/ledger/cmd/mod.rs
jeluard: Where would you see it? Somewhere outside of `cmd` ?

KtorZ: Probably yes? It doesn't really seem cmd-specific; although I get the point of not finding a better place...

jeluard: I have some other PR involving `API` changes. I'd rather tackle this there as I'm starting to confuse both.

crates/amaru/src/bin/ledger/cmd/sync.rs
jeluard: I would actually rather tackle this in the API branch too as I suspect some other API changes would be helpful so that API users can get a hold of download progress.

crates/amaru/src/bin/ledger/cmd/mithril.rs
KtorZ: There's probably only one aggregator being deployed; but I believe the intent is to have multiple of those.

jeluard: @jpraynaud Maybe you have some insights here? What's the best practice to manage aggregator details?

PR #408, opened 2025-08-26
fix: store and use delegation and pool 1st registration dates

KtorZ:
This is necessary to prevent pools that unregister and re-register to preserve their delegators. The other option would be to iterate over all delegators of a pool at the moment of the retirement; yet, these can be arbitrarily large. So, like for DRep, I opted for filtering them out when re-building the stake distribution based on those certificates.

With this fix, we can now finally synchronize the preview network back to the tip.


crates/amaru-ledger/src/summary/stake_distribution.rs
```diff
@@ -173,12 +126,67 @@ impl StakeDistribution {
...
+                    AccountState {
+                        lovelace: account.rewards,
+                        pool: account.pool.and_then(|(pool, since)| {
+                            let PoolState { registered_at, .. } = pools.get(&pool)?;
+                            if &since >= registered_at {
+                                Some(pool)
+                            } else {
+                                None
+                            }
+                        }),
```
KtorZ, line 147: This is the interesting change ☝️, but I had to move the account processing after the pools' one so we could perform the lookup (`pool.get`). So the diff looks a bit poor.

jeluard (approved):

PR #409, opened 2025-08-27
Auto prune ledger db snapshots

KtorZ:
- :round_pushpin: **fix: use protocol parameters from loaded snapshot when recovering stake distr.**

- :round_pushpin: **feat: prune ledger db epoch snapshots on each epoch transitions.**
    We are reaching moments where those snapshots are numerous, and the ledger.db gets real big. On mainnet, in particular, it has become a problem (~2GB per snapshot). The next commit will push the definition of this parameter up the user via the command-line.


crates/amaru/src/bin/amaru/cmd/daemon.rs
```diff
@@ -123,5 +136,6 @@ fn parse_args(args: Args) -> Result<Config, Box<dyn std::error::Error>> {
         network_magic: args.network.to_network_magic(),
         listen_address: args.listen_address,
         max_downstream_peers: args.max_downstream_peers,
+        max_extra_ledger_snapshots: args.max_extra_ledger_snapshots,
```
jeluard, line 139: Do we want to fail if `max_extra_ledger_snapshots`is bellow `MIN_LEDGER_SNAPSHOTS` ?

jeluard (approved):

PR #410, opened 2025-08-27
Peerless bootstrap

pgrange:
This PR allows one to bootstrap an Amaru node without needing any cardano peer.

To do so, we store the needed chain headers in the data directory of this repository, as they are really small files. That way, they're next to all the other metadata that we currently store. They need to be kept in sync with the rest of the data directory.

`data/README.md` explains how to generate these files to keep them in sync (`make fetch-chain-headers`).

The bootstrap process itself has been changed to pick chain headers information from this data directory instead of trying to connect to a peer.

Now with this new fetch-chain-header command we have some duplication with the existing import-headers command. So the `import-headers` command is deleted as it does not seem to be of use anymore.

## Summary by CodeRabbit

- New Features
  - Added a CLI subcommand to fetch chain headers and save them as CBOR files.
  - Introduced a Makefile target (fetch-chain-headers) to generate headers into data/<network>/headers.

- Improvements
  - Simplified bootstrap and import workflows to use a config directory instead of a peer address.
  - Import headers now reads locally stored headers, improving setup consistency.

- Documentation
  - Updated data/README with a new step to dump chain headers and where they are stored.


crates/amaru/src/bin/amaru/cmd/bootstrap.rs
```diff
@@ -116,34 +107,41 @@ pub async fn run(args: Args) -> Result<(), Box<dyn Error>> {
...
+                    .inspect_err(|reason| tracing::error!(file = %path.display(), reason = %reason, "Failed to open header file"))
+                    .map_err(|_| WorkerError::Panic)?;
+            let mut cbor_data = Vec::new();
+            file.read_to_end(&mut cbor_data).await
+                    .inspect_err(|reason| tracing::error!(file = %path.display(), reason = %reason, "Failed to read header file"))
+                    .map_err(|_| WorkerError::Panic)?;
+            let header_from_file: Header = from_cbor(&cbor_data).unwrap();
+            let hash = header_from_file.hash();
+            db.store_header(&hash, &header_from_file)
+                .map_err(|_| WorkerError::Panic)?;
```
jeluard: Why WorkerError?

crates/amaru/src/bin/amaru/cmd/fetch_chain_headers.rs
```diff
@@ -0,0 +1,265 @@
...
+    network: NetworkName,
+
+    /// Path to directory containing per-network bootstrap configuration files.
+    ///
+    /// This path will be used as a prefix to resolve per-network configuration files
+    /// needed for bootstrapping and to store fetched chain headers. Given a source
+    /// directory `data`, and a network name of `preview`, the expected layout for
+    /// configuration files would be:
+    ///
+    /// * `data/preview/snapshots.json`: a list of `Snapshot` vaalues,
```
jeluard: typo

crates/amaru/src/bin/amaru/cmd/fetch_chain_headers.rs
```diff
@@ -0,0 +1,265 @@
...
+    config_dir: &Path,
+) -> Result<(), Box<dyn Error>> {
+    let headers_file: PathBuf = config_dir.join("headers.json");
+    let content = tokio::fs::read_to_string(headers_file).await?;
+    let points: Vec<String> = serde_json::from_str(&content)?;
+    let mut initial_headers = Vec::new();
+    for point in points {
+        match Point::try_from(point.as_str()) {
+            Ok(point) => initial_headers.push(point),
+            Err(e) => tracing::warn!("Ignoring malformed header point '{}': {}", point, e),
```
jeluard: Probably should crash here?

crates/amaru/src/bin/amaru/cmd/fetch_chain_headers.rs
```diff
@@ -0,0 +1,265 @@
...
+// limitations under the License.
+
+use crate::cmd::{DEFAULT_NETWORK, connect_to_peer};
+use amaru::stages::pull;
+use amaru_consensus::IsHeader;
+use amaru_kernel::{Header, Point, from_cbor, network::NetworkName, peer::Peer};
+use amaru_network::session::PeerSession;
+use amaru_progress_bar::{ProgressBar, new_terminal_progress_bar};
+use clap::{Parser, arg};
+use gasket::framework::{AsWorkError, WorkerError};
```
jeluard, line 21: We don't want gasket dependency here

pgrange: I must admit, I just copy/pasted what we had in import_headers without giving it a lot of though. I'm not sure why but it has this dependency. Maybe I can see how to replace the code to not use gasket here?

jeluard: I see you didn't introduce it, now it would make sense to remove it. Or maybe as a separate PR?

data/README.md
```diff
@@ -84,3 +83,13 @@ For example, for PreProd, that means removing all immutable chunks after `03149`
...
+
+4. Dump chain headers as CBOR
+
+From the repository root, run:
+
+```
+make fetch-chain-headers
+```
+
+This will add `chain.<slot>.<hash>.cbor` files to data/<network>/headers, which the bootstrap command consumes to initialize the Amaru node.
```
jeluard: Pourquoi le prefix `chain`?

data/preview/headers/chain.55987200.7aaf1c7f89bbe29cef8c4c0098d52717e1d56c13dbe768d314a6414a27fb79f1.cbor
jeluard, line 1: We only miss `mainnet` :)

pgrange: Just added mainnet, sorry :)

data/README.md
pgrange: These are the chain headers and it's prefixed by chain in a headers directory... well ok, I can see why it would be confusing. Let me rename it header.*

jeluard: In most places we simply follow the pattern `<slot>.<hash>.cbor`

jeluard:
@abailly That is something we discussed with @pgrange and will soon start looking into. Also see https://github.com/pragma-org/amaru/discussions/327

data/README.md
pgrange: I've renamed it header.<slot>.<hash>.cbor as it's the header of the block. To be honest, I'm not sure it's super helpful to put in inside a headers directory so, in case we want to change a flatten the directory structure, at least we'll still remember that this file contains a header.

wdyt? Let me know if you need me to remove `header` from the file prefix, no problem.

crates/amaru/src/bin/amaru/cmd/fetch_chain_headers.rs
pgrange: If that's ok, I'd like to not do it inside this PR.

crates/amaru/src/bin/amaru/cmd/bootstrap.rs
pgrange: Because we're using gasket. If that's ok for you, can you resolve this conversation and we keep your comment above about not using gasket?

crates/amaru/src/bin/amaru/cmd/fetch_chain_headers.rs
pgrange: Fixed 👍

crates/amaru/src/bin/amaru/cmd/fetch_chain_headers.rs
pgrange: This is copy/pasted from previous code. We could, indeed, crash here unless there's a good reason not to.

@abailly, I noticed you introduced this in commit [881b58f9f](https://github.com/pragma-org/amaru/commit/881b58f9f). Did you have any strong reason to not fail here or is it just ok to crash?

I'm not sure there would be any valid reason for our files here to be corrupted so I tend to lean towards the `let's crash` approach. wdyt?

jeluard: Didn't realize it was move code. Let's address that later on then.

pgrange: @jeluard I changed the code so that it will crash if a header file in our source code is malformed. It feels like the right thing to do 👍

I tested with all the networks and I did not see any error (as all our files are, indeed, well-formed).

@abailly if you can think of any good reason to not do this and keep the old code, just let me know.

abailly:
Indeed, I am aware of this initiative which was started in Paris back in April. Yet it will take some time to come to maturity and we'll need to deploy and bootstrap amaru nodes before.

data/README.md
jeluard: Ok. We also have some `cbor` files in `snapshots/NETWORK/<slot>.<block>.header` so at some point it will make sense to align those.

pgrange:
> @pgrange actually `import-headers` is needed due to our CI setup (see https://github.com/pragma-org/amaru/blob/main/.github/workflows/continuous-integration.yml#L228). It will need to be restored but in a way that it relies on your local files probably.
> 
> The alternative is to also cache `chain.db`. Maybe it's not cached due to size constraints cc @KtorZ

@jeluard I fixed the issue by factorizing the way the headers are imported in `bootstrap` and in `import-headers`.

PR #411, opened 2025-08-28
feat: allow fetch to work with custom network and without pre-defined config.

KtorZ:
When set to 'custom', fetch will simply adapt to whatever network is available, and fetch snapshots on epoch boundaries (auto-magically detected).

## Summary by CodeRabbit

* New Features
  * Support for a “custom” network and OGMIOS_URL override.
  * Optional per-network ledger snapshot dumping.

* Improvements
  * Automatic live fetching when no config is present.
  * Faster concurrent fetching (up to 10 simultaneous fetches).
  * Spinner progress indicator and TTY-aware progress logging.
  * Network-specific configuration and storage to avoid refetching.
  * Safer handling when crossing epoch boundaries.

* Reliability
  * Improved retry and error handling for transient connection/data issues.

* Refactor
  * Clear separation between config-driven and continuous fetching flows.

### DIFF
--- a/data/fetch.mjs
+++ b/data/fetch.mjs
@@ -23,6 +23,8 @@ const network = (process.argv[2] ?? "").toLowerCase();
 
 const includeSnapshots = (process.argv[3] ?? "false").toLowerCase() == "true";
 
+const ogmiosUrl = process.env.OGMIOS_URL ?? "ws://127.0.0.1:1337"
+
 if (!["preview", "preprod", "mainnet", "custom"].includes(network)) {
   console.log(`Missing or invalid network.
 Usage:
@@ -156,7 +158,7 @@ async function fetchContinuously() {
       }
 
       done(true);
-    });
+    }, ogmiosUrl);
 
     if (exit === undefined) {
       process.stderr.cursorTo(0, 0);
--- a/data/fetch.mjs
+++ b/data/fetch.mjs
@@ -23,7 +23,7 @@ const network = (process.argv[2] ?? "").toLowerCase();
 
 const includeSnapshots = (process.argv[3] ?? "false").toLowerCase() == "true";
 
-const ogmiosUrl = process.env.OGMIOS_URL ?? "ws://127.0.0.1:1337"
+const ogmiosUrl = process.env.OGMIOS_URL ?? "ws://127.0.0.1:1337";
 
 if (!["preview", "preprod", "mainnet", "custom"].includes(network)) {
   console.log(`Missing or invalid network.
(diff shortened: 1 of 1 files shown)


data/fetch.mjs
```diff
@@ -17,58 +17,24 @@ import * as path from "node:path";
 import { bech32 } from 'bech32';
 import { ogmios, Json } from "@cardano-ogmios/mdk";
 
+const spinner = ["⠋", "⠙", "⠹", "⠸", "⠼", "⠴", "⠦", "⠧", "⠇", "⠏"];
```
abailly, line 20: Cute 😄

data/fetch.mjs
```diff
@@ -102,73 +68,175 @@ const queries = [
...
+        const point = {
+          id: block.id,
+          slot: block.slot,
+          epoch: slotToEpoch(eraSummaries, block.slot)
+        };
+
+        if (previousTip != null) {
+          process.stderr.cursorTo(0, 1);
+          process.stderr.clearLine(0);
+	  const num = relativeSlot(eraSummaries, point.slot);
```
abailly: nitpick: alignment seems off at various places 🤔

abailly:
Hopefully I should be able to test that with testnets tomorrow

data/fetch.mjs
KtorZ: Ah yes. My bad, I haven't configured my formatter and space/tab for js on this repository.

abailly:
I am trying to test this for a cluster of 5 nodes (eg. what we run in [antithesis](https://app.radicle.xyz/nodes/seed.hydra.bzh/rad:z2a7Te5b28CX5YyPQ7ihrdG2EEUsC/tree/compose/testnets/cardano_node_10.5.1/README.md)) and so far managed to run the cluster with ogmios connected to one of the nodes. Next step is to find a way to run the script _inside_ the ogmios container.

However, the approach we sketched of generating snapshots in advance for test networks feels more and more problematic: 
* It takes ages: at 86400 slots per epoch and 1s slot length, it will take 3 days to generate the needed 3 snapshots
* It's very sensitive to minor changes: Validity of blocks and ledger depends on the network parameters hence any attempt to change those (number of nodes, initial stake distribution, k, f, etc...) will fail, if not in amaru at least in cardano-node
* It's somewhat complicated.

abailly:
How do I connect to an Ogmios instance _not_ running on `localhost:1337`? I am packaging and running fetcher inside its own container and can only access Ogmios remotely.

abailly:
> Couldn't we leverage Antithesis for that? Accelerating time, and capturing snapshots at the right steps?

Sure, but this complicates the test runs and setting significantly as we need to run the whole cluster plus ogmios plus the data fetcher for a while, _then_ only start amaru. Also, extracting data from test runs is possible but not easy, and only available in case of test failure. 

I will continue trying to use fetch but I really think for testnets we should find another solution.

jeluard:
@abailly not a tty? Interestingly the rabbit suggested to guard those calls.

abailly:
@KtorZ If you think the commits I pushed are fine we can merge this but I think this is not the way to go for testnets. I wrote some thoughts about it in the logbook, we can discuss at the dev meeting or at any other time.

PR #419, opened 2025-09-02
feat: allow passing more arguments through env variables

abailly:
While 12 factor apps is not anymore considered state of the art given the pervasiveness of k8s-based deployments, having the possibility to use environment variables instead of explicit argumetns on the command-line is useful esp. in headless/remote server deployment scenarios.

## Summary by CodeRabbit

* **New Features**
  * Commands now accept AMARU_* environment variables for peer/listen addresses, network (default: preprod), ledger/chain/config dirs, and max extra ledger snapshots.
  * Demo script accepts a build profile and uses AMARU_* env vars with updated defaults, usage, and examples.

* **Chores**
  * Build, run, and test targets switched to the AMARU_* namespace; new AMARU_* variables exported and .PHONY updated to include dev.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,12 +28,12 @@ help:
 
 snapshots/$(AMARU_NETWORK): ## Download snapshots
 	@if [ ! -f "${SNAPSHOTS_FILE}" ]; then echo "SNAPSHOTS_FILE not found: ${SNAPSHOTS_FILE}"; exit 1; fi;
-	mkdir -p $@
+	mkdir -p "$@"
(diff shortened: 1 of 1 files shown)


jeluard:
Makes complete sense! One thing to consider is that all env variables in `Makefile` lack the `AMARU_` prefix. Shall we settle for one? Also we should double check that same var have the same name.

jeluard:
Looks like the rabbit made some relevant suggestions @abailly

PR #420, opened 2025-09-02
feat: introduce p2p networking stack

scarmuega:
⚠️ This is WIP

This pull request introduces the new P2P network machinery from pallas. The new network stack lives in the `pallas-network2` crate and is independent from the previous network code, meaning, they can co-exist side-by-side in the same crate.

I won't go into details here of how the internals of the p2p machinery work, but I'll explain the specific integration strategy:

- The new "pull" stage (`pull_p2p.rs`) now represents an abstraction for pulling data from a whole network of peers instead of a single upstream peer like before. This stage owns the whole network interface and is responsible of mediating between downstream stages and the external peers. Please note that there are no background threads / stack introduced by the network stack, the "tick" of the stage is what progresses the state of the network (by calling `network.poll_next()`).
- The config passed to the stage defines the initial state for bootsrapping the network. From this initial state, further interactions are designed using an event-driven approach, the stage handles events coming from the network and introduces extrinsic behavior by dispatching commands.
- Low-level mini-protocols are in "auto" mode. Handshake, keep-alive, peer-sharing, etc don't have to be manually controlled, the network stack assumes an unambiguous behavior and automates the whole interaction across connected peers.
- High-level mini-protocols that involve data flows provide specific network-level events and commands allowing external consumers to interact with them. This includes chain-sync, block-fetch, tx-submit, etc. There're no assumptions regarding the control-flow, but some of the details are abstracted way so that consumer can interact with the network as a whole instead of dealing with fine-grained interactions of particular peers.
- The "pull" stage main concern is to mediate between downstream stages and the state of the network. It provides a "request" input port where stages can ask for specific actions and data output ports for dispatching headers and blocks.

### DIFF
--- a/crates/amaru/src/stages/pull_p2p.rs
+++ b/crates/amaru/src/stages/pull_p2p.rs
@@ -1,0 +1,348 @@
+// Copyright 2024 PRAGMA
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::send;
+use amaru_consensus::consensus::ChainSyncEvent;
+use amaru_kernel::Point;
+use gasket::framework::*;
+use pallas_network2::{
+    Manager, PeerId,
+    behavior::{
+        InitiatorBehavior, InitiatorCommand, InitiatorEvent, PromotionBehavior, PromotionConfig,
+    },
+};
+use std::time::Duration;
+use tokio::select;
+use tracing::{Level, Span, instrument};
+
+pub type BlockBodyData = Vec<u8>;
+pub type RequestPort = gasket::messaging::InputPort<DownstreamRequest>;
+pub type SyncDataPort = gasket::messaging::OutputPort<ChainSyncEvent>;
+pub type BlockDataPort = gasket::messaging::OutputPort<BlockBodyData>;
+pub type NetworkMessage = pallas_network2::behavior::AnyMessage;
+pub type TcpInterface = pallas_network2::interface::TcpInterface<NetworkMessage>;
+pub type BlockRange = (Point, Point);
+
+const DEFAULT_HOUSEKEEPING_INTERVAL: Duration = Duration::from_secs(3);
+
+// preferring conversion via ad-hoc functions instead of implementing `From` to
+// avoid coupling other amaru crates with pallas-network2.
+
+fn amaru_to_pallas_point(point: &Point) -> pallas_network2::protocol::Point {
+    match point {
+        Point::Origin => pallas_network2::protocol::Point::Origin,
+        Point::Specific(slot, hash) => {
+            pallas_network2::protocol::Point::Specific(*slot, hash.clone())
+        }
+    }
+}
+
+fn pallas_to_amaru_point(point: &pallas_network2::protocol::Point) -> Point {
+    match point {
+        pallas_network2::protocol::Point::Origin => Point::Origin,
+        pallas_network2::protocol::Point::Specific(slot, hash) => {
+            Point::Specific(*slot, hash.clone())
+        }
+    }
+}
+
+fn pallas_to_amaru_peer(peer: &PeerId) -> amaru_kernel::peer::Peer {
+    amaru_kernel::peer::Peer::new(&peer.to_string())
+}
+
+#[derive(Default)]
+pub struct WorkUnit {
+    /// if set, instruct the network machinery to execute a particular command.
+    pub network_cmd: Option<InitiatorCommand>,
+
+    /// if set, dispatch a particular chain sync event to downstream stages.
+    pub sync_data: Option<ChainSyncEvent>,
+
+    /// if set, dispatch a particular block body data to downstream stages.
+    pub block_data: Option<BlockBodyData>,
+}
+
+impl From<InitiatorCommand> for WorkUnit {
+    fn from(command: InitiatorCommand) -> Self {
+        Self {
+            network_cmd: Some(command),
+            sync_data: None,
+            block_data: None,
+        }
+    }
+}
+
+/// A request from a downstream stage
+#[derive(Debug, Clone)]
+pub enum DownstreamRequest {
+    /// A request to fetch a particular block range.
+    FetchBlock(BlockRange),
+}
+
+pub struct Config {
+    /// The chain-sync intersection to ask to every hot peer.
+    pub intersection: Vec<Point>,
+
+    /// To configure the "promotion" behavior (eg: hot, warm, cold peers).
+    pub promotion: PromotionConfig,
+
+    /// The initial set of peers to connect to (aka: bootstrap nodes).
+    pub initial_peers: Vec<PeerId>,
+
+    /// The interval at which the housekeeping task is run (defaults to 3 secs).
+    pub housekeeping: Option<Duration>,
+}
+
+#[derive(Stage)]
+// TODO: revisit stage name. I would prefer something like "network", "p2p", etc. I'll leave the
+// decision for maintainers since it might affect tracing.
+#[stage(name = "stage.chain_sync_client", unit = "WorkUnit", worker = "Worker")]
+pub struct Stage {
+    /// To receive downstream requests.
+    pub requests: RequestPort,
+
+    /// To output chain sync data.
+    pub sync_data: SyncDataPort,
+
+    /// To output block body data.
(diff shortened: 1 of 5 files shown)


crates/amaru/src/stages/pull_p2p.rs
```diff
@@ -0,0 +1,348 @@
...
+        }
+    }
+}
+
+fn pallas_to_amaru_peer(peer: &PeerId) -> amaru_kernel::peer::Peer {
+    amaru_kernel::peer::Peer::new(&peer.to_string())
+}
+
+#[derive(Default)]
+pub struct WorkUnit {
```
abailly, line 65: Instead of stitching together `Option`s would it not make sense to have a composable enum, something like:

```
pub enum WorkUnit {
  Network(InitiatorCommand),
  Sync(ChainSyncEvent),
  Block(BlockBodyData),
  More(Box<WorkUnit>)
} 
```

crates/amaru/src/stages/pull_p2p.rs
```diff
@@ -0,0 +1,348 @@
...
+    pub initial_peers: Vec<PeerId>,
+
+    /// The interval at which the housekeeping task is run (defaults to 3 secs).
+    pub housekeeping: Option<Duration>,
+}
+
+#[derive(Stage)]
+// TODO: revisit stage name. I would prefer something like "network", "p2p", etc. I'll leave the
+// decision for maintainers since it might affect tracing.
+#[stage(name = "stage.chain_sync_client", unit = "WorkUnit", worker = "Worker")]
```
abailly, line 110: `stage.network.upstream` would seems to be a good fit

crates/amaru/src/stages/pull_p2p.rs
```diff
@@ -0,0 +1,348 @@
...
+    }
+}
+
+pub struct Worker {
+    pub network: Manager<TcpInterface, InitiatorBehavior, NetworkMessage>,
+    pub housekeeping: tokio::time::Interval,
+}
+
+impl Worker {
+    fn handle_network(&self, event: Option<InitiatorEvent>) -> WorkSchedule<WorkUnit> {
```
abailly, line 141: passing an `Option<>` here feels a bit weird, why not avoid calling the function altogether if we don't have an event to handle?

crates/amaru/src/stages/pull_p2p.rs
```diff
@@ -0,0 +1,348 @@
...
+            // The downside of this approach is that we don't have much information about
+            // the block at this point, just the payload. We can't infer which point in the chain or
+            // to which downstream request it belongs to. The behavior here is to just dispatch the
+            // block body downstream and let downstream stages decide how to handle it.
+            //
+            // There's no need for explicit commands back to the network machinery since the
+            // continuation of the mini-protocol is unambiguous. If the fetch request involved more
+            // blocks, the network machinery will keep receiving and notifying us of new
+            // block bodies.
+            InitiatorEvent::BlockBodyReceived(pid, body) => {
```
abailly, line 229: What's `pid` here? the `peerId`? This would seem useful to correlate with the request

abailly: Or maybe not if don't care where the block body came from.

crates/amaru/src/stages/pull_p2p.rs
```diff
@@ -0,0 +1,348 @@
...
+
+            // There're other events that the network can trigger, but we don't care about them for
+            // now.
+            _ => WorkSchedule::Idle,
+        }
+    }
+
+    fn handle_request(&self, request: &DownstreamRequest) -> WorkSchedule<WorkUnit> {
+        match request {
+            DownstreamRequest::FetchBlock(range) => {
```
abailly, line 248: How do we know which peer to send the request to? This seems important because there are some security implications in a peer not being able to deliver a block it's supposed to have (eg. they sent a header but fail to deliver block)

abailly (commented):
This looks surprisingly simple and I think this would fit into `pure_stage` framework, but we would need a bit of experimenting to see how that would work. Is pallas2 in a workable stage so that we can experiment integrating this code in a real environment?
also, can you point us at other example use of the framework to get some insights on how far it goes and how to handle the subtle cases you highlighted in comments?

crates/amaru/src/stages/pull_p2p.rs
scarmuega: The problem is that commands and data dispatch are orthogonal to each other. In any given tick you might want to send a network command AND dispatch data. Eg: after a header, you will want to dispatch the header and continue with the sync.

scarmuega: I now see that you have a continuation in your enum example, that could work to concatenate both actions.

Personally, I find the other struct more straight forward, but feel free to change it if you opt for the other one.

crates/amaru/src/stages/pull_p2p.rs
scarmuega: Yeah, I kind of agree. I did this to improve readability on the `select!` branches and emphasize the fact that returning `idle` for an empty event is a business logic decision (as opposed to glue code between functions).

One could make use of empty network ticks to provide metrics about the absence of network activity (eg: a counter of idle event).

Having said that, feel free to change it if the alternative feels more suitable.

crates/amaru/src/stages/pull_p2p.rs
scarmuega: `pid` is short for `peer_id`. Saved me thousands of keystroke when building the network crate :)

Almost all network events have a `pid` tag. Consensus will certainly need `pid` on header events, but probably not for block fetch. It could be useful if you want to keep peer-level stats (eg: blocks received per peer, latency per peer, etc)

crates/amaru/src/stages/pull_p2p.rs
scarmuega: The assumption here is that choosing a particular peer from the hot peers is a low-level detail that could be hidden from the consumer. If this design has implication on consensus security, we can certainly modify the api so that the specific peer is specified by the caller.

scarmuega:
The `pallas-network2` is in a _usable_ state, yes.

Not battle tested, not the best test coverage, but ready for beta. Although dev is still ongoing, my plan is to deploy it on some low-risk environments to start gaining some real-usage feedback.

This [example](https://github.com/txpipe/pallas/tree/main/examples/p2p) contains a working version of a naive node doing chainsync + blockfetch (plus all of the low-level: peer discovery, handshake, keep-alive, etc).

I don't have examples for the subtleties (back-pressure, config sweet spot, etc) that I mention in the comments; mainly because they depend on the architecture of the consumer, and we don't have many consumers ATM.

rkuhn:
Thanks @scarmuega, this is a nice improvement that will make it much easier to integrate with different execution models (like pure-stage). While studying the code of pallas_network2 I found some aspects that I’d like to ask about.

## Initiator / Responder

Currently only the initiator role is implemented, with only the initiator sides of the respective miniprotocols. What is your vision for how to offer the responder sides of the protocols and also the responder role of the whole endpoint? Reading the [network spec](https://ouroboros-network.cardano.intersectmbo.org/pdfs/network-spec/network-spec.pdf) §1.4.1 it seems that the initiator might also run the responder side of the miniprotocols, so would you have a separate ResponderBehavior or extend the InitiatorBehavior such that it serves both parts and should be renamed to N2NBehavior or similar?

On this note: we’ll need to have the responder functionality ready before we can switch from pallas_network to pallas_network2.

## Flow control

It seems to me that the flow control mechanism described in §2.1.2 and §2.1.3 has not been implemented yet, correct? In particular, the per-channel receive buffers might grow past fixed bounds unless I’m missing something. It seems difficult to add the intended behaviour of disconnecting the node upon TCP back pressure within the chosen implementation approach.

## Unbounded buffers

The use of FuturesUnordered to collect outstanding work does not prevent unbounded growth of the task queue managed inside. How do you intend to keep all work in progress bounded in size within the network stack? Bounding these queues would require for example that sending a block fetch request might result in a QueueOverflowError, which is currently not foreseen in the API.

## Peer selection for block fetch

As you stated, the block fetch operation shall be dispatched to the appropriate peers by the p2p network stack, thus simplifying the Amaru consensus codebase. Currently, it seems that a block fetch request will be sent to one node only, namely the one whose visitor happens to be called first during a housekeeping operation. We’ll need to make this more selective and also ask a configurable number of peers instead of one — typically we ask the peer that sent the corresponding header, plus a small number of nodes that have proven to respond quickly in the past. How would you foresee to implement such logic given the current internal design?

## Performance concerns

The design of the behaviour is such that it will dispatch each event to all sub-behaviours in sequence, while for housekeeping it will invoke all behaviours of all peer connections. This shouldn’t be an issue as long as the number of connections is small, but it may become an issue when managing 500 connections (which we think should be within the operational parameter range of Amaru). Currently, the housekeeping task only runs every 3sec by default (as per this PR), but I also saw that fetching a block is only triggered by the next housekeeping tick, which means that we’ll want to run housekeeping every couple of milliseconds or so — otherwise block fetch latency would impact the Amaru node’s timeliness constraints.

What is your intended evolution of this aspect of the implementation? It seems to me that focusing event-driven updates only on those tasks that need polling will become necessary at a certain usage intensity of this network stack.

PR #421, opened 2025-09-03
refactor: remove an unnecessary type parameter for stage refs

etorreborre:
This PR removes the `State` type parameter on `StageRef` and `Effects` since arguably the state of a stage should only be an implementation concern.

There is still a need for a `StageStateRef<Msg, St>` though, for example to use the `get_state` function (this is why `wire_up` returns that type). As a consequence I have updated the `Effect` interface to either accept `StageRef` or `StageStateRef`.


crates/pure-stage/src/effect.rs
```diff
@@ -587,11 +609,14 @@ impl Effect {
...
 
-    pub fn assert_external<Msg, St, Eff: ExternalEffect + PartialEq>(
+    pub fn assert_external<Msg, S, Eff: ExternalEffect + PartialEq>(
         &self,
-        at_stage: &StageRef<Msg, St>,
+        at_stage: S,
         effect: &Eff,
-    ) {
+    ) where
+        S: Into<StageRef<Msg>>,
```
rkuhn, line 617: This feels a little too heavy, my first thought of which trait to use was `AsRef<StageRef<Msg>>`. Then `StateStageRef` would just wrap a `StageRef` with an additional type parameter.

crates/pure-stage/src/stage_ref.rs
```diff
@@ -55,20 +55,82 @@ impl<Msg, State> Clone for StageRef<Msg, State> {
...
         self.name.clone()
     }
 
-    pub fn without_state(&self) -> StageRef<Msg, Void> {
+    pub fn without_state(&self) -> StageRef<Msg> {
+        StageRef {
+            name: self.name.clone(),
+            _ph: PhantomData,
+        }
+    }
```
rkuhn, line 76: this function no longer makes sense

PR #422, opened 2025-09-03
chore: use expect for warnings and remove unused ones

etorreborre:
`expect` raises a warning when the annotation is not necessary anymore.


jeluard (approved):
Nice! Didn't know about that

PR #427, opened 2025-09-05
fix: stake address prefix on mainnet for fetch and generate scripts.

KtorZ:
Follow-up from #411, after trying this on mainnet.


data/fetch.mjs
```diff
@@ -326,7 +326,7 @@ function fetchRewardsProvenance(ws) {
   return ws.queryLedgerState("rewardsProvenance")
 }
 
-async function fetchDReps(ws, { stakePools }) {
+async function fetchDReps(ws, stakePools = {}) {
```
jeluard, line 347: Is `stakePools` parameter ever provided? Following your change in `query` call I don't see it anymore.

KtorZ: No it isn't anymore. I didn't want to further trim the code below _for now_; since to properly remove it; one PR must first be merged into the Haskell's consensus code, and then released. 

So for now, this requires running a forked cardano-node; passing the result as argument on mainnet causes a heap-overflow on node.js for mainnet 🙈... Which is why I ended up tweaking the consensus directly. I should probably leave all this comment as a comment *in the code*.

PR #430, opened 2025-09-05
Adjust verbosity of chain tip logs

KtorZ:
I also slightly altered the `SyncingTracker` implementation, so the check for `is_caught_up` is really not expensive and doesn't require traversing all peers. It appears that we never update that list; and if we do (peers added or remove), I'd simply suggest to create a new one. 

## Summary by CodeRabbit

- New Features
  - None.

- Refactor
  - Simplified internal peer sync-tracking; no changes to public APIs.

- Style / Logging
  - Adjusted log levels and wording for fork/switch/“no change” events; trace vs debug now varies by peer catch-up state.

- Observability / UX
  - Deferred environment-variable warnings and unified telemetry/JSON warning handling so warnings are emitted after logging is initialized.

### DIFF
--- a/crates/amaru/src/bin/amaru/observability.rs
+++ b/crates/amaru/src/bin/amaru/observability.rs
@@ -140,24 +140,8 @@ impl TracingSubscriber<Registry> {
                         .with_filter(log_filter()),
                 )
                 .init(),
-            TracingSubscriber::WithJson(layered) => layered
-                .with(
-                    tracing_subscriber::fmt::layer()
-                        .with_writer(log_writer())
-                        .event_format(log_format())
-                        .with_span_events(log_events())
-                        .with_filter(log_filter()),
-                )
-                .init(),
-            TracingSubscriber::WithBoth(layered) => layered
-                .with(
-                    tracing_subscriber::fmt::layer()
-                        .with_writer(log_writer())
-                        .event_format(log_format())
-                        .with_span_events(log_events())
-                        .with_filter(log_filter()),
-                )
-                .init(),
+            TracingSubscriber::WithJson(layered) => layered.init(),
+            TracingSubscriber::WithBoth(layered) => layered.init(),
         };
 
         if let Some(notify) = warning {


crates/amaru-consensus/src/consensus/select_chain.rs
```diff
@@ -120,21 +111,23 @@ impl SelectChain {
 
         let events = match result {
             ForwardChainSelection::NewTip { peer, tip } => {
-                debug!(target: EVENT_TARGET, hash = %tip.hash(), slot = %tip.slot(), "new_tip");
-
+                if self.sync_tracker.is_caught_up() {
+                    debug!(target: EVENT_TARGET, %peer, hash = %tip.hash(), slot = %tip.slot(), "new tip");
+                } else {
+                    trace!(target: EVENT_TARGET, %peer, hash = %tip.hash(), slot = %tip.slot(), "new tip");
+                }
```
KtorZ, line 118: This is otherwise way too verbose with our default settings; so while syncing, those traces are better down at the `TRACE` level.

data/mainnet/snapshots.json
KtorZ, line 1: Whoops, wrong commit.

crates/amaru-consensus/src/consensus/select_chain.rs
abailly: I disagree and I am surprised: I thought we discussed that together and you agreed it was better to have a simple log every time we changed the tip, whether or not you are syncing.

abailly (changes requested):
I would rather keep the log as is

KtorZ:
> @abailly 
> 
> I disagree and I am surprised: I thought we discussed that together and you agreed it was better to have a simple log every time we changed the tip, whether or not you are syncing.

Yes, but not at the default verbosity! These are millions of lines when syncing x_x ... anything else is completely drawn under those tips. So, at the `trace` level, when aggregated with other traces and processed in a pipeline, sure. But not in the dev logs.

abailly:
I disagree as I think this information is important in dev even when syncing and I don't really look at the logs the way you do. I am running nodes all the time too, and I don't have any problems with that. On the contrary I find it useful to witness the tip is updating correctly, and when it crashes to know where it stopped easily.

crates/amaru-consensus/src/consensus/select_chain.rs
abailly: I particularly dislike having different kind of level of logs depending on the state of the system.

crates/amaru-consensus/src/consensus/select_chain.rs
```diff
@@ -37,34 +39,23 @@ pub struct SelectChain {
...
-                SyncState::CaughtUp => (),
-            }
-        } else {
+        if !self.syncing_peers.remove(peer) {
             warn!("unknown caught-up peer {}", peer);
         }
     }
 
     pub fn is_caught_up(&self) -> bool {
-        self.peers_state.values().all(|s| *s == SyncState::CaughtUp)
```
abailly, line 67: I would rather keep the map and cache the overall state, but that's only an itch that we'll need it later. YAGNI and all that so 🤷

jeluard:
@abailly I see 3 states here: bootstrap phase, catching up phase and following the chain. At least for bootstrap it's a lot of noise.

jeluard:
How shall we proceed with this one?

abailly:
@jeluard There's only really one phase because that's how the protocol works: syncing up is an integral part of the process and can happen if for example your node shutdown for voluntary (eg. you are running on intermittent devices) or involuntary (cutoff from network) reasons. I contend that logging the current best chain when it changes is a critically important information to know. I fail to see why the amount of logs produced is a problem for a piece of software that's supposed to run as a headless server: modern execution environments know how to take care of log rotation, log file size, and logs are usually shipped to other services anyway.

In development, filtering the logs you want to see is trivial and can be done through an environment variable or a simple `grep`. I also don't understand why we need 2 different environment variables to filter logs, why I cannot have _both_ json logs and open-telemetry, why there are logs who go to stderr and others to stdout... I don't know how to proceed, I have been voicing for a while my disagreement with the logging system we have. 

Perhaps I am overreacting here, I am not the lead architect of this project, this responsibility lies on @KtorZ's shoulders so I will let him decide on this matter. In the meantime I will remove my change request and if you want to merge this, go for it.

KtorZ:
> @abailly: I contend that logging the current best chain when it changes is a critically important information to know. 

Agreed. That's why we log it! 

> @abailly: I fail to see why the amount of logs produced is a problem for a piece of software that's supposed to run as a headless server: modern execution environments know how to take care of log rotation, log file size, and logs are usually shipped to other services anyway.

It isn't, that's why we have them (no logs are removed here, they are simply logged at different severities). The problem is what appears in the terminal by default.

> @abailly: In development, filtering the logs you want to see is trivial and can be done through an environment variable or a simple grep. 

Yes, precisely. 

> @abailly: I also don't understand why we need 2 different environment variables to filter logs, why I cannot have both json logs and open-telemetry, why there are logs who go to stderr and others to stdout... 

That's an orthogonal question, and I actually have a PR I've been wanting to do on that matter for a while now.  I think you'll like it. Because it's been quite apparent that no one uses both JSON logs or OTLP; I am thinking about making these a mutually exclusive switch:

- When nothing is provided, we only get (pretty) dev logs on the terminal (stderr).
- When the OLTP flag is present, we get both the (pretty) dev logs on the terminal (stderr), and traces via OTLP. 
- When the JSON flag is present, we get traces in JSON on the terminal (stdout).
- When both flags are present, it's an error. 

There will still be an environment variable; that is for controlling the traces vs the dev logs. 

> @abailly  I don't know how to proceed, I have been voicing for a while my disagreement with the logging system we have. Perhaps I am overreacting here;

Honestly, yes; Insomuch as it's not like we've not been listening to feedback and disagreement and we keep iterating on that piece. It's hard to find the sweet spot where everyone is happy. But I am confident we'll get there. I think your latest bit of annoyance is really that when JSON logs are activated, you still get the dual traces on stderr/stdout. So by fixing that, I believe we reach a good status quo.

crates/amaru-consensus/src/consensus/select_chain.rs
KtorZ: I do think it's actually nice to adjust logs' severity based on the state of the system; when we do have that insight as application developer. For example, if a request that should be quick takes a significant amount of time (e.g. validating block takes suddenly 30s), we should be able to flag that for whomever is observing the system. 

Maybe that is something which should entirely lie on the operational side; although there are cases (like block validation) where these become application concerns as well. 

Also; while in the protocol, there is only one state (syncing); it isn't how people perceived this in practice. I've made that mistake once with Kupo for example. And I got users telling me how it's annoying that there's no distinction between syncing vs _somewhat synchronized_. Because syncing is a one time thing, that one can do on a different (beefy) machine, while then deploying the long-running instance on a different sub-system. 

People even asked for specific metrics capturing the syncing percentage, and special mode that would disable indexing while syncing making the indexer somewhat inoperative because no large queries would really work in a reasonable amount of time. But that did not matter because people treat it as two separate deployments even. There's effectively a not-synced vs synced state. So it is quite natural that we would also reflect that. 

I am happy to be proven wrong later; but so far, this is where the past learnings take me.

KtorZ: Truth is that figuring out whether a node is synchronized isn't an easy question, and even requires some extensive knowledge of the protocol / consensus somehow. So, that signal is extremely valuable to observers of the system.

crates/amaru-consensus/src/consensus/select_chain.rs
KtorZ: Yes, that's an arguable choice; but since we don't really need the map at this point, I got rid of it. If we ever need one, that's easy to re-introduce (although likely with some `Rc` :D).

abailly:
And I really don't understand where this misunderstanding (!) comes from 🤷 Perhaps because I don't care about showing "nice logs" in a terminal?

jeluard:
It appears that people have different opinions and we try to reconciliate that?

crates/amaru-consensus/src/consensus/select_chain.rs
abailly: > if a request that should be quick takes a significant amount of time (e.g. validating block takes suddenly 30s), we should be able to flag that for whomever is observing the system.

That has nothing to do with adjusting the _log level_ depending on the state of the system, it's really just an event about something important happening in the system, eg:
* One event is _request X happened_
* Another event is _request X took an inordinate amount of time_

I would indeed argue that filtering/choosing what an inordinate amount of time is is an operational concern, but I can agree that it makes sense in some cases to bake that in the system. Yet, for the sake of simplicity and because it's hard to tell which thresholds for which metrics are important, I would leave that aside for now.

> while in the protocol, there is only one state (syncing);

_Caught-up_ has a very precise definition in the Haskell node (in the context of Genesis) actually, something I linked in the journal IIRC: 
1. The node is in `Wait` state with all its upstream peers,
2. There's no better chain than the one it's currently on.

We currently signal when we reach `CaughtUp` state based on 1., while 2. is implicit in the `select_chain` stage. I have not put any efforts in tracking the _other_ direction: one you're `CaughtUp` you stay so, but that's something we should probably do in a later stage, eg. if we realise we are suddenly lagging behind one or more of our peers something we can possibly detect at the `Chainsync` client level.

> syncing is a one time thing, that one can do on a different (beefy) machine, while then deploying the long-running instance on a different sub-system.

There's a whole continuum there, which gets even more complicated when you include the concept of snapshots, Mithril, etc.

abailly:
Thanks a lot @KtorZ, that's perfect for me.

PR #431, opened 2025-09-05
refactor: split the stage api differently

etorreborre:
@rkuhn this is an exploration on how we could decouple the making of stage references from the registration of the transition function in order to create `Stage` components that are instantiated right away with their downstream references, instead of passing them as state. What do you think?

### DIFF
--- a/crates/pure-stage/src/stagegraph.rs
+++ b/crates/pure-stage/src/stagegraph.rs
@@ -17,9 +17,12 @@ use crate::{
     BoxFuture, Effects, Instant, Name, OutputEffect, Receiver, Resources, SendData, Sender,
     StageBuildRef, StageRef, types::MpscSender,
 };
+use async_trait::async_trait;
+use serde::Serialize;
+use std::pin::Pin;
+use std::sync::Arc;
 use std::{
     fmt::Debug,
-    future::Future,
     marker::PhantomData,
     sync::atomic::{AtomicU64, Ordering},
 };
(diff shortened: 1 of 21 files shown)


crates/pure-stage/src/stagegraph.rs
```diff
@@ -232,3 +370,10 @@ pub trait StageGraphRunning {
     /// A future that resolves once the stage graph has terminated.
     fn termination(&self) -> BoxFuture<'static, ()>;
 }
+
+#[async_trait]
+pub trait Stage<Msg, State> {
+    fn initial_state(&self) -> State;
```
rkuhn, line 376: This implies that it is rather ugly/inconvenient to use something as State that can’t be cloned — I don’t see a reason why we should introduce this restriction.

crates/pure-stage/src/stagegraph.rs
```diff
@@ -158,19 +194,24 @@ pub trait StageGraph {
...
+    ///
+    /// The returned [`StageRef`] can be used to send messages to this stage from other stages.
+    ///
+    fn register<Msg, St>(
         &mut self,
-        name: impl AsRef<str>,
-        f: F,
-    ) -> StageBuildRef<Msg, St, Self::RefAux<Msg, St>>
+        stage: &StageStateRef<Msg, St>,
+        stageable: impl Stage<Msg, St> + 'static + Send + Clone,
```
rkuhn, line 203: What is gained by requiring an instance of a trait here? This means adding a `struct` that has no other use than to temporarily hold the initial state.

rkuhn (commented):
I think I’ll need some more background and motivation for these changes, they don’t look like an overall improvement to me. This PR basically moves the transition function from one place to the other and bundles it up with the initial state, which doesn’t feel natural to me because it requires more plumbing.

I placed the transition function on the `stage()` method in order to fix the type parameters of the returned reference at this point. Rust type inference is not Haskell type inference, perhaps we have different expectations and experiences in this regard :-) One thing that could be improved on the existing API is that all the type parameter constraints should be present on `stage()` already — I’ve seen cases where the compiler only complained at `wire_up` time that the provided transition function was no good.

The naming of `make_stage` / `register` also doesn’t ring true with me: the first one doesn’t really make a stage and the second one is not about registering, it is the one that actually makes the stage.

etorreborre:
Thanks for the feedback @rkuhn!

> The naming of make_stage / register also doesn’t ring true with me: the first one doesn’t really make a stage and the second one is not about registering, it is the one that actually makes the stage.

Yes, those names are not great but the intent was indeed to name the stages before assigning their transition function.

> What is gained by requiring an instance of a trait here? This means adding a struct that has no other use than to temporarily hold the initial state.

Then my other motivation was to avoid having to pass those handles around at runtime as a piece of state since they are already known before we start anything and never change at runtime. The struct was only there to support this.

> This implies that it is rather ugly/inconvenient to use something as State that can’t be cloned — I don’t see a reason why we should introduce this restriction.

Yes, that's not great, I wish I could avoid cloning there. That being said, if we move everything to using effects there will be no state left as far as I can see. So eventually we will just be passing handles around unless I missed something.

Anyway, thanks for having a look, I just wanted to have an opinion on this idea. I'll close this draft.

PR #435, opened 2025-09-06
Convert cardano-node serialised ledger state to amaru snapshot

### LINKED ISSUE #277: Provide a simple to use `docker-compose` script that creates a small testnet composed of several amaru nodes.
Provide a simple to use `docker-compose` script that creates a small testnet composed of several amaru nodes.
`monitoring` should be properly setup so that once running every is available to get a glimpse of the whole network.

_Originally posted by @jeluard in https://github.com/pragma-org/amaru/discussions/256_

abailly:
This PR adds a `convert-ledger-state` command that takes one or more ledger state as serialised by the cardano-node and "convert" into a format suitable for `import-ledger-state` command. The former contains more information that needed by Amaru, so we need to dig into the CBOR produced for a consensus' `ExtLedgerState ShelleyBlock` to extract the part we care about, namely the `Ledger.NewEpochState`. 

This PR comes with a test that checks we can indeed import the 3 snapshots generated for a testnet by a custom extension of [db-synthesizer](https://github.com/abailly/ouroboros-consensus/blob/1508638f832772d21874e18e48b908fcb791cd49/ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Tools/DBSynthesizer/Forging.hs#L252) tool, which paves the way to answers #256 and fixes #277


crates/amaru/src/bin/amaru/cmd/convert_ledger_state.rs
```diff
@@ -0,0 +1,195 @@
...
+    let slot = d.u64()?;
+    let _height = d.u64()?;
+    let hash: Hash<HEADER_HASH_SIZE> = d.decode()?;
+
+    let p = d.position();
+
+    let target_path = target_dir.join(format!("{}.{}.cbor", slot, hash));
+
+    fs::write(&target_path, &bytes[p..bytes.len() - 1]).await?;
+
```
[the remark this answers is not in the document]
abailly: @coderabbit Why would the last byte be `0xff`? This does not make much sense

jeluard (approved):

abailly:
https://github.com/IntersectMBO/ouroboros-consensus/issues/1668

crates/amaru/src/bin/amaru/cmd/convert_ledger_state.rs
```diff
@@ -0,0 +1,372 @@
...
+    let mut eras: Vec<Summary> = decode_eras(&mut d, network)?;
+
+    // current era lower bound
+    d.array()?;
+    let start: Bound = d.decode()?;
+    eras.push(Summary {
+        start,
+        end: None,
+        params: EraParams {
+            epoch_size_slots: network.default_epoch_size_in_slots(),
```
KtorZ, line 111: Shouldn't this be inferred from the `eras` instead of being hardwired to the network? It probably doesn't matter for well-known networks, but for devnets which can have arbitrary globals, that might not end well?

crates/amaru/src/bin/amaru/cmd/convert_ledger_state.rs
```diff
@@ -0,0 +1,372 @@
...
+    d.array()?;
+    d.array()?;
+    let slot = d.u64()?;
+    let _height = d.u64()?;
+    let hash: Hash<HEADER_HASH_SIZE> = d.decode()?;
+
+    // ledger state
+    let begin = d.position();
+    d.skip()?;
+    let end = d.position();
```
KtorZ, line 137: Smart.

crates/amaru/src/bin/amaru/cmd/convert_ledger_state.rs
```diff
@@ -0,0 +1,372 @@
...
+    d.array()?;
+    // NOTE: The encoding of an AnnTip is not consistent with the encoding of a Tip
+    let tip_slot = d.u64()?;
+    let _tip_hash: Hash<HEADER_HASH_SIZE> = d.decode()?;
+    let _tip_height = d.u64()?;
+
+    // ChainDepState for Praos
+    d.array()?;
+
+    // more HFC telescope encoding
```
KtorZ, line 163: 🙃

crates/amaru/src/bin/amaru/cmd/convert_ledger_state.rs
```diff
@@ -0,0 +1,372 @@
...
+    write_ledger_snapshot(target_dir, slot, hash, &bytes[begin..end]).await
+}
+
+fn decode_eras(
+    d: &mut minicbor::Decoder<'_>,
+    network: &NetworkName,
+) -> Result<Vec<Summary>, Box<dyn std::error::Error>> {
+    let mut eras = Vec::new();
+
+    for _ in 0..6 {
```
KtorZ: 6 ? 

I reckon you probably want to use: [minicbor#array_iter](https://docs.rs/minicbor/latest/minicbor/decode/struct.Decoder.html#method.array_iter) here.

crates/amaru/src/bin/amaru/cmd/convert_ledger_state.rs
```diff
@@ -0,0 +1,372 @@
...
+                let path = entry.path();
+                result.push(path.clone());
+                if path.is_dir() {
+                    acc.push(path.clone());
+                }
+            }
+        }
+        Ok(result)
+    }
+}
```
KtorZ, line 421: Alternatively: [walkdir](https://docs.rs/walkdir/latest/walkdir/index.html#example)

KtorZ (commented):
I like where this is going. One thought though: shouldn't that (eventually) entirely replace our bootstrap snapshots? Instead of being a separate convert command?

PR #437, opened 2025-09-08
fix: incompatible encoding for era bounds

abailly:
Haskell node outputs a fixed length array of 3 elements for `Bound` and not a variable length array as it was previously defined. I stumbled upon this while working on #435 and looking at the output of `writeExtLedgerState` function from Haskell node. 

For a testnet the relevant part looks like:
```
82                                      # array(2)
   01                                   # unsigned(1)
   82                                   # array(2)
      87                                # array(7)
         82                             # array(2)
            83                          # array(3)
               00                       # unsigned(0)
               00                       # unsigned(0)
               00                       # unsigned(0)
            83                          # array(3)
               00                       # unsigned(0)
               00                       # unsigned(0)
               00                       # unsigned(0)
         82                             # array(2)
            83                          # array(3)
               00                       # unsigned(0)
               00                       # unsigned(0)
               00                       # unsigned(0)
            83                          # array(3)
               00                       # unsigned(0)
               00                       # unsigned(0)
               00                       # unsigned(0)
         82                             # array(2)
            83                          # array(3)
               00                       # unsigned(0)
               00                       # unsigned(0)
               00                       # unsigned(0)
            83                          # array(3)
               00                       # unsigned(0)
               00                       # unsigned(0)
               00                       # unsigned(0)
         82                             # array(2)
            83                          # array(3)
               00                       # unsigned(0)
               00                       # unsigned(0)
               00                       # unsigned(0)
            83                          # array(3)
               00                       # unsigned(0)
               00                       # unsigned(0)
               00                       # unsigned(0)
         82                             # array(2)
            83                          # array(3)
               00                       # unsigned(0)
               00                       # unsigned(0)
               00                       # unsigned(0)
            83                          # array(3)
               00                       # unsigned(0)
               00                       # unsigned(0)
               00                       # unsigned(0)
         82                             # array(2)
            83                          # array(3)
               00                       # unsigned(0)
               00                       # unsigned(0)
               00                       # unsigned(0)
            83                          # array(3)
               00                       # unsigned(0)
               00                       # unsigned(0)
               00                       # unsigned(0)
         82                             # array(2)
            83                          # array(3)
               00                       # unsigned(0)
               00                       # unsigned(0)
               00                       # unsigned(0)
            82                          # array(2)
               02                       # unsigned(2)
               83                       # array(3)
                  81                    # array(1)
                     83                 # array(3)
                        1A 0003F466     # unsigned(259174)
                        19 32AC         # unsigned(12972)
                        58 20           # bytes(32)
                           A07DA7616822A1CCB4811E907B1F3A3C5274365908A241F4D5FFAB2A69EB8802 # "\xA0}\xA7ah\"\xA1̴\x81\u001E\x90{\u001F:<Rt6Y\b\xA2A\xF4\xD5\xFF\xAB*i\xEB\x88\u0002"
```

## Summary by CodeRabbit

- New Features
  - Added a dedicated millisecond-based time type across era history and APIs; improved handling and encoding of very large timestamps.

- Refactor
  - Migrated time fields and related APIs to the new millisecond type without altering era boundaries or parameters.

- Bug Fixes
  - Updated error messaging to correctly indicate underflow.

- Chores
  - Added workspace dependencies to support new time type and serialization.

- Tests
  - Updated tests to use the new millisecond time representation.

### DIFF
--- a/crates/slot-arithmetic/src/lib.rs
+++ b/crates/slot-arithmetic/src/lib.rs
@@ -241,8 +241,13 @@ impl<'b, C> Decode<'b, C> for TimeMs {
         let datatype = d.datatype();
         (match datatype {
             Ok(Type::Tag) => {
-                // consume tag!!
-                d.tag()?;
+                // consume and validate tag
+                let tag = d.tag()?;
+                if tag != IanaTag::PosBignum.into() {
+                    return Err(minicbor::decode::Error::message(format!(
+                        "Unexpected CBOR tag decoding TimeMs: {tag} (expected positive bignum)"
+                    )));
+                };
                 // Haskell node defines timestamp as a Fixed E12
                 // (wrapped in a couple of newtypes), a fixed-decimal
                 // number with $10^{12}$ precision, eg. picoseconds
@@ -1138,6 +1143,16 @@ mod tests {
         // [558144000000000000001234567890000000000, 55814400, 646]
         let buffer =
             hex::decode("83c25101a3e69fd156bd141cccb9fb74768db4001a0353a900190286").unwrap();
+        let result = minicbor::decode::<Bound>(&buffer);
+        assert!(result.is_err());
+    }
+
+    #[test]
+    fn cannot_decode_bounds_with_invalid_tag() {
+        // CBOR encoding for
+        // [-558144000000000000001234567890000000001, 55814400, 646]
+        let buffer =
+            hex::decode("83c35101a3e69fd156bd141cccb9fb74768db4001a0353a900190286").unwrap();
         let result = minicbor::decode::<Bound>(&buffer);
         assert!(result.is_err());
     }


crates/slot-arithmetic/src/lib.rs
```diff
@@ -207,22 +265,20 @@ impl<C> Encode<C> for Bound {
...
         self.slot.encode(e, ctx)?;
         self.epoch.encode(e, ctx)?;
-        e.end()?;
         Ok(())
     }
 }
 
 impl<'b, C> Decode<'b, C> for Bound {
     fn decode(d: &mut Decoder<'b>, ctx: &mut C) -> Result<Self, minicbor::decode::Error> {
         let _ = d.array()?;
```
jeluard: Worth validating that the length returned by `d.array()?` is always `3`?

crates/slot-arithmetic/src/lib.rs
```diff
@@ -183,9 +184,66 @@ impl Step for Epoch {
...
+        e: &mut minicbor::Encoder<W>,
+        ctx: &mut C,
+    ) -> Result<(), minicbor::encode::Error<W::Error>> {
+        self.0.encode(e, ctx)
+    }
+}
+
+impl<'b, C> Decode<'b, C> for TimeMs {
+    fn decode(d: &mut Decoder<'b>, _ctx: &mut C) -> Result<Self, minicbor::decode::Error> {
+        d.u64()
```
jeluard: Does this really work? `d.u64` in turns calls `read` that advances the internal `pos`. So I would expect that the subsequent `d.bytes` (in `or_else` branch) would start reading a different slice of data.
Can we inspect the bytes beforehand to distinguish both cases?

crates/slot-arithmetic/src/lib.rs
```diff
@@ -183,9 +184,66 @@ impl Step for Epoch {
...
+            .or_else(|_| {
+                let bytes: &[u8] = d.bytes()?;
+                let num = BigUint::from_bytes_be(bytes);
+                let num_ms = num / 1_000_000_000u64;
```
jeluard, line 266: Are we fine cropping to millis? Why not keep the full resolution?

crates/slot-arithmetic/src/lib.rs
```diff
@@ -183,9 +184,66 @@ impl Step for Epoch {
...
+                if num_ms > BigUint::from(u64::MAX) {
```
jeluard, line 267: I would simply just `try_into ` and let it fail. Not sure the slightly more better error message is worth repeatedly calling `BigUint::from(u64::MAX)` (or then make it a global?)

crates/slot-arithmetic/src/lib.rs
KtorZ: Likely yes; and also, worth supporting both definite and indefinite here for flexibility. This type isn't sent over the wire. 

I'd suggest using: https://github.com/pragma-org/amaru/blob/main/crates/minicbor-extra/src/decode.rs#L265-L266

crates/slot-arithmetic/src/lib.rs
KtorZ: Yes; the resolution is Pico _by accident_ in the Haskell code; when the value is never larger than ms. See https://github.com/pragma-org/amaru/wiki/log-::-2025-09#converting-extledgerstate


I think that's still worth a comment in the code here 😅

crates/slot-arithmetic/src/lib.rs
abailly: There is a test for that, but perhaps it's not doing what I think it does?

crates/slot-arithmetic/src/lib.rs
abailly: I do think the finer grained error message is worthwhile lest we end up running into a very cryptic error down the road should someone decide it's a good idea to put such large numbers there. I agree rebuilding the `BigUint` is pretty inefficient, on the other hand we are talking about code that's run quite rarely.

crates/slot-arithmetic/src/lib.rs
abailly: > This type isn't sent over the wire.

What do you mean?

crates/slot-arithmetic/src/lib.rs
abailly: I would argue the value does not even need to have a finer grained resolution than _seconds_ but yes, I will add some comment.

crates/slot-arithmetic/src/lib.rs
KtorZ: It doesn't appear in _blocks_, and is only found in client APIs / user interfaces. Which means that we can be a little bit flexible about the decoder. For anything that appear in blocks, we MUST match precisely the Haskell's behavior. If the Haskell impl expects an definite array and doesn't check the length, then we must do exactly that even if it means accepting potentially invalid CBOR.

crates/slot-arithmetic/src/lib.rs
abailly: Thanks for the suggestion, I did not fully understand your comment and I wasted a couple of hours figuring out why my tests were failing.

PR #442, opened 2025-09-10
feat: added bootstrapping from mithril

jeluard:
**WIP**

**bootstrap** amaru from a mithril snapshot


jeluard:
Can't be used directly as `amaru` expects snapshots taken at specific `point`s.

PR #445, opened 2025-09-12
feat: will not start if ulimit is too low on fd

pgrange:
EDIT

When starting Amaru daemon, if file descriptor ulimit is too
low, Amaru will crash on mainnet:

* RocksDB can consume some amount of FDs for its internal operations;
* System metrics collection with sysinfo also consumes FDs;
* And of course we still need some FDs for network connections and so on.

From empirical tests, 1 000 file descriptors seem to be enough to
catch-up with mainnet so setting a pre-flight check to that number,
we can still increase it later if we see that we need more.

~~When running Amaru on mainnet, I stumble upon file descriptor limit issues. Amaru crashed and the ledger database was _slightly_ corrupted (removing the last epochs fixed the issue but still). This change introduces a limit to how many file descriptors RocksDb can use to prevent such crash.~~

~~When starting Amaru daemon, if file descriptor ulimit is lower than 65535 (because why not 65535?), Amaru will display a warning to suggest increasing it.~~

~~Then, we take the soft limit, we remove 100 to let room for other file descriptors, we divide by 4, because the daemon usually opens 4 rocksDb instance, and allocate that amount of file descriptors to each rocksdb instance.~~

## Summary by CodeRabbit

* **New Features**
  * Added a configurable RocksDB max-open-files option with an auto-estimated default based on system limits; applied across chain, ledger, and historical stores.
  * Startup now logs warnings if file-descriptor limits are low or cannot be determined.
  * Config initialization now fills unspecified fields from defaults automatically.

* **Chores**
  * Added a Unix-only dependency to read system file-descriptor limits.

### DIFF
--- a/crates/amaru/src/bin/amaru/metrics.rs
+++ b/crates/amaru/src/bin/amaru/metrics.rs
@@ -98,6 +98,11 @@ mod internals {
         /// It includes physical RAM, allocated but not used regions, swapped-out regions, and even
         /// memory associated with memory-mapped files.
         memory_available_virtual_bytes: Gauge<u64>,
+
+        /// The total number of opened file descriptors.
+        ///
+        /// On Windows, this metric is not available as file descriptors per se do not exist.
+        open_files: Gauge<u64>,
     }
 
     impl ProcessMetrics {
@@ -154,6 +159,11 @@ mod internals {
                 .with_unit("bytes")
                 .build();
 
+            let open_files = meter
+                .u64_gauge("process_open_files")
+                .with_description("Total number of file descriptors.")
+                .build();
+
             Self {
                 number_of_cpus,
                 runtime_seconds,
@@ -164,6 +174,7 @@ mod internals {
                 cpu_live_percent,
                 memory_live_resident_bytes,
                 memory_available_virtual_bytes,
+                open_files,
             }
         }
 
@@ -186,6 +197,9 @@ mod internals {
 
             self.memory_available_virtual_bytes
                 .record(proc.virtual_memory(), &[]);
+
+            self.open_files
+                .record(proc.open_files().map_or(0, |files| files as u64), &[]);
         }
     }
 }


crates/amaru/Cargo.toml
```diff
@@ -56,6 +56,9 @@ amaru-progress-bar = { workspace = true, features = ["terminal"] }
 amaru-slot-arithmetic.workspace = true
 pure-stage.workspace = true
 
+[target.'cfg(unix)'.dependencies]
+rlimit = "0.10.2"
```
jeluard, line 60: I'd rather find something that works on all platform we support :)

crates/amaru/src/stages/mod.rs
```diff
@@ -106,10 +109,47 @@ impl Default for Config {
...
+            const MIN_FD_PER_INSTANCE: u64 = 10;
+
+            let max_open_files = (soft_limit.saturating_sub(FD_RESERVE_FOR_OTHERS)
+                / ESTIMATED_NUMBER_OF_ROCKSDB_INSTANCES)
+                .max(MIN_FD_PER_INSTANCE);
+            let max_open_files = max_open_files.min(i32::MAX as u64) as i32;
+
+            if soft_limit < 65_535 {
+                warn!(
+                    "Setting max open files per RocksDB instance to {}. Consider increasing ulimit for better performance. Current (soft/hard): {}/{}",
```
jeluard: How about crashing? It's more than perf, it will prevent the node to operate.
We should also find a way to have a better estimate for this value.

pgrange: Without this PR, Amaru will crash on mainnet because it will hit file-descriptors limit. But with this code, I tested Amaru, on mainnet, with only 39 file descriptors per rocksdb instance. It did not crash. I think tocksdb manages to deal with this limited supply of file-descriptor, at the price of lower performance (having to open/close more often its data files) but without crashing.

Regarding the estimate of the value, you're right, it would be awesome to find a better way. I can't think of any strategy, though. It's a very global, process wise setup.

Each time you instanciates rocksDb in the code, you have to ensure that it will not exhaust the number of file descriptors of the whole process. The only way is to give it a hard limit: _do not use more than that amount of file descriptors_. But then, rocksDb will do whatever it wants regarding FD, staying inside this boundary. That means that we have to estimate how much, maximum, rocksDb instances will co-exist at any time during the process execution and use that to limit each of them fairly.

That's the only strategy I've found: at the _main_ level, decide how much rocksdb instances you want, divide the number of file desriptors available by this number and distribute them fairly between the instances.

I though of an other strategy to have something less global: every time we instantiate rocksDb, we look at how much file descriptors are available, right now, to the process and we _allocate_ that number, or a part of that to that new instance. But that would not work for, at least, two reasons:
1. How could we estimate how much file descriptors we should reserve for the future instances that we don't know about yet, in this part of the program?
2. More importantly, it's not because a rocksDb instance has been allowed n FD that it will use them so, when we look at how much FD are available, this doesn't consider FD that existing instance may use in the future but don't yet and so we end with over allocating FD and risking a crash at some point.

pgrange: Hence my best guess: let's keep 100 FD for other use and allow rocksDb to use the remaining ones and since, in that process, we instanciate 4 of them, lets divide by 4 🤷

crates/amaru/Cargo.toml
pgrange: I'm not a Windows master but it is my understanding that this limit notion per process  is really unix only, Windows does not have this.

In fact, form what I can see, Windows does not even has a notion of FD but of Handles instead and they're not limited per process but globally, depending on memory available and other things like that.

So the problem here is really unix-y and that's why this package is only supported for unix systems. I've paid attention to ensure that the code is not impacted for Windows.

jeluard: Good!

crates/amaru/src/stages/mod.rs
jeluard: Hmm ok I read too quickly.

The `- 100 / 4` part feels a bit too magical / brittle somehow. Maybe we should just settle on some fixed number? Also maybe a candidate for some config available to the end user? Now feels weird to expose some `rocksdb` specifics.

Quantumplation: would it be possible to instrument this as a metric, too, if it's not already? that way we could monitor how many actually get used and adjust the limit over time until we find the sweet spot.

KtorZ: I think the overall strategy makes sense here; and I also agree with @Quantumplation; that would be a good one to have a metric to better inform users about a good value. We might decide to crash if `ulimit` is simply set _too low_ because indeed, performances become a security concern for a block producing node. So we need to ensure a certain perf target. 

The only thing that puzzles me here in this code is ... why `ESTIMATED_NUMBER_OF_ROCKSDB_INSTANCES: u64 = 4` ? 

I don't think we ever have more than 2 instances of rocksdb running; are we?

crates/amaru/src/stages/mod.rs
```diff
@@ -105,10 +108,47 @@ impl Default for Config {
...
+                    max_open_files, soft_limit, hard_limit
+                );
```
KtorZ:
```suggestion
                warn!(
                    %max_open_files,
                    %soft_limit,
                    %hard_limit,
                    "Setting max open files per RocksDB instance; consider increasing ulimit for better performance.",
                );
```

Please prefer structured traces 🙏

crates/amaru/src/stages/mod.rs
pgrange: @KtorZ I should have taken better notes about that. For some reason I stumbled upon this number, 4, before but looking into the code to find it again it seems I'm wrong BUT.

From my understanding, we open up at lease 3 rocksDb instances in the daemon.

In [make_ledger](https://github.com/pgrange/amaru/blob/29a1b281b2e9a2f42e49c25f6eb1a13f48f120f0/crates/amaru/src/stages/mod.rs#L408) we create two instances through `RocksDB::new()` and `RocksDBHistoricalStores::new()`.

In [make_chain_store](https://github.com/pgrange/amaru/blob/29a1b281b2e9a2f42e49c25f6eb1a13f48f120f0/crates/amaru/src/stages/mod.rs#L326) we create a third instance with `RocksDBStore::new()`.

KtorZ: Ho. Yes, I forgot about the chain store for a second. I don't think this one is very greedy though in terms of open files. The worst is probably the historical store, since we go over the *entire* store every epoch.

pgrange: Metric added

pgrange: Also, you convinced me that, actually, the correct behavior is to crash and not try some smart computation.

Indeed, we can have 3 instances of RocksDb that need some file descriptors but we also have metrics collection that can eat a bunch of them.

At the end of the day, I've tried and played with Amaru on mainnet with and without metrics collection and decided to pick 1000 as the lowest acceptable limit regarding file descriptor.

I've added a file descriptor metric and seen that it never goes higher than 420 so I think we're ok with 1000... until one day,  maybe, we'll have to change it :)

pgrange: Keeping the PR as draft as I need to tinker with the Windows build and probably other CI runs. I'll keep you posted on Discord when it's ready for review.

pgrange: Ready for review

PR #453, opened 2025-09-15
Nview Metrics

yHSJ:
This PR introduces exporting the metrics that nview consumes

## Summary by CodeRabbit

* **New Features**
  * Emits OpenTelemetry ledger metrics (block height, txs processed, slot, slot‑in‑epoch, epoch, chain density) during block validation to a configured Meter.
* **Refactor**
  * Replaces legacy metrics stage with a Meter/resource-based flow and adds a shared amaru-metrics crate to centralize metric types and recording.
* **Documentation**
  * Adds README for the metrics crate and an engineering decision record explaining the new design.
* **Style**
  * Minor formatting cleanup in the Node.js example.


crates/amaru-ledger/src/state.rs
```diff
@@ -541,6 +541,26 @@ impl<S: Store, HS: HistoricalStores> State<S, HS> {
...
+            .view_front()
+            .map(|state| &state.anchor.0)
+            .unwrap_or(&Point::Origin)
+            .slot_or_default();
+
+        if k_slot > latest_slot {
+            0f64
+        } else {
+            // Add one to the number of blocks in the volatileDB because we are including the `Point` in the chain density
+            (self.volatile.len() as f64 + 1f64)
```
KtorZ: Note that apart from _just after the startup_ (literally within the second that follows), self.volatile.len() is always equal to `k`, by construction.

crates/amaru-ledger/src/state.rs
```diff
@@ -541,6 +541,26 @@ impl<S: Store, HS: HistoricalStores> State<S, HS> {
         Ok(result)
     }
 
+    /// Calcualte chain density over the last `k` blocks (or oldest block in the volatileDB) given some `Point`.
```
```suggestion
    /// Calculate chain density over the last `k` blocks (or oldest block in the volatileDB) given some `Point`.
```

crates/amaru-ledger/src/state.rs
```diff
@@ -541,6 +541,26 @@ impl<S: Store, HS: HistoricalStores> State<S, HS> {
...
+    pub fn chain_density(&self, point: &Point) -> f64 {
+        let latest_slot = point.slot_or_default();
+        let k_slot = self
+            .volatile
+            .view_front()
+            .map(|state| &state.anchor.0)
+            .unwrap_or(&Point::Origin)
+            .slot_or_default();
+
+        if k_slot > latest_slot {
```
```suggestion
        if k_slot >= latest_slot {
```

KtorZ: Seems like the `else` branch will panic if they're equal.

crates/amaru-ledger/src/state.rs
```diff
@@ -541,6 +541,26 @@ impl<S: Store, HS: HistoricalStores> State<S, HS> {
...
+            0f64
+        } else {
+            // Add one to the number of blocks in the volatileDB because we are including the `Point` in the chain density
+            (self.volatile.len() as f64 + 1f64)
+                / (u64::from(latest_slot) as f64 - u64::from(k_slot) as f64)
```
KtorZ: The dimensional analysis doesn't quite check for this one; this is a ratio of number of _blocks_ over a number of _slots_. I believe the numerator should be:

- k  (or self.volatile.len(); same thing)

and the denominator:

- (latest_slot - k_slot) * active_slot_coeff_inverse 

So said differently, we compute the actual number of blocks created over the "maximum" that could have been created during that time. Ideally, we want the numerator to be as close to k * active_slot_coeff_inverse (43200 on mainnet) as possible.

No sure about the `+1` either.

crates/amaru-metrics/src/ledger.rs
```diff
@@ -0,0 +1,108 @@
...
+        let block_height = BLOCK_HEIGHT.get_or_init(|| {
+            meter
+                .u64_gauge("cardano_node_metrics_blockNum_int")
+                .with_description("block height of latest processed block")
+                .with_unit("int")
+                .build()
+        });
+        let txs_processed = TXS_PROCESSED.get_or_init(|| {
+            meter
+                .u64_counter("cardano_node_metrics_txsProcessedNum_int")
```
KtorZ, line 75: it's really a mix of snake case and camel case, is it?

crates/amaru/src/stages/ledger/mod.rs
```diff
@@ -161,8 +168,22 @@ impl<S: Store + Send, HS: HistoricalStores + Send> ValidateBlockStage<S, HS> {
...
-                Ok(Ok(*block_height))
+
+                Ok(Ok(LedgerMetrics {
+                    block_height: *block_height,
+                    txs_processed,
+                    slot,
+                    slot_in_epoch,
+                    epoch,
+                    density,
+                }))
```
KtorZ: Why return anything? Seems like we could just record metrics right here?

crates/amaru-ledger/src/state.rs
yHSJ: I cleaned up the logic and removed the `+1` (doesn't makes sense you're right).

Your statement makes sense, but that is not typically how chain density is calcula
ted from what I've seen. Chain density is a measure of how many slots are "filled" out of all slots, as a percentage.

So it would be `blocks / slots`, within some interval (`k`, or the longest we have).

[Reference Implementation](https://github.com/IntersectMBO/cardano-node/blob/master/cardano-node/src/Cardano/Tracing/Tracers.hs#L1776-L1799)

crates/amaru-ledger/src/state.rs
yHSJ: Yeah, that shouldn't be an issue

crates/amaru-metrics/src/ledger.rs
yHSJ: Yes, yes it is 🙃

crates/amaru/src/stages/ledger/mod.rs
yHSJ: We could, but I was trying to avoid passing around the metrics downstream port into the ledger as much as I could, and trying to avoid just making everything async.

crates/amaru-ledger/src/state.rs
KtorZ: _sigh_, I guess that's one way to look at it. Though that means an expected "ideal" value for the chain density is ~5%, which just sounds extremely weird to me. 

I could trace back where this ... oddity comes from; and it seems that back in 2019, people just translated the Ouroboros Classic "formula" to Praos'. 

https://github.com/IntersectMBO/cardano-node/issues/250

This worked in Classic because you expect (in nominal conditions) one block per slot; and so an ideal density is equal to `1`. But with Praos and the notion of active slot, that doesn't hold anymore and it seems like no one bothered about doing a dimensional analysis. 

I'll likely open an issue on the cardano-node... though I don't know if I have the energy to fight for it.

crates/amaru-metrics/src/ledger.rs
```diff
@@ -0,0 +1,111 @@
...
+                .build()
+        });
+
+        block_height.record(self.block_height, &[]);
+        txs_processed.add(self.txs_processed, &[]);
+        slot_num.record(self.slot, &[]);
+        epoch.record(self.epoch, &[]);
+        slot_in_epoch.record(self.slot_in_epoch, &[]);
+        density.record(self.density, &[]);
+    }
```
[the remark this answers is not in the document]
yHSJ: This is not true, As you can see in `amaru-ledger/src/state`, `tx_processed` is just the number in a block, instead of the current total, so it's growing as it should
```rs
                let txs_processed = block.transaction_bodies.len() as u64;
```

crates/amaru-consensus/src/consensus/effects/metrics_effects.rs
```diff
@@ -0,0 +1,46 @@
...
+impl RecordMetricsEffect {
+    pub fn new(event: MetricsEvent) -> Self {
+        Self { event }
+    }
+}
+
+impl ExternalEffect for RecordMetricsEffect {
+    fn run(self: Box<Self>, resources: Resources) -> BoxFuture<'static, Box<dyn SendData>> {
+        Self::wrap(async move {
+            if let Ok(meter) = resources.get::<Arc<Meter>>() {
```
etorreborre: The convention is to declare a type alias for resources. Here that would be `type ResourceMeter`.

crates/amaru-consensus/src/consensus/effects/metrics_effects.rs
```diff
@@ -0,0 +1,46 @@
...
+                self.event.record_to_meter(&meter);
+            }
+            // No-op if there is no meter, since metrics collecting is optional
+            // TODO: do we want to have some state so we can error
+            // if there is no meter when we expect to have one?
```
etorreborre: We should probably leave it as it is here to avoid paying the cost of a check every time we don't need to send metrics.

crates/amaru-ledger/src/state.rs
```diff
@@ -657,6 +685,23 @@ impl<S: Store, HS: HistoricalStores> State<S, HS> {
...
+    pub fn chain_density(&self, point: &Point) -> f64 {
+        let latest_slot = point.slot_or_default();
+        let k_slot = self
+            .volatile
+            .view_front()
+            .map(|state| &state.anchor.0)
+            .unwrap_or(&Point::Origin)
+            .slot_or_default();
+
+        if k_slot >= latest_slot {
```
etorreborre, line 699: Could we add a unit test for this code?

crates/amaru-metrics/README.md
```diff
@@ -0,0 +1,3 @@
+# Amaru Metrics
+
+Amaru Metrics is a small crate that contains interfaces and shared logic related to collecting and emitting metrics throighout Amaru crates.
```
etorreborre: typo: `s/throighout/throughout`

engineering-decision-records/015-recording-cardano-metrics.md
```diff
@@ -13,26 +13,10 @@ nView reads Prometheus metrics from a `/metrics` endpoint, so if we collect and
...
 
-```rs
-pub trait Metric: Send + Sync {
-    /// Update underlying OTel instruments (counters, histograms, gauges).
-    fn record(&self, meter: &opentelemetry::metrics::Meter);
-}
-```
+## Discussion
 
+We had previously discused using a generic `Metric` trait that a metrics stage could consume without carying about specifics using dynamic dispatch. This allowed the metrics to be owned by the relevant stage and modifying said metrics required no changes to any other pieces of logic.
```
etorreborre: typos: `s/discused/discussed`, `s/carying/caring`.

engineering-decision-records/015-recording-cardano-metrics.md
```diff
@@ -13,26 +13,10 @@ nView reads Prometheus metrics from a `/metrics` endpoint, so if we collect and
 
 ## Decision
 
-We will propogate metrics via stage events instead of passing around a large record to each stage to be mutated. This metrics stage will be agnostic to the specific metrics themselves, instead relying on a `Metrics` trait, similar to the following:
+We will propogate metrics via a stage effect instead of passing around a large record to each stage to be mutated. There is a new `amaru-metrics` crate that holds the structs and enums relevant for the metrics collection.There is a `MetricsEvent` enum, with a variant for each specific type of metric (`LedgerMetrics`, `ConsensusMetrics`, etc.). Each of those metrics implements a `MetricRecorder` trait, which is called from `RecordMetricsEffect` effect.
```
etorreborre: typo: add a `.` before `There is a ...`

crates/amaru-ledger/src/state.rs
yHSJ: Yes, I'll work on some unit tests this week, but I don't think it's blocking this PR.

engineering-decision-records/015-recording-cardano-metrics.md
```diff
@@ -13,26 +13,10 @@ nView reads Prometheus metrics from a `/metrics` endpoint, so if we collect and
...
-## Consequences
-
-- Each component of the Amaru node can fully own its own metrics
-
-- It's very simple to introduce new metrics, and the metrics stage doesn't have to change frequently
-
-- We can expose a set of metrics that has some intersection with the metrics exposed by the cardano-node to prometheus, allowing existing tooling to support Amaru.
-
-- One difficulty introduced by this trait decision is that it makes static analysis extremely difficult, whether by a human or a tool. The convention of using the `metrics.rs` file is an attempt to alleviate that, so one does not have to search the whole codebase for metrics.
+While this worked well with the `gasket` framework, it was incompatible with `Pure Stage` due to the required `pure_stage::SendData` trait. As a result, we had to compromise and fallback to the `MetricsEvent` enum solution, which still limits the amount of changes needed to add/remove metrics.
```
rkuhn, line 22: So basically dynamic dispatch became more cumbersome to funnel it through the effects hatch, while on the flip side the injection of the metrics collector now comes for free, does that sound like a fair summary?

PR #457, opened 2025-09-16
fix: clear current AND past delegations on DRep deregistration in v9

KtorZ:
See also: https://github.com/pragma-org/amaru/wiki/log-::-2025%E2%80%9009#voting-stake-inconsistency-and-drep-delegation-disappearances

## Summary by CodeRabbit

* **New Features**
  * Persist and index delegations to delegate-representatives during import/save; reports processed delegation counts.
  * Much faster bootstrap/imports via larger batch processing.
  * New transactional API to submit DRep delegations.

* **Improvements**
  * Enhanced runtime logging and instrumentation across startup, header import, governance and ratification flows.
  * Better display formatting for transaction/certificate pointers.
  * Stricter collateral script-lock detection for validation.

* **Data**
  * Extensive preprod DRep delegator additions and several new/updated preprod pools.

* **Chores**
  * CI: bumped ledger-cache key version.


conformance-tests/scripts/generate.mjs
```diff
@@ -93,6 +97,7 @@ const poolIds = Object.keys(distr.stakePools).sort();
 withStream(`summary__stake_distribution_${epoch}.snap`, (stream) => {
   stream.write("---\n")
   stream.write(`source: ${source}\n`)
+  stream.write(`assertion_line: 100\n`)
```
jeluard, line 100: Leftover?

crates/amaru-ledger/src/bootstrap.rs
```diff
@@ -315,11 +314,21 @@ pub fn import_initial_snapshot(
...
 
-    import_dreps(db, point, era_history, &protocol_parameters, epoch, dreps)?;
+    import_proposals(db, point, era_history, &protocol_parameters, &proposals)?;
 
-    import_proposals(db, point, era_history, &protocol_parameters, proposals)?;
+    import_votes(db, point, era_history, &protocol_parameters, proposals)?;
 
-    import_votes(db, point, era_history, &protocol_parameters, votes)?;
+    // NOTE: It's important to import dreps *after* votes, because voting dreps from imported votes
+    // will be get their expiry updated, However:
```
jeluard:
```suggestion
    // will get their expiry updated, However:
```

crates/amaru-ledger/src/bootstrap.rs
```diff
@@ -531,6 +542,17 @@ fn import_dreps<S: Store>(
                     registered_at,
                 };
 
+                state.delegators.to_vec().into_iter().for_each(|delegator| {
```
jeluard: `for_each` with mutation? Can't it be a `map` `extend`ing `delegations` ?

crates/amaru-ledger/src/governance/ratification/proposals_forest.rs
```diff
@@ -421,14 +421,15 @@ impl ProposalsForestCompass {
         //   or nice polls; but as soon as one of the other is encountered; EVERYTHING (including
         //   treasury withdrawals and parameters changes) is postponed until the next epoch.
         if forest.is_interrupted {
+            debug!("high-impact proposal was enacted; pausing ratification for this epoch");
             return None;
         }
 
         loop {
-            let result: Option<(
+            let mut process_next = || -> Option<(
```
jeluard: Can't it be defined outside the loop? Unclear if there is some perf impact.

crates/amaru-ledger/src/rules/transaction/collateral.rs
```diff
@@ -225,7 +223,7 @@ where
             .lookup(collateral)
             .ok_or_else(|| InvalidCollateral::UnknownInput(collateral.clone().into()))?;
 
-        if output.address.has_script() {
+        if is_locked_by_script(&output.address) {
```
jeluard, line 227: Hopefully it will be caught by the tests that are being worked on.

crates/amaru-ledger/src/summary/stake_distribution.rs
```diff
@@ -149,18 +147,27 @@ impl StakeDistribution {
...
                                 } = dreps.get(&drep)?;
 
-                                if protocol_parameters.protocol_version <= PROTOCOL_VERSION_9 {
-                                    if &Some(since) > previous_deregistration {
-                                        Some(drep)
-                                    } else {
-                                        None
-                                    }
-                                } else if &since >= registered_at {
+                                // NOTE: This is subtle. Delegation to a non-existing DRep was
```
jeluard: 🤯

crates/amaru-ledger/src/summary/stake_distribution.rs
```diff
@@ -236,20 +243,16 @@ impl StakeDistribution {
                     return match pools.get_mut(&pool_id) {
                         None => false,
                         Some(pool) => {
+                            let stake = account.lovelace;
                             // NOTE: Governance deposits do not count towards the pools' stake.
                             // They are only counted as part of the voting power.
-                            active_stake += &account.lovelace;
-                            pool.stake += &account.lovelace;
-
-                            // NOTE: Because votes are ratified with an epoch delay and using the
```
jeluard, line 244: removal: Worth a comment?

crates/amaru-stores/src/rocksdb/ledger/columns/accounts.rs
```diff
@@ -60,20 +157,37 @@ pub fn add<DB>(
...
-            )
-        };
+            unreachable!(
+                "attempted to create an account without a deposit: account={:?}",
+                credential
+            );
+        }?;
+
+        // NOTE:
+        // In protocol version 9, a subtle bug causes registered DRep to retain their past
```
jeluard, line 176: All those might be hard to track the day it will be removed. Maybe we should standardize some tags easily grepable e.g. PROTOCOL_VERSION_9 (even when th const actually isn't used)

crates/amaru-stores/src/rocksdb/ledger/columns/dreps_delegations.rs
```diff
@@ -0,0 +1,228 @@
...
+/// Name prefixed used for storing Account -> DRep & DRep -> Account entries. UTF-8 encoding for
+/// "dlg".
+///
+/// We leave one byte up, so that RocksDB will partition files into ~256 buckets. This allows to
+/// reduce the burden of iterating over *all* dreps, but only a subset of those. On mainnet, we
+/// count about 200k delegators, spread over 1.5k dreps. So iterating on the entire column can be a
+/// little expensive.
+///
+/// With this extra byte, we should have amortised performances equivalent to iterating over ~1K
+/// entrie, which seems much more reasonable. Besides, that work is *bounded* anyway, because it
```
jeluard:
```suggestion
/// entries, which seems much more reasonable. Besides, that work is *bounded* anyway, because it
```

crates/amaru-stores/src/rocksdb/ledger/columns/dreps_delegations.rs
```diff
@@ -0,0 +1,228 @@
...
+    stake_credential_hash,
+};
+use amaru_ledger::store::{StoreError, columns::unsafe_decode};
+use std::collections::BTreeSet;
+use tracing::{Level, debug, instrument, warn};
+
+/// Name prefixed used for storing Account -> DRep & DRep -> Account entries. UTF-8 encoding for
+/// "dlg".
+///
+/// We leave one byte up, so that RocksDB will partition files into ~256 buckets. This allows to
```
jeluard, line 28: Don't quite follow this trick, but looks like worth adding to our (non-existent) pattern FAQ for RocksDB.

crates/amaru-stores/src/rocksdb/ledger/columns/dreps_delegations.rs
```diff
@@ -0,0 +1,228 @@
...
+        // However, Alice unregistration has already happened and so, we must not record this
+        // binding. It suffices to check whether Alice's last de-registration pointer, if any,
+        // came after the delegation. In which case, we have technically already cleared up the
+        // DRep relation due to the v9 bug. Note that:
+        //
+        // 1. We never reset the last de-registration pointer to None. Even when the DRep
+        //    re-registered. So even the case where Alice only re-register prior to the delegation
+        //    to Bob, we would not retain the link.
+        //
+        // 2. It doesn't matter if Alice re-regiters and unregisters multiple times since any new
```
jeluard:
```suggestion
        // 2. It doesn't matter if Alice re-resgiters and unregisters multiple times since any new
```

crates/amaru-stores/src/rocksdb/ledger/columns/dreps_delegations.rs
```diff
@@ -0,0 +1,228 @@
...
+    drep: &DRep,
+    delegator: &StakeCredential,
+) -> Result<(), StoreError> {
+    let drep = match drep {
+        DRep::Key(hash) => StakeCredential::AddrKeyhash(*hash),
+        DRep::Script(hash) => StakeCredential::ScriptHash(*hash),
+        DRep::Abstain | DRep::NoConfidence => return Ok(()),
+    };
+
+    let mut key = PREFIX.to_vec();
```
jeluard: Duplicated code, extract as a function?

crates/amaru-stores/src/rocksdb/ledger/columns/dreps_delegations.rs
```diff
@@ -0,0 +1,228 @@
...
+    let mut delegators = BTreeSet::new();
+
+    let mut prefix = PREFIX.to_vec();
+    prefix.extend_from_slice(as_value(drep).as_slice());
+
+    // NOTE: The invariant is preserved because the first 3 bytes are PREFIX, which is
+    // not all 0xFF.
+    let upper_bound = into_next_prefix(prefix.clone());
+
+    // NOTE: We must set an explicit upper bound on this one, without what we end up sometimes
```
jeluard: Looks similar to what Pascal was mentioning today, hopefully we can get clarity on that.

conformance-tests/scripts/generate.mjs
KtorZ: Nope. That's also generated by `insta` nowadays; and I was simply trying to better match the generated outputs (hoping it would ease the diff; but with mainnet data, insta is just struggling to provide meaningful diffs if there more than 5-6 mismatches; it just prints the ENTIRE snapshot on error... so 6.8M lines)

crates/amaru-ledger/src/bootstrap.rs
KtorZ: Most likely yes!

crates/amaru-ledger/src/governance/ratification/proposals_forest.rs
KtorZ: I believe it can; modulo arguments passing. I would expect the compiler to inline it anyway, so I don't think perf is really a valid argument here.

crates/amaru-ledger/src/rules/transaction/collateral.rs
KtorZ: I notified @yHSJ already, and he's on it!

crates/amaru-stores/src/rocksdb/ledger/columns/accounts.rs
KtorZ: I was actually thinking down this line the other day: we should _name_ our NOTEs. Having notes with titles or tags, would allow extracting them and possibly organizing them in an automated fashion. So I don't think that's a bad idea here indeed.

crates/amaru-stores/src/rocksdb/ledger/columns/dreps_delegations.rs
KtorZ: The prefix size is *fixed per db*, and in our case is set to 4. And RocksDB partitions files based on prefixes. So with 3 bytes fixed, it lives us with 1 byte that will vary -> 256 values. So it behaves _as if_ we had 256 different namespaces from RocksDB perspective. And it should create 256 partitions so that. Iterations will never go over multiple partitions (since we need only to iterate over all accounts of a drep; so the first 3 + 28 bytes will be fixed and live all in the same partition).

Also... I haven't checked that behavior in practice, so it's mostly theoretical.

crates/amaru-ledger/src/summary/stake_distribution.rs
KtorZ: Erf. I am still not sure whether this one is sound or not. It's not easy to reason about. And we have a similar issue for DReps. Here I was mostly trying to match the data from the generated snapshots; although that may not be what we want... Here's an attempt to explain:

- Throughout all this code, I've been trying to provide a *clear* (lol!) definition of a "stake distribution snapshot for epoch e": it is the stake distribution exactly at the end of `e`, before the beginning of `e + 1`. 

- Specific operations such as pool reaping (removing now-retired pools) happen at the beginning of the epoch. Which means that a pool that retires in `e + 1` is still technically part of the snapshot at epoch `e`. 

- Governance makes it harder, because it mixes data from multiple epochs creating behaviors that are complex to reason about:

1. First, ratification happens with a delay of one epoch and using data from (at least) 2 epochs. The ratification of proposals & votes submitted in epoch `e` happens during the transition from `e + 1` to `e + 2`. The stake distribution for pools that is used for determining voting power is the one at epoch `e`. However, it is the delegation state at the beginning of `e + 1` that is used. 

2. Because of that, a pool that retires in `e + 1` will still be present in the stake distribution of `e`, but will have a virtual voting power of `0` (because there's no longer any account delegated to it in `e + 1`). 

So, TL; DR; that change is very likely _wrong_ here, and the problem comes from *the original data* which simply return the raw stake distribution without acknowledge the retirement. We shall fix the data generation instead, and revert this change here. 

I will also try to produce more ratification data; so that we can compare the *actual voting stake distribution* during a ratification procedure (taking into account retirements and other events happening at epoch boundaries).

PR #463, opened 2025-09-18
Docker testnet feedbacks

pgrange:
Some proposals to improve the docker/testnet part.


docker/testnet/docker-compose.yaml
```diff
@@ -99,6 +99,11 @@ services:
       - AMARU_NETWORK=testnet_42
     volumes:
       - a1-state:/srv/amaru
+    depends_on:
+      cardano-loader:
+        condition: service_completed_successfully
+      p1:
```
abailly, line 105: Actually, I purposely don't want `amaru` to depend on a well-behaving cardano-node: It should start even if connections fail and keep trying.

docker/testnet/docker-compose.yaml
```diff
@@ -107,6 +112,11 @@ services:
       - AMARU_NETWORK=testnet_42
     volumes:
       - a2-state:/srv/amaru
+    depends_on:
+      cardano-loader:
+        condition: service_completed_successfully
+      amaru-1:
```
abailly, line 118: Same remark.

docker/testnet/docker-compose.yaml
```diff
@@ -99,6 +99,11 @@ services:
       - AMARU_NETWORK=testnet_42
     volumes:
       - a1-state:/srv/amaru
+    depends_on:
+      cardano-loader:
```
abailly, line 103: This is not needed as it's implied in the `x-amaru` fragment: the `Amaru-loader` depends on `Cardano-loader`.

docker/testnet/docker-compose.yaml
```diff
@@ -107,6 +112,11 @@ services:
       - AMARU_NETWORK=testnet_42
     volumes:
       - a2-state:/srv/amaru
+    depends_on:
+      cardano-loader:
```
abailly, line 116: same as above

abailly (commented):
Thanks for the suggestions but I don't think we should apply them

PR #464, opened 2025-09-18
refactor: move forward chain to pure stage

etorreborre:


etorreborre:
@coderabbitai review please

crates/amaru/src/stages/consensus/forward_chain/tcp_forward_chain_server.rs
```diff
@@ -0,0 +1,204 @@
...
+            }
+            ForwardEvent::Backward(tip) => {
+                self.clients.send(ClientMsg::Op(ClientOp::Backward(Tip(
+                    tip.point().pallas_point(),
+                    tip.block_height(),
+                ))));
+                Ok(())
+            }
+        }
+    }
```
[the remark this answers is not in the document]
etorreborre: Done

crates/amaru/src/stages/consensus/forward_chain/tcp_forward_chain_server.rs
```diff
@@ -0,0 +1,211 @@
...
+        runtime: Arc<AcTokio>,
+        tcp_listener: TcpListener,
+        network_magic: u64,
+        max_peers: usize,
+        our_tip: HeaderTip,
+    ) -> anyhow::Result<Self> {
+        let clients = runtime
+            // FIXME: This is a temporary stop gap solution while we wait
+            // to refactor to use pure-stage. Acto library has as a capped
+            // size for the mailbox and drops incoming messages when it's
```
rkuhn, line 75: There’s [`send_wait`](https://docs.rs/acto/latest/acto/struct.ActoRef.html#method.send_wait) as well (just FYI).

simulation/amaru-sim/src/simulator/run.rs
```diff
@@ -315,3 +284,72 @@ impl BlockFetcher for FakeBlockFetcher {
...
+/// This implementation of ForwardEventListener sends the received events a Sender that can collect them
+/// to check them later.
+///
+/// A message id is maintained for each received event and each message is duplicated to the number of downstream peers.
+#[derive(Clone, Debug)]
+pub struct MockForwardEventListener {
+    node_id: String,
+    number_of_downstream_peers: u8,
+    sender: mpsc::Sender<Envelope<ChainSyncMessage>>,
+    msg_id: Arc<Mutex<u64>>,
```
rkuhn: should use `Arc<AtomicU64>` for such functionality (with `Ordering::Relaxed` because only the counter needs to be consistent)

simulation/amaru-sim/src/simulator/run.rs
```diff
@@ -315,3 +284,72 @@ impl BlockFetcher for FakeBlockFetcher {
...
+        Self {
+            node_id,
+            number_of_downstream_peers,
+            msg_id: Arc::new(Mutex::new(0)),
+            sender,
+        }
+    }
+}
+
+#[async_trait]
```
rkuhn, line 316: this is no longer necessary since a couple of Rust versions

simulation/amaru-sim/src/simulator/run.rs
```diff
@@ -315,3 +284,72 @@ impl BlockFetcher for FakeBlockFetcher {
...
+                    hash: Bytes {
+                        bytes: tip.hash().as_slice().to_vec(),
+                    },
+                },
+            }
+        }
+
+        for i in 1..=self.number_of_downstream_peers {
+            let envelope = {
+                let mut msg_id = self.msg_id.lock().await;
```
rkuhn: this would then become `self.msg_id.fetch_add(1, Ordering::Relaxed)`; if consecutive numbers are important, add `self.number_of_downstream_peers` before the loop and add `i - 1` to the result

crates/amaru/src/stages/consensus/forward_chain/tcp_forward_chain_server.rs
```diff
@@ -0,0 +1,211 @@
...
+#[async_trait]
+impl ForwardEventListener for TcpForwardChainServer {
+    async fn send(&self, event: ForwardEvent) -> anyhow::Result<()> {
+        match event {
+            ForwardEvent::Forward(header) => {
+                let mut our_tip = self
+                    .our_tip
+                    .lock()
+                    .map_err(|e| anyhow::anyhow!("Mutex poisoned: {}", e))?;
+                *our_tip = header.as_header_tip();
```
rkuhn: probably irrelevant in this case but keeping this mutex locked after this line feels dirty

crates/amaru/src/stages/consensus/forward_chain/tcp_forward_chain_server.rs
```diff
@@ -0,0 +1,211 @@
...
+use std::sync::Mutex;
+use tokio::net::TcpListener;
+
+pub const EVENT_TARGET: &str = "amaru::consensus::forward_chain";
+
+/// The TcpForwardChainServer listens for incoming TCP connections from peers
+/// and spawns a client protocol handler for each accepted connection.
+/// It also implements the ForwardEventListener trait to receive forward events
+/// and forward them to all connected peers.
+#[derive(Clone)]
```
rkuhn: Why does this need to support cloning? From the description it seems like it should exist only once.

crates/amaru-consensus/src/consensus/effects/network_effects.rs
```diff
@@ -0,0 +1,102 @@
...
+use pure_stage::{ExternalEffect, ExternalEffectAPI, Resources};
+use serde::{Deserialize, Serialize};
+use std::fmt::Display;
+use std::sync::Arc;
+
+pub type ResourceForwardEventListener = Arc<dyn ForwardEventListener + Send + Sync>;
+
+/// A listener interface for forward events (new headers or rollbacks).
+/// These events are either caught for tests or forwarded to downstream peers (see the TcpForwardEventListener implementation).
+#[async_trait]
```
rkuhn, line 31: not needed in our Rust version anymore

crates/amaru-consensus/src/consensus/effects/network_effects.rs
```diff
@@ -0,0 +1,102 @@
...
+pub trait ForwardEventListener {
+    async fn send(&self, event: ForwardEvent) -> anyhow::Result<()>;
```
rkuhn, line 33: It might make sense to explore whether this should take `&mut self`, which would get rid of the `Arc<Mutex<...>>` in implementations. At least that’s what I’d expect for a listener — and it might not even have state in the production case because it’ll just use the interior mutability of some channel endpoints.

crates/amaru/src/stages/consensus/forward_chain/tcp_forward_chain_server.rs
```diff
@@ -0,0 +1,211 @@
...
+        runtime.with_rt(|rt| {
+            let our_tip_clone = our_tip.clone();
+            let clients_clone = clients.clone();
+            rt.spawn(async move {
+                loop {
+                    // due to the signature of TcpListener::accept, this is the only way to use this API
+                    // in particular, it isn’t possible to poll for new peers within the `schedule` method
+                    match PeerServer::accept(&tcp_listener, network_magic).await {
+                        Ok(peer) => {
+                            let our_tip = our_tip_clone.lock().unwrap().clone();
```
rkuhn: This seems off: why would we inject into the connection notification the current state of the actor (or something close to it), when the client will tell us later where it is at? Let the information flow naturally to the one place where the pieces can be combined and acted upon.

crates/amaru/src/stages/consensus/forward_chain/tcp_forward_chain_server.rs
```diff
@@ -0,0 +1,211 @@
...
+    }
+}
+
+async fn client_supervisor(
+    mut cell: ActoCell<ClientMsg, impl ActoRuntime, anyhow::Result<()>>,
+    store: Arc<dyn ChainStore<Header>>,
+    max_peers: usize,
+) {
+    let mut clients = BTreeMap::new();
+    while let Some(msg) = cell.recv().await.has_senders() {
```
rkuhn, line 162: `has_senders` will always remain true because of the ActoRef held in the listener task. If we want to shut this down, we’ll need a dedicated ClientMsg for that purpose.

crates/amaru/src/stages/consensus/forward_chain/tcp_forward_chain_server.rs
```diff
@@ -0,0 +1,211 @@
...
+}
+
+impl TcpForwardChainServer {
+    /// Creates a new TcpForwardChainServer instance:
+    ///
+    ///  - Start an Acto runtime
+    ///  - Bind a TCP listener to the given address
+    ///  - Spawn the client supervisor actor
+    #[expect(clippy::expect_used)]
+    #[expect(clippy::unwrap_used)]
```
rkuhn: in production we should always use `.expect` to better mark the context of the error

rkuhn (commented):
it looked a bit large at first, but is mostly straight-forward — looks good apart from some nitpicks

crates/amaru/src/stages/consensus/forward_chain/tcp_forward_chain_server.rs
etorreborre: Isn't that the place where we communicate our tip to the client upon connection? Are you saying that this is not necessary?

crates/amaru-consensus/src/consensus/effects/network_effects.rs
etorreborre: See my comment above.

crates/amaru/src/stages/consensus/forward_chain/tcp_forward_chain_server.rs
etorreborre: Right, not necessary.

crates/amaru/src/stages/consensus/forward_chain/tcp_forward_chain_server.rs
etorreborre: Fixed.

simulation/amaru-sim/src/simulator/run.rs
etorreborre: Done.

crates/amaru/src/stages/consensus/forward_chain/tcp_forward_chain_server.rs
etorreborre: Makes sense, fixed.

simulation/amaru-sim/src/simulator/run.rs
etorreborre: If I don't use the annotation I'm getting an error telling me that "the trait `ForwardEventListener` is not dyn compatible". I could revert to using a boxed future but since that what the annotation does under the hood, we might as well keep it, right?

simulation/amaru-sim/src/simulator/run.rs
etorreborre: Done.

crates/amaru-consensus/src/consensus/effects/network_effects.rs
etorreborre: I tried that but that fails to compile when used as a network effect:
```
error[E0596]: cannot borrow data in an `Arc` as mutable
  --> crates/amaru-consensus/src/consensus/effects/network_effects.rs:89:17
   |
89 |                 listener.send(self.event).await.map_err(|e| {
   |                 ^^^^^^^^ cannot borrow as mutable
```
The workaround would be to declare the `ResourceForwardEventListener` as an `Arc<Mutex<...>>` but then we're back to square one. What do you think?

crates/amaru/src/stages/consensus/forward_chain/tcp_forward_chain_server.rs
etorreborre: I don't think I changed anything compared to the previous logic in this PR. I propose that we keep it as it is since this is something that you are going to revisit anyway.

simulation/amaru-sim/src/simulator/run.rs
rkuhn: ah, I didn’t check what it does exactly, assumed that it adds associated types; good to know (I still prefer to make BoxFuture explicit in the signatures, but that’s a matter of taste)

crates/amaru-consensus/src/consensus/effects/network_effects.rs
rkuhn: Ah, right, I was in the wrong mental space at the time.

crates/amaru/src/stages/consensus/forward_chain/tcp_forward_chain_server.rs
rkuhn: okay

PR #465, opened 2025-09-19
fix: adjust testnet docker cluster to run antithesis

abailly:
This PR does some adjustments to the `docker/testnet` configuration for the sake of running it in Antithesis using the [anti command-line interface](https://app.radicle.xyz/nodes/seed.hydra.bzh/rad:z2a7Te5b28CX5YyPQ7ihrdG2EEUsC/tree/cli/README.md) tool developed by the HAL Team at CF.

* Trim down logging (in our first run we generated 150GB of logs...)
* Add `testnet.yaml` (the `anti cli` requires this file to be present atm)
* Remove non standard markdown from `README.md` (the `anti cli` parses this file 🤷 )
* adjust image names to match [published packages](https://github.com/orgs/pragma-org/packages)

## Summary by CodeRabbit

* **Refactor**
  * Unified logging into a single environment variable and simplified daemon startup behavior.

* **Chores**
  * Updated testnet docker-compose by removing the healthcheck and adding runtime performance options to the node.
  * Minor script formatting cleanup.

* **Documentation**
  * Tidied warning formatting in the testnet README.

### DIFF
--- a/docker/start_amaru.sh
+++ b/docker/start_amaru.sh
@@ -26,10 +26,10 @@ fi
 
 # keep stack traces for troubleshooting purposes
 export RUST_BACKTRACE=full
-export AMARU_LOG=debug
-export AMARU_TRACE=gasket=error,debug
 
-exec cargo run --profile dev -- --with-json-traces daemon \
+export AMARU_LOG=gasket=error,amaru=debug,amaru::stages::consensus::forward_chain=info,info
+
+exec cargo run --profile dev -- daemon \
       --peer-address "${AMARU_PEER_ADDRESS}" \
       --ledger-dir "${AMARU_LEDGER_DIR}" \
       --chain-dir "${AMARU_CHAIN_DIR}"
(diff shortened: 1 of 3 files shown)


docker/start_amaru.sh
```diff
@@ -26,10 +26,10 @@ fi
 
 # keep stack traces for troubleshooting purposes
 export RUST_BACKTRACE=full
-export AMARU_LOG=debug
-export AMARU_TRACE=gasket=error,debug
 
-exec cargo run --profile dev -- --with-json-traces daemon \
+export AMARU_LOG=gasket=error,amaru=debug,amaru::stages::consensus::forward_chain=info,info
```
jeluard, line 30: Isn't explicitly setting `amaru::stages::consensus::forward_chain` redundant?

jeluard (commented):
I didn't see this `adjust image names` in the changes

docker/start_amaru.sh
abailly: I don't think so, I tried locally and they were still very chatty.

abailly:
> I didn't see this `adjust image names` in the changes

Good point, I probably did it before.

PR #469, opened 2025-09-22
unpack test vectors

rrruko:
Second attempt at evaluating a set of test vectors derived from cardano-ledger's "ImpTest" suite in amaru.

Each vector contains an initial ledger state, a final ledger state, and a list of transactions, each tagged with a boolean flag indicating whether the transaction should pass validation and a slot timestamp:
```
[
  initialLedgerState : ledger_state,
  finalLedgerState : ledger_state,
  txes : [ * [transaction, bool, slot] ],
  description : string
]
```

Currently, 203 tests pass and 117 fail. As far as I can tell, all 117 failures are cases where we expect a transaction to fail but amaru accepts it.

## Summary by CodeRabbit

* **New Features**
  * Added a ledger conformance test framework to evaluate ledger state transitions and validate execution results.

* **Improvements**
  * Simplified ledger snapshot import flow by removing directory-based parameter selection.
  * Expanded public codec exports for broader API accessibility.
  * Build scripts now read NETWORK env var (defaults to "preprod").

* **Chores**
  * Added cardano-blueprint as a submodule.
  * Added dev-deps and test-generation build logic; updated .gitignore for test data.

### DIFF
--- a/crates/amaru-ledger/build.rs
+++ b/crates/amaru-ledger/build.rs
@@ -20,11 +20,11 @@ fn main() {
 
 #[allow(clippy::unwrap_used)]
 fn get_conformance_test_vectors() {
-    let test_dir = "tests/data/rules-conformance";
-    println!("cargo:rerun-if-changed={}", test_dir);
+    let test_dir = Path::new("tests/data/rules-conformance");
+    println!("cargo:rerun-if-changed={}", test_dir.to_str().unwrap());
 
     let mut files = Vec::new();
-    visit_dirs(Path::new(test_dir), &mut files);
+    visit_dirs(test_dir, &mut files);
 
     let out_dir = env::var("OUT_DIR").unwrap();
     let dest_path = Path::new(&out_dir).join("test_cases.rs");
@@ -35,20 +35,25 @@ fn get_conformance_test_vectors() {
         );
     }
 
+    let pparams_dir = test_dir.join("eras/conway/impl/dump/pparams-by-hash");
+
     let mut output = String::new();
     for path in files {
         if path.contains("pparams-by-hash") {
             continue;
         }
 
-        output.push_str(&format!("#[test_case(\"{}\")]\n", path));
+        output.push_str(&format!(
+            "#[test_case::test_case(\"{}\", \"{}\")]\n",
+            path,
+            pparams_dir.to_str().unwrap()
+        ));
     }
 
     output.push_str(
         r#"
-#[allow(clippy::unwrap_used)]
-pub fn rules_conformance_test_case(snapshot: &str) {
-    evaluate_vector(snapshot).unwrap()
+pub fn rules_conformance_test_case(snapshot: &str, pparams_dir: &str) -> Result<(), Box<dyn std::error::Error>> {
+    import_and_evaluate_vector(snapshot, pparams_dir)
 }
 "#,
     );
--- a/crates/amaru-ledger/tests/evaluate_ledger_states.rs
+++ b/crates/amaru-ledger/tests/evaluate_ledger_states.rs
@@ -22,47 +22,21 @@ pub mod tests {
     use amaru_ledger::{
         self, context::DefaultValidationContext, rules::transaction, store::GovernanceActivity,
     };
-    use once_cell::sync::Lazy;
-    use std::{collections::BTreeMap, env, fs, ops::Deref, path::PathBuf, sync::Mutex};
-
-    struct TestContext {
-        pparams_dir: PathBuf,
-    }
-
-    fn get_test_context() -> Result<TestContext, Box<dyn std::error::Error>> {
-        let working_dir = env::current_dir()?;
-        Ok(TestContext {
-            pparams_dir: working_dir.join(PathBuf::from("../../cardano-blueprint/src/ledger/conformance-test-vectors/eras/conway/impl/dump/pparams-by-hash/")),
-        })
-    }
-
-    static TEST_MUTEX: Lazy<Mutex<()>> = Lazy::new(Mutex::default);
-
-    #[allow(unused_imports)]
-    use test_case::test_case;
+    use std::{collections::BTreeMap, env, fs, ops::Deref, path::Path};
 
     // Tests cases are constructed in build.rs, which generates the test_cases.rs file
     include!(concat!(env!("OUT_DIR"), "/test_cases.rs"));
 
-    fn evaluate_vector(snapshot: &str) -> Result<(), Box<dyn std::error::Error>> {
-        let _shared = TEST_MUTEX.lock()?;
+    fn import_and_evaluate_vector(
+        snapshot: &str,
+        pparams_dir: &str,
+    ) -> Result<(), Box<dyn std::error::Error>> {
         let network = NetworkName::Testnet(1);
         let era_history = network.into();
-        match get_test_context() {
-            Ok(tc) => {
-                let vector_file = fs::read(snapshot)?;
-                let record: TestVector = cbor::decode(&vector_file)?;
-                let () = import_vector(record, era_history, &tc.pparams_dir)?;
-                Ok(())
-            }
-            Err(e) => {
-                tracing::warn!(
-                    "skipping vector evaluation because some env vars are missing: {}",
-                    e
-                );
-                Ok(())
-            }
-        }
+        let vector_file = fs::read(snapshot)?;
+        let record: TestVector = cbor::decode(&vector_file)?;
+        evaluate_vector(record, era_history, Path::new(pparams_dir))?;
+        Ok(())
     }
 
     #[derive(cbor::Decode)]
@@ -187,7 +161,7 @@ pub mod tests {
     }
 
     fn decode_segregated_parameters(
-        dir: &PathBuf,
+        dir: &Path,
         hash: &cbor::bytes::ByteSlice,
     ) -> Result<ProtocolParameters, Box<dyn std::error::Error>> {
         let pparams_file_path = fs::read_dir(dir)?
@@ -206,10 +180,10 @@ pub mod tests {
         Ok(pparams)
     }
 
-    fn import_vector(
+    fn evaluate_vector(
         record: TestVector,
         era_history: &EraHistory,
-        pparams_dir: &PathBuf,
+        pparams_dir: &Path,
     ) -> Result<(), Box<dyn std::error::Error>> {
         let mut decoder = cbor::Decoder::new(&record.initial_state);
         let (mut validation_context, pparams_hash, governance_activity) =


crates/amaru/Cargo.toml
```diff
@@ -28,6 +28,8 @@ minicbor.workspace = true
 opentelemetry-otlp.workspace = true
 opentelemetry.workspace = true
 opentelemetry_sdk.workspace = true
+pallas-addresses.workspace = true
+pallas-codec.workspace = true
```
KtorZ: Aren't those only test dependencies? Also, I believe they aren't needed because everything is in the kernel already. If they aren't, then re-export them from the kernel instead.

crates/amaru/tests/evaluate_ledger_states.rs
```diff
@@ -0,0 +1,223 @@
...
+
+fn peek_epoch(nes: &AnyCbor) -> Result<u64, Box<dyn std::error::Error>> {
+    let mut d = cbor::decode::Decoder::new(nes.raw_bytes());
+    let _ = d.array()?;
+    Ok(d.u64()?)
+}
+
+#[derive(cbor::Decode)]
+#[allow(dead_code)]
+struct NESVector {
```
KtorZ:
```suggestion
struct TestVector {
```

crates/amaru/tests/evaluate_ledger_states.rs
```diff
@@ -0,0 +1,223 @@
...
+    #[n(0)]
+    initial_nes: AnyCbor,
+    #[n(1)]
+    final_nes: AnyCbor,
+    #[n(2)]
+    transactions: Vec<AnyCbor>,
+    #[n(3)]
+    test_state: String,
```
KtorZ:
```suggestion
    initial_state: AnyCbor,
    #[n(1)]
    final_state: AnyCbor,
    #[n(2)]
    transactions: Vec<AnyCbor>,
    #[n(3)]
    title: String,
```

Please avoid "cryptic" acronyms. "New Epoch State" is Haskell's jargon.

crates/amaru/tests/evaluate_ledger_states.rs
```diff
@@ -0,0 +1,223 @@
...
+
+fn import_vector(
+    record: NESVector,
+    ledger_dir: &PathBuf,
+    era_history: &EraHistory,
+    pparams_dir: &PathBuf,
+) -> Result<(), Box<dyn std::error::Error>> {
+    if record.initial_nes[0] == 0x80 {
+        return Ok(());
+    }
```
KtorZ: What's going on here? Why do we compare against a magic empty CBOR array?

crates/amaru/tests/evaluate_ledger_states.rs
```diff
@@ -0,0 +1,223 @@
...
+    let epoch = peek_epoch(&record.initial_nes)?;
+    let point = amaru_kernel::Point::try_from(
+        format!(
+            "{}.0000000000000000000000000000000000000000000000000000000000000000",
+            epoch * 86400
+        )
+        .as_str(),
+    )?;
```
KtorZ:
```suggestion
  let point = Point::Specific(
    epoch * 86400, 
    hex::decode("0000000000000000000000000000000000000000000000000000000000000000")?,
  );
```

crates/amaru/tests/evaluate_ledger_states.rs
```diff
@@ -0,0 +1,223 @@
...
+
+    fs::create_dir_all(ledger_dir)?;
+    let db = RocksDB::empty(ledger_dir)?;
+
+    let network = NetworkName::Testnet(1);
```
KtorZ: Seems like this should come from the arguments, since the import data are already network-dependent.

crates/amaru/tests/evaluate_ledger_states.rs
```diff
@@ -0,0 +1,223 @@
...
+    let epoch = crate::amaru_ledger::bootstrap::import_initial_snapshot(
```
KtorZ:
```suggestion
    let epoch = import_initial_snapshot(
```

prefer unqualified import here; the namespace doesn't help disambiguate the function call and is therefore superfluous.

crates/amaru/tests/evaluate_ledger_states.rs
```diff
@@ -0,0 +1,223 @@
...
+        &point,
+        None,
+        Default::default(),
+        Default::default(),
+        iter::empty(),
+    )?;
+    transaction.commit()?;
+
+    for entry in record.transactions {
+        let (tx_bytes, success, slot): (pallas_codec::utils::Bytes, bool, minicbor::data::Int) =
```
KtorZ: `Bytes` is already exported from amaru_kernel, you shouldn't need pallas_codec here.

crates/amaru/tests/evaluate_ledger_states.rs
```diff
@@ -0,0 +1,223 @@
...
+};
+use amaru_ledger::{
+    self,
+    context::DefaultValidationContext,
+    rules::transaction,
+    state::{self, volatile_db::StoreUpdate},
+    store::{EpochTransitionProgress, ReadStore, Store, TransactionalContext},
+};
+use amaru_stores::rocksdb::RocksDB;
+use pallas_addresses::Network;
```
KtorZ: This is already re-exported from amaru_kernel, you shouldn't introduce pallas_addresses here.

crates/amaru/tests/evaluate_ledger_states.rs
```diff
@@ -0,0 +1,223 @@
...
+            &mut validation_context,
+            &Network::Testnet,
+            &protocol_parameters,
+            era_history,
+            &governance_activity,
+            pointer,
+            true,
+            tx_body,
+            &tx_witness_set,
+            None, // tx_auxiliary_data.as_ref(),
```
KtorZ: ?

crates/amaru/tests/evaluate_ledger_states.rs
```diff
@@ -0,0 +1,223 @@
...
+        );
+
+        let vs = state::VolatileState::from(validation_context);
+
+        // Can we re-use this point? We should be able to treat all txes as if they're in the same block
+        let p: Point = point.clone();
+        let issuer: PoolId = [7; 28].into();
```
KtorZ: ?

crates/amaru/tests/evaluate_ledger_states.rs
```diff
@@ -0,0 +1,223 @@
...
+        era_history,
+        &protocol_parameters,
+        &mut governance_activity,
+        &point,
+        None,
+        Default::default(),
+        Default::default(),
+        iter::empty(),
+    )?;
+    transaction.commit()?;
```
KtorZ: What is happening here? I can only assume this is meant to set the _point_ as the tip of the ledger db, though that is already done by the `import_initial_snapshot` function. So that is probably redundant/pointless here.

Unless I am missing something, you can likely safely remove.

crates/amaru/tests/evaluate_ledger_states.rs
```diff
@@ -0,0 +1,223 @@
...
+
+        let tx_body: KeepRaw<'_, MintedTransactionBody<'_>> = tx.transaction_body.clone();
+        let tx_witness_set: MintedWitnessSet<'_> = tx.transaction_witness_set.deref().clone();
+        let _tx_auxiliary_data =
+            Into::<Option<KeepRaw<'_, AuxiliaryData>>>::into(tx.auxiliary_data.clone())
+                .map(|aux_data| Hasher::<256>::hash(aux_data.raw_cbor()));
+
+        let pointer = TransactionPointer {
+            slot: point.slot_or_default(),
+            transaction_index: 0,
```
KtorZ: That shall likely come from the loop index, right?

crates/amaru/tests/evaluate_ledger_states.rs
Quantumplation: There are several test vectors that are gated to be no-ops in certain protocol versions, but which run anyway, and in the on-disk files, they turn into weirdly encoded empty arrays like this;

I suggested that Ethan just delete those files instead, which we've done, but looks like this check hung around and can be removed.

crates/amaru/tests/evaluate_ledger_states.rs
yHSJ: We have a `slot` being read from the event, why are we not using that?

crates/amaru-ledger/src/bootstrap.rs
```diff
@@ -50,267 +47,211 @@ static DEFAULT_CERTIFICATE_POINTER: LazyLock<CertificatePointer> =
...
 pub fn import_initial_snapshot(
     db: &(impl Store + 'static),
-    file: &mut std::fs::File,
+    bytes: &[u8],
     point: &Point,
-    era_history: &EraHistory,
     network: NetworkName,
     // A way to notify progress while importing. The second argument is a template argument, which
     // follows the format described in:
     //
```
yHSJ, line 73: I assume we actually don't want to remove this...

crates/amaru/tests/evaluate_ledger_states.rs
```diff
@@ -0,0 +1,274 @@
...
+struct TestContext {
+    ledger_dir: PathBuf,
+    pparams_dir: PathBuf,
+}
+
+fn get_test_context() -> Result<TestContext, Box<dyn std::error::Error>> {
+    let working_dir = env::current_dir()?;
+    Ok(TestContext {
+        ledger_dir: working_dir.join(PathBuf::from("../../ledger.db")),
+        pparams_dir: working_dir.join(PathBuf::from("../../cardano-blueprint/src/ledger/conformance-test-vectors/eras/conway/impl/dump/pparams-by-hash/")),
```
yHSJ: Worth noting that, from what I can tell, the conformance test vectors are in a tarball, so this will fail for most people.

crates/amaru/tests/evaluate_ledger_states.rs
```diff
@@ -0,0 +1,274 @@
...
+
+fn peek_epoch(nes: &AnyCbor) -> Result<u64, Box<dyn std::error::Error>> {
+    let mut d = cbor::decode::Decoder::new(nes.raw_bytes());
+    let _ = d.array()?;
+    Ok(d.u64()?)
+}
+
+#[derive(cbor::Decode)]
+#[allow(dead_code)]
+struct TestVector {
```
yHSJ: *Eventually*, though perhaps it's not needed right now, I'd love to get these `AnyCbor` types replaced with more strict types. Is there anything holding us back from that?

crates/amaru/tests/evaluate_ledger_states.rs
```diff
@@ -0,0 +1,274 @@
...
+    let epoch = peek_epoch(&record.initial_state)?;
+    let point = Point::Specific(
+        epoch * 86400,
+        hex::decode("0000000000000000000000000000000000000000000000000000000000000000")?
+    );
+
+    let mut decoder = minicbor::Decoder::new(&record.initial_state);
+    let (mut validation_context, pparams_hash, governance_activity) = decode_ledger_state(&mut decoder)?;
+
+    let protocol_parameters = decode_segregated_parameters(pparams_dir, &pparams_hash)?;
```
yHSJ: Why are we passing in protocol params here separately from the test vector? Could they/should they not be included in the `TestVector` itself?

crates/amaru/tests/evaluate_ledger_states.rs
```diff
@@ -0,0 +1,274 @@
...
+            Into::<Option<KeepRaw<'_, AuxiliaryData>>>::into(tx.auxiliary_data.clone())
+                .map(|aux_data| Hasher::<256>::hash(aux_data.raw_cbor()));
+
+        let pointer = TransactionPointer {
+            slot: point.slot_or_default(),
+            transaction_index: ix,
+        };
+
+        // Run the transaction against the imported ledger state
+        let result = transaction::execute(
```
yHSJ: As far as I understand, there are no test vectors that currently have more than one transaction, so this makes sense. I worry that rather quickly as these test vectors are adopted more widely, that will change and these tests will break.

May be worth making some small adjustments to handle multiple transactions in a test vector

crates/amaru/tests/evaluate_ledger_states.rs
```diff
@@ -0,0 +1,274 @@
...
+            pointer,
+            true,
+            tx_body,
+            &tx_witness_set,
+            tx_auxiliary_data,
+        );
+
+        match result {
+            Ok(()) => {
+                if success {
```
yHSJ: Again, this seems to be _acceptable_ for a first iteration of it, but it's important we are actually comparing the final state instead of just checking for the absence of an error.

A transaction could be executed without an error but provide the incorrect state transitions for the ledger, and that's what these tests should be validating for us.

yHSJ (commented):
Much better I think! We're getting much closer.

crates/amaru/tests/evaluate_ledger_states.rs
rrruko: Yes, this code assumes that you have decompressed and extracted the archive.

crates/amaru/tests/evaluate_ledger_states.rs
rrruko: The protocol parameters account for a significant portion of the encoded size of each ledger state. As an optimization, we use a modified NES format where the parameters are encoded by hash. The full pparams records are included in the test vectors tarball.

crates/amaru/tests/evaluate_ledger_states.rs
rrruko: >  there are no test vectors that currently have more than one transaction

In a previous iteration, this was true. Now, there is exactly one test vector for each cardano-ledger test case. So many of them have lots of transactions.

What specific adjustments would you make here?

crates/amaru/tests/evaluate_ledger_states.rs
Quantumplation: They're interned: the test vector refers to them by hash, because otherwise they get duplicated for *every single* test vector and balloon the size quite dramatically

crates/amaru/tests/evaluate_ledger_states.rs
rrruko: Yes; in my mind the scope of this PR was simply to integrate the test vectors into amaru. The omission of the final state check was intentional, to keep the scope small, with the expectation that this PR would be immediately followed by a robust implementation pass.

crates/amaru-ledger/src/bootstrap.rs
rrruko: In my previous "test vectors" PR for amaru, I added this segregated pparams handling to `import_initial_snapshot` so that I could re-use it to consume the custom NES record format used by the test vectors. Since I made up that format for this use-case, this feature is not used by any other code in amaru. And since we decided to decouple from the DB, it is just extra complexity now.

crates/amaru/tests/evaluate_ledger_states.rs
rrruko: fixed

crates/amaru-ledger/src/bootstrap.rs
yHSJ: Ah, I saw it credited to Matthias, so I wasn't sure. If that's the case, this is totally fine.

crates/amaru/tests/evaluate_ledger_states.rs
yHSJ: I might argue that the "integration of the test vectors" includes the implementation, but I get your point.

KtorZ:
*rebased

PR #472, opened 2025-09-25
Optionally maintain a PID file for monitoring applications

yHSJ:
see title

## Summary by CodeRabbit

* **New Features**
  * Optional PID file support for the daemon lifecycle.
  * New command-line flag and environment variable to specify PID file location.
  * Prevents startup if an existing daemon instance is detected.
  * Writes current PID at startup and removes the PID file on shutdown.
  * Cross-platform process detection (Unix and Windows).
  * Logs warnings if PID file creation fails.

### DIFF
--- a/crates/amaru/src/bin/amaru/cmd/daemon.rs
+++ b/crates/amaru/src/bin/amaru/cmd/daemon.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crate::pid::ProcessIdHandle;
+use crate::pid::with_optional_pid_file;
 use crate::{cmd::connect_to_peer, metrics::track_system_metrics};
 use amaru::stages::{Config, MaxExtraLedgerSnapshots, StoreType, bootstrap};
 use amaru_kernel::{default_chain_dir, default_ledger_dir, network::NetworkName};
@@ -23,7 +23,7 @@ use pallas_network::facades::PeerClient;
 use std::{path::PathBuf, time::Duration};
 use thiserror::Error;
 use tokio_util::sync::CancellationToken;
-use tracing::{debug, error, trace, warn};
+use tracing::{error, trace, warn};
 
 #[derive(Debug, Parser)]
 pub struct Args {
@@ -86,47 +86,34 @@ pub async fn run(
     args: Args,
     meter_provider: Option<SdkMeterProvider>,
 ) -> Result<(), Box<dyn std::error::Error>> {
-    // This is being kept-alive so that the `drop` function
-    // is called when the `ProcessIdHandler` goes out of scope.
-    // That allows the file to be cleaned up before Amaru gracefully exits.
-    // It is not vital that this cleanup happens, but it's nice to have :)
-    let _pid_file = args.pid_file.as_ref().map(|path| {
-        ProcessIdHandle::new(path)
-            .inspect(|pid_file| {
-                debug!(
-                    "created PID File {}, current PID: {}",
-                    pid_file,
-                    pid_file.pid()
-                )
-            })
-            .inspect_err(|e| warn!("failed to create or write to PID file: {} ", e))
-    });
+    with_optional_pid_file(args.pid_file.clone(), async |_pid_file| {
+        let config = parse_args(args)?;
+        pre_flight_checks()?;
 
-    let config = parse_args(args)?;
-    pre_flight_checks()?;
+        let metrics = meter_provider
+            .clone()
+            .map(track_system_metrics)
+            .transpose()?;
 
-    let metrics = meter_provider
-        .clone()
-        .map(track_system_metrics)
-        .transpose()?;
+        let mut clients: Vec<(String, PeerClient)> = vec![];
+        for peer in &config.upstream_peers {
+            let client = connect_to_peer(peer, &config.network).await?;
+            clients.push((peer.clone(), client));
+        }
 
-    let mut clients: Vec<(String, PeerClient)> = vec![];
-    for peer in &config.upstream_peers {
-        let client = connect_to_peer(peer, &config.network).await?;
-        clients.push((peer.clone(), client));
-    }
+        let exit = amaru::exit::hook_exit_token();
 
-    let exit = amaru::exit::hook_exit_token();
+        let sync = bootstrap(config, clients, exit.clone(), meter_provider).await?;
 
-    let sync = bootstrap(config, clients, exit.clone(), meter_provider).await?;
+        run_pipeline(gasket::daemon::Daemon::new(sync), exit).await;
 
-    run_pipeline(gasket::daemon::Daemon::new(sync), exit).await;
+        if let Some(handle) = metrics {
+            handle.abort();
+        }
 
-    if let Some(handle) = metrics {
-        handle.abort();
-    }
-
-    Ok(())
+        Ok(())
+    })
+    .await
 }
 
 pub async fn run_pipeline(pipeline: gasket::daemon::Daemon, exit: CancellationToken) {
--- a/crates/amaru/src/bin/amaru/pid.rs
+++ b/crates/amaru/src/bin/amaru/pid.rs
@@ -20,6 +20,8 @@ use std::{
     process::{self, Command},
 };
 
+use tracing::{debug, warn};
+
 pub struct ProcessIdHandle {
     path: PathBuf,
     pid: u32,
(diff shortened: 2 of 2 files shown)


crates/amaru/src/bin/amaru/cmd/daemon.rs
```diff
@@ -76,14 +75,29 @@ pub struct Args {
         default_value_t = MaxExtraLedgerSnapshots::default(),
     )]
     max_extra_ledger_snapshots: MaxExtraLedgerSnapshots,
+
+    /// Path to the PID file, managed by Amaru.
+    #[arg(long, value_name = "DIR", env = "AMARU_PID_FILE")]
+    pid_file: Option<PathBuf>,
```
KtorZ: Is that meant to be a directory?

crates/amaru/src/bin/amaru/pid.rs
```diff
@@ -0,0 +1,86 @@
...
+
+        if let Some(parent) = path.parent() {
+            fs::create_dir_all(parent)?;
+        }
+
+        if path.exists()
+            && let Ok(existing_pid) = fs::read_to_string(&path)
+            && let Ok(existing_pid) = existing_pid.trim().parse::<u32>()
+            && process_exists(existing_pid)
+        {
```
KtorZ: That is very strange. The process id is unique to our process; so even if _Amaru_ is already running, there's no chance that another instance of Amaru will have the same process id. 

All-in-all, I am not sure what this is trying to prevent?

crates/amaru/src/bin/amaru/pid.rs
```diff
@@ -0,0 +1,86 @@
...
+// limitations under the License.
+//
+use std::{
+    fmt::Display,
+    fs,
+    path::{Path, PathBuf},
+    process::{self, Command},
+};
+
+pub struct PIDFile {
```
KtorZ:
```suggestion
pub struct ProcessIdHandle {
```

This (a runtime reference to an open file) is typically called: a handle.

crates/amaru/src/bin/amaru/pid.rs
Quantumplation: My thought is that this is to prevent clobbering another processes PID file;

For example, consider:
- You start Amaru with PID path `a/b/c.pid`
- Amaru crashes without a chance to clean up the PID file, and gets automatically restarted
- It starts back up, sees that the file exists (ok...), parses it (ok...), but the PID inside the file isn't running, so we're ok to overwrite this

On the other hand, consider:
- You start Amaru with PID path `a/b/c.pid`
- You accidentally start amaru again with the same command line arguments
- It starts up, sees that the file exists (ok...), parses it (ok...), but the PID inside the file *is* running (it's the previous copy of amaru), so we need to report an error to prevent clobbering that PID file

Finally, consider:
- You start {some other process} with PID path `a/b/c.pid`
- You start Amaru with PID path `a/b/c.pid`
- It starts up, sees that the file exists (ok...), parses it (ok...), but the PID inside the file *is* running (it's {some other process}), so we need to report an error to prevent clobbering that other processes PID file

crates/amaru/src/bin/amaru/cmd/daemon.rs
yHSJ: No, it should be a file.

crates/amaru/src/bin/amaru/pid.rs
yHSJ: Yes, you've got it precisely right Pi. It would be really easy to cause issues overwriting PIDs from another running process. The goal here is to help users from stepping on their own toes.

crates/amaru/src/bin/amaru/cmd/daemon.rs
```diff
@@ -77,14 +76,29 @@ pub struct Args {
         default_value_t = MaxExtraLedgerSnapshots::default(),
     )]
     max_extra_ledger_snapshots: MaxExtraLedgerSnapshots,
+
+    /// Path to the PID file, managed by Amaru.
+    #[arg(long, value_name = "FILE", env = "AMARU_PID_FILE")]
+    pid_file: Option<PathBuf>,
```
jeluard, line 82: Could you describe how this feature is intended to be used? Do we foresee use cases where the pid file location could be manually set by the end user to different location?

crates/amaru/src/bin/amaru/cmd/daemon.rs
```diff
@@ -77,14 +76,29 @@ pub struct Args {
...
 }
 
 pub async fn run(
     args: Args,
     meter_provider: Option<SdkMeterProvider>,
 ) -> Result<(), Box<dyn std::error::Error>> {
-    let config = parse_args(args)?;
+    let _pid_file = args.pid_file.as_ref().map(|path| {
```
jeluard: `map` looks weird here as you don't use the mapped value. `if let Some()` would look more natural to me.

jeluard: Or I guess the intention is to keep it alive so that `Drop` is then called when `run` ends? 
Feels dangerous, as someone refactoring this might break this behavior.

Some comment might help ensuring this contract is not broken.

crates/amaru/src/bin/amaru/cmd/daemon.rs
yHSJ: If I am running more than one amaru process on my machine and I want to monitor them both, they'd have to have different PID files.

crates/amaru/src/bin/amaru/cmd/daemon.rs
yHSJ: Yes, the `map` is keeping it alive so that `Drop` is executed when it falls out of scope. It's not the end of the world if we don't clean up the file (as we check the processes for the PID in the file anyway), it's just nice to have. I will leave a comment. But the implementation of `ProcessIdHandler` was written such that it can handle existing PID files.

crates/amaru/src/bin/amaru/cmd/daemon.rs
abailly: Looking at [nView source code](https://github.com/blinklabs-io/nview/blob/fbd000ac51091659a371f52330d064a53d6bf61a/config.yaml.example#L63) I find that nView uses PID files only for monitoring Amaru. Why not use the [same mechanism](https://github.com/blinklabs-io/nview/blob/fbd000ac51091659a371f52330d064a53d6bf61a/main.go#L1264) than for `cardano-node`s which would alleviate the need to maintain a PID file for Amaru?

crates/amaru/src/bin/amaru/cmd/daemon.rs
abailly: The typical way to handle that situation is to pass the result to another function thus enforcing "bracket-like" semantics for the lifecycle of the data denoted by the variable, perhaps this would make things clearer?

abailly (commented):
I am mostly curious as to why nView needs that feature, not much to say about the implementation itself.

crates/amaru/src/bin/amaru/cmd/daemon.rs
yHSJ: *Currently*, it's only used for Amaru. It *will* be used for other nodes, like Dingo.

Amaru doesn't take a `--port` flag like the `cardano-node` does, so it would require a change to Amaru in that sense (which was an option) since Chris asked me to leave the `caradno-node` logic untouched since it's not tested very well and is rather brittle.

I also think in general maintaining a PID file is a useful feature to have for the daemon.

abailly: What do you mean by 
> Amaru doesn't take a --port flag like the cardano-node does, 

There's 
```
     --listen-address <LISTEN_ADDRESS>
          The address to listen on for incoming connections

          [env: AMARU_LISTEN_ADDRESS=]
          [default: 0.0.0.0:3000]
```
which groups both host and port part. Wouldn't that be enough to discriminate which instance of Amaru you care about? Or are you talking about another port?

abailly: > I also think in general maintaining a PID file is a useful feature to have for the daemon.

My Ops experience is not deep enough to have a strong opinion on this, my gut feeling is that in a world of `systemd` managed services and containers orchestration, this might not be as useful.

crates/amaru/src/bin/amaru/cmd/daemon.rs
yHSJ: Sure, that is acceptable.

crates/amaru/src/bin/amaru/cmd/daemon.rs
yHSJ: > What do you mean by
> 
> > Amaru doesn't take a --port flag like the cardano-node does,
> 
> There's
> 
> ```
>      --listen-address <LISTEN_ADDRESS>
>           The address to listen on for incoming connections
> 
>           [env: AMARU_LISTEN_ADDRESS=]
>           [default: 0.0.0.0:3000]
> ```
> 
> which groups both host and port part. Wouldn't that be enough to discriminate which instance of Amaru you care about? Or are you talking about another port?

Yes, the `--listen-address` flag would work fine. But the way `nView` works is it looks specifically for the `--port` flag in the CLI command executed. So, it would require changes to `nView` anyway to make this work. Again, since Chris said that they want to use the PID file for Dingo, this logic would have to exist in `nView` anyway, so I felt it made more sense to do it this way.

> My Ops experience is not deep enough to have a strong opinion on this, my gut feeling is that in a world of systemd managed services and containers orchestration, this might not be as useful.

I am not an Ops guy myself, but from what I've seen this is a very standard thing that is used by many external monitoring tools. Worse case scenario, you just don't have to use it (hence it being optional)

abailly: > the way nView works is it looks specifically for the --port flag in the CLI command executed

🙄

abailly (approved):

PR #444, opened 2025-09-12
refactor: use a chain store inside the headers tree

etorreborre:
This PR backs up the HeadersTree implementation with a `ChainStore` instead of in-memory data:

 - The `RocksDB` database is used to persist:
    - The headers.
    - The parent / child relationship between them.
    - The headers tree anchor.
    - The headers tree best chain tip.
 - The `HeadersTree` mostly reads from the `ChainStore`, except for setting the chain anchor and tip (this is done transactionally  when necessary).
 - The code removing headers that are unreachable (in the previous `prune_headers` function) has been removed so that we can diagnose issues in the future. This also speeds up execution.
 - The `InMemConsensusStore` has been extended to support the same operations.
 - The unit tests are executed against both implementations
 - The `HeadersTree` benchmark can also now run both in-memory and on-disk (profiling show that the difference in performance is just due to writes).
 - I also had to move the `ChainStore` traits to a shared crate in order to allow `amaru-consensus` to compile under wasm (otherwise the `getrandom` dependency gets pulled).
  
Additionally:

 * An incorrect header cbor file was removed.
 * A separate branch containing a new `dump-chain-db` command has been merged in order to diagnose an end to end snapshot failure.  
    

## Summary by CodeRabbit

- New Features
  - In-memory consensus store added; RocksDB store enhanced with best-chain/anchor tracking and parent–child queries.
  - CLI: header import now requires network and config paths.
  - Network client exposes current intersections for inspection; improved intersection logging.

- Refactor
  - HeadersTree now backed by a pluggable store; updated constructor and tree building.
  - Unified ChainStore usage across components via Arc (no Mutex), simplifying concurrency.
  - Nonce evolution routed through a new PraosChainStore; trait updated to take &self.
  - Public re-exports simplify imports of consensus/traits.

PR #473, opened 2025-09-26
fix: remove the use of column families in rocksdb

etorreborre:
This PR removes the use of column families that was introduced in #444.
Instead, when querying by prefix, we stop iterating when keys with a different prefix are returned. 

## Summary by CodeRabbit

* **Refactor**
  * Simplified storage to a single keyspace using prefix-based keys instead of separate column-family layers, streamlining open/read/write paths while keeping public APIs unchanged.
  * Aligned control flow and error handling for more consistent behavior.
* **Bug Fixes**
  * Added early-exit on prefix scans to prevent cross-key reads and improve data correctness.

### DIFF
--- a/crates/amaru-stores/src/rocksdb/consensus.rs
+++ b/crates/amaru-stores/src/rocksdb/consensus.rs
@@ -17,10 +17,7 @@ use amaru_kernel::{HEADER_HASH_SIZE, Hash, RawBlock, cbor, from_cbor, to_cbor};
 use amaru_ouroboros_traits::is_header::IsHeader;
 use amaru_ouroboros_traits::{ChainStore, Nonces, ReadOnlyChainStore, StoreError};
 use amaru_slot_arithmetic::EraHistory;
-use rocksdb::{
-    ColumnFamilyDescriptor, DB, Direction, IteratorMode, OptimisticTransactionDB, Options,
-    SliceTransform,
-};
+use rocksdb::{DB, OptimisticTransactionDB, Options};
 use std::ops::Deref;
 use std::path::PathBuf;
 use tracing::{Level, instrument};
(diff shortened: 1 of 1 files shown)


crates/amaru-stores/src/rocksdb/consensus.rs
```diff
@@ -225,10 +191,13 @@ macro_rules! impl_ReadOnlyChainStore {
             fn get_children(&self, hash: &Hash<32>) -> Vec<Hash<32>> {
                 let mut result = Vec::new();
                 let prefix = [&CHILD_PREFIX[..], &hash[..]].concat();
-                let cf = self.db.cf_handle(CHILDREN_CF).expect("missing children CF");
-                for res in self.db.prefix_iterator_cf(&cf, &prefix) {
+                for res in self.db.prefix_iterator(&prefix) {
                     match res {
                         Ok((key, _value)) => {
+                            // Stop once we get keys that don't start with the required prefix
+                            if !key.starts_with(&prefix) {
```
rkuhn, line 197: It seems to me from casual inspection of db_iterator.rs as well as a grok answer that this check shouldn’t be necessary. Do you have evidence to the contrary? Then we’d need to audit all RocksDB usage.

rkuhn (commented):
Looks okay to me, but I lack context why column families were used before and why they are removed now.

crates/amaru-stores/src/rocksdb/consensus.rs
pgrange: ~~The problem seems to occur because we don't set any prefix extractor.~~

~~Adding this solves the issue and allows to get rid of this line:~~
```rust
        opts.set_prefix_extractor(rocksdb::SliceTransform::create_fixed_prefix(
            CONSENSUS_PREFIX_LEN,
        ));
```

pgrange: My bad! We need this check here, I added a comment to explain why: the prefix extractor only picks the first 5 bytes of the key so we iterate through _all_ of the keys with type `CHILD_PREFIX`. But actually, since we asked for a longer prefix, rocksdb will start iterating at the first key matching `CHILD_PREFIX + hash` but then, it will not stop automatically but, instead, continue to iterate until we stop it. You can verify by commenting out this check and run the tests that will fail.

The other rocksdb usage are not concerned (at least all the tests pass) because we always iterate with a five byte prefix (`CONSENSUS_PREFIX_LEN`).

pgrange (approved):
Thanks for making the effort @etorreborre. I think we're improving our understanding and usage of rocksdb which is cool.

I've noted that some robots are rambling a bunch of stuff but I have strictly no opinion on these comments.

crates/amaru-stores/src/rocksdb/consensus.rs
KtorZ: Alternatively: https://github.com/pragma-org/amaru/blob/main/crates/amaru-stores/src/rocksdb/ledger/columns/dreps_delegations.rs#L188

pgrange: Nice, thanks @KtorZ . So this `set_iterate_upper_bound` is what makes rocksdb stop automatically with no need for us to ckeck for invalid values inside the iteration 👍

PR #485, opened 2025-10-01
ScriptContext & ToPlutusData

yHSJ:
In order to implement phase 2 validation, we need to create a `ScriptContext` that is passed to the VM when executing a script. We need to encode data in that `ScriptContext` as `PlutusData`, which is an (unspecified) in-memory representation that is deserialized from CBOR. 

This PR introduces a new `amaru-plutus` for logic that is related to plutus scripts but doesn't neatly fit into the `uplc` repository nor the `amaru-ledger` crate.  Notably, it provides a new `ScriptContext` struct for each plutus version and implementations to serialize it as PlutusData.

## Summary by CodeRabbit

* **New Features**
  * Full Plutus support (v1–v3): versioned script contexts, script/context types, values/mint/withdrawals/datums, rich PlutusData serialization and exported constructor macros.

* **Documentation**
  * Adds README describing the new Plutus capabilities.

* **Tests**
  * Adds JSON test fixtures covering script-context scenarios across versions.

* **Chores**
  * Introduces a new workspace crate for Plutus logic and a serde helper for hex deserialization.

* **Bug Fixes**
  * Cleans up kernel public exports to correct duplicated/legacy declarations.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -283,7 +283,6 @@ dependencies = [
  "amaru-kernel",
  "amaru-slot-arithmetic",
  "hex",
- "insta",
  "itertools 0.14.0",
  "minicbor",
  "num",
--- a/crates/amaru-plutus/Cargo.toml
+++ b/crates/amaru-plutus/Cargo.toml
@@ -17,7 +17,7 @@ thiserror.workspace = true
 itertools.workspace = true
 hex.workspace = true
 minicbor.workspace = true
-insta.workspace = true
+
 # Internal dependencies
 amaru-kernel.workspace = true
 amaru-slot-arithmetic.workspace = true


crates/amaru-kernel/src/lib.rs
```diff
@@ -939,6 +939,45 @@ impl HasRedeemers for Redeemers {
...
 }
 
+pub fn from_alonzo_value(value: &AlonzoValue) -> Result<Value, Box<dyn std::error::Error>> {
+    match value {
+        AlonzoValue::Coin(coin) => Ok(Value::Coin(*coin)),
+        AlonzoValue::Multiasset(coin, assets) if assets.is_empty() => Ok(Value::Coin(*coin)),
+        AlonzoValue::Multiasset(coin, assets) => {
+            let multiasset = assets
+                .iter()
+                .cloned()
```
KtorZ: It'll be better to make the clones explicit; if everything down below is going to be cloned, the signature shall just take a plain `AlonzoValue`, and have the caller do the cloning.

crates/amaru-plutus/src/lib.rs
```diff
@@ -0,0 +1,352 @@
...
+        }
+    }
+}
+
+#[macro_export]
+macro_rules! constr {
+    ($index:expr) => {{
+        let maybe_constr_tag = $index.to_constr_tag();
+        PlutusData::Constr(Constr {
+            tag: maybe_constr_tag.unwrap_or(DEFAULT_TAG),
```
KtorZ: That is ... not right. If "ToConstrTag" returns `None`, then that's a glaring mistake from the caller and whomever defines the tag. That shouldn't be silently swallowed with a default.

crates/amaru-plutus/src/lib.rs
```diff
@@ -0,0 +1,352 @@
...
+                        pointer.cert_idx(),
+                    ))
+                    .to_plutus_data(),
+                    ShelleyDelegationPart::Null => None::<StakeCredential>.to_plutus_data(),
+                };
+
+                constr!(0, payment_part_plutus_data, stake_part_plutus_data)
+            }
+            Address::Stake(stake_address) => stake_address.to_plutus_data(),
+            Address::Byron(_) => unreachable!("unable to encode Byron address in PlutusData"),
```
KtorZ: That is possible in V1 (and possibly V2, I don't remember)

crates/amaru-plutus/src/script_context/v1.rs
```diff
@@ -0,0 +1,169 @@
...
+
+impl ToPlutusData<1> for ScriptContext {
+    fn to_plutus_data(&self) -> PlutusData {
+        constr!(v: 1, 0, self.tx_info, self.purpose)
+    }
+}
+
+impl ToPlutusData<1> for TxInfo {
+    fn to_plutus_data(&self) -> PlutusData {
+        constr!(v: 1, 0, self.inputs, self.outputs, self.fee, self.mint, self.certificates, self.withdrawals.iter().map(|(address, coin)| (constr!(v: 1, 0, address), *coin)).collect::<Vec<_>>(), self.validity_range, self.signatories, self.data, constr!(v:1,0, self.id))
```
KtorZ: What's going on here 😅

crates/amaru-plutus/src/lib.rs
yHSJ: This I borrowed directly from the Aiken code. As you know, the tag for any "alternative" outside of 0-127 is 102 ([see here](https://github.com/IntersectMBO/plutus/blob/1f31e640e8a258185db01fa899da63f9018c0e85/plutus-core/plutus-core/src/PlutusCore/Data.hs#L61)). In Aiken, the constructor is only included when there is an alternative within the range of 0-127. This could obviously be done another way, but I wanted to make sure the logic was clear.

crates/amaru-plutus/src/script_context/v1.rs
yHSJ: `constr` is the macro I defined. `v` is the plutus version to use when calling `to_plutus_data` on the fields (sometimes it's needed explicitly), the second argument is the index, and the rest of the arguments are the fields, which get encoded as `PlutusData`. Probably could be a bit more obvious though

Quantumplation: @SupernaviX is pretty good at rust macros, he might be able to suggest some prettification

KtorZ: Sorry; my comment lacked context. I was mostly picking on the lack of formatting here. Seems like your editor don't format within macros?

crates/amaru-plutus/src/lib.rs
KtorZ: Ho I see. I misread that; forgot for a second that Pallas represents this tag as two separate fields. So, no problem.

crates/amaru-plutus/src/script_context/v1.rs
yHSJ: Yeah, for some reason `cargo fmt` wasn't formatting the macros when I had named arguments in them. I've updated the macros for readability and now they're formatted as well.

crates/amaru-kernel/src/lib.rs
```diff
@@ -939,6 +939,45 @@ impl HasRedeemers for Redeemers {
     }
 }
 
+pub fn from_alonzo_value(value: AlonzoValue) -> Result<Value, Box<dyn std::error::Error>> {
```
KtorZ: I feel like I've written / seen this function 50 times already.

crates/amaru-plutus/src/script_context/v1.rs
```diff
@@ -0,0 +1,226 @@
...
+        PlutusData::Map(KeyValuePairs::Def(entries))
+    }
+}
+
+impl<const V: u8> ToPlutusData<V> for TransactionInput
+where
+    PlutusVersion<V>: IsKnownPlutusVersion + IsPrePlutusVersion3,
+{
+    fn to_plutus_data(&self) -> PlutusData {
+        constr!(v: V, 0, [constr!(v: V, 0, [self.transaction_id]), self.index])
```
KtorZ: This is a bit weird; I would expect the extra constr wrapping on transaction id to live in the `for TransactionId` instance. Otherwise, this particular chunk of logic gets duplicated; for example in both `for TxInfo` instances for V1 and V2.

yHSJ: Here `transaction_id` is a `Hash<32>`. If we provide an implementation that wraps it for `Hash<32>`, it would break things like datum hashes.

crates/amaru-kernel/src/lib.rs
yHSJ: I had the same feeling, but I couldn't find it. Doesn't mean it doesn't exist though.

crates/amaru-plutus/src/script_context/v1.rs
KtorZ: Hmmm. This isn't ideal indeed. I suppose it's a hash<32> thoughout. So we would have to wrap/unwrap every inputs too.

KtorZ: Although in a certain way, we already have to manually fiddle with all inputs. So having a newtype to capture this behaviour would be the right move regardless.

yHSJ: I'm actually not opposed to creating new types for the `ScriptContext` fields. We *do* need to sort things inside the our existing structs already, so that could be a part of the type definitions. Though I'd rather confirm that this stuff works™️ at all

PR #489, opened 2025-10-06
refactor: move validate_header effects to pure-stage

etorreborre:
This PR moves the effects used by the `validate_header` stage to `pure-stage`. Those effects are:

 - Store effects, to access nonces.
 - Ledger effects to access a pool summary.

The most notable changes involve:

 - Creating a `ConsensusParameters` data type encapsulating all the parameters necessary to validate a block header:
   - Some parameters taken from `GlobalParameters`.
   - The `EraHistory`.
   - A list of opcert sequence numbers (apparently not used in production code for now, it's only used in tests)
 - As a result, it is not necessary to thread the `EraHistory` through all the `ChainStore` instances.
 - Similarly, it is also not possible to pass the `GlobalParameters` to the `StakeDistributionObserver` backing up the `HasStakeDistribution` instance, which simplifies that interface to just one function.

## Summary by CodeRabbit

- New Features
  - Added configurable ConsensusParameters used across validation and nonce evolution.
  - Ledger exposes pool-aware queries; PoolSummary now includes active_stake and stake.

- Refactor
  - Simplified chain storage and RocksDB initialization (no era history required).
  - Unified ledger access via shared handles and streamlined header/block validation stages.

- CLI
  - Import/bootstrap commands no longer require an explicit network argument; paths determine context.

- Chores
  - Added serialization for expanded PoolSummary and adjusted workspace dependencies.


crates/amaru-consensus/src/consensus/effects/consensus_effects.rs
```diff
@@ -31,39 +30,34 @@ pub trait ConsensusOps: Send + Sync + Clone {
     fn network(&self) -> impl NetworkOps;
     /// Return a LedgerOps implementation to access ledger operations, considering that it is a sub-system
     /// external to consensus.
-    fn ledger(&self) -> impl LedgerOps;
+    fn ledger(&self) -> Arc<dyn LedgerOps>;
```
abailly, line 33: Perhaps the change to an `Arc<>` would warrant a change in the comment? Why do we need an Arc here?

crates/amaru-consensus/src/consensus/stages/validate_header.rs
```diff
@@ -12,104 +12,103 @@
...
+        DecodedChainSyncEvent::CaughtUp { .. } => {
+            eff.base().send(&downstream, msg).await;
+            return (consensus_parameters, downstream, errors);
         }
+    };
 
-        fn max_kes_evolutions(&self) -> u64 {
-            unimplemented!()
-        }
+    let span = tracing::trace_span!(
```
abailly, line 57: Is this span really useful here?

abailly: It would probably be useful to look at how the "waterfall" looks like with those changes?

crates/amaru-kernel/src/protocol_parameters.rs
```diff
@@ -433,6 +436,104 @@ pub struct GlobalParameters {
     pub randomness_stabilization_window: u64,
 }
 
+/// This data type encapsulates the parameters needed by the consensus layer to operate.
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub struct ConsensusParameters {
```
abailly, line 441: This change is useful but I think it would warrant its own PR rather than being mixed with other concerns in the same PR.

crates/amaru-ouroboros-traits/src/praos/mod.rs
```diff
@@ -103,9 +103,5 @@ pub trait Praos<H: IsHeader>: Send + Sync {
     ///
     /// Once the stability window has been reached, the candidate is fixed for the epoch and will
     /// be used once crossing the epoch boundary to produce the next epoch nonce.
-    fn evolve_nonce(
-        &self,
-        header: &H,
-        global_parameters: &GlobalParameters,
-    ) -> Result<Nonces, Self::Error>;
+    fn evolve_nonce(&self, header: &H) -> Result<Nonces, Self::Error>;
```
abailly, line 106: Does the comment makes sense now? We have removed the `global_parameters` parameter to encapsulate nonces evolution logic somewhere else. It somehow makes sense, but it also obscures the nonces evolution logic somewhat. 
From an API design perspective, having explicit parameters makes it clearer what the nonce evolution depends on 🤔

crates/amaru-ouroboros-traits/src/has_stake_distribution/mod.rs
```diff
@@ -32,16 +33,4 @@ pub trait HasStakeDistribution: Send + Sync {
     /// Obtain information about a pool such as its VRF key hash and its stake. The information is
     /// fetched from the ledger based on the given slot.
     fn get_pool(&self, slot: Slot, pool: &PoolId) -> Option<PoolSummary>;
-
-    /// Calculate the KES period given an absolute slot and some shelley-genesis values
-    fn slot_to_kes_period(&self, slot: Slot) -> u64;
-
-    /// Get the maximum number of KES evolutions from the ledger state
-    fn max_kes_evolutions(&self) -> u64;
```
abailly, line 40: If I am not mistaken, this is a parameter that can be changed over the course of the life of the network, so it's something that we'll need to get from the ledger somehow

crates/amaru-ouroboros/src/praos/header.rs
```diff
@@ -83,11 +85,11 @@ impl PartialEq for AssertHeaderError {
 pub type Assertion<'a> = Box<dyn Fn() -> Result<(), AssertHeaderError> + Send + Sync + 'a>;
 
 pub fn assert_all<'a>(
+    consensus_parameters: Arc<ConsensusParameters>,
     header: &'a Header,
     raw_header_body: &'a [u8],
-    ledger_state: &'a dyn HasStakeDistribution,
+    ledger_state: Arc<dyn HasStakeDistribution>,
```
abailly, line 91: Do we need an `Arc` here, or would an `Rc` suffice?

crates/amaru-ouroboros/src/praos/header.rs
```diff
@@ -83,11 +85,11 @@ impl PartialEq for AssertHeaderError {
 pub type Assertion<'a> = Box<dyn Fn() -> Result<(), AssertHeaderError> + Send + Sync + 'a>;
 
 pub fn assert_all<'a>(
+    consensus_parameters: Arc<ConsensusParameters>,
```
abailly, line 88: The idea behind the `HasStakeDistribution` interface was to enrich it with more information the consensus needs from the ledger, hence the seemingly useless methods you removed. We probably would want to have a discussion with @KtorZ over this design

abailly (commented):
I understand it's not always easy nor convenient when on is working on something and needs to move things around to split it, but it seems to me this PR really should have been split at least in 2 as it mixes 1/ changes on how we pass some parameters needed for header validation through the introduction of `ConsensusParameters` and 2/ actual addition of an effect to handle `header_validation`.
I personally find it hard to review a PR, and even harder when it touches 20+ files.

crates/amaru-consensus/src/consensus/effects/consensus_effects.rs
etorreborre: The reason is that we use store this value inside the `ValidateHeader` structure, which is the same situation as the one we have with the `HeadersTree` where we have an `Arc<dyn ChainStore<H>>`.

crates/amaru-consensus/src/consensus/stages/validate_header.rs
etorreborre: I wondered the same. That span was already there before that PR.

crates/amaru-ouroboros/src/praos/header.rs
etorreborre: I haven't tried but possibly yes downgrading some of these `Arc`s to `Rc`s would work. I propose to do it at once in another PR then.

crates/amaru-ouroboros-traits/src/praos/mod.rs
etorreborre: > From an API design perspective, having explicit parameters makes it clearer what the nonce evolution depends on

I personally prefer to separate static / configuration parameters, from dynamic ones when possible. Because then there's only one place where the static parameters are configured and:

1. It reduces the risk of changing them accidentally by passing modified global parameters for some reason.
2. It reduces the amount of data in flux when trying to understand what comes from where.

crates/amaru-ouroboros-traits/src/has_stake_distribution/mod.rs
etorreborre: Ok, that was indeed not obvious from the current implementation :-).

crates/amaru-ouroboros/src/praos/header.rs
etorreborre: @KtorZ I can put those methods back with a comment saying that there will be some additional ledger logic besides just using the global parameters if you want.

crates/amaru-kernel/src/protocol_parameters.rs
etorreborre: Ok, I can revert this change then and make a smaller PR.

crates/amaru-ouroboros-traits/src/praos/mod.rs
abailly: The point is: those parameters are probably not static

crates/amaru-ouroboros/src/praos/header.rs
abailly: Well, no you're already changing the type and it's just a one letter change

crates/amaru-ouroboros/src/praos/header.rs
abailly: I think your proposal makes sense given the name of the interface and we are mostly not using anyway. Yet it's not really YAGNI because we _know_ we'll need those.

crates/amaru-ouroboros-traits/src/has_stake_distribution/mod.rs
KtorZ: That is correct; although this "interface" here pre-dates a lot of the codebase. It's one of the first thing we wrote in Amaru in anticipation of what the consensus <-> ledger interaction would be. 

Note that we still haven't implemented the operational certificate validation in the header validation. Which is why the `latest_opcert_sequence_number` is mostly unused. As for the max_kes_evolution; I thought we did implement it; but if you've managed to remove it, it may not (or no longer) be the case.

crates/amaru-ouroboros/src/praos/header.rs
KtorZ: For now those methods mostly served as reminder of the _unimplemented bits_ in the header validation. Now we better make sure we remember to implement those bits, eventually 😬 ... (there might be one or two FIXME as well disseminated in the code)

PR #491, opened 2025-10-07
feat: move select chain to the end

etorreborre:
This PR moves the `select_chain` stage to the end of the processing graph in order to make sure that we only select a best chain when a new block is valid:

```mermaid
graph TD
    receive_header[receive header]
    store_header[store header]
    validate_header[validate header]
    fetch_block[fetch block]
    store_block[store block]
    validate_block[validate block]
    select_chain[select chain]
    forward_header[forward header]

    downstream([downstream])
    store@{ shape: cyl, label: "Chain store" }
    net@{ shape: delay, label: "Peers" }

    upstream([upstream]) -.-> net
    upstream -- chain sync --> pull
    pull -- ChainSyncEvent(Vec(u8)) --> receive_header

    receive_header -- NewHeader --> store_header
    receive_header -- ValidateBlock::Rollback --> validate_block
    receive_header -- ChainSyncEvent::CaughtUp --> select_chain

    store_header -.-> store
    store_header -- NewHeader --> validate_header

    validate_header -- FetchBlock --> fetch_block

    fetch_block -- StoreBlock --> store_block
    fetch_block -.-> net

    store_block -.-> store
    store_block -- ValidateBlock::RollForward --> validate_block

    validate_block -.-> store
    validate_block -- ChainSyncEvent::RollForward --> select_chain
    validate_block -- ChainSyncEvent::Rollback --> select_chain

    select_chain -- ForwardHeader::RollForward --> forward_header
    select_chain -- ForwardHeader::Rollback --> forward_header

    forward_header --> downstream
    downstream -.-> net
```

The messages sent between stages have also been reworked so that we send messages directly to the relevant stages without having them needlessly pass-through some stages. For example, if we receive a rollback event we can directly send that event to the ledger to rollback the corresponding block (in the `validate_block` stage), without having to go through the `fetch_block` stage.

For clarity the failures and errors in each stage have been extracted out to a separate diagram:
```mermaid
graph LR
    receive_header[receive header]
    store_header[store header]
    validate_header[validate header]
    fetch_block[fetch block]
    store_block[store block]
    validate_block[validate block]
    select_chain[select chain]

    disconnect_peer[disconnect peer]
    panic[panic]

    receive_header -- malformed header --> disconnect_peer
    validate_header -- invalid header --> disconnect_peer
    fetch_block -- fetch error --> disconnect_peer
    validate_block -- invalid block --> disconnect_peer
    select_chain -- invalid chain --> disconnect_peer

    store_header -- storage failure --> panic
    store_block -- storage failure --> panic
```

Note: as discussed with @rkuhn I added a commit to put `StageRef::named` under a `test-utils` feature flag.


crates/amaru-consensus/src/consensus/events.rs
```diff
@@ -75,237 +75,306 @@ impl fmt::Debug for ChainSyncEvent {
...
+impl PartialEq for NewHeader {
+    fn eq(&self, other: &Self) -> bool {
+        self.header == other.header && self.peer == other.peer
+    }
 }
 
 #[derive(Clone, Serialize, Deserialize)]
-pub enum ValidateBlockEvent {
-    Validated {
+pub enum ForwardHeader {
```
abailly, line 130: this is in effect changing the semantics of this structure from being an _event_ (something happen) to being a _Command_ (do something). That's not how it was intended and I personally would not want to do it that way. To be fair the original name was incorrect it should have been `ValidatedBlock` possibly dropping the `Event` suffix.

crates/amaru-consensus/src/consensus/events.rs
```diff
@@ -75,237 +75,306 @@ impl fmt::Debug for ChainSyncEvent {
...
             _ => false,
         }
     }
 }
 
-#[derive(Debug, Clone, Serialize, Deserialize)]
-pub enum BlockValidationResult {
-    BlockValidated {
+#[derive(Clone, Serialize, Deserialize)]
+pub struct FetchBlock {
```
abailly, line 195: Same remark as above: instead of having a stage producing an event representing the result of its computation (`BlockValidated` or `BlockInvalid`) this is instructing some downstream stage to  `FetchBlock`

crates/amaru-consensus/src/consensus/events.rs
```diff
@@ -75,237 +75,306 @@ impl fmt::Debug for ChainSyncEvent {
...
+}
+
+impl PartialEq for FetchBlock {
+    fn eq(&self, other: &Self) -> bool {
+        self.header == other.header && self.peer == other.peer
+    }
+}
+
+#[derive(Clone, Serialize, Deserialize)]
+pub struct StoreBlock {
```
abailly, line 240: Same remark as above.

crates/amaru-consensus/src/consensus/events.rs
```diff
@@ -75,237 +75,306 @@ impl fmt::Debug for ChainSyncEvent {
...
+}
+
+impl PartialEq for StoreBlock {
+    fn eq(&self, other: &Self) -> bool {
+        self.header == other.header && self.peer == other.peer
+    }
+}
+
+#[derive(Clone, Serialize, Deserialize)]
+pub enum ValidateBlock {
```
abailly, line 306: Same remark as above.

abailly (changes requested):
Again this PR contains 2 orthogonal changes, one of which I disagree with: 
* moving select chain at the end 👍 
* renaming events and changing their semantics 👎 

I would appreciate if the latter was actually discussed, perhaps over another PR, before enacting this change.

crates/amaru-consensus/src/consensus/events.rs
etorreborre: I totally agree with the awkward command naming I introduced. In that case would `ValidatedBlock` even be the proper name? Wouldn't something like `NewChainSelectionResultAvailable`, be more appropriate, knowing that in the present graph, this also means that, if a block had to be validated, it is now valid?

But fundamentally I think I introduced those names because they decouple stages. When we arrive at the `fetch_block` stage, from that stage point of view, it does not really matter that its header was already stored or validated. That doesn't change the work it has to do. It feels to me that the only things that look like events are:

 - `receive_header` because it receives information about different things that happened outside of the node.
 - `select_chain` for the same reason (with some storing/fetching/validating in the middle). 

In any case I can revert back to a more eventful naming, if you agree on the structure of the graph, which I think is the most important part of this PR.

abailly: I would like this PR to be about one and only one thing, namely to move `select_chain` at the end of the processing pipelining, _without_ any other change. 
Then I am happy to have a work session with you on redesigning the structure of the graph and the naming of events.

etorreborre: OK I can limit the changes but there needs to be _some_ event changes since we reshuffle stages.

PR #492, opened 2025-10-07
feat: print command argument details 

jeluard:
Rename `daemon` to `run` (as this command doesn't spawn a daemon).

Also prints global and command argument details to improve transparency.

e.g.

```rust
2025-10-07T15:52:04.589680Z  INFO amaru: With global parameters:
- otlp_metric_url: "http://localhost:4318/v1/metrics"  
- otlp_service_name: "amaru"  
- otlp_span_url: "http://localhost:4317"  
- with_json_traces: "false"  
- with_open_telemetry: "false"
2025-10-07T15:52:04.589728Z  INFO amaru: Running command: 'run' with arguments
- listen_address: "0.0.0.0:3000"  
- max_downstream_peers: "10"  
- max_extra_ledger_snapshots: "0"  
- network: "mainnet"  
- peer_address: "localhost:3001"
```


crates/amaru/src/bin/amaru/main.rs
```diff
@@ -104,6 +108,37 @@ struct Cli {
     otlp_metric_url: String,
 }
 
+struct Arguments(BTreeMap<String, String>);
+
+impl Display for Arguments {
```
jeluard: I couldn't find an easy way to add dynamic fields so I just format them as a single line.

KtorZ: They don't have to be dynamic, though? We know all of them at build time

KtorZ: We can use "field::empty" as a default, and pluck the right values when necessary.

jeluard: That would defeat the idea of having a generic way to handle commands and/or make things more complex.

Any reason to have those arguments as fields of the trace? It looks like it would not change much the output.

KtorZ: I would expect traces to be independent of their representations. Otherwise, we loose the ability to pick a different output format (e.g. JSON). Here, the trace is being tailored to the pretty-printing output; which isn't a good tradeoff. Traces are _structured data_ before they are printed strings. 

So I rather have non-generic traces to be honest. I don't see what the generic aspect gets us here beyond a bit less "boilerplate" when defining new arguments / options. But hey, if it makes people think twice before cluttering the command-line, I think that's even a good thing.

jeluard: Now moved to all manual

crates/amaru/src/bin/amaru/main.rs
```diff
@@ -112,9 +114,9 @@ struct Cli {
     #[clap(long, action, env("AMARU_WITH_JSON_TRACES"))]
     with_json_traces: bool,
 
-    #[arg(long, value_name = "STRING", env("AMARU_SERVICE_NAME"), default_value_t = DEFAULT_SERVICE_NAME.to_string()
+    #[arg(long, value_name = "STRING", env("AMARU_OTLP_SERVICE_NAME"), default_value_t = DEFAULT_SERVICE_NAME.to_string()
```
abailly: Shouldn't we also rename `DEFAULT_SERVICE_NAME` for consistency's sake?

abailly: I am not sure about pinning the name to otlp use only, the concept of a "service name" can be useful in a lot of contexts, eg. OS-level service configuration

jeluard: It replicates the arg name `--otlp-service-name` so I feel like for now that's probably the more logical path?
Note that there are [existing standard env var](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/) for those that we might want to reuse (e.g. `OTEL_SERVICE_NAME`). They might actually work out of the box.

PR #493, opened 2025-10-08
test: introduce a separate config for the simulated nodes

etorreborre:
The current simulation setup is a bit awkward since we:

 - Get arguments, `Args`
 - Extract a `SimulateConfig` from `Args`
 - But still pass `Args` down the line in addition to that config.

This PR extract another config data type, `NodeConfig` that contains the configuration that is relevant to a single simulated node, whereas `SimulateConfig` contains the configuration required for the simulation as a whole.

## Summary by CodeRabbit

- **New Features**
  - Added a dedicated per-node configuration with sensible defaults, including peer counts and file path settings.
  - Centralized chain/block file path handling within node configuration.

- **Refactor**
  - Simulator now uses the new node configuration across node spawning, initialization, and input generation.
  - Simulation configuration no longer includes peer count settings; those are managed per node.
  - Public API updated to accept the new node configuration where applicable.

- **Tests**
  - Updated simulation tests to initialize and use both simulation and per-node configuration.

### DIFF
--- a/simulation/amaru-sim/src/simulator/data_generation/generate.rs
+++ b/simulation/amaru-sim/src/simulator/data_generation/generate.rs
@@ -24,7 +24,7 @@ use crate::echo::Envelope;
 use crate::simulator::bytes::Bytes;
 use crate::simulator::data_generation::base_generators::generate_arrival_times;
 use crate::simulator::data_generation::chain::{Block, Chain};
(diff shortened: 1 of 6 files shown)


etorreborre:
> and one or more NodeConfigs?

For now just one, but eventually there could be different node configuration instances for different nodes, yes.

PR #494, opened 2025-10-09
"Strict" txbody POC

rrruko:
Proof-of-concept implementation of the "strict wrapper" idea that we discussed for pedantic decoding, applied just to the TransactionBody type. Note that the decoder for the pallas type discards duplicate entries, so we need to decode the same bytes a second time if we want to check for duplicates (of course, we could also choose to not reuse the pallas decoder at all).

Note also that we have to explicitly traverse all of the fields of the transaction body and then explicitly traverse the sub-fields of those fields etc. in order to make sure the whole thing is valid. We previously discussed a potential PR to pallas to implement validations as optional data in the minicbor context so that amaru's strict wrapper decoder can simply check for those validations. I don't think we reached a consensus on whether to do that, but in my opinion it is a better and more composable solution, whereas (exhaustively, recursively) validating structures from the top down as in this PR seems more error prone to me.

Since this is a POC I didn't try to exhaustively implement every validation. In particular, the address validation is not as strict as the haskell node's conway CBOR decoder, because the Address decoder from pallas-addresses does not implement the "rigorous" decoder used from conway onward, where extra garbage bytes are not allowed at the end of an address.

## Summary by CodeRabbit

- New Features
  - Introduces strict decoding and validation for transactions, enforcing required fields, detecting duplicates, validating addresses and multi-asset sizes, and rejecting empty-but-present optional sections. This results in clearer error messages and improved safety and reliability when parsing and validating transaction data.

- Tests
  - Adds comprehensive test coverage for successful decoding and a range of failure scenarios, including missing fields, duplicates, malformed addresses, and oversized multi-asset cases.

### DIFF
--- a/crates/amaru-kernel/src/strict_txbody.rs
+++ b/crates/amaru-kernel/src/strict_txbody.rs
@@ -1,6 +1,7 @@
 use pallas_codec::minicbor::{Decode, Decoder, decode::Error};
-use pallas_codec::utils::AnyCbor;
-use pallas_codec::utils::KeyValuePairs;
+use pallas_codec::utils::{AnyCbor, Bytes, KeyValuePairs};
+use pallas_addresses::Address;
+use pallas_primitives::alonzo;
 use pallas_primitives::conway::{
     Multiasset, PseudoTransactionOutput, TransactionBody, TransactionOutput, Value,
 };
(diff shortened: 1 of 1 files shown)


crates/amaru-kernel/src/strict_txbody.rs
```diff
@@ -0,0 +1,547 @@
...
+
+    let policy_count = ma.deref().len();
+    let mut asset_count = 0;
+    for (_policy, assets) in ma.deref().iter() {
+        asset_count += assets.len();
+    }
+
+    let size = per_asset_size * asset_count + per_policy_size * policy_count;
+    size <= 65535
+}
```
KtorZ, line 38: Is this checking that the size of a serialised multi-asset is smaller than 65K bytes? Any reference to the Haskell code here? The check seems quite redundant with the transaction max size.

yHSJ:
We have agreed this is not the solution. Thanks for doing this work @rrruko, it was very helpful. I'm closing this PR since the decision was finalized.

PR #500, opened 2025-10-13
refactor: remove store_header and store_block as separate stages

etorreborre:
This PR removes the `store_header` and `store_block` stages since they just encapsulate effects that can be executed within `receive_header` and `fetch_block` respectively.

## Summary by CodeRabbit

* **New Features**
  * Clear separation of validation vs. processing failures for more precise error reporting.
* **Refactor**
  * Streamlined consensus pipeline by removing intermediate storage stages and rewiring stage flow for simpler, more direct processing.
  * Improved error routing to dedicated failure paths, enhancing robustness and observability.
* **Tests**
  * Updated simulation traces to reflect the simplified pipeline (removed intermediate store stages).
* **Chores**
  * Reduced public surface of internal modules to minimize exposure and tighten module boundaries.


crates/amaru-consensus/src/consensus/stages/fetch_block.rs
```diff
@@ -183,7 +197,8 @@ mod tests {
 
     fn make_state() -> State {
         let downstream: StageRef<ValidateBlockEvent> = StageRef::named("downstream");
-        let errors: StageRef<ValidationFailed> = StageRef::named("errors");
-        (downstream, errors)
+        let failures: StageRef<ValidationFailed> = StageRef::named("failures");
+        let errors: StageRef<ProcessingFailed> = StageRef::named("errors");
```
abailly, line 201: Do we really need to distinguish those 2 stages at this point?

abailly: This seems to me it introduces additional complexity for a unclear benefits.

crates/amaru-consensus/src/consensus/stages/fetch_block.rs
```diff
@@ -183,7 +197,8 @@ mod tests {
...
+        (downstream, failures, errors)
```
abailly, line 202: At this point, we should probably return a `struct` rather than an anonymous tuple

crates/amaru-consensus/src/consensus/stages/fetch_block.rs
etorreborre: The idea is to distinguish failures, that don't stop the node, from errors, that should stop it. At the moment we don't really deal with either, we just terminate.

abailly: I understand the idea, and as you noticed, for now we just always terminate so we don't need this distinction (yet?).

etorreborre: Maybe an alternative would be to have `stage` return a `Result<State>` instead of just `State` in order to deal with fatal errors? cc @rkuhn

crates/amaru-consensus/src/consensus/stages/fetch_block.rs
etorreborre: I would even prefer to have a stage being a struct with those values that never change, instead of being an anonymous function that passes them around unchanged :-).

crates/amaru-consensus/src/consensus/stages/fetch_block.rs
rkuhn: I’d prefer a single stage that dynamically either shuts down the node or disconnects a peer (or whatever else is necessary). We just need to send clearly distinguishable error types to that stage.

Regarding returning `Result` from a stage transition function: that’s what I started out with, but it was annoying to work with, the types got in the way. We can create an extension method on Result and Option to extract the value or terminate the stage; this would be a few characters more than a `?`-based approach (which is more than ? because you want to log or provide error context anyway).

crates/amaru-consensus/src/consensus/stages/fetch_block.rs
rkuhn: The state can of course be a struct, and in that case you’d just pass it back without moving the properties out of it (unless you want to, for some reason). The reason why the state needs to be passed by value is that otherwise the Future isn’t `'static` and all hell breaks loose. It is of course possible to work with lifetimes, but it requires `unsafe` and great care. And if you want to make a trait with a method that takes `self` and returns `Self`, then that isn’t dyn-compatible, defeating the whole purpose. Please believe me when I say that the design space for such abstractions is pretty narrow and of fractal dimension for Rust — stay out of that rabbit hole if you want to retain your sanity.

PR #503, opened 2025-10-14
feat: make a full effect for validating headers

etorreborre:
This PR adds a full external effect for validating a header. This allows the simulation to easily override the `validate_header` stage and allow headers to be validated even though the generated data does not produce fully consistent headers.

Note: @abailly this pushes us in a direction where 2 stages, `validate_block` and `validate_header` are mere wrappers on top of external effects. Which begs the question: should they be fused with `receive_header` and `fetch_block` like we did for `store_header` and `store_block`?

## Summary by CodeRabbit

- New Features
  - Added header-centric validation flow with a dedicated ValidateHeader effect and HeaderValidationError.
  - Simplified node bootstrap: stage graph no longer requires global parameters or era history.

- Refactor
  - Reworked ledger validation APIs and resources to a header-first interface; updated error mapping and stage wiring.
  - Adjusted stage and effect names and public signatures to reflect header validation.

- Tests
  - Added mock header validator and updated traces/tests to match new validation path.

- Chores
  - Removed legacy stake-distribution wiring; annotated unused simulator code as dead code.

### DIFF
--- a/crates/amaru-consensus/src/consensus/effects/ledger_effects.rs
+++ b/crates/amaru-consensus/src/consensus/effects/ledger_effects.rs
@@ -14,19 +14,23 @@
 
 use crate::consensus::errors::ProcessingFailed;
 use amaru_kernel::peer::Peer;
-use amaru_kernel::{Point, PoolId, RawBlock};
+use amaru_kernel::{Point, RawBlock};
 use amaru_metrics::ledger::LedgerMetrics;
-use amaru_ouroboros_traits::{
-    BlockValidationError, CanValidateBlocks, HasStakeDistribution, PoolSummary,
-};
-use amaru_slot_arithmetic::Slot;
+use amaru_ouroboros_traits::can_validate_blocks::{CanValidateHeaders, HeaderValidationError};
+use amaru_ouroboros_traits::{BlockHeader, BlockValidationError, CanValidateBlocks};
 use pure_stage::{BoxFuture, Effects, ExternalEffect, ExternalEffectAPI, Resources, SendData};
 use std::sync::Arc;
 
 /// Ledger operations available to a stage.
 /// This trait can have mock implementations for unit testing a stage.
-pub trait LedgerOps: HasStakeDistribution {
-    fn validate(
+pub trait LedgerOps: Send + Sync {
+    fn validate_header(
+        &self,
+        point: &Point,
+        header: &BlockHeader,
+    ) -> BoxFuture<'_, Result<(), HeaderValidationError>>;
+
+    fn validate_block(
         &self,
         peer: &Peer,
         point: &Point,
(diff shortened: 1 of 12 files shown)


crates/amaru-consensus/src/consensus/effects/ledger_effects.rs
```diff
@@ -117,64 +123,67 @@ impl ExternalEffectAPI for ValidateBlockEffect {
...
-                .get::<ResourceBlockValidation>()
-                .expect("RollbackBlockEffect requires a ResourceBlockValidation resource")
+                .get::<ResourceHeaderValidation>()
+                .expect("ValidateBlockEffect requires a ResourceHeaderValidation resource")
                 .clone();
-            validator
-                .rollback_block(&self.point)
-                .map_err(|e| ProcessingFailed::new(&self.peer, e.to_anyhow()))
+            validator.validate_header(&self.point, &self.header)
         })
```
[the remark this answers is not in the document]
etorreborre: Good rabbit

crates/amaru-ouroboros-traits/src/can_validate_blocks/mod.rs
```diff
@@ -91,3 +92,73 @@ impl PartialEq for BlockValidationError {
...
+    fn from(err: anyhow::Error) -> Self {
+        HeaderValidationError::new(err)
+    }
+}
+
+impl Display for HeaderValidationError {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(f, "BlockValidationError: {}", self.0)
+    }
+}
```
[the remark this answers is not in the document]
etorreborre: ok

crates/amaru-ouroboros-traits/src/can_validate_blocks/mod.rs
```diff
@@ -91,3 +92,73 @@ impl PartialEq for BlockValidationError {
         self.0.to_string() == other.0.to_string()
     }
 }
+
+pub trait CanValidateHeaders: Send + Sync {
```
abailly, line 96: Could you elaborate on why this new trait is needed in lieu of `HasStakeDistribution`? It seems they are serving the same purpose, perhaps not at the same level in the system? Maybe there's something to document here?

etorreborre: `HasStakeDistribution` is used deeply in the behavior of `ValidateHeaders`. It is difficult to run the header validation and make it succeed by just being able to override that trait during the simulation. Hence the `CanValidateHeaders` trait. Then the `HasStakeDistribution` trait is still useful if we want to write a test at the level of `ValidateHeader` where we can inject a `ChainStore` and some consensus parameters to check the actual logic.

abailly: I see. At this point I feel we need to start maintaining a slightly more sophisticated architecture diagram than the one in the `amaru-consensus/README` file :)

abailly: But possibly in another PR

PR #510, opened 2025-10-17
test: improve the data generation for the consensus simulation

etorreborre:
This PR uses the data generation support testing the `HeadersTree` in the simulation runs:

 * New arguments have been added to control the data generation (rollback ratio, branching ratio).
 * A bug is fixed to allow the simulation to run with several downstream peers.
 * Validation errors now don't terminate the node (only processing errors, like not being able to write in the database, do).
 * The whole downstream chain is now tested instead of just its tip.

## Summary by CodeRabbit

* **New Features**
  * Configurable chain generation with per-peer branching and rollback ratios; CLI accepts ratio strings.
  * Message debug output now includes parent-hash for easier troubleshooting.
  * Exposed parent-hash accessor for header inspection.

* **Bug Fixes**
  * Added round-trip header serialization test to catch decode regressions.
  * Validation errors are logged instead of aborting execution.

* **Refactoring**
  * Switched from file-based to generated chain inputs and consolidated simulator config around ratio-based generation.

### DIFF
--- a/crates/amaru-consensus/src/consensus/headers_tree/data_generation/actions.rs
+++ b/crates/amaru-consensus/src/consensus/headers_tree/data_generation/actions.rs
@@ -23,7 +23,7 @@
 use crate::consensus::errors::ConsensusError;
 use crate::consensus::headers_tree::Tracker::{Me, SomePeer};
 use crate::consensus::headers_tree::data_generation::SelectionResult::{Back, Forward};
-use crate::consensus::headers_tree::data_generation::any_tree_of_headers;
+use crate::consensus::headers_tree::data_generation::any_tree_of_headers_and_best_chain;
 use crate::consensus::headers_tree::tree::Tree;
 use crate::consensus::headers_tree::{HeadersTree, Tracker};
 use crate::consensus::stages::select_chain::RollbackChainSelection::RollbackBeyondLimit;
@@ -377,13 +377,13 @@ pub fn generate_random_walks(
         .collect()
 }
 
-/// Generate a random list of actions, for a fixed number of peers, with:
+/// Generate a random list of actions, for a number of peers.
 ///
-///  - A tree of headers of depth `depth`.
-///  - A `max_length` for how far rollback actions can go in the past.
+/// We first generate a tree of headers of depth `depth` with a given `branching_ratio`
+/// (the probability that a branch is created from an original spine of headers).
 ///
-/// If we use `max_length` < depth we will generate test cases where the `HeadersTree` will
-/// have to be truncated after a number of roll forwards.
+/// Then we execute a random walk on that tree for `peers_nb` peers, with a given `rollback_ratio`,
+/// which is how often a given peer will rollback.
 ///
 /// Important note: this function returns a prop test strategy but the random walks are generated
 /// using a `StdRng` generator. This makes the generator reproducible, because the `StdGenerator`
@@ -394,11 +394,17 @@ pub fn any_select_chains(
     depth: usize,
     rollback_ratio: Ratio,
     branching_ratio: Ratio,
-) -> impl Strategy<Value = Vec<Action>> {
-    any_tree_of_headers(depth, branching_ratio).prop_flat_map(move |tree| {
-        (1..u64::MAX)
-            .prop_map(move |seed| generate_random_walks(&tree, peers_nb, rollback_ratio, seed))
-    })
+) -> impl Strategy<Value = (Vec<Action>, Chain)> {
+    any_tree_of_headers_and_best_chain(depth, branching_ratio).prop_flat_map(
+        move |(tree, best_chain)| {
+            (1..u64::MAX).prop_map(move |seed| {
+                (
+                    generate_random_walks(&tree, peers_nb, rollback_ratio, seed),
+                    best_chain.clone(),
+                )
+            })
+        },
+    )
 }
 
 /// Create an empty `HeadersTree` handling chains of maximum length `max_length` and
--- a/simulation/amaru-sim/src/echo/simulate.rs
+++ b/simulation/amaru-sim/src/echo/simulate.rs
@@ -69,10 +69,10 @@ pub fn spawn_echo_node(_node_id: String) -> NodeHandle<EchoMessage> {
 }
 
 /// Generate some input echo messages at different arrival times.
-pub fn echo_generator(rng: &mut StdRng) -> Vec<Entry<EchoMessage>> {
+pub fn echo_generator(rng: &mut StdRng) -> (Vec<Entry<EchoMessage>>, ()) {
     let now = Instant::at_offset(Duration::from_secs(0));
     let size = 20;
-    generate_zip_with(
+    let entries = generate_zip_with(
         size,
         generate_vec(generate_u8(0, 128)),
         generate_arrival_times(now, 200.0),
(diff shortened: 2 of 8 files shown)


crates/amaru-consensus/src/consensus/headers_tree/data_generation/actions.rs
```diff
@@ -315,6 +316,30 @@ pub fn random_walk(
...
+        }
+        let numerator = parts[0]
+            .parse::<u32>()
+            .map_err(|_| format!("Invalid numerator in ratio: {}", parts[0]))?;
+        let denominator = parts[1]
+            .parse::<u32>()
+            .map_err(|_| format!("Invalid denominator in ratio: {}", parts[1]))?;
+        Ok(Ratio(numerator, denominator))
+    }
+}
```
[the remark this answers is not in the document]
etorreborre: Thanks

simulation/amaru-sim/src/simulator/run.rs
```diff
@@ -240,39 +237,107 @@ fn make_chain_selector(
...
+    let mut current_chains: BTreeMap<String, Chain> = BTreeMap::new();
+
+    // also keep track of the best chain seen so far
+    let best_chain_tracker = "best".to_string();
+    current_chains.insert(best_chain_tracker.clone(), vec![]);
+    for message in messages {
+        // only consider messages from the peers
+        if !message.src.starts_with("c") {
+            continue;
+        };
```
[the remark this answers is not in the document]
etorreborre: Will fix later

simulation/amaru-sim/src/simulator/run.rs
```diff
@@ -240,39 +237,107 @@ fn make_chain_selector(
...
+}
+
+/// Build the best chain from messages sent to downstream peers.
+pub fn make_best_chain_from_downstream_messages(history: &History<ChainSyncMessage>) -> Chain {
+    let mut best_chain = vec![];
+    for (i, message) in history.0.iter().enumerate() {
+        // only consider messages sent to the first peer
+        if !message.dest.starts_with("c1") {
+            continue;
+        };
```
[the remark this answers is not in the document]
etorreborre: Will fix later

simulation/amaru-sim/src/simulator/run.rs
```diff
@@ -240,39 +237,107 @@ fn make_chain_selector(
...
+        match &message.body {
+            msg @ ChainSyncMessage::Fwd { .. } => {
+                best_chain.push(msg.decode_block_header().unwrap())
```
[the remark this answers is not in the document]
etorreborre: Good idea

simulation/amaru-sim/src/simulator/run.rs
```diff
@@ -240,39 +237,107 @@ fn make_chain_selector(
...
-            } else {
-                Ok(())
+        // update the best chain seen so far
+        let best_length = current_chains.values().map(|c| c.len()).max().unwrap();
+        let best_chain = current_chains
+            .clone()
+            .into_values()
+            .filter(|c| c.len() == best_length)
+            .collect::<Vec<Chain>>();
+        current_chains.insert(best_chain_tracker.clone(), best_chain[0].clone());
```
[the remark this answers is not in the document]
etorreborre: Ok for now.

simulation/amaru-sim/src/simulator/data_generation/generate.rs
```diff
@@ -12,699 +12,72 @@
...
-                                dest: "n1".to_string(),
-                                body: msg,
-                            },
-                        })
-                    })
-                    .collect::<Vec<_>>(),
-            );
+) -> impl Fn(&mut R) -> Vec<Entry<ChainSyncMessage>> + use<'_, R> {
+    move |rng: &mut R| {
+        let actions = run_with_rng(
```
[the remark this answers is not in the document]
etorreborre: Done.

crates/amaru-consensus/src/consensus/stages/receive_header.rs
```diff
@@ -44,6 +44,8 @@ pub async fn stage(
             raw_header,
             span,
         } => {
+            // TODO: check the point vs the deserialized header point and invalidate if they don't match
+            // then simplify and don't pass the point separately
```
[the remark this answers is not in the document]
etorreborre: This will be done in a separate PR

crates/amaru-consensus/src/consensus/headers_tree/data_generation/actions.rs
```diff
@@ -315,6 +316,40 @@ pub fn random_walk(
 #[derive(Copy, Clone, PartialEq, Eq, Debug)]
 pub struct Ratio(pub u32, pub u32);
```
abailly, line 317: I am surprised there's not already a proper rational type in one of the libs we are using 🤔

simulation/amaru-sim/src/echo/simulate.rs
```diff
@@ -70,33 +69,34 @@ pub fn spawn_echo_node(_node_id: String) -> NodeHandle<EchoMessage> {
 }
 
 /// Generate some input echo messages at different arrival times.
-pub fn echo_generator(rng: &mut StdRng) -> Vec<Reverse<Entry<EchoMessage>>> {
+pub fn echo_generator(rng: &mut StdRng) -> Vec<Entry<EchoMessage>> {
```
abailly: At this stage I wonder if we should not get rid of the `Echo` stuff

simulation/amaru-sim/src/simulator/run.rs
```diff
@@ -240,39 +237,112 @@ fn make_chain_selector(
...
+            actual, expected,
+            "\nThe actual chain\n{}\n\nis not the best chain\n\n{}\n\nThe history is:\n{:?}",
+            actual_chain, expected_chain, history,
+        );
+        Ok(())
+    }
+}
+
+/// Build the best chain from messages incoming from upstream peers.
+pub fn make_best_chain_from_upstream_messages(messages: &[Envelope<ChainSyncMessage>]) -> Chain {
```
abailly: It seems odd we have to write this complicated logic to compute the best chain from upstream messages, are we not duplicating some of the logic of the chain selection? Shouldn't the chain generator already know what the best chain is? I think we already had that discussion 😓

simulation/amaru-sim/src/echo/simulate.rs
etorreborre: This is used to test the simulation framework on a simple system but I also can't tell how useful this is by now.

crates/amaru-consensus/src/consensus/headers_tree/data_generation/actions.rs
etorreborre: We indeed have `num-rational` as a transitive dependency but here we impose `numerator <= denominator` (so maybe the name is misleading, it's more like a probability represented as a fraction?)

abailly: Couldn't we use directly a probability then? I understand the desire to have a more precise type, just feel that this seems such a common thing we should not write it ourselves :)

simulation/amaru-sim/src/echo/simulate.rs
abailly: Let's keep it for now, and remove it in a later version.

simulation/amaru-sim/src/simulator/run.rs
etorreborre: This is simpler logic than chain selection but you're right we can keep the original spine for the generated tree of headers and pass it around. I pushed a commit for this.

simulation/amaru-sim/src/echo/simulate.rs
etorreborre: @abailly Would you rather suppress it now?

abailly: Your call!

PR #511, opened 2025-10-17
feat: check that a received header has the correct point

etorreborre:
This PR checks that a received header, when deserialized, has the same point as the point declared in the chain sync event.

## Summary by CodeRabbit

* **Refactor**
  * Simplified header validation interfaces and reduced payloads passed between stages and events.
* **Bug Fixes**
  * Added explicit detection and a new error for header/identifier mismatches; validation now fails earlier to avoid storing invalid headers.
* **Tests**
  * Updated tests to match the streamlined flow and added a helper to generate headers with guaranteed parents.

### DIFF
--- a/crates/amaru-consensus/src/consensus/stages/select_chain.rs
+++ b/crates/amaru-consensus/src/consensus/stages/select_chain.rs
@@ -344,6 +344,9 @@ mod tests {
         let consensus_ops = mock_consensus_ops();
 
         let store = consensus_ops.store();
+        if let Some(parent_hash) = header.parent() {
+            store.set_anchor_hash(&parent_hash)?;
+        }
         let anchor = store.get_anchor_hash();
         let (select_chain, _, _) = stage(
             make_state(store.clone(), &peer, &anchor),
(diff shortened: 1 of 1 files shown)


crates/amaru-ouroboros-traits/src/is_header/tests.rs
```diff
@@ -85,13 +85,18 @@ pub fn any_header() -> impl Strategy<Value = BlockHeader> {
...
 pub fn any_header_with_parent(parent: HeaderHash) -> impl Strategy<Value = BlockHeader> {
     (0u64..=1_000_000, 0u64..=1_000_000).prop_map(move |(block_number, slot)| {
         BlockHeader::from(make_header(block_number, slot, Some(parent)))
     })
 }
 
+/// Create an arbitrary BlockHeader, with an arbitrary parent that is guaranteed to be Some
+pub fn any_header_with_some_parent() -> impl Strategy<Value = BlockHeader> {
+    any_header().prop_flat_map(|h| any_header_with_parent(h.hash))
+}
```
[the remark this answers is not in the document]
etorreborre: Done

crates/amaru-consensus/src/consensus/stages/receive_header.rs
```diff
@@ -46,14 +46,33 @@ pub async fn stage(
...
-                    if let Err(error) = result {
+                    if header.point() != point {
+                        tracing::error!(%point, %peer, "Header point {} does not match expected point", header.point());
                         eff.base()
-                            .send(&errors, ProcessingFailed::new(&peer, anyhow!(error)))
+                            .send(
+                                &failures,
+                                ValidationFailed::new(
+                                    &peer,
+                                    ConsensusError::CannotDecodeHeader {
```
abailly: I would rather have an explicit variant for this specific error, rather than having to unpack the `reason` field to know exactly why the validation failed.

etorreborre: @abailly good idea, I pushed a new commit for this.

PR #513, opened 2025-10-17
make upstream connections dynamic

rkuhn:
This PR replaces the pre-created gasket stages for upstream peers with a NetworkResource
that is used by a very simple pull stage that uses the new ChainSyncEffect based on this resource.

I just noticed that I haven’t yet wired up the FetchBlock and Disconnect effects, which probably
means that e2e tests will fail - funny (or rather sad) that none of the local test cases catch this.
Should probably add a better simulation test case for this that doesn’t abstract the communication
quite as much. The eventual plan is to lower the simulation testing setup to actual network effects
instead of the current high-level chain sync communication.

## Summary by CodeRabbit

* **New Features**
  * Actor-based peer connection manager and network resource for improved peer syncing and explicit disconnect handling
  * Simulation: preload messages into stages before runtime; runtime-aware simulation node spawning

* **Refactor**
  * Consolidated consensus/network event types under shared kernel abstractions
  * Replaced legacy fetch plumbing with unified network operations and simplified bootstrap/peer lifecycle

* **Chores**
  * Workspace and dependency reorganizations; public API surface adjustments and module reshaping

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -116,7 +116,6 @@ dependencies = [
 name = "amaru-consensus"
 version = "0.1.0"
 dependencies = [
- "acto",
  "amaru-kernel",
  "amaru-metrics",
  "amaru-ouroboros",
@@ -128,7 +127,6 @@ dependencies = [
  "hex",
  "itertools 0.14.0",
  "pallas-crypto",
- "parking_lot",
  "pprof",
  "proptest",
  "pure-stage",
--- a/crates/amaru-consensus/Cargo.toml
+++ b/crates/amaru-consensus/Cargo.toml
@@ -18,13 +18,11 @@ profiling = ["dep:pprof"]
 test-utils = ["proptest", "dep:rand", "serde_json"]
 
 [dependencies]
-acto.workspace = true
 anyhow.workspace = true
 async-trait.workspace = true
 hex.workspace = true
 itertools.workspace = true
 pallas-crypto.workspace = true
-parking_lot.workspace = true
 proptest = { workspace = true, optional = true }
 rand = { workspace = true, optional = true }
 rayon.workspace = true


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "rust-analyzer.cargo.features": ["amaru-upstream", "test-utils", "profiling"]
```
rkuhn: due to the way some crates are set up, not setting these will lead to errors in my IDE — let me know if you want me to fix the code instead

crates/amaru-consensus/src/consensus/effects/network_effects.rs
```diff
@@ -204,3 +175,213 @@ impl ExternalEffect for ForwardEventEffect {
...
+                .get::<NetworkResource>()
+                .expect("DisconnectEffect requires a NetworkResource")
+                .clone();
+            network.disconnect(&self.peer).await
+        })
+    }
+}
+
+#[derive(Clone)]
+pub struct NetworkResource {
```
rkuhn: this (and below) is the actual new code — the rest of the PR resulted from a rebase from hell, mostly due to imports

crates/amaru-consensus/src/consensus/stages/pull.rs
```diff
@@ -0,0 +1,30 @@
...
+use pure_stage::{Effects, StageRef};
+
+#[derive(Debug, PartialEq, serde::Serialize, serde::Deserialize)]
+pub struct NextSync;
+
+pub async fn stage(
+    downstream: StageRef<ChainSyncEvent>,
+    _msg: NextSync,
+    eff: Effects<NextSync>,
+) -> StageRef<ChainSyncEvent> {
```
rkuhn, line 26: This is the new upstream stage design, will remain like this also after any network layer overhaul: just keep pulling chain sync updates one after the other. My only question is whether we’ll want to put the header deduplication into the pure-stage part or into the resource part.

crates/amaru-consensus/src/consensus/stages/select_chain.rs
```diff
@@ -148,19 +178,31 @@ impl SelectChain {
...
+            };
+
+            if ret.as_ref().map(|evs| !evs.is_empty()).unwrap_or(false) {
+                // FIXME add some more points in between?
+                let hashes = [store.get_anchor_hash(), store.get_best_chain_hash()];
+                let points = hashes
+                    .iter()
+                    .filter_map(|h| store.load_header(h).map(|h| h.point()))
+                    .collect();
+                *self.intersection.write() = points;
```
rkuhn: @abailly This is the place where chain updates are communicated to the upstream stages for better intersection after reconnect.

crates/amaru/src/bin/amaru/cmd/mod.rs
```diff
@@ -37,3 +38,24 @@ pub(crate) async fn connect_to_peer(
...
+        match self {
+            WorkerError::Recv => write!(f, "error receiving work unit through input port"),
+            WorkerError::Panic => write!(f, "operation panic, stage should stop"),
+            WorkerError::Restart => write!(f, "operation requires a restart of the stage"),
+            WorkerError::Retry => write!(f, "operation should be retried"),
+        }
+    }
+}
+
+impl std::error::Error for WorkerError {}
```
rkuhn, line 62: copied this error type from gasket as the only remaining used part

crates/amaru/src/stages/mod.rs
```diff
@@ -260,17 +223,19 @@ pub async fn bootstrap(
...
+    let acto_runtime = AcTokio::from_handle("network", Handle::current().clone());
+
+    let receive_header_stage =
+        build_stage_graph(chain_selector, sync_tracker, our_tip, &mut network);
 
-    let graph_input = build_stage_graph(chain_selector, sync_tracker, our_tip, &mut network);
-    let graph_input = network.input(&graph_input);
+    let pull_stage = network.stage("pull", pull::stage);
+    let pull_stage = network.wire_up(pull_stage, receive_header_stage);
+    assert!(network.preload(pull_stage, vec![pull::NextSync]));
```
rkuhn, line 233: This is a new pure-stage feature: the pull stage needs the initial message to get going, it will then keep going by sending to itself.

crates/amaru-consensus/src/consensus/stages/pull.rs
```diff
@@ -0,0 +1,30 @@
...
+    let msg = eff.external(ChainSyncEffect).await;
+    eff.send(&downstream, msg).await;
+    eff.send(eff.me_ref(), NextSync).await;
```
abailly, line 29: This is looping by sending `NextSync` message to ourselves?

crates/amaru-consensus/src/consensus/stages/select_chain.rs
abailly: Would it not be better to use the `ReadOnlyChainStore` for that? Seems to me this would make the overall interactions simpler to understand

crates/amaru/src/bin/amaru/cmd/mod.rs
abailly: Seems like the only place we are using this is in the `fetch_block_headers` command? We should use our own error I guess

crates/amaru-consensus/src/consensus/effects/network_effects.rs
```diff
@@ -204,3 +176,213 @@ impl ExternalEffect for ForwardEventEffect {
...
+
+impl ExternalEffect for FetchBlockEffect {
+    fn run(self: Box<Self>, resources: Resources) -> BoxFuture<'static, Box<dyn SendData>> {
+        Self::wrap(async move {
+            #[expect(clippy::expect_used)]
+            let network = resources
+                .get::<NetworkResource>()
+                .expect("BlockFetchEffect requires a NetworkResource")
+                .clone();
+            let point = self.point.clone();
```
abailly, line 216: do you really need to define a variable here? the `clone()` could be in the ` fetch_block` call and then use directly `self.point` in the error?

crates/amaru-consensus/src/consensus/effects/network_effects.rs
```diff
@@ -204,3 +176,213 @@ impl ExternalEffect for ForwardEventEffect {
...
+        peer: &Peer,
+        point: Point,
+    ) -> Result<Vec<u8>, BlockFetchClientError> {
+        let (tx, rx) = oneshot::channel();
+        let tx = Arc::new(Mutex::new(Some(tx)));
+        if let Some(peer) = self.connections.get(peer) {
+            peer.send(ConnMsg::FetchBlock(point.clone(), tx.clone()));
+        }
+        #[cfg(feature = "amaru-upstream")]
+        let others = rand::seq::IteratorRandom::choose_multiple(
```
abailly: What's this for?

crates/amaru-consensus/src/consensus/effects/network_effects.rs
```diff
@@ -204,3 +176,213 @@ impl ExternalEffect for ForwardEventEffect {
...
+            self.connections.keys(),
+            &mut rand::rng(),
+            3,
+        );
+        #[cfg(not(feature = "amaru-upstream"))]
+        let others = self.connections.keys().take(3);
+        for p in others {
```
abailly: I see: we are fetching the block from multiple peers in parallel and taking the first one?

crates/amaru-consensus/src/consensus/effects/network_effects/connection.rs
```diff
@@ -0,0 +1,200 @@
...
+                            NextResponse::Await => unreachable!(),
+                        }
+                    }
+                }
+            }
+        });
+
+        loop {
+            let mut may_fetch = if let State::Running(f) = &mut fetch {
+                f.left_future()
```
abailly, line 132: 😱 What's left and right here?

crates/amaru-consensus/src/consensus/effects/network_effects/connection.rs
```diff
@@ -0,0 +1,200 @@
...
+        blockfetch,
+        chainsync::{HeaderContent, NextResponse},
+    },
+};
+use pure_stage::BoxFuture;
+use std::{collections::VecDeque, future::pending, time::Duration};
+use tokio::{select, sync::mpsc, task::JoinHandle, time::sleep};
+use tracing::Span;
+
+pub async fn actor(
```
abailly, line 52: I think we will need some explanation/documentation here to

crates/amaru-consensus/src/consensus/stages/select_chain.rs
rkuhn: While this interaction will always be non-obvious, you’re right that my approach duplicates existing facilities; somehow I hadn’t considered this. Will update.

crates/amaru/src/bin/amaru/cmd/mod.rs
rkuhn: No, it is used in other cmdline tools as well, I wanted to keep this small.

crates/amaru-consensus/src/consensus/effects/network_effects.rs
```diff
@@ -85,51 +90,17 @@ impl<T: SendData + Sync> NetworkOps for Network<'_, T> {
             ForwardEvent::Backward(header_tip),
         ))
     }
+
+    fn disconnect(&self, peer: Peer) -> BoxFuture<'_, Result<(), ProcessingFailed>> {
+        let f = self.0.external(DisconnectEffect::new(peer));
+        Box::pin(async move { Ok(f.await) })
+    }
```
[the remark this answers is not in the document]
rkuhn: Passing the function return value to the `Ok(…)` constructor ensures that the types align. The proposed change would keep working if `f` returned a different type, which is not what I want.

crates/amaru-consensus/src/consensus/stages/pull.rs
rkuhn: Yes, indeed.

crates/amaru-consensus/src/consensus/effects/network_effects.rs
rkuhn: Yes, as I was told to do :-) (the number 3 should probably be configurable — or it should be included with the protocol parameters?)

crates/amaru-consensus/src/consensus/effects/network_effects/connection.rs
rkuhn: You mean politically?

Jokes apart: I want a future that can be either this type or that, and the old `futures` crate provides exactly such a facility. The alternative would be boxing, but that isn’t necessary here.

crates/amaru-consensus/src/consensus/effects/network_effects/connection.rs
rkuhn: Well, famous last words, but: this is supposed to be the stop-gap measure for the next demo. The proper implementation will use pure-stage throughout and will be properly documented.

crates/amaru-consensus/src/consensus/effects/network_effects.rs
rkuhn: The difference lies in what is moved into the `.map_err` closure: just a `point` or `self`? `network.fetch_block` shouldn’t take a `&Peer` because it unconditionally needs a `Peer` anyway, and once I fix that the difference will become a compiler error in your proposal.

crates/amaru-consensus/src/consensus/effects/network_effects.rs
abailly: Yeah, I forgot :) Don't know what the right number is, we should check what the haskell node is doing perhaps? I am not even 100% sure it does that in practice.

crates/amaru-consensus/src/consensus/effects/network_effects/connection.rs
abailly: Would you mind still writing a few sentences explaining the whole thing, even pointing out this is s stop gap solution? As you know very well, temporary solutions have a natural tendency to stick around :)

crates/amaru-consensus/src/consensus/effects/network_effects.rs
rkuhn: Ok, will remove for now.

PR #518, opened 2025-10-22
fix: always start with a `Rollback` message when downstream client connect successfully

abailly:
According to the [Network specification](https://ouroboros-network.cardano.intersectmbo.org/pdfs/network-spec/network-spec.pdf) section 3.7.2:

> Whenever the server replies with MsgIntersectFound, the client can expect the next update (i.e. a reply to MsgRequestNext) to be MsgRollBackward, either to the specified `point_intersect` or an earlier point if the producer switched to a different fork in the meantime.

This PR also prepares work to serve headers from the store by encapsulating the whole logic in a `ChainFollower` structure (used to be a `ChainState`).

## Summary by CodeRabbit

* **Improvements**
  * Simplified and centralized chain selection and roll-forward logic with clearer handling for initial inserts and forks.
  * Updated consensus client flow to use a follower-based approach for intersection and message sequencing.

* **Refactor**
  * Reworked client-state internals and renamed test helper types to better reflect forwarding semantics.

* **Tests**
  * Added iterator tests for DB iteration and updated forward-chain tests to match new messaging behavior.

### DIFF
--- a/crates/amaru/src/stages/consensus/forward_chain/client_protocol.rs
+++ b/crates/amaru/src/stages/consensus/forward_chain/client_protocol.rs
@@ -186,16 +186,16 @@ async fn chain_sync<H: IsHeader + 'static + Clone + Send>(
     };
 
     tracing::debug!(
-        "finding headers between {:?} and {:?}",
-        our_tip.0,
-        requested_points
+        points = ?requested_points,
+        tip = our_tip.1,
+        "request interception",
     );
     let Some(mut chain_follower) = ChainFollower::new(store, &our_tip.0, &requested_points) else {
         tracing::debug!("no intersection found");
         server.send_intersect_not_found(our_tip).await?;
         return Err(ClientError::NoIntersection.into());
     };
-    tracing::debug!("intersection found: {:?}", chain_follower.tip);
+    tracing::debug!(intersection = ?chain_follower.tip, "intersection found");
     server
         .send_intersect_found(chain_follower.tip.0.clone(), our_tip.clone())
         .await?;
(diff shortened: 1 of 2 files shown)


crates/amaru-stores/src/rocksdb/consensus/mod.rs
```diff
@@ -370,11 +370,16 @@ pub mod test {
...
     };
+    use rocksdb::Direction;
     use std::collections::BTreeMap;
     use std::path::Path;
     use std::sync::Arc;
     use std::{fs, io};
 
+    /// "chain"
+    /// TODO: move to util once we are ready to implement chain tracking
+    const CHAIN_PREFIX: [u8; CONSENSUS_PREFIX_LEN] = *b"chain";
```
rkuhn, line 381: oh nice, I didn’t that that this works! OTOH you only use `&CHAIN_PREFIX[..]`, so typing as `&[u8]` would be more direct (you can add an assertion about the length in the initializer, if I’m not mistaken)

crates/amaru-stores/src/rocksdb/consensus/mod.rs
```diff
@@ -724,6 +729,57 @@ pub mod test {
...
+
+        // iterator continues from where it left off, skipping deleted keys
+        assert_eq!(
+            *(iter.next().unwrap().unwrap().0),
+            [&CHAIN_PREFIX[..], &slot8].concat()
+        );
+        assert_eq!(
+            *(iter.next().unwrap().unwrap().0),
+            [&CHAIN_PREFIX[..], &slot9].concat()
+        );
```
rkuhn, line 780: perhaps add that the iterator is now done?

crates/amaru/src/stages/consensus/forward_chain/client_protocol.rs
```diff
@@ -180,60 +180,62 @@ async fn chain_sync<H: IsHeader + 'static + Clone + Send>(
...
         // need an intersection point to start
         return Err(ClientError::EarlyRequestNext.into());
     };
 
-    tracing::debug!("finding headers between {:?} and {:?}", our_tip.0, req);
-    let Some((catch_up, client_at)) = find_headers_between(store, &our_tip.0, &req) else {
+    tracing::debug!(
+        points = ?requested_points,
+        tip = our_tip.1,
+        "request interception",
```
rkuhn, line 191: either inception or intersection :-)

rkuhn (approved):
Overall looking good to me, with 2–3 nitpicks that might be worth fixing.

crates/amaru-stores/src/rocksdb/consensus/mod.rs
abailly: Credit goes to @KtorZ

crates/amaru/src/stages/consensus/forward_chain/client_protocol.rs
abailly: Good catch!

PR #520, opened 2025-10-23
chore: simplify Read-only interface for ChainStore

abailly:
The previous interface contained 'diagnostic' functions which are only used for the `dump-chain-store` command, and contain `panic` or `expect` calls. 

This PR moves those functions to a (monomorphic) `DiagnosticChainStore` interface which has the consequence of simplifying the various stores implementation (RocksDB, InMemory, Effects).

### DIFF
--- a/crates/amaru-stores/src/rocksdb/consensus/mod.rs
+++ b/crates/amaru-stores/src/rocksdb/consensus/mod.rs
@@ -21,7 +21,9 @@ use amaru_kernel::{
     HEADER_HASH_SIZE, Hash, HeaderHash, ORIGIN_HASH, RawBlock, cbor, from_cbor, to_cbor,
 };
 use amaru_ouroboros_traits::is_header::IsHeader;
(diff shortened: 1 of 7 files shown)


crates/amaru-stores/src/rocksdb/consensus/mod.rs
```diff
@@ -492,7 +502,7 @@ pub mod test {
 
     #[test]
     fn load_all_headers() {
-        with_db(|db| {
+        with_db_path(|(db, path)| {
```
etorreborre, line 505: Since this is done in a few places, we could add a `with_test_db` function that uses a `|db, test_db|` closure that has already done the initialization (with 2 parameters instead of a pair while we're at it).

abailly: Not sure how to do the latter part

abailly: Actually, that does not work because the read only DB takes a snapshot at the moment of its creation which means all the changes we do in the callback are not visible.

PR #524, opened 2025-10-25
feat: decode initial snapshot lazily, to limit memory usage.

KtorZ:
This commit introduce as reusable 'LazyDecoder', built on top of minicbor and the io::Read interface.

It allows decoding large data incrementally by reading bytes only when needed. It keeps an internal state with bytes read but not consumed, as well as handle to the reader.

Currently, the implementation reads by chunks of 2MiB, and only requests more bytes when needed (i.e. when hitting a 'end_of_input' error while decoding).

In *most cases*, this is already sufficient to decode the NewEpochState incrementally while storing (and de-allocating) objects as we go. Even accounts "only" demand ~150MB of allocation.

UTxOs, however, must be dealt with differently because we can afford building a BTreeMap buffer before flushing them. So we decode-and-flush in chunks, which allows to decode the entire UTxO set without really allocating much more than the chunk size.

Overall, **this approach allows decoding the initial mainnet snapshots with <750MB of resident memory** ... instead of 22GB previously.

@jeluard and @geofflittle will be pleased, I suppose.

## Summary by CodeRabbit

* **New Features**
  * Streaming snapshot import using a lazy decoder for large ledger files.
  * Network-aware import option to apply environment-specific rules.

* **Improvements**
  * Safer, incremental import with finer-grained progress reporting and guarded state checks.
  * Optional tracing feature to enable richer runtime diagnostics.

* **Tests**
  * Improved test failure messages to include error details for easier debugging.

### DIFF
--- a/crates/amaru-ledger/src/bootstrap.rs
+++ b/crates/amaru-ledger/src/bootstrap.rs
@@ -30,7 +30,6 @@ use amaru_kernel::{
     protocol_parameters::{PREPROD_INITIAL_PROTOCOL_PARAMETERS, ProtocolParameters},
 };
 use amaru_progress_bar::ProgressBar;
-use anyhow::anyhow;
 use std::{
     collections::{BTreeMap, BTreeSet},
     fs, iter,
@@ -38,7 +37,7 @@ use std::{
     rc::Rc,
     sync::LazyLock,
 };
-use tracing::info;
+use tracing::{info, warn};
 
 const BATCH_SIZE: usize = 1000;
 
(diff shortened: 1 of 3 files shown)


crates/amaru-minicbor-extra/src/decode/lazy.rs
```diff
@@ -0,0 +1,99 @@
...
+    /// fails due to a lack of bytes. And error otherwise.
+    pub fn with_decoder<T>(
+        &mut self,
+        decode: impl Fn(&mut cbor::decode::Decoder<'_>) -> Result<T, Box<dyn std::error::Error>>,
+    ) -> Result<T, Box<dyn std::error::Error>> {
+        let mut should_read_more = self.bytes.is_empty();
+        let mut can_read_more = true;
+        loop {
+            if should_read_more {
+                let mut buf = [0; Self::CHUNK_SIZE];
```
yHSJ, line 62: what happens if there is less than 2 MiB left for a chunk? Is there an issue when there is unused zero bytes at the end of the buffer?

KtorZ: It reads `up to 2MiB`; and returns the actual size read. In case where there's no bytes left, it returns `0`. 

> Is there an issue when there is unused zero bytes at the end of the buffer

And, no. I did not alter the behavior or minicbor in this regard. If there's leftovers that the decoder do not consume, then it just returns the decoded value and ignore the surplus.

PR #526, opened 2025-10-26
add short-circuit facilities for pure-stage

rkuhn:
The new trait `TryInStage` offers methods `.or_terminate` and `.or_return` (the latter is only
useful on nightly) to allow concise short-circuiting in pure-stage stages.

## Summary by CodeRabbit

* **New Features**
  * Added convenient constructor methods to the Effect API for creating effect variants (receive, send, clock, wait, respond, external, terminate).
  * Added assertion helpers to Effect for runtime validation and extraction of effect variants.
  * Introduced TryInStage extension trait enabling cleaner error handling and stage termination patterns.
  * Added constructor methods to TraceEntry for streamlined trace event creation.

* **Chores**
  * Updated workspace dependencies.


crates/pure-stage/Cargo.toml
```diff
@@ -8,7 +8,10 @@ authors.workspace = true
 repository.workspace = true
 homepage.workspace = true
 documentation = "https://docs.rs/pure-stage"
-rust-version.workspace = true
+# rust-version.workspace = true
```
etorreborre: Remove this line?

crates/pure-stage/src/types.rs
```diff
@@ -222,14 +225,14 @@ impl<T> DerefMut for MpscSender<T> {
     }
 }
 
-#[expect(dead_code)]
+#[allow(dead_code)]
```
etorreborre: Why change `expect` to `allow`? The purpose of `expect` is to get notified when an annotation does not make sense anymore and clean-up irrelevant annotations.

crates/pure-stage/Cargo.toml
rkuhn: Yes, that’s a left-over of trying to compile with Rust 1.86 (to see whether users outside Amaru on stable Rust can use it).

crates/pure-stage/src/types.rs
rkuhn: `expect` makes sense within Amaru (being an application), but my worry is that `expect` will break dependent code outside Amaru once Rust stops warning … probably paranoid in this particular instance, but less implausible for the warning regarding using `async fn` in traits.

rkuhn:
@etorreborre My plan was to demonstrate the usage in response to your comment, but unfortunately I found a wrinkle that seems fundamental (Rust's ownership types: it is only possible to conditionally move a value with the branching language constructs, this cannot be used for functional abstraction). I'll keep thinking today about whether `.or_return` can be salvaged. Will be back in amaru-land tomorrow.

PR #527, opened 2025-10-26
fix: add debug traces to demo output

abailly:
## Summary by CodeRabbit

* **Chores**
  * Demo script now echoes each daemon trace line immediately for real-time visibility before parsing, improving diagnostic output during development and testing.
  * Tracing verbosity and event handling remain unchanged.
  * CI workflow cache keys updated to a new version to refresh build cache behavior.


abailly:
Shall we keep the log lines @KtorZ ?

PR #529, opened 2025-10-28
Amaru performances

KtorZ:
- :round_pushpin: **perf: box a few inner large types in struct.**
    Rust allocate struct fields based on the size of their largest field. So struct with many small fields and a large one pay a significant cost. This is cheap to do, and an easy win.

- :round_pushpin: **perf: intern stake distribution map keys**
    There's an enormous repetition of keys between the stake distribution maps; by interning the keys, we avoid copying and re-allocating most of those elements.

  This interning is currently done via an adhoc structure held by the ledger (and is thus, local to the ledger). However, I've considered using something more holistic such as: https://docs.rs/internment/latest/internment/struct.ArcIntern.html

  The issue with the latter is that it uses HashMap under the hood, and thus requires keys to be Hashshable. StakeCredential aren't at the moment. But easily done should we want to push the change in Pallas.

  One gotcha though: because the interning is local, we need to explicit manage the "lifetime" of the interned object. When entities are no longer referenced, they must be cleaned up from the interned struct. This is done once per epoch, and is relatively quick.

- :round_pushpin: **perf: conditionally enable different allocators: jemalloc / mimalloc**
    Jemalloc seems to perform generally better for us... but it is not longer maintained as of recently. No one stood up to pick it up *yet*; so mimalloc (from Microsoft) might be a better choice in the long-run.

## Summary by CodeRabbit

* **New Features**
  * Optional memory allocator features (mimalloc, jemalloc) to tune runtime allocation.
  * In-memory interner to deduplicate recurring keys and reduce memory use.

* **Chores**
  * Memory- and data-structure optimizations across summaries, rewards, and stores.
  * CI/build updates and added Zlib to the license allowlist.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -43,7 +43,7 @@ jobs:
           - runner: windows-latest
             target: x86_64-pc-windows-msvc
             title: x86_64/windows
-            command: test-amaru --profile dev-debug
+            command: test --workspace
 
           - runner: ubuntu-latest
             target: wasm32-unknown-unknown
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,6 +49,7 @@ pallas-crypto = "0.33.0"
 pallas-math = "0.33.0"
 pallas-network = "0.33.0"
 pallas-primitives = "0.33.0"
+mimalloc = "0.1.48"
 pallas-traverse = "0.33.0"
 parking_lot = "0.12.3"
 rand = "0.9.2"
@@ -75,6 +76,7 @@ tracing-subscriber = { version = "0.3.20", features = [
   "std",
   "json",
 ] }
+tikv-jemallocator = "0.6.1"
 typetag = "0.2.21"
 
 # Internal dependencies
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,6 +30,7 @@ either = "1.15"
 futures-util = "0.3.31"
 gasket = { version = "0.8.0", features = ["derive"] } # More recent versions have regression (e.g. https://github.com/construkts/gasket-rs/pull/34)
 hex = "0.4.3"
+hashbrown = "0.16.0"
 indicatif = "0.18.0"
 indoc = "2.0"
 itertools = "0.14.0"
--- a/crates/amaru-ledger/src/summary/arc_interner.rs
+++ b/crates/amaru-ledger/src/summary/arc_interner.rs
@@ -1,0 +1,83 @@
+// Copyright 2024 PRAGMA
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use amaru_kernel::{DRep, PoolId, StakeCredential};
+use std::{collections::BTreeSet, sync::Arc};
+
+/// A structure for [interning](https://en.wikipedia.org/wiki/String_interning) recurring keys held
+/// in memory. This allows to drastically reduce the number of allocated elements for keys that are
+/// often found over many places.
+///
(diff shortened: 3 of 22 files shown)


KtorZ:
@rkuhn pointed out to: https://docs.rs/intern-arc/latest/intern_arc/ 

which we might consider instead of the local interner?

.github/workflows/continuous-integration.yml
```diff
@@ -43,7 +43,7 @@ jobs:
           - runner: windows-latest
             target: x86_64-pc-windows-msvc
             title: x86_64/windows
-            command: test-amaru --profile dev-debug
```
jeluard, line 46: Shall we get rid of `test-amaru` then?

Cargo.toml
```diff
@@ -48,6 +49,7 @@ pallas-crypto = "0.33.0"
 pallas-math = "0.33.0"
 pallas-network = "0.33.0"
 pallas-primitives = "0.33.0"
+mimalloc = "0.1.48"
```
jeluard, line 52: Those should be gated behind a feature?

jeluard (commented):
Looks like worth it. Given the intrusive (as in, not transparent) nature of this approach we probably need some guideline to figure out where/when to follow this pattern.

Cargo.toml
KtorZ: Not in the workspace no (not possible). It's gated at the crate level.

crates/amaru-ledger/src/summary/arc_interner.rs
```diff
@@ -0,0 +1,83 @@
...
+    /// This must be called periodically (e.g. once every epoch) to free-up previously allocated
+    /// keys that are no longer used (because pool have fully unregistered, for example);
+    /// otherwise, we're just leaking memory forever...
+    ///
+    /// Fortunately, the Arc makes this easy because they keep a count of references to them. So it
+    /// suffices to look for items that have don't have references anymore and drop them.
+    pub fn free_orphans(&mut self) {
+        self.accounts.retain(|arc| Arc::strong_count(arc) > 0);
+        self.pools.retain(|arc| Arc::strong_count(arc) > 0);
+        self.dreps.retain(|arc| Arc::strong_count(arc) > 0);
```
Eh2406, line 43: I think there's an off by one error here. Arc drops its inner contents automatically when the strong count gets to zero. We know that hasn't happened in this case because the `HashSet` still has a copy. I think we want to retain if `>1`, throw out the things where we have the only copy.

crates/amaru/Cargo.toml
```diff
@@ -13,6 +13,11 @@ documentation.workspace = true
 rust-version.workspace = true
 build = "build.rs"
 
+[features]
+default = []
+mimalloc = ["dep:mimalloc"]
+jemalloc = ["dep:tikv-jemallocator"]
```
Eh2406, line 19: Should this feature also enable the `jemalloc` feature of `rocksdb`? by doing so, the statistics and profiling supported by `jemalloc` also include data from the `rocksdb`.

crates/amaru/build.rs
```diff
@@ -23,4 +23,13 @@ fn main() {
...
     println!("cargo:rustc-env=AMARU_NETWORK={}", network);
+
+    let jemalloc = std::env::var("CARGO_FEATURE_JEMALLOC").is_ok();
+    let mimalloc = std::env::var("CARGO_FEATURE_MIMALLOC").is_ok();
+
+    if jemalloc && mimalloc {
+        println!(
+            "cargo:warning=Features `jemalloc` and `mimalloc` are mutually exclusive yet both enabled; the default allocator will be used instead."
+        );
+    }
```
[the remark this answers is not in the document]
Eh2406: I believe the bot is just wrong here.

crates/amaru-ledger/src/summary/arc_interner.rs
KtorZ: You're right; I noticed this a while back and fixed it but somehow never pushed the change to the branch 🤔

.github/workflows/continuous-integration.yml
KtorZ: Possibly.. only the windows command seems to use it; and it's turning `--all-features` which isn't good... Especially on Windows 😅

crates/amaru/Cargo.toml
KtorZ: Good point, didn't know about that feature from rocksdb.

KtorZ:
Will revisit this branch/PR using a global interner.

PR #537, opened 2025-11-03
feat: serve chain from store to followers

### LINKED ISSUE #426: Downstream clients connecting entail huge memory spike and stalls process
When a downstream client connects to Amaru to sync up and the intersection points it provides are "old" or even `Genesis`, Amaru memory consumption surges (from <300MB to ~1.5GB on `preview`) and there's a significant amount of time spent finding relevant headers. 

Image

The following trace shows that it takes about 50s to find the headers to send to client. 

```
Sep 04 14:34:50 clermont amaru[1329417]: 2025-09-04T12:34:50.137731Z DEBUG amaru::stages::consensus::forward_chain::client_protocol: finding headers between (90333249, 4fb9cff59ef56a3661d45ab7e445593d743241a95f5f9ecbd6f7751ce67472cb) and [(61789671, 846addc03cf9b816cf956daf0e8ae029aa3a466ac3e4ba52b8d526eac7118aab)]
Sep 04 14:35:36 clermont amaru[1329417]: 2025-09-04T12:35:36.657826Z DEBUG amaru::stages::consensus::forward_chain::client_protocol: intersection found: Tip((61789671, 846addc03cf9b816cf956daf0e8ae029aa3a466ac3e4ba52b8d526eac7118aab), 2559288)
```

Moreover, while the client is synchronising the upstream amaru node is stalled and does not synchronise with its own upstream peer.

While the migration to pallas-network2 (see #420) should make our network stack more robust, it will take some time so we probably want a stop-gap solution before that.

abailly:
This PR introduces a `ChainFollower` structure that keeps track of a single downstream peer's synchronization state with two modalities:
- while the synchronisation point (initially from `intersection` and then evolving) is behind the `anchor` of the chain, the forwarded headers are loaded from the store,
- when the synchronisation reaches the anchor, the peer is sent the `ClientOp` messages accumulated in a queue while it was synchronising. This queue is initially filled with the "volatile" part of the chain, e.g the part between the `anchor` and the `tip`.

The code is surprisingly convoluted and there are a few edge cases one needs to deal with that should be covered by the store's invariant, e.g the store should guarantee that if there's a child relationship between 2 headers, then both of them are in the store. However it's really not obvious to me how to enforce that and expose it in the API.

Note that ideally I would have each `ChainFollower` keep an `Iterator` over the underlying `ChainStore` but as `Arc<T>` is not `Send` it's seemingly not possible. I would be interested in knowing from Rust experts whether or not this is feasible as I think it would somewhat simplify the design and possibly improve the performances.

Fix #426 

## Summary by CodeRabbit

* **New Features**
  * Enhanced chain store with improved next-chain lookup capability for efficient chain traversal.

* **Improvements**
  * Added contextual location-aware logging to consensus validation for better error diagnostics and debugging.

* **Infrastructure**
  * Added multi-client launch script to support testing multiple node deployments simultaneously.

### DIFF
--- a/crates/amaru/src/stages/consensus/forward_chain/chain_follower.rs
+++ b/crates/amaru/src/stages/consensus/forward_chain/chain_follower.rs
@@ -145,6 +145,7 @@ impl<H: IsHeader + Clone + Send> ChainFollower<H> {
         })
     }
 
+    #[allow(clippy::panic)]
     pub fn next_op(&mut self, store: Arc<dyn ReadOnlyChainStore<H>>) -> Option<ClientOp<H>> {
         // is this initial rollback?
         if let Some(ref init_tip) = self.initial {
@@ -167,12 +168,10 @@ impl<H: IsHeader + Clone + Send> ChainFollower<H> {
                             trace!(forwarded = %child.point(), anchor = ?self.anchor, "forwarding from store at origin");
                             return Some(ClientOp::Forward(child));
                         }
-                        None => {
-                            // FIXME: this seems possible in some circumstances given how our DB is structured
-                            // but this should never happen in practice. Perhaps turn into a proper `panic!`?
-                            warn!(intersection = ?self.intersection, child = %point, anchor = ?self.anchor, "child not found in store");
-                            return None;
-                        }
+                        None => panic!(
+                            "Store invariant violated:\nnext_best_chain returned {} but header not in store (intersection: {:?}, anchor: {:?})",
+                            point, self.intersection, self.anchor
+                        ),
                     }
                 }
                 None => {
@@ -182,7 +181,12 @@ impl<H: IsHeader + Clone + Send> ChainFollower<H> {
             }
         }
 
-        self.ops.pop_front()
+        if let Some(op) = self.ops.pop_front() {
+            self.intersection = op.tip();
+            Some(op)
+        } else {
+            None
+        }
     }
 
     pub fn add_op(&mut self, op: ClientOp<H>) {
@@ -248,11 +252,11 @@ pub(crate) mod tests {
     fn find_headers_starting_at_tip() {
         let store = mk_in_memory_store(CHAIN_47);
 
-        let tip = store.get_point(TIP_47);
+        let tip_point = store.get_point(TIP_47);
         let points = [store.get_point(TIP_47)];
-        let start = Tip(tip.clone(), store.get_height(TIP_47));
+        let start = Tip(tip_point.clone(), store.get_height(TIP_47));
 
-        let mut chain_follower = ChainFollower::new(store.clone(), &tip, &points).unwrap();
+        let mut chain_follower = ChainFollower::new(store.clone(), &tip_point, &points).unwrap();
 
         assert_eq!(
             chain_follower.next_op(store.clone()),
@@ -343,6 +347,7 @@ pub(crate) mod tests {
     }
 
     #[test]
+    #[should_panic(expected = "Store invariant violated")]
     fn next_op_returns_none_given_it_fails_to_load_child_header() {
         let store = mk_in_memory_store(CHAIN_47);
         let unstored_header = run(any_header_with_parent(Hash::from(
@@ -364,7 +369,8 @@ pub(crate) mod tests {
         let mut chain_follower = ChainFollower::new(store.clone(), &tip, &points).unwrap();
 
         let _ = chain_follower.next_op(store.clone()); // initial rollback
-        assert_eq!(chain_follower.next_op(store.clone()), None);
+
+        chain_follower.next_op(store.clone());
     }
 
     // HELPERS


crates/amaru-ouroboros-traits/src/stores/consensus/in_memory_consensus_store.rs
```diff
@@ -125,6 +125,25 @@ impl<H: IsHeader + Clone + Send + Sync + 'static> ReadOnlyChainStore<H> for InMe
...
+        let min_slot = point.slot_or_default();
+
+        let next: Vec<&Point> = inner
+            .chain
+            .iter()
+            .filter(move |p| p.slot_or_default() > min_slot)
+            .take(1)
+            .collect();
+
+        if next.is_empty() {
```
etorreborre, line 141: `next.first().cloned()`?

crates/amaru/src/stages/consensus/forward_chain/chain_follower.rs
```diff
@@ -0,0 +1,425 @@
...
+            .unwrap_or(Tip(Point::Origin, 0));
+        Some(Self {
+            initial: Some(best_tip.clone()),
+            ops: headers.into(),
+            intersection: best_tip,
+            anchor,
+        })
+    }
+
+    pub fn next_op(&mut self, store: Arc<dyn ReadOnlyChainStore<H>>) -> Option<ClientOp<H>> {
```
etorreborre, line 149: Why pass a `store` reference again here since we already have `self.store`?

crates/amaru/src/stages/consensus/forward_chain/chain_follower.rs
```diff
@@ -0,0 +1,425 @@
...
+                    let child_header =
+                        store.load_header(&hash_point(&to_network_point(point.clone())));
+                    match child_header {
+                        Some(child) => {
+                            self.intersection = child.as_tip();
+                            trace!(forwarded = %child.point(), anchor = ?self.anchor, "forwarding from store at origin");
+                            return Some(ClientOp::Forward(child));
+                        }
+                        None => {
+                            // FIXME: this seems possible in some circumstances given how our DB is structured
```
etorreborre: I would panic here. There are similar cases in the `HeadersTree` with the `unsafe_get_header` method.

crates/amaru/src/stages/consensus/forward_chain/chain_follower.rs
```diff
@@ -0,0 +1,425 @@
...
+        )
+    }
+
+    #[test]
+    fn find_headers_starting_at_tip() {
+        let store = mk_in_memory_store(CHAIN_47);
+
+        let tip = store.get_point(TIP_47);
+        let points = [store.get_point(TIP_47)];
+        let start = Tip(tip.clone(), store.get_height(TIP_47));
```
etorreborre: Isn't it possible to use `.as_tip()` here?

crates/amaru/src/stages/consensus/forward_chain/chain_follower.rs
```diff
@@ -0,0 +1,425 @@
...
+        let points = [
+            store.get_point(FORK_47),   // this will lose to the (taller) winner
+            store.get_point(WINNER_47), // this is the winner after the branch
+        ];
+        let expected = store.get_point(WINNER_47);
+
+        let mut chain_follower = ChainFollower::new(store.clone(), &tip, &points).unwrap();
+        assert_eq!(
+            chain_follower.next_op(store.clone()),
+            Some(ClientOp::Backward(Tip(expected, 8)))
```
etorreborre, line 301: Isn't it possible to use `.as_tip()` here?

crates/amaru/src/stages/consensus/forward_chain/chain_follower.rs
```diff
@@ -0,0 +1,425 @@
...
+        let points = [store.get_point(LOST_47)];
+        let first = store
+            .load_header(&Hash::from(hex::decode(FIRST_HEADER).unwrap().as_slice()))
+            .expect("could not load header");
+
+        let mut chain_follower = ChainFollower::new(store.clone(), &tip, &points).unwrap();
+
+        assert_eq!(
+            chain_follower.next_op(store.clone()),
+            Some(ClientOp::Backward(Tip(Point::Origin, 0)))
```
etorreborre, line 319: Isn't it possible to use `.as_tip()` here?

etorreborre (approved):
⛓️ I would be good to be able to test that logic in the simulation and see, for example, that the nodes are converging on the same chain, even they start with different states.

crates/amaru/src/stages/consensus/forward_chain/chain_follower.rs
abailly: I don't think we have it in `self`, or at least I removed it! I need to pass it because `Arc<dyn ChainStore>` is not `Send` and therefore I cannot store it in `self`, that's the solution I found: pass it to `next_op`

abailly:
I ran a small experiment with a remote node running `preview` and starting up 5 clients at a 1 minute interval on a local machine and monitoring RAM consumption of remote process. It seems that this PR mostly solves the memory consumption problem.

Screenshot 2025-11-04 at 11 02 02

crates/amaru/src/stages/consensus/forward_chain/chain_follower.rs
etorreborre: I tried it and that works fine. See my commit in `abailly/follow-chain-from-store-eric`.

crates/amaru/src/stages/mod.rs
```diff
@@ -299,6 +282,16 @@ fn open_chain_store(config: &Config) -> anyhow::Result<Arc<dyn ChainStore<BlockH
...
+    if *tip != ORIGIN_HASH && chain_store.load_header(tip).is_none() {
+        panic!(
+            "Tip {} not found in chain database '{}'",
+            tip, config.chain_store
+        )
+    };
+
+    chain_store.set_anchor_hash(tip)?;
+    chain_store.set_best_chain_hash(tip)?;
     Ok(chain_store)
```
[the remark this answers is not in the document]
abailly: This is not possible as Amaru explicit does not support starting from `ORIGIN`

crates/amaru/src/stages/consensus/forward_chain/chain_follower.rs
abailly: Not without some significant restructuring: We need the `Header` to use `as_tip()` but in the tests we have only a hex-string representing the hash. It does not seem to simplify the code.

crates/amaru/src/stages/consensus/forward_chain/chain_follower.rs
abailly: Ditto

crates/amaru/src/stages/consensus/forward_chain/chain_follower.rs
abailly: `Origin` is a `Point` and we don't have `as_tip` for a `Point` as it requires the height of  the header which is not part of the point.

PR #549, opened 2025-11-11
feat: improve naming for consensus logs and spans

abailly:
This PR is a first attempt at rationalising the spans and logs we emit from Amaru, mostly focusing on the consensus part.
* Removed text interpolation (exception from commands) in favour of structured logs
* Use "semantic convention" for the message part, eg. dotted identifiers
* Remove `stages` prefix from spans in favour of names explicitly mentioning the purpose of the stage
* Update documentation

The trace now looks like the following:

jaeger-trace

## Release Notes

* **Documentation**
  * Updated consensus pipeline documentation with clearer descriptions of chain diffusion, synchronization, and selection processes, including new tracing spans reference guide for debugging.

* **Refactor**
  * Standardized internal tracing instrumentation naming and logging messages across consensus and network modules for improved observability and consistency.
  * Reorganized test utilities for headers tree validation.

### DIFF
--- a/crates/amaru-consensus/src/consensus/headers_tree/headers_tree.rs
+++ b/crates/amaru-consensus/src/consensus/headers_tree/headers_tree.rs
@@ -160,7 +160,6 @@ impl<H: IsHeader + Debug + Clone + Display + PartialEq + Eq + 'static> Display f
 
 /// Debugging and formatting methods.
 impl<H: IsHeader + Debug + Clone + PartialEq + Eq + 'static> HeadersTree<H> {
-
     /// Human-readable formatting of the headers tree.
     fn format(
         &self,
@@ -195,7 +194,6 @@ impl<H: IsHeader + Debug + Clone + PartialEq + Eq + 'static> HeadersTree<H> {
 }
 
 impl<H: IsHeader + Clone + Debug + PartialEq + Eq + Send + Sync + 'static> HeadersTree<H> {
-
     /// Create a new HeadersTree
     pub(crate) fn create(
         chain_store: Arc<dyn ChainStore<H>>,


crates/amaru-consensus/src/consensus/stages/track_peers.rs
```diff
@@ -30,15 +30,15 @@ pub fn stage(
...
 ) -> impl Future<Output = State> {
-    let span = tracing::trace_span!(parent: msg.span(), "stage.track_peers");
+    let span = tracing::trace_span!(parent: msg.span(), "chain_sync.track_peers");
     async move {
         let (mut tracker, downstream) = state;
         match msg {
             Tracked::Wrapped(e @ DecodedChainSyncEvent::RollForward { .. }) => {
                 if tracker.is_caught_up() {
-                    debug!(target: EVENT_TARGET, peer = %e.peer(), point = %e.point(), "new tip");
+                    debug!(target: EVENT_TARGET, peer = %e.peer(), point = %e.point(), "track_peers.caught_up.new_tip");
```
jeluard, line 39: Would be great to normalize the parameters we pass also. e.g. `tip_changed` pass the point as `tip`

abailly: Sorry, I don't understand, what's `tip_changed`? 
But I agree: sometimes we use points, sometimes hash and slot, sometimes hash only...

jeluard: I mean [this trace](https://github.com/pragma-org/amaru/blob/main/crates/amaru-consensus/src/consensus/stages/forward_chain.rs#L57) that is logically related.

PR #553, opened 2025-11-12
Phase 2 Validation

yHSJ:
This PR introduces a subset of Phase 2 validations, targeting Plutus V3 script execution.

In order to achieve that, we also need to introduce an `ArenaPool`, which contains a pool of arenas (go figure), used for script evaluation.

## Summary by CodeRabbit

* **New Features**
  * Concurrent arena pool for VM allocations used during transaction validation
  * Transaction validation split into Phase One and Phase Two with distinct error types
  * Ordered redeemer type to enable deterministic script ordering and wiring

* **Improvements**
  * Expanded Plutus evaluation across V1–V3 with richer error reporting and cost-model support
  * Exposed configuration to tune VM arena count/size; ledger/validator wiring updated to use the arena pool
  * Tests updated to use the new validation phases and network naming

* **Chores**
  * CI enhanced for RISC‑V cross-builds; workspace dependencies and license allowlist updated

### DIFF
--- a/crates/amaru-ledger/src/rules/transaction/phase_two/mod.rs
+++ b/crates/amaru-ledger/src/rules/transaction/phase_two/mod.rs
@@ -205,28 +205,45 @@ where
 
             let result = program.eval_with_params(&arena, plutus_version, cost_model, uplc_budget);
 
-            match result.term {
-                Ok(term) => match term {
-                    Term::Error => Err(PhaseTwoError::UplcMachineError(UplcMachineError {
+            match plutus_version {
+                PlutusVersion::V1 | PlutusVersion::V2 => match result.term {
+                    Ok(term) => match term {
+                        Term::Error => Err(PhaseTwoError::UplcMachineError(UplcMachineError {
+                            plutus_version,
+                            info: result.info,
+                            err: "Error term evaluated".into(),
+                        })),
+                        Term::Var(_)
+                        | Term::Lambda { .. }
+                        | Term::Apply { .. }
+                        | Term::Delay(_)
+                        | Term::Force(_)
+                        | Term::Case { .. }
+                        | Term::Constr { .. }
+                        | Term::Constant(_)
+                        | Term::Builtin(_) => Ok(()),
+                    },
+                    Err(e) => Err(PhaseTwoError::UplcMachineError(UplcMachineError {
                         plutus_version,
                         info: result.info,
-                        err: "Error term evaluated".into(),
+                        err: e.to_string(),
                     })),
-                    Term::Var(_)
-                    | Term::Lambda { .. }
-                    | Term::Apply { .. }
-                    | Term::Delay(_)
-                    | Term::Force(_)
-                    | Term::Case { .. }
-                    | Term::Constr { .. }
-                    | Term::Constant(_)
-                    | Term::Builtin(_) => Ok(()),
                 },
-                Err(e) => Err(PhaseTwoError::UplcMachineError(UplcMachineError {
-                    plutus_version,
-                    info: result.info,
-                    err: e.to_string(),
-                })),
+
+                // According to [CIP-117](https://cips.cardano.org/cip/CIP-0117), Plutus V3 scripts must evaluate to a constant term of unit
+                PlutusVersion::V3 => match result.term {
+                    Ok(Term::Constant(Constant::Unit)) => Ok(()),
+                    Ok(_) => Err(PhaseTwoError::UplcMachineError(UplcMachineError {
+                        plutus_version,
+                        info: result.info,
+                        err: "evaluated to a non-unit term".to_string(),
+                    })),
+                    Err(e) => Err(PhaseTwoError::UplcMachineError(UplcMachineError {
+                        plutus_version,
+                        info: result.info,
+                        err: e.to_string(),
+                    })),
+                },
             }
         })
         .collect::<Result<Vec<_>, _>>()


crates/amaru-ledger/src/state.rs
```diff
@@ -602,6 +602,7 @@ impl<S: Store, HS: HistoricalStores> State<S, HS> {
         &mut self,
         point: &Point,
         raw_block: &RawBlock,
+        arena_pool: &ArenaPool,
```
yHSJ, line 607: I am not a *huge* fan of the drilling of the arena_pool through so many methods/functions. I just couldn't quite find the right place to have the arena_pool live. It is not state, it isn't quite the validation context either.

Would love some thoughts about the best place this arena pool should live.

.github/workflows/continuous-integration.yml
```diff
@@ -58,8 +58,12 @@ jobs:
...
             extra-args: +nightly -Zbuild-std=std,panic_abort
             command: build
-            setup: rustup component add rust-src --toolchain nightly
-
+            setup: |
+              rustup component add rust-src --toolchain nightly
+              curl -L -o riscv32-glibc-llvm.tar.xz \
+                "https://github.com/riscv-collab/riscv-gnu-toolchain/releases/download/2025.11.04/riscv32-glibc-ubuntu-24.04-llvm.tar.xz"
+              tar -xvf riscv32-glibc-llvm.tar.xz
+              export CC_riscv32gc_unknown_linux_gnu="$(pwd)/riscv/bin/riscv32-unknown-linux-gnu-clang"
```
KtorZ, line 80: What's up here 🤔 ?

crates/amaru-kernel/src/arena_pool.rs
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    /// Acquire an arena from the pool
+    ///
+    /// Blocks if all arenas are in use, waiting for one to be returned.
+    #[allow(unused_assignments)]
+    pub fn acquire(&self) -> PooledArena {
+        let arena = loop {
+            // FIXME: We may need to not ignore a poisoned mutex here
+            let mut guard = self.inner.arenas.lock().unwrap_or_else(|p| p.into_inner());
```
KtorZ, line 62: Seems to me that there's no value in acquiring and releasing the guard on each loop? Why not acquire the mutex outside of the loop, and drop it once at the end.

crates/amaru-kernel/src/arena_pool.rs
```diff
@@ -0,0 +1,191 @@
...
+impl AsRef<Bump> for PooledArena {
+    fn as_ref(&self) -> &Bump {
+        &self.arena
+    }
+}
+
+impl Drop for PooledArena {
+    fn drop(&mut self) {
+        // SAFETY NOTE: It's important we only take the arena once, here in the drop implementation
+        let mut arena = unsafe { ManuallyDrop::take(&mut self.arena) };
```
KtorZ, line 111: Why is the unsafe necessary here 🤔 ?

crates/amaru-kernel/src/arena_pool.rs
```diff
@@ -0,0 +1,191 @@
...
+    /// At most `max_size` arenas will exist simultaneously.
+    /// If all arenas are in use, `acquire()` will block.
+    pub fn new(max_size: usize, initial_capacity: usize) -> Self {
+        Self {
+            inner: Arc::new(Inner {
+                arenas: Mutex::new(VecDeque::with_capacity(max_size)),
+                condvar: Condvar::new(),
+                max_size,
+                initial_capacity,
+                total_created: Mutex::new(0),
```
KtorZ: I feel like this could just be an [`AtomicUsize`](https://doc.rust-lang.org/std/sync/atomic/struct.AtomicUsize.html). 

But also... why not simply use `arenas.len()`?  

It sounds way easier (and more reliable from an observability standpoint) to create all the arenas upfront when the pool is created; and have them ready to be acquired. This allows:

- reserving all resources we need for runtime _upfront_ (so that if we must fail due to a lack of memory, we fail early, and not later down the line when we cannot allocate a new arena). 
- simplify the logic of having to _dynamically_ create the arenas
- make the execution time more deterministic (as otherwise, we may *sometimes* pay the extra cost of creating and allocating the arena).

crates/amaru-kernel/src/arena_pool.rs
KtorZ: On second thoughts, I just realized what's going on here. Happy to see a `Condvar` here to avoid building an active waiting loop exhausting the CPU.

.github/workflows/continuous-integration.yml
yHSJ: This feels like ages ago 😅 

We were unable to compile the UPLC crate to WASM because of something missing in the clang compiler I think? This is a solution that I found with @SupernaviX to allow things to compile

crates/amaru-kernel/src/arena_pool.rs
yHSJ: This kind of concurrency in Rust is something I don't have a ton of experience with yet. So a lot of this is me doing my best to make a compiler happy and relying on documentation and forum discussions about this stuff. Happy to sit down to learn better ways to manage this?

crates/amaru-kernel/src/arena_pool.rs
yHSJ: Because `ManuallyDrop::take` is `unsafe`. So the compiler is not happy when we don't have it

crates/amaru-kernel/src/arena_pool.rs
yHSJ: Yeah, I think it may make more sense to create them all at startup. The goal with this solution was just to make sure we're not using unneeded resources, but obviously that is on the user to determine how much memory they want to allocate for this.

And finding the "right" number will take some tuning

.github/workflows/continuous-integration.yml
KtorZ: I suppose you meant `RISC-V`, not `WASM`, right? Because this seems to be explicitly fetching some riscv toolchain 😅 ... and it's the risc-5 job. Maybe add a relevant comment here? But I am curious about what the problem was. I'll do some experiment on my end.

crates/amaru-ledger/src/block_validator.rs
```diff
@@ -32,11 +33,13 @@ where
     HS: HistoricalStores + Send,
 {
     pub state: Arc<Mutex<state::State<S, HS>>>,
+    pub arena_pool: ArenaPool,
```
KtorZ: small note: would be better to get a name here that provides context; we know it's an arena pool already from the type, so the name better describe _what for_: (e.g. `vm_eval_pools`)

crates/amaru-ledger/src/rules/block.rs
```diff
@@ -14,26 +14,35 @@
...
+use thiserror::Error;
 use tracing::{Level, instrument};
 
 pub mod body_size;
 pub mod ex_units;
 pub mod header_size;
 
+#[derive(Debug, Error)]
+pub enum TransactionInvalid {
+    #[error("tranasction failed phase one validation: {0}")]
```
KtorZ:
```suggestion
    #[error("transaction failed phase one validation: {0}")]
```

crates/amaru-ledger/src/rules/block.rs
```diff
@@ -203,6 +213,11 @@ where
     for (i, transaction) in (0u32..).zip(transactions.into_iter()) {
         let transaction_hash = transaction.original_hash();
 
+        let is_valid = !block
+            .invalid_transactions
+            .as_ref()
+            .is_some_and(|invalid_txs| invalid_txs.contains(&i));
```
KtorZ:
```suggestion
            .is_some_and(|invalid_txs| invalid_txs.contains(&i));
```

Not sure about the `is_some_and` here. Semantically: if the field `invalid_transactions` is missing, then it likely means that there's no invalid transactions whatsoever. I am not sure whether a missing field is a valid encoding but I see room for a good test here (if it's not a valid encoding, we should fail to decode it; and if it is, then this logic is likely broken).

yHSJ: You're right, this logic is almost certainly wrong. The field in Pallas is optional, so I assume that it will decode, but another question is whether we *should*.

I'll change this logic and leave a note.

.github/workflows/continuous-integration.yml
yHSJ: Sorry, yes. I meant `RISC-V`.

crates/amaru-plutus/src/to_plutus_data.rs
```diff
@@ -158,7 +156,22 @@ where
...
+                2 => {
+                    <amaru_kernel::StakeAddress as ToPlutusData<2>>::to_plutus_data(stake_address)?
+                        .to_plutus_data()
+                }
+                3 => {
+                    <amaru_kernel::StakeAddress as ToPlutusData<3>>::to_plutus_data(stake_address)?
+                        .to_plutus_data()
+                }
+                _ => unreachable!(),
+            },
```
KtorZ: What's going on here?

yHSJ: I just can't seem to make the compiler happy any other way...

jeluard:
@yHSJ rebasing should help having the e2e tests pass

.claude/settings.local.json
KtorZ, line 1: Please don't commit this; this is specific to your setup, not to the project. 

I recommend using a system-wide .gitignore in your gitconfig 

<p align="right"><sub><strong>~/.gitconfig</strong></sub></p>

```toml
[core]
  excludesfile = "/Users/jshy/.gitignore"
```

yHSJ: Yeah, didn't mean to include this. Was using it to help step through some messy rebases, it turns out Claude is *very* good at that.

PR #554, opened 2025-11-14
fix: release pipeline

abailly:
Not sure if this is what you had in mind @KtorZ 

## Release Notes

* **Chores**
  * Updated release artifact naming to include version and environment information for improved organization and traceability
  * Enhanced build and deployment pipeline with integrated testing and conditional service management

* **Documentation**
  * Updated project documentation

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -89,4 +89,5 @@ jobs:
 
       - uses: actions/upload-pages-artifact@v3
         with:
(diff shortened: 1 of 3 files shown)


KtorZ:
Not quite. I don't think this will work. We need to publish _a single page artifact_. Otherwise I believe this will simply override previous ones.

KtorZ:
Also, one important factor is that I want the URL to be _static_; otherwise it's extra-painful to link to it on the website. The nightly version contains a git sha, so it's not suitable.

PR #556, opened 2025-11-14
chore: remove completely gasket

abailly:
The only remaining use of gasket was in the simulation, and only to implement the `echo` "simulator" which is seldom run and used.

## Summary by CodeRabbit

* **Chores**
  * Added a shell-based workflow to build/install the service and restart it when on main.
  * Removed the echo simulation functionality and related binaries.
  * Removed an outdated workspace dependency.

* **Refactor**
  * Simplified chain-sync error handling.
  * Reorganized simulator messaging types (Envelope moved/centralized).

* **Documentation**
  * Minor README update and simulation docs revised with alternative tooling guidance.


.radicle/native.yaml
```diff
@@ -0,0 +1,31 @@
+shell: |
```
jeluard: Is this expected to be part of this PR?

abailly: No

jeluard (approved):

Issue #577, opened 2025-11-21
Slow build time with no code changes


jeluard:
A `build.rs` file has been added recently to the `amaru-ledger` crate. It forces some compilation/linking even when no code changes has been introduced. This is especially problematic when recompiling in `release` mode.

Note that I don't believe the issue is the `build.rs` file itself, but what it does. There is already a `build.rs` file in `amaru` crate and there is no issue.

```shell
❯ cargo build
    Finished `dev` profile [optimized] target(s) in 0.16s
```

with current build.rs:

```
❯ cargo build
   Compiling amaru-ledger v0.1.0 (/Users/julien/Documents/Projects/pragma-org/amaru/crates/amaru-ledger)
warning: amaru-ledger@0.1.0: no conformance ledger test vectors found; unpack them in amaru-ledger/tests/data/conformance
   Compiling amaru-stores v0.1.0 (/Users/julien/Documents/Projects/pragma-org/amaru/crates/amaru-stores)
   Compiling amaru-consensus v0.1.0 (/Users/julien/Documents/Projects/pragma-org/amaru/crates/amaru-consensus)
   Compiling amaru v0.1.0 (/Users/julien/Documents/Projects/pragma-org/amaru/crates/amaru)
    Finished `dev` profile [optimized] target(s) in 4.48s
```

current build.rs and `release` mode:

```shell
❯ cargo build --release
   Compiling amaru-ledger v0.1.0 (/Users/julien/Documents/Projects/pragma-org/amaru/crates/amaru-ledger)
   Compiling amaru-stores v0.1.0 (/Users/julien/Documents/Projects/pragma-org/amaru/crates/amaru-stores)
   Compiling amaru-consensus v0.1.0 (/Users/julien/Documents/Projects/pragma-org/amaru/crates/amaru-consensus)
   Compiling amaru v0.1.0 (/Users/julien/Documents/Projects/pragma-org/amaru/crates/amaru)
    Finished `release` profile [optimized] target(s) in 1m 54s
```

PR #585, opened 2025-12-03
fix: stop unnecessary rebuilds caused by missing test folder

geofflittle:
The `amaru-ledger` build script was telling Cargo to watch the `tests/data/rules-conformance` folder for changes, even if that folder didn't exist.

Because the folder was missing, Cargo always thought the code was out of date. This forced the project to rebuild, starting from `amaru-ledger` and up, every run. This update fixes the script so it only watches that folder if it exists.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a potential build failure when a tests directory is missing by adding a conditional check before emitting a build directive.


jeluard:
Looks like it does help with #577 
Also surprisingly I have to run twice the build commands after the initial one, the first extra one is still slow. Any clue why?

PR #586, opened 2025-12-05
Implements 'reset to epoch'

Quantumplation:
Allows you to truncate your database to the start of some epoch;

Works by manipulating the rocksdb database, deleting all epochs after the target epoch, and then copying the previous one as "live" to "start" at the beginning of the epoch.

## Summary by CodeRabbit

* **New Features**
  * Added a new CLI command "reset_to_epoch" to reset ledger snapshots to a specified epoch.
  * Command accepts a ledger directory and target epoch, validates safety (min history requirement, no future resets, epoch bounds), preserves the penultimate epoch for restoration, and restores it as the live snapshot while removing others.
  * The command is now integrated into the main CLI.

### DIFF
--- a/crates/amaru/src/bin/amaru/cmd/reset_to_epoch.rs
+++ b/crates/amaru/src/bin/amaru/cmd/reset_to_epoch.rs
@@ -12,110 +12,22 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use amaru_ledger::state::MIN_LEDGER_SNAPSHOTS;
 use clap::Parser;
-use std::fs;
-use std::io;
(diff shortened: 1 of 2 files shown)


jeluard:
This makes several assumptions about internal implementation details tied to our current usage of `rocksdb`. Mainly, that snapshots are complete live state copies and can be swapped around.
This makes it problematic IMO to enshrine in the API as those might not hold anymore if we optimize snapshots and will probably break with non-rocksdb implementations (we don't strictly need full copies of the current epoch data for previous epochs).
In gral I think we should just stick with what is offered at the trait `Store` level.

Specifically here, I think that an extra script would be more sensible.

PR #587, opened 2025-12-05
feat: tx submission protocol - end to end

etorreborre:
This PR implements some preliminary support for the [transaction submission protocol](https://ouroboros-network.cardano.intersectmbo.org/pdfs/network-spec/network-spec.pdf).

The main blocks provided in this PR are:

 - A `Mempool` with a `TxSubmissionMempool` interface dedicated to mempool operations required for the tx submission protocol.
 - Two `pure-stage` stages processing server requests and client replies.
 - The abstraction of interactions with the mempool as `Mempool` effects.
 - The abstraction of interactions with the network as `Network` effects.
 - The implementation of the network communication as:
    - An `UpstreamClient` encapsulating the existing actor code for communicating with upstream peers and extended to the client part of the tx submission protocol.
    - A `DownstreamServer` encapsulating the existing actor code for communicating with downstream peers and extended to the server part of the tx submission protocol.

## Protocol support

The full protocol specification is not implemented yet. In particular the control of in-flight data size is not done at all.
This will be done in a subsequent PR.

## Testing

For now the only testing performed for this protocol in the current PR is limited to:

 - Unit testing of the client side with a mock transport.
 - Unit testing of the server side with a mock transport.
 - Unit testing of the integration of the client and server sides with a mock transport.

In follow-up PRs there will be:

 - Simulation testing for the new stages.
 - End-to-end testing with actual nodes (this can't be done for now because the handshake protocol fails between the Amaru node and the Haskell node at the moment).

### Existing mempool

I left the existing `DummyMempool` and adapted it to the new `Mempool` interface. This code will probably disappear or be merged with the `InMemoryMempool` from this PR.

### `ForwardEventListener`

`ChainSync` events sent downstream were using a specific network effect and this interface. It has been removed to be simply a function on the `NetworkResource` that mediates the actual network access downstream.

### `ChainFollower` code

This code has been moved to the `amaru-network` crate (from the `amaru-consensus` crate in order to have it close to the `DownstreamServer` code where it is used.

### `ClientMsg`

This message type, used in the downstream code, was parametrized by the header type that could be sent downstream.
Since this data type has been extended to more messages types that chain sync messages, the `H` type parameter was a bit confusing and actually not necessary. So I removed it and replaced by a concrete `BlockHeader` type.
 
### Dependency on a Pallas branch

Some instances were missing on a few Pallas data types. There is [a PR opened addressing that on the pallas `lts/v0` branch](https://github.com/txpipe/pallas/pull/715), but it will need to be merged because we can merge this PR (and even better a pallas version `0.34.0` released).

### Dependency on a Pallas crates

More generally the `amaru-consensus` code has been isolated from the `amaru-network` and the `pallas-network` crates so that the future switch to our own networking stack will be easier in the future.
 

## Summary by CodeRabbit

* **New Features**
  * Peer-to-peer transaction submission (client/server) with a configurable in-memory mempool and async sequencing.
  * Pluggable transaction validation API and test validator utilities.

* **Refactor**
  * Network split into upstream (peers) and downstream (server) components; clearer tx/chain responsibilities.
  * Consensus and tx-submission wiring modularized into separate graphs and stages.

* **Chores**
  * Updated dependency sources and dev-dependencies.

* **Tests**
  * Expanded tx-submission test harness, mocks, and helpers for end-to-end scenarios.


etorreborre:
@coderabbitai review please

crates/amaru-consensus/src/consensus/stages/receive_tx_client_reply.rs
```diff
@@ -0,0 +1,140 @@
...
+/// This stage processes incoming tx submission client replies:
+///
+///  - It keeps track of connected peers.
+///  - For each peer, it maintains some state about the requests that have made to the client.
+///
+pub fn stage(
+    (mut servers, errors): State,
+    msg: TxClientReply,
+    eff: impl ConsensusOps,
+) -> impl Future<Output = State> {
```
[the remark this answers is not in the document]
etorreborre: Not necessary.

crates/amaru-consensus/src/consensus/effects/mempool_effects.rs
```diff
@@ -0,0 +1,409 @@
...
+    CanValidateTransactions, MempoolSeqNo, TransactionValidationError, TxId, TxOrigin,
+    TxRejectReason, TxSubmissionMempool,
+};
+use pure_stage::{
+    BoxFuture, Effects, ExternalEffect, ExternalEffectAPI, ExternalEffectSync, Resources, SendData,
+};
+use serde::{Deserialize, Serialize};
+use std::fmt::Debug;
+use std::pin::Pin;
+use std::sync::Arc;
```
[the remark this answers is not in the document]
etorreborre: Not necessary.

crates/amaru-consensus/src/consensus/tx_submission/tests/assertions.rs
```diff
@@ -0,0 +1,128 @@
...
+    rx_messages: &mut Receiver<Message<EraTxId, EraTxBody>>,
+) -> anyhow::Result<()> {
+    let actual = rx_messages.recv().await;
+    assert!(
+        actual.is_none(),
+        "expected no message, but got one: {:?}",
+        actual
+    );
+    Ok(())
+}
```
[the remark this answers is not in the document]
etorreborre: Good point, done.

crates/amaru-consensus/src/consensus/tx_submission/tx_submission_client_state.rs
```diff
@@ -0,0 +1,164 @@
...
+                    return Ok(TxClientResponse::<Tx>::ProtocolError(anyhow!(
+                        "0 transaction ids requested"
+                    )));
+                }
+                if !mempool
+                    .wait_for_at_least(mempool.last_seq_no().add(req as u64))
+                    .await
+                {
+                    return Ok(TxClientResponse::<Tx>::Done);
+                }
```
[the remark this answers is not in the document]
etorreborre: That's the point of this method.

crates/amaru-mempool/src/strategies/dummy.rs
```diff
@@ -12,89 +12,117 @@
...
-        mem::take(&mut self.transactions)
+impl<Tx: Encode<()> + Send + Sync + 'static> TxSubmissionMempool<Tx> for DummyMempool<Tx> {
+    fn insert(&self, tx: Tx, _tx_origin: TxOrigin) -> Result<(TxId, MempoolSeqNo), TxRejectReason> {
+        let tx_id = TxId::from(&tx);
+        self.inner.write().transactions.push(tx);
+        Ok((
+            tx_id,
+            MempoolSeqNo(self.inner.read().transactions.len() as u64 - 1),
+        ))
     }
```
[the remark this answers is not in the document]
etorreborre: Good point.

crates/amaru-mempool/src/strategies/in_memory_mempool.rs
```diff
@@ -0,0 +1,302 @@
...
+        Self: Sized,
+    {
+        let keys_to_remove = BTreeSet::from_iter(keys(tx));
+        let mut inner = self.inner.write();
+        inner.entries_by_id.retain(|_, entry| {
+            !keys(&entry.tx)
+                .into_iter()
+                .any(|k| keys_to_remove.contains(&k))
+        });
+    }
```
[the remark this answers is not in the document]
etorreborre: Done.

crates/amaru-consensus/src/consensus/effects/network_effects.rs
```diff
@@ -188,11 +224,260 @@ impl ExternalEffect for ChainSyncEffect {
...
+            let result: <Self as ExternalEffectAPI>::Response =
+                network_operations.send_tx_server_request(request).await;
+            Box::new(result) as Box<dyn SendData>
         })
     }
 }
 
+impl ExternalEffectAPI for TxIdsRequestEffect {
+    type Response = Result<(), ClientConnectionError>;
+}
```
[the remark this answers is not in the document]
etorreborre: Done

crates/amaru-consensus/src/consensus/tx_submission/tests/assertions.rs
```diff
@@ -0,0 +1,128 @@
...
+                .collect::<Vec<_>>()
+                .join(", "),
+            tx_ids
+                .iter()
+                .map(|id| format!("{}", id))
+                .collect::<Vec<_>>()
+                .join(", ")
+        );
+    }
+}
```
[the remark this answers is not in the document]
etorreborre: Done.

crates/amaru-network/src/server/downstream_server.rs
```diff
@@ -116,76 +114,114 @@ impl<H: IsHeader + 'static + Clone + Send> TcpForwardChainServer<H> {
...
-#[async_trait]
-impl ForwardEventListener for TcpForwardChainServer<BlockHeader> {
-    async fn send(&self, event: ForwardEvent) -> anyhow::Result<()> {
+    pub async fn next_tx_reply(&self) -> Result<TxClientReply, ClientConnectionError> {
+        if let Some(tx_reply) = self.tx_client_reply_receiver.lock().await.recv().await {
+            Ok(tx_reply)
+        } else {
+            Err(anyhow!("tx reply channel closed").into())
+        }
+    }
```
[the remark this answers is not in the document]
etorreborre: Done.

crates/amaru-consensus/src/consensus/effects/network_effects.rs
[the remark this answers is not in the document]
etorreborre: Done.

crates/amaru-network/src/server/downstream_server.rs
```diff
@@ -116,76 +114,114 @@ impl<H: IsHeader + 'static + Clone + Send> TcpForwardChainServer<H> {
...
+            ActoMsgSuper::Message(ClientMsg::TxSubmission(request)) => {
+                if let Some(sender) = clients_by_peer.get(request.peer()) {
+                    let peer = request.peer().clone();
+                    if !sender.send(ClientProtocolMsg::TxSubmission(request)) {
+                        tracing::warn!(target: EVENT_TARGET, "failed to send tx request to peer: {}", peer);
+                    }
+                } else {
+                    tracing::warn!(target: EVENT_TARGET, "no connected peers to send tx request");
                 }
             }
```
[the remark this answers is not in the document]
etorreborre: Done.

crates/amaru-consensus/src/consensus/tx_submission/tests/sized_mempool.rs
```diff
@@ -0,0 +1,113 @@
...
+        seq_no: MempoolSeqNo,
+    ) -> Pin<Box<dyn Future<Output = bool> + Send + '_>> {
+        if seq_no <= MempoolSeqNo(self.capacity)
+            && self.inner_mempool.last_seq_no() >= MempoolSeqNo(self.capacity)
+        {
+            self.inner_mempool.wait_for_at_least(seq_no)
+        } else {
+            Box::pin(async move { true })
+        }
+    }
```
etorreborre, line 90: Fair point. I updated the logic to make all of this clearer

crates/amaru-network/src/server/downstream_server.rs
```diff
@@ -116,76 +114,114 @@ impl<H: IsHeader + 'static + Clone + Send> TcpForwardChainServer<H> {
...
+        }
+    }
+
+    pub async fn send_tx_request(
+        &self,
+        request: TxServerRequest,
+    ) -> Result<(), ClientConnectionError> {
+        self.clients.send(ClientMsg::TxSubmission(request));
+        Ok(())
+    }
```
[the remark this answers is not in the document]
etorreborre: Good idea

crates/amaru-ouroboros-traits/src/tx_submission_events.rs
```diff
@@ -0,0 +1,288 @@
...
+            TxClientReply::TxIds { peer, tx_ids, .. } => f
+                .debug_struct("TxIds")
+                .field("peer", peer)
+                .field(
+                    "tx_ids",
+                    &tx_ids
+                        .iter()
+                        .map(|(tx_id, size)| format!("{} (size: {}))", tx_id, size))
+                        .join(", "),
+                )
```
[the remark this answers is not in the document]
etorreborre: Done.

crates/amaru-consensus/src/consensus/effects/network_effects.rs
```diff
@@ -188,11 +224,260 @@ impl ExternalEffect for ChainSyncEffect {
...
+                })
+                .await;
+            Box::new(result) as Box<dyn SendData>
+        })
+    }
+}
+
+impl ExternalEffectAPI for TxsRequestEffect {
+    type Response = Result<(), ClientConnectionError>;
+}
```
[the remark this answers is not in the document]
etorreborre: Done.

crates/amaru-network/src/server/downstream_server.rs
```diff
@@ -116,76 +114,114 @@ impl<H: IsHeader + 'static + Clone + Send> TcpForwardChainServer<H> {
...
             ActoMsgSuper::Supervision { id, name, result } => {
                 tracing::info!(target: EVENT_TARGET, "client {} terminated: {:?}", name, result);
-                clients.remove(&id);
+                if let Some(client_peer) = peers_by_client_id.remove(&id) {
+                    clients_by_peer.remove(&client_peer);
+                };
+                peers_by_client_id.remove(&id);
             }
         }
     }
```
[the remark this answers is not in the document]
etorreborre: 👍

crates/amaru-consensus/src/consensus/effects/network_effects.rs
```diff
@@ -188,11 +224,260 @@ impl ExternalEffect for ChainSyncEffect {
...
+    fn run(self: Box<Self>, resources: Resources) -> BoxFuture<'static, Box<dyn SendData>> {
+        Self::wrap(async move {
+            #[expect(clippy::expect_used)]
+            let network = resources
+                .get::<ResourceNetworkOperations>()
+                .expect("TxRequestEffect requires a NetworkOperations")
+                .clone();
+            network.next_tx_server_request().await
+        })
+    }
```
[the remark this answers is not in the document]
etorreborre: Done.

crates/amaru-mempool/src/strategies/in_memory_mempool.rs
```diff
@@ -0,0 +1,317 @@
...
+use amaru_kernel::to_cbor;
+use amaru_ouroboros_traits::mempool::Mempool;
+use amaru_ouroboros_traits::{
+    CanValidateTransactions, MempoolSeqNo, TransactionValidationError, TxId, TxOrigin,
+    TxRejectReason, TxSubmissionMempool,
+};
+use std::collections::{BTreeMap, BTreeSet};
+use std::mem;
+use std::pin::Pin;
+use std::sync::Arc;
```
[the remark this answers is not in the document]
etorreborre: Not necessary

crates/amaru-ouroboros-traits/src/tx_submission_events.rs
```diff
@@ -0,0 +1,288 @@
...
+        #[serde(skip, default = "Span::none")]
+        span: Span,
+    },
+    Txs {
+        peer: Peer,
+        txs: Vec<Tx>,
+        #[serde(skip, default = "Span::none")]
+        span: Span,
+    },
+}
```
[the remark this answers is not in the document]
etorreborre: Done.

crates/amaru-mempool/src/strategies/dummy.rs
```diff
@@ -12,89 +12,116 @@
...
-        }
+    fn get_txs_for_ids(&self, _ids: &[TxId]) -> Vec<Arc<Tx>> {
+        vec![]
     }
 
-    impl Deref for FakeTx<'_> {
-        type Target = Self;
+    fn last_seq_no(&self) -> MempoolSeqNo {
+        MempoolSeqNo(self.inner.read().transactions.len() as u64)
+    }
```
etorreborre, line 78: Done.

crates/amaru-mempool/src/strategies/in_memory_mempool.rs
```diff
@@ -0,0 +1,317 @@
...
+impl<Tx: Send + Sync + 'static + Encode<()>> Mempool<Tx> for InMemoryMempool<Tx> {
+    fn take(&self) -> Vec<Tx> {
+        let mut inner = self.inner.write();
+        let entries = mem::take(&mut inner.entries_by_id);
+        let _ = mem::take(&mut inner.entries_by_seq);
+        entries
+            .into_values()
+            .filter_map(|entry| Arc::into_inner(entry.tx))
+            .collect()
+    }
```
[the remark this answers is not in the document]
etorreborre: I removed the use of `Arc` now.

crates/amaru/src/stages/build_tx_submission_graph.rs
```diff
@@ -0,0 +1,84 @@
...
+    network: &mut impl StageGraph,
+) -> (StageRef<TxServerRequest>, StageRef<TxClientReply>) {
+    let receive_tx_server_request_stage = network.stage(
+        "receive_tx_server_request",
+        with_tx_effects(receive_tx_server_request::stage),
+    );
+    let receive_tx_client_reply_stage = network.stage(
+        "receive_tx_reply",
+        with_tx_effects(receive_tx_client_reply::stage),
+    );
```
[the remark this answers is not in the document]
etorreborre: Done.

crates/amaru-consensus/src/consensus/tx_submission/tests/tx_submission_server.rs
```diff
@@ -0,0 +1,259 @@
...
+        }
+        Ok(())
+    }
+}
+
+mod tests {
+    use super::*;
+    use crate::consensus::tx_submission::tests::{
+        MockServerTransport, assert_next_message, create_transaction, create_transactions,
+    };
```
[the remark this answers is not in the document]
etorreborre: This module is already in another module gated with `#[cfg(test)]`

crates/amaru-network/src/server/client_protocol.rs
```diff
@@ -344,16 +318,66 @@ async fn block_fetch<H: IsHeader>(
...
+                    ActoInput::NoMoreSenders | ActoInput::Supervision { .. }  => {
+                        tracing::debug!("parent terminated");
+                        break;
+
+                   }
+                }
+            }
+        }
+    }
     Ok(())
```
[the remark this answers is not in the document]
etorreborre: This will be entirely reworked

crates/amaru-network/src/network_resource.rs
```diff
@@ -12,101 +12,224 @@
...
+        if let Some(receiver) = self.tx_client_reply_senders.get(reply.peer()) {
+            receiver
+                .send(reply)
+                .await
+                .map_err(|e| ClientConnectionError::new(e.into()))
+        } else {
+            Ok(())
+        }
+    }
+
```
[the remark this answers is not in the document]
etorreborre: I added a warning.

crates/amaru-mempool/src/strategies/in_memory_mempool.rs
```diff
@@ -0,0 +1,316 @@
...
+                    panic!(
+                        "Inconsistent mempool state: entry missing for tx_id {:?}",
+                        tx_id
+                    );
+                }
+            })
+            .collect();
+        result.sort_by_key(|(_, _, seq_no)| *seq_no);
+        result
+    }
```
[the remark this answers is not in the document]
etorreborre: I changed the busy wait.

crates/amaru-consensus/src/consensus/effects/consensus_effects.rs
```diff
@@ -27,6 +28,9 @@ pub trait ConsensusOps: Send + Sync + Clone {
     fn store(&self) -> Arc<dyn ChainStore<BlockHeader>>;
     /// Return a NetworkOps implementation to access network operations, like fetch_block
     fn network(&self) -> impl NetworkOps;
+    /// Return a NetworkOps implementation to access mempool operations, like get_tx to retrieve a transaction
+    /// from the mempool.
+    fn mempool(&self) -> Arc<dyn TxSubmissionMempool<Tx>>;
```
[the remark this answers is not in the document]
etorreborre: Done.

crates/amaru-mempool/src/strategies/in_memory_mempool.rs
```diff
@@ -0,0 +1,343 @@
...
+                let current_seq = { self.inner.read().next_seq };
+                if current_seq >= seq_no.0 {
+                    return true;
+                }
+
+                // Wait until someone inserts a new transaction and notifies us
+                self.notify.notified().await;
+            }
+        })
+    }
```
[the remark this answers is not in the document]
etorreborre: Done.

crates/amaru-consensus/src/consensus/tx_submission/tests/tx_client_transport.rs
```diff
@@ -0,0 +1,126 @@
...
+    async fn reply_txs(&mut self, txs: Vec<EraTxBody>) -> Result<(), TransportError>;
+}
+
+/// For now the transport error is either a pallas Error or a generic anyhow error.
+#[derive(thiserror::Error, Debug)]
+#[error(transparent)]
+pub enum TransportError {
+    PallasError(#[from] Error),
+    Other(#[from] anyhow::Error),
+}
```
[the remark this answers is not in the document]
etorreborre: ok

crates/amaru-consensus/src/consensus/tx_submission/tx_submission_client_state.rs
```diff
@@ -0,0 +1,170 @@
...
+                let txs = mempool.get_txs_for_ids(tx_ids.as_slice());
+                if txs.is_empty() {
+                    Ok(TxClientResponse::<Tx>::ProtocolError(anyhow!(
+                        "unknown transactions requested: {:?}",
+                        tx_ids
+                    )))
+                } else {
+                    Ok(TxClientResponse::NextTxs(txs))
+                }
+            }
```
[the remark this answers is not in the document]
etorreborre: Let's leave it for now.

crates/amaru-consensus/src/consensus/effects/mempool_effects.rs
```diff
@@ -0,0 +1,393 @@
...
+    effects: Effects<T>,
+}
+
+impl<T> MemoryPool<T> {
+    pub fn new(effects: Effects<T>) -> MemoryPool<T> {
+        MemoryPool { effects }
+    }
+
+    /// This function runs an external effect synchronously.
+    pub fn external_sync<E: ExternalEffectSync + serde::Serialize + 'static>(
```
rkuhn, line 46: ExternalEffectSync is already `'static`, so this bound should be unnecessary

crates/amaru-consensus/src/consensus/effects/mempool_effects.rs
```diff
@@ -0,0 +1,393 @@
...
+    ) -> E::Response
+    where
+        T: SendData + Sync,
+    {
+        self.effects.external_sync(effect)
+    }
+}
+
+impl<T: SendData + Sync> CanValidateTransactions<Tx> for MemoryPool<T> {
+    fn validate_transaction(&self, tx: &Tx) -> Result<(), TransactionValidationError> {
```
rkuhn: If the called function will unconditionally need to clone the tx, then it should accept a `Tx` instead of `&Tx`. Otherwise the caller needs to keep a Tx around even when not needed.

crates/amaru-consensus/src/consensus/effects/mempool_effects.rs
```diff
@@ -0,0 +1,393 @@
...
+        self.effects.external_sync(ValidateTransaction(tx.clone()))
+    }
+}
+
+impl<T: SendData + Sync> TxSubmissionMempool<Tx> for MemoryPool<T> {
+    fn insert(&self, tx: Tx, tx_origin: TxOrigin) -> Result<(TxId, MempoolSeqNo), TxRejectReason> {
+        self.external_sync(Insert::new(tx, tx_origin))
+    }
+
+    fn get_tx(&self, tx_id: &TxId) -> Option<Arc<Tx>> {
```
rkuhn: I’d hope that `TxId` can be Copy, which would remove the need for cloning here.

crates/amaru-consensus/src/consensus/effects/mempool_effects.rs
```diff
@@ -0,0 +1,393 @@
...
+        self.external_sync(GetTx::new(tx_id.clone()))
+    }
+
+    fn tx_ids_since(&self, from_seq: MempoolSeqNo, limit: u16) -> Vec<(TxId, u32, MempoolSeqNo)> {
```
rkuhn, line 79: The `limit: u16` looks suspicious: if it is the maximum number of transactions the why not use `usize`?

crates/amaru-consensus/src/consensus/effects/mempool_effects.rs
```diff
@@ -0,0 +1,393 @@
...
+    }
+
+    fn last_seq_no(&self) -> MempoolSeqNo {
+        self.external_sync(LastSeqNo)
+    }
+}
+
+// EXTERNAL EFFECTS DEFINITIONS
+
+pub type ResourceMempool = Arc<dyn TxSubmissionMempool<Tx>>;
```
rkuhn: I haven’t yet seen where this is used, but it feels wrong: resources are used by effect handlers, but this shape of resource is just a factory for effect handlers.

crates/amaru-consensus/src/consensus/effects/mempool_effects.rs
```diff
@@ -0,0 +1,393 @@
...
+
+impl ExternalEffect for Insert {
+    #[expect(clippy::expect_used)]
+    fn run(self: Box<Self>, resources: Resources) -> BoxFuture<'static, Box<dyn SendData>> {
+        Self::wrap_sync({
+            let mempool = resources
+                .get::<ResourceMempool>()
+                .expect("ResourceMempool requires a mempool")
+                .clone();
+            mempool.insert(self.tx, self.tx_origin)
```
rkuhn, line 125: I see, so you’re using `trait TxSubmissionMempool<Tx>` for two very different purposes. My gut feeling is that this will lead to confusion, the API trait and the resource implementation should be different types. And the resource implementation doesn’t need to go through a trait because effects can already be replaced in pure-stage.

crates/amaru-consensus/src/consensus/stages/pull_tx_replies.rs
```diff
@@ -0,0 +1,53 @@
...
+
+        drop(entered);
+        async {
+            eff.send(&downstream, msg).await;
+            eff.send(eff.me_ref(), NextTxReply).await;
+        }
+        .instrument(span)
+        .await;
+        downstream
+    } else {
```
rkuhn, line 45: an error should probably terminate the connection — I guess this will be done when integrating with the proper networking layer, right?

crates/amaru-consensus/src/consensus/stages/pull_tx_replies.rs
```diff
@@ -0,0 +1,53 @@
...
+        .await
+    {
+        let span = tracing::trace_span!("diffusion.tx");
+        span.set_parent(msg.span().context()).ok();
+        let entered = span.enter();
+
+        msg.set_span(span.clone());
+
+        drop(entered);
+        async {
```
rkuhn: How about this approach? Reads a bit simpler and should do the same thing (unless I’m missing something).
```suggestion
        let span2 = span.clone();
        async move {
            msg.set_span(span2);
```
(same below)

crates/amaru-consensus/src/consensus/tx_submission/tests/assertions.rs
```diff
@@ -0,0 +1,114 @@
...
+use std::time::Duration;
+use tokio::sync::mpsc::Receiver;
+use tokio::time::{sleep, timeout};
+
+/// Check that all the given transactions are eventually present in the server mempool.
+#[track_caller]
+pub async fn expect_server_transactions(txs: Vec<Tx>, node_handle: &NodeHandle) {
+    let server_mempool: Arc<dyn Mempool<Tx>> = node_handle.server_mempool.clone();
+    let tx_ids: Vec<_> = txs.iter().map(TxId::from).collect();
+    let mut actual: Vec<_> = tx_ids
```
rkuhn: given that the next iteration will be done unconditionally, I’d start with an empty vector here

crates/amaru-consensus/src/consensus/tx_submission/tests/assertions.rs
```diff
@@ -0,0 +1,114 @@
...
+    Ok(())
+}
+
+/// Check that the next message matches the expected one.
+#[track_caller]
+pub async fn assert_next_message(
+    rx_messages: &mut Receiver<Message<EraTxId, EraTxBody>>,
+    expected: Message<EraTxId, EraTxBody>,
+) -> anyhow::Result<()> {
+    let actual = rx_messages
```
rkuhn: there should be a timeout here so that a test failure can be more easily located

crates/amaru-consensus/src/consensus/tx_submission/tests/assertions.rs
```diff
@@ -0,0 +1,114 @@
...
+        .recv()
+        .await
+        .ok_or_else(|| anyhow::anyhow!("no message received"))?;
```
rkuhn: this is rather “channel closed”, right?

crates/amaru-consensus/src/consensus/tx_submission/tests/nodes.rs
```diff
@@ -0,0 +1,197 @@
...
+            mut server,
+            transport,
+        } = self;
+        let MockTransport {
+            client_transport,
+            server_transport,
+            rx_observe_messages,
+        } = transport;
+        let server_handle =
+            tokio::spawn(async move { server.start_server_with_transport(server_transport).await });
```
rkuhn:
```suggestion
            tokio::spawn(server.start_server_with_transport(server_transport));
```

crates/amaru-consensus/src/consensus/tx_submission/tests/node.rs
```diff
@@ -0,0 +1,197 @@
...
+use amaru_ouroboros_traits::{CanValidateTransactions, Mempool};
+use pallas_network::miniprotocols::txsubmission::{EraTxBody, EraTxId, Message};
+use pallas_primitives::conway::Tx;
+use std::sync::Arc;
+use tokio::sync::mpsc::Receiver;
+use tokio::task::JoinHandle;
+
+/// A test node consisting of a TxSubmissionClient and TxSubmissionServer,
+/// their respective mempools, connected via a MockTransport.
+pub struct Node {
```
rkuhn: This isn’t really a node, it is parts of two nodes, correct? (the client and the server)

crates/amaru-consensus/src/consensus/tx_submission/tests/sized_mempool.rs
```diff
@@ -0,0 +1,106 @@
...
+}
+
+impl CanValidateTransactions<Tx> for SizedMempool {
+    fn validate_transaction(&self, tx: &Tx) -> Result<(), TransactionValidationError> {
+        self.inner_mempool.validate_transaction(tx)
+    }
+}
+
+impl TxSubmissionMempool<Tx> for SizedMempool {
+    fn insert(&self, tx: Tx, tx_origin: TxOrigin) -> Result<(TxId, MempoolSeqNo), TxRejectReason> {
```
rkuhn, line 67: `self.capacity` seems to suggest that insertion should fail or displace some old entries when the limit is reached — do I misunderstand the purpose of this implementation?

crates/amaru-consensus/src/consensus/tx_submission/tests/test_cases.rs
```diff
@@ -0,0 +1,175 @@
...
+}
+
+#[tokio::test]
+async fn test_client_with_non_blocking_server() -> anyhow::Result<()> {
+    let txs = create_transactions(8);
+    let server_options = ServerOptions::default()
+        .with_mempool_capacity(8)
+        .with_max_window(8)
+        .with_fetch_batch(2)
+        .with_blocking(Blocking::No);
```
rkuhn: Whether to send `MsgRequestTxIdsNonBlocking` or `MsgRequestTxIdsBlocking` is a dynamic choice that must be made as specified in the network spec, so I don’t understand why this is a server configuration parameter.

crates/amaru-consensus/src/consensus/tx_submission/tests/test_cases.rs
```diff
@@ -0,0 +1,175 @@
...
+    let mut expected = vec![];
+    for (i, tx) in txs.iter().enumerate() {
+        if i % 2 == 0 {
+            expected.push(tx.clone());
+        }
+    }
+    expect_server_transactions(expected, &node_handle).await;
+
+    // Check the requests sent by the server do not ask again for the invalid transactions.
+    let actual: Vec<Message<EraTxId, EraTxBody>> = node_handle.observe_messages().await;
```
rkuhn: If I understand correctly, the messages must have been exchanged earlier, before line 104 completed. Why is getting these messages an async operation, then?

rkuhn (commented):
reviewed until tx_server_transport.rs

crates/amaru-consensus/src/consensus/effects/mempool_effects.rs
etorreborre: That's the size defined by the actual messages in the `pallas-network` crate and also the data type (`Word16`) chosen on the Haskell node and used in the serialization.

crates/amaru-consensus/src/consensus/tx_submission/tests/nodes.rs
etorreborre: This is necessary for the `move` part.

rkuhn: Oh, interesting one!

crates/amaru-consensus/src/consensus/effects/mempool_effects.rs
rkuhn: Might be good to document that; can it be zero? If not, NonZeroU16 is our friend.

crates/amaru-consensus/src/consensus/tx_submission/tests/tx_submission_client.rs
```diff
@@ -0,0 +1,311 @@
...
+    pub fn new(mempool: Arc<dyn TxSubmissionMempool<Tx>>, peer: &Peer) -> Self {
+        Self {
+            mempool: mempool.clone(),
+            state: TxSubmissionClientState::new(peer),
+        }
+    }
+
+    /// Core state machine, parameterized over a transport for testability.
+    pub async fn start_client_with_transport<T: TxClientTransport>(
+        &mut self,
```
rkuhn: My removing this `&` and requiring `T: Send + 'static` you could directly spawn the resulting future without wrapping in `async move { ….await }`

crates/amaru-consensus/src/consensus/tx_submission/tx_submission_client_state.rs
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    pub fn peer(&self) -> &Peer {
+        &self.peer
+    }
+
+    /// Process a request from the tx submission server
+    pub async fn process_tx_server_request<Tx: Send + Debug + Sync + 'static>(
+        &mut self,
+        mempool: Arc<dyn TxSubmissionMempool<Tx>>,
```
rkuhn: given that this Future won’t be `'static` anyway, why not take `&dyn TxSubmissionMempool<Tx>` here? Would save some atomic (i.e. synchronising) operations when calling this method.

crates/amaru-consensus/src/consensus/tx_submission/tests/tx_submission_client.rs
```diff
@@ -0,0 +1,311 @@
...
+                    debug!(peer = %self.state.peer(),
+                        "Error receiving next request, terminating tx submission client: {e:?}"
+                    );
+                    break;
+                }
+            };
+            match self
+                .state
+                .process_tx_server_request(
+                    self.mempool.clone(),
```
rkuhn: this would become `&self.mempool` with the suggestion below

crates/amaru-consensus/src/consensus/tx_submission/tests/tx_submission_client.rs
```diff
@@ -0,0 +1,311 @@
...
+                    to_pallas_request(self.state.peer(), request),
```
rkuhn: style: I prefer giving this a named binding one line above, so that the `match ... {` can be on a single line — it always takes me longer to read these long multi-line expressions

tangent: I’d prefer a wider formatting in general, perhaps I’ll play with rustfmt during a quite time at some point (the merge conflicts resulting from this can be a bit painful)

crates/amaru-consensus/src/consensus/tx_submission/tests/tx_submission_client.rs
```diff
@@ -0,0 +1,311 @@
...
+                }
+                TxClientResponse::NextIds(tx_ids) => {
+                    transport
+                        .reply_tx_ids(
+                            tx_ids
+                                .iter()
+                                .map(|(tx_id, tx_size)| {
+                                    TxIdAndSize(era_tx_id(tx_id.clone()), *tx_size)
+                                })
+                                .collect(),
```
rkuhn: I’d also move that into a preceding statement to make the reading flow match the execution.

crates/amaru-consensus/src/consensus/tx_submission/tests/tx_submission_server.rs
```diff
@@ -0,0 +1,259 @@
...
+        &mut self,
+        mut transport: T,
+    ) -> anyhow::Result<()> {
+        info!(peer = %self.state.peer(),
+            "Started tx submission server for peer {}",
+            self.state.peer()
+        );
+        transport.wait_for_init().await?;
+        let (ack, req, _blocking) = self.state.request_tx_ids(self.mempool.clone()).await?;
+        // The first request is always blocking
```
rkuhn: shouldn’t this come out of the mempool as well? feels strange to potentially override it

crates/amaru-consensus/src/consensus/tx_submission/tx_submission_client_state.rs
```diff
@@ -0,0 +1,170 @@
...
+                    return Ok(TxClientResponse::<Tx>::ProtocolError(anyhow!(
+                        "0 transaction ids requested"
+                    )));
+                }
+                // update the window by discarding acknowledged tx ids
+                // and update the last_seq
+                self.discard(ack);
+                if !mempool
+                    .wait_for_at_least(self.last_seq.unwrap_or_default().add(req as u64))
+                    .await
```
rkuhn, line 76: we’ll need to think about what to do when the peer disconnects while this wait is ongoing, but that’s probably better done when targeting the pure-stage networking stack

crates/amaru-consensus/src/consensus/tx_submission/tx_submission_client_state.rs
```diff
@@ -0,0 +1,170 @@
...
+                {
+                    return Ok(TxClientResponse::<Tx>::Done);
+                }
+                let tx_ids = self.get_next_tx_ids(mempool, req).await?;
+                Ok(TxClientResponse::NextIds(tx_ids))
+            }
+            TxServerRequest::TxIdsNonBlocking { ack, req, .. } => {
```
rkuhn, line 83: according to the spec we need to check that exactly one of `ack` and `req` is non-zero (which sounds stupid to me because I can well imagine setting both to positive values)

crates/amaru-consensus/src/consensus/tx_submission/tx_submission_client_state.rs
```diff
@@ -0,0 +1,170 @@
...
+                    )))
+                } else {
+                    Ok(TxClientResponse::NextTxs(txs))
+                }
+            }
+        }
+    }
+
+    /// Take notice of the acknowledged transactions, and send the next batch of tx ids.
+    async fn get_next_tx_ids<Tx: Send + Debug + Sync + 'static>(
```
rkuhn:
```suggestion
    fn get_next_tx_ids<Tx: Send + Debug + Sync + 'static>(
```

crates/amaru-consensus/src/consensus/tx_submission/tx_submission_server_state.rs
```diff
@@ -0,0 +1,189 @@
...
+
+    pub fn peer(&self) -> &Peer {
+        &self.peer
+    }
+
+    pub fn process_tx_ids_reply(
+        &mut self,
+        mempool: Arc<dyn TxSubmissionMempool<Tx>>,
+        tx_ids: Vec<(TxId, u32)>,
+    ) -> anyhow::Result<Option<Vec<TxId>>> {
```
rkuhn: Every `Option<Vec<…>>` should come with an explanation of the semantic difference between `None` and `Some(vec![])`.

crates/amaru-consensus/src/consensus/tx_submission/tx_submission_server_state.rs
```diff
@@ -0,0 +1,189 @@
...
+    pub async fn process_txs_reply(
+        &mut self,
+        mempool: Arc<dyn TxSubmissionMempool<Tx>>,
+        txs: Vec<Tx>,
+    ) -> anyhow::Result<(u16, u16, Blocking)> {
+        self.received_txs(mempool.clone(), txs).await?;
+        self.request_tx_ids(mempool.clone()).await
+    }
+
+    pub async fn request_tx_ids(
```
rkuhn:
```suggestion
    pub fn request_tx_ids(
```

crates/amaru-consensus/src/consensus/tx_submission/tx_submission_server_state.rs
```diff
@@ -0,0 +1,189 @@
...
+        }
+    }
+
+    pub fn peer(&self) -> &Peer {
+        &self.peer
+    }
+
+    pub fn process_tx_ids_reply(
+        &mut self,
+        mempool: Arc<dyn TxSubmissionMempool<Tx>>,
```
rkuhn: whole file: should use reference instead of `Arc<dyn …>`

crates/amaru-consensus/src/consensus/tx_submission/tx_submission_server_state.rs
```diff
@@ -0,0 +1,189 @@
...
+            // and the tx id is not already rejected.
+            if !mempool.contains(&tx_id) && !self.rejected.contains(&tx_id) {
+                self.pending_fetch.push_back(tx_id);
+            }
+        }
+
+        Ok(())
+    }
+
+    pub fn txs_to_request(&mut self) -> Option<Vec<TxId>> {
```
rkuhn: I’d probably return a slice of the `self.inflight_fetch_queue` instead of a fresh vector, and the empty slice when nothing new was added. The caller can then decide to do whatever it wants, including making a vector.

crates/amaru-consensus/src/consensus/tx_submission/tx_submission_server_state.rs
```diff
@@ -0,0 +1,189 @@
...
+        };
+        Ok((ack, req, blocking))
+    }
+
+    pub fn received_tx_ids<Tx: Send + Sync + 'static>(
+        &mut self,
+        mempool: Arc<dyn TxSubmissionMempool<Tx>>,
+        tx_ids: Vec<(TxId, u32)>,
+    ) -> anyhow::Result<()> {
+        if tx_ids.len() > self.params.max_window {
```
rkuhn: shouldn’t this be bounded by how many we requested earlier?

crates/amaru-consensus/src/consensus/tx_submission/tx_submission_server_state.rs
```diff
@@ -0,0 +1,189 @@
...
+            return Err(anyhow::anyhow!("Too many transactions ids received"));
+        }
+
+        for tx_id_and_size in tx_ids {
+            let (tx_id, size) = (tx_id_and_size.0, tx_id_and_size.1);
```
rkuhn:
```suggestion
        for (tx_id, size) in tx_ids {
```

crates/amaru-consensus/src/consensus/tx_submission/tx_submission_server_state.rs
```diff
@@ -0,0 +1,189 @@
...
+        let mut ack = 0_u16;
+
+        while let Some((tx_id, _size)) = self.window.front() {
+            let already_in_mempool = mempool.contains(tx_id);
+            let already_rejected = self.rejected.contains(tx_id);
+
+            if already_in_mempool || already_rejected {
+                // pop from window and ack it
+                if let Some((front_id, _)) = self.window.pop_front() {
+                    // keep rejected set from growing forever
```
rkuhn: Here `rejected` is only a cache for the current same-client window, meaning it won’t reject anything in realistic scenarios — this would be more useful if shared across client connections. Perhaps this should move into the mempool and be managed with a given max size. Here, we could also use probabilistic filters in the future (forgot the name of the current successor of Bloom).

crates/amaru-consensus/src/consensus/tx_submission/tx_submission_server_state.rs
```diff
@@ -0,0 +1,189 @@
...
+                    if already_rejected {
+                        self.rejected.remove(&front_id);
+                    }
+                    ack = ack.saturating_add(1);
```
rkuhn: shouldn’t we stop the loop upon saturating the u16? otherwise the internal state and the network state diverge

crates/amaru-consensus/src/consensus/tx_submission/tx_submission_server_state.rs
```diff
@@ -0,0 +1,189 @@
...
+
+        // Request as many as we can fit in the window.
+        // Note: we cap at u16::MAX because the protocol uses u16 for counts.
+        let req = self
+            .params
+            .max_window
+            .saturating_sub(self.window.len())
+            .min(u16::MAX as usize) as u16;
+
+        // We need to block if there are no more outstanding tx ids.
```
rkuhn, line 105: this also nicely covers the initial case :-)

crates/amaru-consensus/src/consensus/tx_submission/tx_submission_server_state.rs
```diff
@@ -0,0 +1,189 @@
...
+                self.inflight_fetch_set.remove(&requested_id);
+
+                let inserted = mempool.validate_transaction(&tx).is_ok()
+                    && mempool
+                        .insert(tx, TxOrigin::Remote(self.peer.clone()))
+                        .is_ok();
+                if !inserted {
+                    self.rejected.insert(requested_id);
+                }
+            }
```
rkuhn, line 172: This needs an else branch or a good justification or a more fitting test before the loop. In fact, I’d tear the connection down if this come up empty, just in case.

crates/amaru-consensus/src/consensus/tx_submission/tx_submission_server_state.rs
```diff
@@ -0,0 +1,189 @@
...
+                ids.push(id);
+            } else {
+                break;
+            }
+        }
+
+        if ids.is_empty() { None } else { Some(ids) }
+    }
+
+    pub async fn received_txs(
```
rkuhn:
```suggestion
    pub fn received_txs(
```

crates/amaru-mempool/Cargo.toml
```diff
@@ -11,4 +11,14 @@ documentation.workspace = true
...
+amaru-kernel.workspace = true
+amaru-ouroboros-traits.workspace = true
+minicbor.workspace = true
+parking_lot.workspace = true
+tokio.workspace = true
+
+[dev-dependencies]
+anyhow.workspace = true
+assertables.workspace = true
+minicbor.workspace = true
```
rkuhn: duplicated

crates/amaru-mempool/Cargo.toml
```diff
@@ -11,4 +11,14 @@ documentation.workspace = true
...
+tokio = { "workspace" = true, features = ["time", "macros", "rt"] }
```
rkuhn: I’d add those features in the `[dependencies]` section because they’ll be built for production anyway.

crates/amaru-mempool/src/strategies/in_memory_mempool.rs
```diff
@@ -0,0 +1,344 @@
...
+
+    pub fn from_config(config: MempoolConfig) -> Self {
+        Self::new(config, Arc::new(DefaultCanValidateTransactions))
+    }
+}
+
+/// A default transactions validator.
+#[derive(Clone, Debug, Default)]
+pub struct DefaultCanValidateTransactions;
```
rkuhn, line 64: This highly dangerous code should be named more scarily, e.g. `AcceptAnyTransactionWhatsoever`.

crates/amaru-mempool/src/strategies/in_memory_mempool.rs
```diff
@@ -0,0 +1,344 @@
...
+            .map(|(seq, tx_id)| {
+                if let Some(entry) = self.entries_by_id.get(tx_id) {
+                    let tx_size = to_cbor(&entry.tx).len() as u32;
+                    (tx_id.clone(), tx_size, *seq)
+                } else {
+                    panic!(
+                        "Inconsistent mempool state: entry missing for tx_id {:?}",
+                        tx_id
+                    );
+                }
```
rkuhn: I’ve come to like let-else syntax because it reduces indentation:
```rust
let Some(entry) = … else { panic!(); };
do_stuff(entry);
```

crates/amaru-mempool/src/strategies/in_memory_mempool.rs
```diff
@@ -0,0 +1,344 @@
...
+
+    /// Retrieves all the transaction ids since a given sequence number, up to a limit.
+    fn tx_ids_since(&self, from_seq: MempoolSeqNo, limit: u16) -> Vec<(TxId, u32, MempoolSeqNo)> {
+        let mut result: Vec<(TxId, u32, MempoolSeqNo)> = self
+            .entries_by_seq
+            .range(from_seq..)
+            .take(limit as usize)
+            .map(|(seq, tx_id)| {
+                if let Some(entry) = self.entries_by_id.get(tx_id) {
+                    let tx_size = to_cbor(&entry.tx).len() as u32;
```
rkuhn: even if this uses the serialization buffer to avoid allocations, how about storing the size in the map entry as well?

crates/amaru-mempool/src/strategies/in_memory_mempool.rs
```diff
@@ -0,0 +1,344 @@
...
+            .collect()
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub struct MempoolEntry<Tx> {
+    seq_no: MempoolSeqNo,
+    tx_id: TxId,
+    tx: Tx,
+    origin: TxOrigin,
```
rkuhn, line 175: could add `size` here

crates/amaru-consensus/src/consensus/tx_submission/tx_submission_server_state.rs
```diff
@@ -0,0 +1,189 @@
...
+    ) -> anyhow::Result<()> {
+        if txs.len() > self.params.fetch_batch {
+            return Err(anyhow::anyhow!(
+                "Too many transactions received in one batch"
+            ));
+        }
+
+        for tx in txs {
+            // this is the exact id we requested for this body (FIFO)
+            if let Some(requested_id) = self.inflight_fetch_queue.pop_front() {
```
rkuhn, line 165: should also validate that the size matches the advertisement (and kick the peer if not)

crates/amaru-mempool/src/strategies/in_memory_mempool.rs
```diff
@@ -0,0 +1,344 @@
...
+    }
+
+    fn tx_ids_since(&self, from_seq: MempoolSeqNo, limit: u16) -> Vec<(TxId, u32, MempoolSeqNo)> {
+        self.inner.read().tx_ids_since(from_seq, limit)
+    }
+
+    fn wait_for_at_least(
+        &self,
+        seq_no: MempoolSeqNo,
+    ) -> Pin<Box<dyn Future<Output = bool> + Send + '_>> {
```
rkuhn, line 219: since this isn’t `'static` anyway, and we’re on `nightly` anyway, we might as well return `impl Future<...>` here; simpler and less boxing

crates/amaru-mempool/src/strategies/in_memory_mempool.rs
```diff
@@ -0,0 +1,344 @@
...
+        self.inner.read().get_txs_for_ids(ids)
+    }
+
+    fn last_seq_no(&self) -> MempoolSeqNo {
+        MempoolSeqNo(self.inner.read().next_seq - 1)
+    }
+}
+
+impl<Tx: Send + Sync + 'static + Encode<()> + Clone> Mempool<Tx> for InMemoryMempool<Tx> {
+    fn take(&self) -> Vec<Tx> {
```
rkuhn, line 252: rough  —  is this for testing? would be good to have some API docs in this file

crates/amaru-network/src/network_resource.rs
```diff
@@ -12,101 +12,226 @@
...
 #[derive(Clone)]
 pub struct NetworkResource {
-    inner: Arc<NetworkInner>,
+    upstream_client: Arc<UpstreamClient>,
+    downstream_server: Arc<DownstreamServer>,
 }
 
 impl Deref for NetworkResource {
-    type Target = NetworkInner;
+    type Target = UpstreamClient;
```
rkuhn: in general, this Deref should go because there are two viable targets, but I see this as a simplification of working on this while the real network stack isn’t there yet, right?

crates/amaru-network/src/server/client_protocol.rs
```diff
@@ -344,16 +318,66 @@ async fn block_fetch<H: IsHeader>(
...
+    peer: Peer,
+    tx_client_reply_sender: Sender<TxClientReply>,
 ) -> anyhow::Result<()> {
     server.wait_for_init().await?;
-
-    // TODO: Implement tx submission
-
+    loop {
+        select! {
+            reply = server.receive_next_reply() => {
```
rkuhn, line 334: this requires `receive_next_reply()` to be cancellation safe — is that the case? if not, this Future should be held in a mutable variable and `select!` should only use `&mut receive_next`

crates/amaru-network/src/server/client_protocol.rs
```diff
@@ -344,16 +318,66 @@ async fn block_fetch<H: IsHeader>(
...
+                            .await?
+                    },
+                    ActoInput::Message(TxSubmissionMsg::Op(TxServerRequest::Txs { tx_ids, .. })) => {
+                        server
+                            .request_txs(tx_ids.into_iter().map(era_tx_id).collect())
+                            .await?
+                    },
+                    ActoInput::NoMoreSenders | ActoInput::Supervision { .. }  => {
+                        tracing::debug!("parent terminated");
+                        break;
```
rkuhn: I’d use `let ActoInput::Message(message) = message else { break; };` in front of the match to reduce visual clutter.

rkuhn (commented):
The acto parts were just moved/copied from the earlier client impl, correct?

crates/amaru-consensus/src/consensus/effects/mempool_effects.rs
etorreborre: Done.

crates/amaru-consensus/src/consensus/tx_submission/tests/tx_submission_client.rs
etorreborre: Works fine indeed.

crates/amaru-consensus/src/consensus/tx_submission/tx_submission_client_state.rs
etorreborre: Works fine.

crates/amaru-consensus/src/consensus/tx_submission/tests/tx_submission_client.rs
etorreborre: Done.

crates/amaru-consensus/src/consensus/tx_submission/tx_submission_client_state.rs
etorreborre: Indeed

crates/amaru-mempool/src/strategies/in_memory_mempool.rs
etorreborre: This was really a stand-in for some future implementation. This will be necessarily modified as we implemented the real logic.

crates/amaru-consensus/src/consensus/tx_submission/tests/tx_submission_server.rs
etorreborre: yes, I fixed it.

crates/amaru-consensus/src/consensus/tx_submission/tx_submission_client_state.rs
etorreborre: I'm planning to come back to those details after this PR.

crates/amaru-consensus/src/consensus/tx_submission/tx_submission_server_state.rs
etorreborre: Right. That's not necessary, I removed the whole Option thing.

crates/amaru-consensus/src/consensus/tx_submission/tx_submission_server_state.rs
etorreborre: Eventually this will always turn into a vector so I don't really what we gain?

crates/amaru-consensus/src/consensus/tx_submission/tx_submission_server_state.rs
etorreborre: Yes but that's a sanity check against badly behaved clients.

crates/amaru-consensus/src/consensus/tx_submission/tx_submission_server_state.rs
etorreborre: I agree. Let's come back to this when we really tackle the mempool logic.

crates/amaru-consensus/src/consensus/tx_submission/tx_submission_server_state.rs
etorreborre: That should actually never happen because the window size is very small. I'll come back to that code when making the next pass on the spec + the current haskell code.

crates/amaru-consensus/src/consensus/tx_submission/tx_submission_server_state.rs
etorreborre: I'd leave this kind of validation rework to a next pass. It was over-zealous to try to even start this. I've removed the rejection code for now.

crates/amaru-mempool/src/strategies/in_memory_mempool.rs
etorreborre: Yes that's better

crates/amaru-mempool/src/strategies/in_memory_mempool.rs
etorreborre: Done.

crates/amaru-consensus/src/consensus/tx_submission/tx_submission_server_state.rs
etorreborre: I'll check that when coming back to the protocol rules.

crates/amaru-mempool/src/strategies/in_memory_mempool.rs
etorreborre: If we do this then the `TxSubmissionMempool` trait is not dyn-compatible

crates/amaru-mempool/src/strategies/in_memory_mempool.rs
etorreborre: This code mirrors the code that was sketched in the original `DummyMempool`. We need to come to that when we start implementing the Mempool for real.

crates/amaru-network/src/network_resource.rs
etorreborre: I removed that `Deref`, it doesn't make sense with the evolution of `NetworkResource` now.

crates/amaru-network/src/server/client_protocol.rs
etorreborre: Good idea.

crates/amaru-network/src/server/client_protocol.rs
etorreborre: Yes it is cancellation-safe.

PR #461, opened 2025-09-17
feat: test on env with low memory available

jeluard:
Provides foundation to run amaru on simulated constrained environment.

## Summary by CodeRabbit

* **Tests**
  * Added an on-demand “Constrained Environments Testing” workflow to run builds/tests under limited resources.
  * Uses caching to speed setup and includes fail-fast checks when prerequisites are unavailable.

* **Chores**
  * Automated CI environment setup and cleanup to improve consistency and observability of runs.

PR #594, opened 2025-12-09
perf: vec over btree

Eh2406:
`votes` is the single largest allocation at peak memory usage. There are structural changes that should fix this problem, but for now it hasn't been prioritized. Some [back of the envelope math](https://discord.com/channels/1202416088776712253/1437818120830976112/1438675919521972225) suggests this "ALL votes" known issue should take about ~1GB. This is also confirmed by switching to an alternative allocator and seeing the maximum memory usage be <2GB.  So why is it taking comfortably over 3GB on preprod when talking to a fast peer? In poking around the code to try and understand what refactors were possible, I made a small experimental change that seems to have "fixed" it on my machine. With this change the default allocator has the same top memory as `jemalloc` at peek. However, I'm not really sure why it works. Or rather, I have many theories about why it could work, but none of them seem to be justified by the code.

- `BTrees` have a reasonably constant 50% memory overhead. `Vec` over allocates to amortize the cost of allocations. When it's full, it has 0% overhead. Just after it reallocates, it has 100% overhead.       On average they should be about the same. It could be we're just getting lucky. However, this PR reduces maximum memory usage by >1.5GB, but the data itself is 500MB.  So the math does not work there. The savings are just too big.
- `Vec` can use `Iterator::size_hint` do one allocation up front. This would guarantee us the lucky case. But, being lucky is insufficient to explain the problem. AND We are repeatedly using `push` so there is no place for the code to sneak in a `with_capacity`. (Although we could manually do it easily enough, if we want to do follow up micro optimizations.)
- `BTrees` consist of lots of tiny allocations. This provides ample opportunity for memory fragmentation. Which tends to happen when there are a lot of allocations going on with widely different life times. Amos explains this better than I can in https://www.youtube.com/watch?v=YB6LTaGRQJg and https://www.youtube.com/watch?v=DpnXaNkM9_M. Alternating between inserting into a BTree and constructing data that will long outlive the BTree, is a setup for the data being left scattered all across the memory space, preventing the allocator from freeing the memory. Unfortunately, all the allocations I can track that occurred during the lifespan of this BTree are either extremely short lived (`Box::new` in `rocksdb`) or end up owned by the BTree. Also, empirically fixing those other allocations has no effect on the memory usage. So while this is still my leading theory, I do not have a smoking gun to point to to explain the connection.

## Summary by CodeRabbit

* **Refactor**
  * Changed how governance ratification votes are stored: per-proposal votes are now stored as ordered lists instead of nested maps.
  * Updated internal vote-processing logic to operate on these per-proposal lists (call sites adapted accordingly).
  * No user-facing behavior changes; improves internal maintainability and processing consistency.

### DIFF
--- a/crates/amaru-ledger/src/governance/ratification.rs
+++ b/crates/amaru-ledger/src/governance/ratification.rs
@@ -66,7 +66,7 @@ pub struct RatificationContext<'distr> {
     pub constitutional_committee: Option<ConstitutionalCommittee>,
 
     /// All latest votes indexed by proposals and voters.
-    pub votes: BTreeMap<ComparableProposalId, BTreeMap<Voter, Ballot>>,
+    pub votes: BTreeMap<ComparableProposalId, Vec<(Voter, Ballot)>>,
 }
 
 #[derive(Debug, thiserror::Error)]
@@ -394,9 +394,8 @@ impl<'distr> RatificationContext<'distr> {
     ) -> bool {
         let span = tracing::Span::current();
 
-        let empty = BTreeMap::new();
         let (dreps_votes, cc_votes, pool_votes) =
-            partition_votes(self.votes.get(id).unwrap_or(&empty));
+            partition_votes(self.votes.get(id).map(|v| v.as_slice()).unwrap_or(&[]));
 
         // NOTE: because ratification is an expensive operation, and something that we *may* have
         // to replay due to rollbacks, it's important to do the least amount of work possible.
@@ -519,7 +518,7 @@ impl<'distr> RatificationContext<'distr> {
 /// Split all the ballots into sub-maps that are specific to each voter types; so that we ease the
 /// processing of each category down the line.
 fn partition_votes(
-    votes: &BTreeMap<Voter, Ballot>,
+    votes: &[(Voter, Ballot)],
 ) -> (
     BTreeMap<DRep, &Vote>,
     BTreeMap<StakeCredential, &Vote>,
--- a/crates/amaru-ledger/src/state.rs
+++ b/crates/amaru-ledger/src/state.rs
@@ -49,7 +49,7 @@ use anyhow::{Context, anyhow};
 use std::{
     borrow::Cow,
     cmp::max,
-    collections::{BTreeMap, BTreeSet, VecDeque, btree_map},
+    collections::{BTreeMap, BTreeSet, VecDeque},
     ops::Deref,
     sync::{Arc, Mutex, MutexGuard},
 };
@@ -1021,14 +1021,10 @@ fn new_ratification_context<'distr>(
     let votes = snapshot
         .iter_votes()?
         .fold(BTreeMap::new(), |mut votes, (k, v)| {
-            match votes.entry(k.proposal) {
-                btree_map::Entry::Vacant(entry) => {
-                    entry.insert(BTreeMap::from([(k.voter, v)]));
-                }
-                btree_map::Entry::Occupied(mut entry) => {
-                    entry.get_mut().insert(k.voter, v);
-                }
-            }
+            votes
+                .entry(k.proposal)
+                .or_insert_with(Vec::new)
+                .push((k.voter, v));
 
             votes
         });


jeluard:
> I made a small experimental change that seems to have "fixed" it on my machine

I'd love that we can be more specific and track those things more deterministically and have some ways to track improvements/regressions. 
Also I guess we are looking for different algos that would allow to load less data, like introducing streaming. Here we have the same amount of data stored ultimately right? So unclear how mem usage can go down drastically and I would expect that at least at on moment the top mem usage is roughly the same?

Finally I think we need to make some concrete test run by limiting the mem usage and constrained env e.g. using `cgroup`. See https://github.com/pragma-org/amaru/pull/461

Eh2406:
> Here we have the same amount of data stored ultimately right? So unclear how mem usage can go down drastically and I would expect that at least at on moment the top mem usage is roughly the same?

This is a reasonable expectation to have. For most problems, interacting with this level of abstraction is effective and efficient. Most of the tools available for systematically tracking and measuring memory usage would agree with you that this PR make no difference. They would all agree that the program uses <2GB. (Just the way it does with `jemalloc`.) But that is not what is being reported by the tools like `time` and `top`. I suspect, but have not tested that, if you ran this inside `cgroup` limitling it memory to 2GB, the program would crash, even though the tools say that it used far less than that. And more generally, if the memory usage of the program is just the amount of data that is stored, then how could changing the allocator make any difference at all?

Unfortunately, there is another layer of abstraction leaking all over this reasonable expectation. Our program is talking through the "allocator API" to another part of our program we call the allocator, which itself is talking to the operating system through a series of syscalls. Again Amos explains this better than I can. The tools and the reasonable expectation are thinking about this in terms of the "allocator API".  You can think about the memory usage by adding up all the calls to `malloc` and subtracting all the calls to `free`. You can do this automatically by overriding Rust's allocator, like `DHAT-rs`/`jemalloc` does, or by using `LD_PRELOAD` to change the implementation of `malloc` and `free`, like `memusagestat`/`bytehound` does. Most of the time this understanding is good enough, but sometimes the underlying abstraction is leaky.

The operating system is responsible for a lot of fiddly details every time it hands out some memory. It needs to find a physical spot on a memory chip to store the data. It needs to make sure that if any other program tries to access that piece of memory a Exception gets thrown. And it needs to do some renumbering magic to make sure that the program sees the entire block of memory as contiguous, even if they're on different physical memory chips. There's probably a lot more I'm not aware of. TLDR, The OS cannot afford to deal with tiny allocations. It hands them out in blocks, even if that means giving you more than you need.

Let's say we had a program that makes ten `Box<[u8; ((block size) / 2) + 1]>`. Each time A new box is created, A new allocation is required, and the OS is going to give us an entire block. How much memory program use? Our intuition, and most of the tools, says `5 * block size + 10`. But `top` reports that it takes `10 * block size`. Basically it takes twice as much as the intuition! 

A smarter allocator might notice the pattern and ask the operating system for more blocks thanit strictly needs assuming that new requests will come in. But this is also a risky maneuver. If those requests do not come in, you can end up with even higher rates of wasted space.

Let's say we extend our program to also Store ten Strings of length `(block size) / 4`.  If our allocator is smart enough, it will notice that there is room in the Blocs it already received for the strings to fit in. (Actually, noticing this requires a lot of bookkeeping on its own, which might not be worth it. So different allocators use different heuristics for deciding how precisely to keep track of things.) Our tools now say `7.5 * block size + 10`, and `top` still reports `10 * block size`. 75% utilized. Wonderful!

Now our program freeze the boxes and continues to do some complicated work. How much memory is the program using? The tools report that all of the boxes have been freed, so there are only the 10 strings left. So `2.5 * block size`. But each block has one string in it, because we carefully reused the space after the boxes. So `top` still reports that all ten boxes are used. That is 4x!

So there's six hundred and something words explaining how this kind of problem is even theoretically possible. But I do not have a smoking gun proving how our code falls into one of these problematic patterns.

Eh2406:
Here's the way I set up the test.
```cmd
rm -rf *.preprod.db
cargo run --profile profiling -- bootstrap --network=preprod
sudo mkdir -p /sys/fs/cgroup/limited
echo $((0)) | sudo tee /sys/fs/cgroup/limited/memory.swap.max
echo $((MAX_MEM)) | sudo tee /sys/fs/cgroup/limited/memory.max
sudo cgexec -g memory:limited target/profiling/amaru run --network=preprod --peer-address 127.0.0.1:3001
```
I stopped the experiment as a success if we got to epoch 250, otherwise I recorded the epoch at which the program was killed. Checking out the parent commit. It only got to `epoch=212` when `MAX_MEM` was `5000000000`. It was a success when `MAX_MEM` was `6000000000`.  (If we want more specificity, it would not be hard to continue doing binary search.)  With this PR checked out. It only got to `epoch=189` when `MAX_MEM` was `1572864000`. It was a success when `MAX_MEM` was `1835008000`. (Same offer available about binary search.) So a 2GB limit (on my computer doing this benchmark, with the allocator provided by my OS, ect.) works with the PR and does not work on the parent commit.

Eh2406:
It's hard to tell with all the layers of abstraction, but I think that is just pointers. I don't think it allocates anything on the heap. But when dealing with buggy code, it could be anything.

jeluard:
@Eh2406 Please make sure that your commit is signed :)
Also please rebase so that all tests can pass.

Eh2406:
rebased and signed.

PR #601, opened 2025-12-13
feat: get rid on anchor

jeluard:
Get rid of useless `anchor` as it introduces extra memory usage and processing time.

⚠️ This will break the ledger db. It is proposed not to implement any migration for this as the cost would be significant while we are till in development phase.

## Summary by CodeRabbit

* **Refactor**
  * Removed anchor metadata across governance flows (voting, registrations, resignations); simplified ballots, DRep state/rows, and related APIs.
* **Documentation**
  * Ledger docs updated to reflect the reduced DRep schema (anchor removed).
* **Tests**
  * Updated test data, generators, and assertions to omit anchor.
* **UI/Display**
  * Constitution display no longer shows the prior anchor URL.

### DIFF
--- a/crates/amaru-kernel/src/ballot.rs
+++ b/crates/amaru-kernel/src/ballot.rs
@@ -12,12 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crate::{Anchor, Vote, cbor, heterogeneous_array};
+use crate::{Vote, cbor, heterogeneous_array};
 
 #[derive(Clone, Debug, PartialEq)]
 pub struct Ballot {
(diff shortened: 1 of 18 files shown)


KtorZ:
I am not 100% sure about it. 

- It's potentially useful information to end-users. Though it's recoverable from block; it might get cumbersome. 

- I am not against removing them per se, though. It is true that for the sake of block production, they don't really add much. 

- We need to double-check that there aren't any ledger rules that depend on it; from the top of my head, that is indeed the case.

All-in-all, anchors only became a "problem" because we don't prune votes and fetch them all in memory. Perhaps there's more to be done on that front?

PR #602, opened 2025-12-15
feat: faster release pipeline

jeluard:
Simplify release profile, cuts build times in 2 without significant downsides.

## Summary by CodeRabbit

* **Chores**
  * Adjusted build/release configuration to speed up development compilation by changing optimization settings; profiling build remains available and unchanged.
  * No changes to public APIs or exported declarations.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -156,10 +156,6 @@ inherits = "dev"
 opt-level = 0
 debug = true
 
-[profile.release]
-lto = true
-codegen-units = 1
-
 [profile.profiling]
 inherits = "release"
 debug = true


Cargo.toml
```diff
@@ -156,10 +156,6 @@ inherits = "dev"
 opt-level = 0
 debug = true
 
-[profile.release]
-lto = true
-codegen-units = 1
```
KtorZ, line 161: That is disabling potentially interesting optimisation, though? We don't really have good benchmarks in place at this point to guide the decision but given the significant differences we see between release and dev builds at the moment, I would be quite careful about disabling compiler optimisation options...

Perhaps we can try `thin` for `lto`, for the Rust book says the following:

- `thin`: perform ["thin" LTO](http://blog.llvm.org/2016/06/thinlto-scalable-and-incremental-lto.html). This is similar to "fat", but takes substantially less time to run while still achieving performance gains similar to "fat". For larger projects like the Rust compiler, ThinLTO can even result in better performance than fat LTO.

- Not sure about codegen-units; it says that higher number means more parallelisation but potentially slower code 🤷... 

Beside, I am not sure why having some _relatively long_ release build time is a big problem? The `dev` profile is already performant-enough for most (all?) tasks, isn't it? Perhaps we can have an intermediate profile between release and dev that's _more optimised_ than `dev` but not as much as `release`? 

@etorreborre @rkuhn, any thoughts? I reckon this change was sparked by some of the recent conversations regarding Eric's long compilation time on his branch?

jeluard: Right my intention was to stick with `thin` indeed.
Now that said I feel that we should have a more concrete way of figuring out if those options are actually having an impact on performance. Before moving way from defaults. Especially given that rust moves fast and others changes might conflict with those (e.g. using `jemalloc`).

rkuhn: I'm not yet at the stage where I'd use the release build, from my perspective we need quick turnaround times for local tests as well as CI. I think the test and dev-debug profiles are good enough for those purposes (and I maintain that the names are wrong in that `cargo test` should use a profile that does not mangle backtraces, but that is just me </rant>).

PR #603, opened 2025-12-16
feat: add handshake protocol handler

rkuhn:
also add contramap for StageRefs

### DIFF
--- a/simulation/amaru-sim/tests/traces.rs
+++ b/simulation/amaru-sim/tests/traces.rs
@@ -96,35 +96,62 @@ fn run_simulator_with_traces() {
         execute,
         vec![json!(
         {
-         "name": "handle_msg",
-         "children": [
-           {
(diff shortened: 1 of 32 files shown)


simulation/amaru-sim/tests/traces.rs
```diff
@@ -96,35 +96,62 @@ fn run_simulator_with_traces() {
...
-           },
-           {
-             "name": "chain_sync.select_chain"
-           },
-           {
-             "name": "diffusion.forward_chain"
-           }
-         ]
+          "name": "handle_msg",
+          "children": [
```
etorreborre, line 100: Why has this output changed with this PR?

rkuhn: I have absolutely no idea whatsoever. I made the test pass on my machine and am waiting what will happen on gh actions.

PR #614, opened 2025-12-19
Add modify_proposals which yields a proposals stream to the caller

geofflittle:
See #605.

This example implementation demonstrates a proposed Store design change.

By lifting the closure level to yield the whole stream instead of individual items, we allow:
* The primary object being modified/written to be streamed async'ly
* The secondary object to be async'ly looked-up (e.g., reading pools when validating accounts) rather than pre-loading tables
* The caller to drive the iteration, allowing for early exits/breaks to save resources
* Maintaining the implicit, atomic save/commit guarantee of the original design


crates/amaru-ledger/src/state.rs
```diff
@@ -984,6 +985,89 @@ pub fn tick_proposals<'store>(
...
+    // The closure returns the calculations
+    let (refunds, still_active) = db
+        .modify_proposals(move |db_ref, mut stream| {
+            async move {
+                // No Arc or Mutex or clone for aggregation across awaits
+                let mut refunds: BTreeMap<StakeCredential, Lovelace> = BTreeMap::new();
+                let mut still_active = 0;
+
+                // The caller drives the iteration and gets flexibility to short-circuit / break early
+                // and we keep the implicit atomic buffer write
```
KtorZ, line 1031: This seems to be the main argument here: the flexibility and control-flow is given back to the caller. While I agree that it can be nice in certain scenarios; I don't really think we have such scenarios? Updates we have to perform are either on specific entities (single read/write), or entire collections. 

Yet, I can totally imagine a store API where we provide _both_. The suggested approach here has multiple drawbacks though:

1. It's async. RocksDB isn't capable of doing async operations like this on a single connection. It can do async I/O behind the scene for seek and next operations when explicitly enabled, but not so much for transactional writes. So there's not much benefits in introducing asynchronicity here. It is even annoying because we loose even more determinism and we typically force callers to also be async. At the very least, while the core iterator callback could be async for flexiblity; the outer call to .modify_proposals shouldn't, and we shall use a channel/queue to collect and await results. 

2. I don't think this is abstracting over the right thing. Rather than reworking the existing methods as such, it would make more sense to abstract over the _behaviour_. So instead of a `iter_utxos`, `iter_proposals`, `modify_proposals`, etc... we could rely on a `iter_collection` and `modify_collection`, which can be parameterized by a collection type (then leveraging traits to provide decoders, partial decoders, column prefixes, etc...). This way, we reduce boilerplate significantly, and we make the API way more capable by providing a variety of access with different trade-offs.

PR #625, opened 2026-01-02
Use get_pinned for getting data from rocksdb

Eh2406:
A large number of short lived allocations come from copying data out of rocksdb. The `get` and `iterator_opt` methods copy the data out of the database into a temporary `Vec | Box` so that the Rust code can choose to keep the data as long as it wants. In our case, we immediately decode the data into more specific types, and then drop the `Vec | Box`. So it would be more efficient for us to generate the specific types directly from the underlying buffer held by the database.

The `get` can be replaced with a `get_pinned`. This requires changes in a number of places, but they're pretty mechanical.

Unfortunately, there is no standard idiom for a iterator whose items can only be used until the next item is generated. The general problem is known as "lending iterator" and was one of the known limitations when Rust 1.0 was released. Rocksdb sorts this out by providing a raw interface to its iterators. We can write our own loop to read the raw data out of the database and convert it to our specific type. Once we've done that conversion we no longer have a lending problem and can construct a custom iterator type. `std::iter::from_fn` massively reduces the boilerplate.

These kinds of short lived allocations sometimes anger allocators. This one was particularly suspicious because it occurred at the same time as the construction of the big `votes` container disgust in the #594. Having now made the these changes and tested the result, no connection. These changes do not make a measurable difference to peak memory usage or throughput. :-( It does massively reduce the number of allocations and thereby the overhead of using memory profiling tools.

## Summary by CodeRabbit

* **Refactor**
  * Optimized database reads to use pinned, borrowed data to reduce allocations and improve performance.
  * Adjusted decoding to operate on borrowed data while preserving behavior and error handling.
  * Simplified iteration API for more efficient, on-demand decoding during iteration.

### DIFF
--- a/crates/amaru-ledger/src/store/columns/mod.rs
+++ b/crates/amaru-ledger/src/store/columns/mod.rs
@@ -27,11 +27,11 @@ pub mod votes;
 
 #[expect(clippy::panic)]
 pub fn unsafe_decode<T: for<'d> cbor::Decode<'d, ()>>(bytes: &[u8]) -> T {
-    cbor::decode(&bytes).unwrap_or_else(|e| {
+    cbor::decode(bytes).unwrap_or_else(|e| {
         panic!(
             "unable to decode {} from CBOR ({}): {e:?}",
             std::any::type_name::<T>(),
-            hex::encode(&bytes)
+            hex::encode(bytes)
         )
     })
 }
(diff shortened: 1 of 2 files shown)


Eh2406:
Looked into the AI nitpick. Yes, I missed `pools.rs`. But it just passes the buffer to `Row::extend` witch needs a `Vec`, so no point in using `get_pinned`.

I made clippy and fmt happy and rebased. I don't see how my changes are related to the CI `Error response from daemon: No such container: cardano-node` 🤷 

jeluard, That is a very helpful list of rocksdb parameters. Do you want me to play with them in this PR, or can it be a follow up?

jeluard:
@Eh2406 There are some clippy issues to fix

PR #628, opened 2026-01-08
pedantic conway txbody

rrruko:
## Summary by CodeRabbit

* **New Features**
  * Added richer block/transaction/auxiliary data support with CBOR encode/decode, hashing accessors and iteration semantics
  * New validated, non-empty and ordered collection types for key/value and set data
  * Added a non-panicking CBOR include macro

* **Refactor**
  * Replaced legacy "minted" representations with streamlined public types and simplified APIs

* **Tests**
  * Added extensive decoding and round-trip test vectors and tooling for CBOR fixtures

### DIFF
--- a/crates/amaru-kernel/src/lib.rs
+++ b/crates/amaru-kernel/src/lib.rs
@@ -49,15 +49,15 @@ pub use pallas_primitives::{
         AddrKeyhash, AssetName, BigInt, BootstrapWitness, Certificate, Coin, Constitution, Constr,
         CostModel, CostModels, DRep, DRepVotingThresholds, DatumHash, DatumOption, ExUnitPrices,
         ExUnits, GovAction, GovActionId as ProposalId, HeaderBody, KeepRaw, Language,
-        MaybeIndefArray, Mint, MintedDatumOption, MintedScriptRef, MintedTransactionOutput,
-        Multiasset, NativeScript, NetworkId, NonEmptyKeyValuePairs as PallasNonEmptyKeyValuePairs,
-        NonZeroInt, PlutusData, PlutusScript, PolicyId, PoolMetadata, PoolVotingThresholds,
-        PositiveCoin, PostAlonzoTransactionOutput, ProposalProcedure as Proposal,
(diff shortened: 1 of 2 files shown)


crates/amaru-kernel/src/non_empty_key_value_pairs.rs
```diff
@@ -0,0 +1,219 @@
...
+        }
+    }
+}
+
+impl<K, V> TryFrom<Vec<(K, V)>> for NonEmptyKeyValuePairs<K, V>
+where
+    K: Clone,
+    V: Clone,
+{
+    type Error = String;
```
SupernaviX: I'm not familiar with amaru error conventions, but `String` as error is a bit awkward to work with (callers have to convert it to their error type, and it forces an allocation). I think a better alternative would be to define an error type which implemented `Error` (like https://doc.rust-lang.org/src/core/num/error.rs.html#10)

jeluard: See the error EDR: https://github.com/pragma-org/amaru/blob/main/engineering-decision-records/013-error-handling-strategies.md

crates/amaru-kernel/src/transaction_body.rs
```diff
@@ -0,0 +1,388 @@
...
+
+        Ok(TransactionBody {
+            hash: Hasher::<{ TransactionBody::HASH_SIZE * 8 }>::hash(
+                &original_bytes[start_position..end_position],
+            ),
+            original_size: (end_position - start_position) as u64, // from usize
+            inputs: expect_field(mem::take(&mut state.required.inputs), 0, "inputs")?,
+            outputs: expect_field(mem::take(&mut state.required.outputs), 1, "outputs")?,
+            fee: expect_field(mem::take(&mut state.required.fee), 2, "fee")?,
+            ..state.optional
```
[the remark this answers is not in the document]
KtorZ: You're wrong here; empty inputs and empty outputs are allowed. What's not allowed is for the fields to be missing; which is already enforced by the decoder here.

PR #637, opened 2026-01-19
downstream connections and integration tests

rkuhn:
This PR:

 - Adds capabilities for creating TCP listeners and accepting connections.
 - Provides those capabilities to the `Manager` as `pure-stage` effects.
 - Adds a test showing the execution of miniprotocols for 2 `Managers`, one acting as an initiator and another one as a responder, over TCP. The test then checks that the 2 "nodes" eventually get the same best chain, via the chainsync protocol, and the same transactions via the txsubmission protocol (the blockfetch protocol will be tested next).

## Release Notes

* **New Features**
  * Added server-side connection acceptance with listening capability and socket reuse options.
  * Improved connection lifecycle management with peer address tracking.
  * Enhanced protocol handling for initiator and responder roles.

* **Tests**
  * Added end-to-end integration tests for protocol interaction verification.

* **Chores**
  * Updated dependencies and logging levels for better diagnostics.


crates/amaru-network/src/connection.rs
```diff
@@ -14,26 +14,50 @@
...
+    peer_addr: SocketAddr,
     reader: Arc<AsyncMutex<(OwnedReadHalf, BytesMut)>>,
     writer: Arc<AsyncMutex<OwnedWriteHalf>>,
 }
 
+impl Connection {
+    pub fn new(
+        peer_addr: SocketAddr,
+        reader: OwnedReadHalf,
+        writer: OwnedWriteHalf,
```
rkuhn: might just take a TcpStream and call `.into_split` inside this function

crates/amaru-network/src/connection.rs
```diff
@@ -72,15 +96,24 @@ impl Connections {
 pub struct TokioConnections {
     connections: Arc<Mutex<Connections>>,
     read_buf_size: usize,
+    listener: Arc<AsyncMutex<Option<TcpListener>>>,
```
rkuhn: There will very likely be many listeners (at least one per address family, likely specific ones for specific network interfaces).

crates/amaru-network/src/connection.rs
```diff
@@ -72,15 +96,24 @@ impl Connections {
...
     pub fn new(read_buf_size: usize) -> Self {
         Self {
             connections: Arc::new(Mutex::new(Connections::new())),
             read_buf_size,
+            listener: Arc::new(AsyncMutex::new(None)),
+            accept_sender: None,
         }
     }
+
+    pub fn with_accept_sender(mut self, accept_sender: Sender<(Peer, ConnectionId)>) -> Self {
```
rkuhn: Instead of sending from the effects side I’d prefer offering an `accept` effect that will wait for the next connection. This can then be used to build a stage that processes incoming connections.

crates/amaru-network/src/connection.rs
```diff
@@ -14,26 +14,50 @@
...
 
 pub struct Connection {
+    peer_addr: SocketAddr,
     reader: Arc<AsyncMutex<(OwnedReadHalf, BytesMut)>>,
     writer: Arc<AsyncMutex<OwnedWriteHalf>>,
 }
 
+impl Connection {
+    pub fn new(
+        peer_addr: SocketAddr,
```
rkuhn: should just call `.peer_addr()` on the passed-in TcpStream

crates/amaru-network/src/connection.rs
```diff
@@ -94,17 +127,89 @@ async fn connect(
...
+        let listener = self.listener.clone();
+        Box::pin(
+            async move {
+                let mut guard = listener.lock().await;
+                if guard.is_some() {
+                    return Err(std::io::Error::other(
+                        "listener already created; cannot bind multiple times",
+                    ));
+                }
+                let bound = TcpListener::bind(addr).await?;
```
rkuhn: need to set SO_REUSEADDR and SO_NODELAY to avoid annoying admins

crates/amaru-network/src/connection.rs
```diff
@@ -94,17 +127,89 @@ async fn connect(
...
+                let addr = bound.local_addr()?;
+                tracing::debug!(%addr, "listening");
+                *guard = Some(bound);
+                Ok(addr)
+            }
+            .instrument(tracing::debug_span!("bind_listener", %addr)),
+        )
+    }
+
+    fn listening_port(&self) -> BoxFuture<'static, std::io::Result<u16>> {
```
rkuhn: This feels like it shouldn’t be necessary since the information has already been given to the caller of `bind` (which should probably be called `listen`).

crates/amaru-network/src/connection.rs
```diff
@@ -94,17 +127,89 @@ async fn connect(
...
+                        writer,
+                        read_buf_size,
+                    ))
+                };
+
+                if let Some(sender) = accept_sender {
+                    let peer = Peer::new(peer_addr.to_string().as_str());
+                    sender.send((peer, id)).await.map_err(|_| {
+                        std::io::Error::other("failed to send accepted connection to accept stage")
+                    })?;
```
rkuhn: In my proposal sketched above, there would be an mpmc-channel into which these new connections are sent (incl. back-pressure as you’ve done here) and from which the `accept` effect would pull. The accept loop itself would be in a separate Tokio task (that needs to be aborted once the `TokioConnections` resource is dropped).

crates/amaru-network/src/connection.rs
```diff
@@ -212,3 +317,97 @@ impl ConnectionProvider for TokioConnections {
...
+    async fn bind_and_accept_a_client_connection() -> std::io::Result<()> {
+        // Create a TokioConnections instance and bind a TCP listener
+        // to an ephemeral port.
+        let connections = TokioConnections::new(1024);
+
+        connections
+            .bind(SocketAddr::from(([127, 0, 0, 1], 0)))
+            .await?;
+
+        let addr = {
```
rkuhn: why not use to return value of `connections.bind`? or better, assert that it is equal?

crates/amaru-network/src/connection.rs
```diff
@@ -212,3 +317,97 @@ impl ConnectionProvider for TokioConnections {
...
+        assert_eq!(reply.as_ref(), b"pong");
+
+        connections.close(connection_id).await?;
+        server.await.expect("server task panicked")?;
+
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn bind_and_accept_a_client_connection() -> std::io::Result<()> {
```
rkuhn: This is okay since we’re using nightly rust. On stable, there is no useful source code location printed when returning Result instead of just using `.unwrap()` within the test body.

EDIT: this would only be okay if returning `anyhow::Result` because that will take a stack trace upon creating the `anyhow::Error`.

crates/amaru-network/src/connection.rs
```diff
@@ -212,3 +317,97 @@ impl ConnectionProvider for TokioConnections {
...
+            stream.read_exact(&mut buf).await?;
+            assert_eq!(&buf, b"world");
+
+            Ok(())
+        });
+
+        // Receive "hello" from the client and respond with "world".
+        let connection_id = connections.accept(Duration::from_secs(1)).await?;
+        let result = connections
+            .recv(connection_id, NonZeroUsize::new(5).expect("non-zero"))
```
rkuhn: fun fact: could use `const { NonZeroUsize::new(5).unwrap() }` to statically ensure no panics

crates/amaru-network/src/connection.rs
```diff
@@ -212,3 +317,97 @@ impl ConnectionProvider for TokioConnections {
...
+        };
+
+        // Start a client that connects to the listener and
+        // sends "hello", expecting "world" in response.
+        let client: JoinHandle<std::io::Result<()>> = tokio::spawn(async move {
+            let mut stream = TcpStream::connect(addr).await?;
+            stream.write_all(b"hello").await?;
+
+            let mut buf = [0u8; 5];
+            stream.read_exact(&mut buf).await?;
```
rkuhn:
```suggestion
            let mut buf = String::new();
            stream.read_to_string(&mut buf).await?;
```
With this change we’re also testing that closing the connection works properly.

crates/amaru-protocols/src/blockfetch/mod.rs
```diff
@@ -116,7 +116,7 @@ pub async fn register_blockfetch_responder<M>(
     eff.send(
         muxer,
         MuxMessage::Register {
-            protocol: PROTO_N2N_BLOCK_FETCH.erase(),
+            protocol: PROTO_N2N_BLOCK_FETCH.responder().erase(),
```
rkuhn, line 119: ah yes, that is a jarring remnant — we have fully type-checked protocol ID available at the Miniprotocol level, perhaps we can use that somehow

crates/amaru-protocols/src/manager.rs
```diff
@@ -156,6 +158,55 @@ pub async fn stage(
             eff.send(&connection, ConnectionMessage::Initialize).await;
             *entry = ConnectionState::Connected(conn_id, connection);
         }
+        ManagerMessage::Accept => {
+            if Network::new(&eff)
+                .accept(Duration::from_secs(1))
+                .await
```
rkuhn: This will block the manager for a second at a time. My proposed scheme would run the accept loop inside a separate stage and then send new connections to the manager as soon as they are ready.

crates/amaru-protocols/src/manager.rs
```diff
@@ -156,6 +158,55 @@ pub async fn stage(
...
+        }
+        ManagerMessage::Accepted(peer, conn_id) => {
+            match manager.peers.get_mut(&peer) {
+                Some(ConnectionState::Connected(..)) => {
+                    tracing::debug!(%peer, "discarding connection request, already connected");
+                    return manager;
+                }
+                // TODO: not sure about the scheduled case
+                Some(ConnectionState::Disconnecting | ConnectionState::Scheduled) => {
+                    tracing::debug!(%peer, "discarding connection request, already disconnecting");
```
rkuhn: The scheduled case ostensibly means that we want to be connected, so should work with the new connection (the scheduled connection attempt will then be ignored, which is fine).

In the Disconnecting case, the newly accepted connection still needs to be closed.

crates/amaru-protocols/src/manager.rs
```diff
@@ -156,6 +158,55 @@ pub async fn stage(
...
+                .is_err()
+            {
+                eff.schedule_after(ManagerMessage::Accept, Duration::from_secs(1))
+                    .await;
+                return manager;
+            }
+            eff.send(eff.me_ref(), ManagerMessage::Accept).await;
+        }
+        ManagerMessage::Accepted(peer, conn_id) => {
+            match manager.peers.get_mut(&peer) {
```
rkuhn:
```suggestion
            match manager.peers.get(&peer) {
```

crates/amaru-protocols/src/tests.rs
```diff
@@ -0,0 +1,324 @@
...
+}
+
+/// Create and start a responder node that listens for incoming connections on a free port found at
+/// runtime.
+async fn start_responder() -> anyhow::Result<TokioRunning> {
+    tracing::info!("Creating the responder");
+    let mut responder_network = TokioBuilder::default();
+    let responder_manager = Manager::new(NetworkMagic::PREPROD, StageRef::blackhole());
+    let responder_stage = responder_network.stage("responder", manager::stage);
+    let responder_stage = responder_network.wire_up(responder_stage, responder_manager);
```
rkuhn, line 92: This would be more readable if there was a blank line after each stage setup.

crates/amaru-protocols/src/tests.rs
```diff
@@ -0,0 +1,324 @@
...
+        &mut responder_network,
+        Role::Responder,
+        responder_connections,
+    )?;
+    tracing::info!("Responder listening on {}", peer_addr);
+
+    tracing::info!("Start the responder");
+    let running_responder = responder_network.run(Handle::current());
+    responder_sender.send(ManagerMessage::Accept).await.unwrap();
+    Ok(running_responder)
```
rkuhn: could return `peer_addr` here as well to get rid of the `get_listening_port()` workaround

crates/amaru-protocols/src/tests.rs
```diff
@@ -0,0 +1,324 @@
...
+}
+
+/// Create and start an initiator node that connects to the given port.
+/// ChainSync events are sent to a stage that stores them in the in-memory store without further processing.
+async fn start_initiator(port: Option<u16>) -> anyhow::Result<TokioRunning> {
```
rkuhn: I’d just pass around SocketAddr instead of u16

crates/amaru-protocols/src/tests.rs
```diff
@@ -0,0 +1,324 @@
...
+    );
+
+    // Verify that the 2 nodes have the same transactions
+    let tx_ids = get_tx_ids();
+    assert_eq!(
+        responder_mempool.get_txs_for_ids(tx_ids.as_slice()),
+        initiator_mempool.get_txs_for_ids(tx_ids.as_slice())
+    );
+
+    Ok(())
```
rkuhn, line 81: NOICE!

crates/amaru-network/src/connection.rs
etorreborre: That's doable indeed.

crates/amaru-network/src/connection.rs
etorreborre: Done.

crates/amaru-network/src/connection.rs
etorreborre: Good idea.

crates/amaru-network/src/connection.rs
etorreborre: Right, I'm just going to use the returned addr. I don't think it's necessary to check that it's the same as `local_addr` since that would be testing the `TcpListener`.

crates/amaru-network/src/connection.rs
etorreborre: Nice!

crates/amaru-network/src/connection.rs
etorreborre: Done.

crates/amaru-network/src/connection.rs
etorreborre: Good point.

crates/amaru-protocols/src/blockfetch/mod.rs
etorreborre: Yes I wondered about that too. It's a bit sad that we have typechecked protocol ids but that this kind of error can occur. I didn't see any quick fix though.

crates/amaru-protocols/src/manager.rs
etorreborre: Copy and paste.

crates/amaru-protocols/src/tests.rs
etorreborre: Done with the removal of `TokioConnections::listening_port`

crates/amaru-network/src/connection.rs
etorreborre: Done.

crates/amaru-protocols/src/tests.rs
etorreborre: Done.

crates/amaru-protocols/src/tests.rs
etorreborre: Done.

crates/amaru-network/src/connection.rs
```diff
@@ -302,6 +304,29 @@ impl ConnectionProvider for TokioConnections {
...
+/// `SO_REUSEADDR` (and `SO_REUSEPORT` on Linux) enabled.
+fn bind_address(addr: SocketAddr) -> std::io::Result<TcpListener> {
+    let domain = match addr {
+        SocketAddr::V4(_) => Domain::IPV4,
+        SocketAddr::V6(_) => Domain::IPV6,
+    };
+
+    let socket = Socket::new(domain, Type::STREAM, None)?;
+
+    // This makes sure that ports can be reused immediately after the program exits.
```
rkuhn: No, this is needed by the new process when it wants to use a port that has recently been closed.

crates/amaru-network/src/connection.rs
```diff
@@ -302,6 +304,29 @@ impl ConnectionProvider for TokioConnections {
...
+    socket.set_reuse_address(true)?;
+
+    #[cfg(target_os = "linux")]
+    socket.set_reuse_port(true)?;
```
rkuhn: I don’t think we want this: Amaru assumes that it is the only service listening on this port.

crates/amaru-protocols/src/network_effects.rs
```diff
@@ -20,6 +20,8 @@ use std::{num::NonZeroUsize, time::Duration};
 
 pub fn register_deserializers() -> pure_stage::DeserializerGuards {
     vec![
+        pure_stage::register_data_deserializer::<ListenEffect>().boxed(),
+        pure_stage::register_data_deserializer::<AcceptEffect>().boxed(),
```
rkuhn, line 25: ah, I didn’t catch this earlier: effects need to use register_effect_deserializer because the infrastructure for those is separate

crates/amaru-network/src/connection.rs
etorreborre: I modified the comment.

crates/amaru-network/src/connection.rs
```diff
@@ -212,3 +315,117 @@ impl ConnectionProvider for TokioConnections {
...
+
+    // Allow rebinding to a port that was recently used (e.g., still in TIME_WAIT).
+    socket.set_reuse_address(true)?;
+
+    socket.bind(&addr.into())?;
+    socket.listen(1024)?;
+
+    socket.set_nonblocking(true)?;
+    TcpListener::from_std(socket.into())
+}
```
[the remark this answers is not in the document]
etorreborre: Done.

crates/amaru-protocols/src/tests.rs
```diff
@@ -0,0 +1,337 @@
...
+        initiator_chain_store.get_best_chain_hash(),
+        responder_chain_store.get_best_chain_hash()
+    );
+
+    // Verify that the 2 nodes have the same transactions
+    let tx_ids = get_tx_ids();
+    assert_eq!(
+        responder_mempool.get_txs_for_ids(tx_ids.as_slice()),
+        initiator_mempool.get_txs_for_ids(tx_ids.as_slice())
+    );
```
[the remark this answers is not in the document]
rkuhn: no, leaving it running for 5sec without crashing is part of the test, and the convergence is checked either way

rkuhn:
e2e test failure is on `main` and being tackled elsewhere (possibly #641, let’s see), so I propose to merge — @etorreborre WDYT?

PR #638, opened 2026-01-20
feat: improve observability

jeluard:
General cleanup/improvements of the observability stack. For more context about those changes please refer to this [document](https://gist.github.com/jeluard/d4034da7290e3702f0166e6bfec539f3).

* introduced a dedicated crate `amaru-observability`
* introduced a `schemas` concept alongside `trace` and `augment-trace` macros (see `schemas.rs`)
* `docs/TRACES.md` documentation is automatically generated and checked by CI
* `docs/traces-schema.json` JSON Schema is automatically generated and checked by CI

Defining `AMARU_TRACE_NOOP` allows to disable all trace generations, and removes all associated code generation.

# Syntax

(from the EDR)

```rust
#[trace(ledger::state::APPLY_BLOCK)]
fn apply_block(block: &Block) {
    // Record additional fields with schema context
    trace_record!(ledger::state::APPLY_BLOCK, block_size = block.size(), tx_count = block.transactions.len());

    // or alternatively
    trace_span!(ledger::state::APPLY_BLOCK);
}
```

This PR leverages significantly complex macros to generate meaningful `compile-time` errors. It received some AI love in the process.

Fixes https://github.com/pragma-org/amaru/issues/640

## Summary by CodeRabbit

* **New Features**
  * Introduces a schema-driven tracing system with macros, a runtime registry, examples, and a CLI command to dump/export trace schemas.

* **Documentation**
  * Adds comprehensive TRACES.md, a generated traces JSON Schema, and a script to regenerate trace docs.

* **Tests**
  * Adds extensive unit, UI (compile-fail), and integration tests validating macros, schema validation, and trace recording.

* **Chores**
  * CI/workflow updates, workspace/manifest and toolchain adjustments, line-ending rule, and build/test command standardization.

### DIFF
--- a/crates/amaru-ledger/src/rules.rs
+++ b/crates/amaru-ledger/src/rules.rs
@@ -17,7 +17,6 @@ use std::{fmt, fmt::Display};
 use amaru_kernel::Block;
 use amaru_observability::trace;
 pub use block::execute as validate_block;
-use tracing::warn;
 
 use crate::context::PreparationContext;
 
--- a/crates/pure-stage/src/logging/layer.rs
+++ b/crates/pure-stage/src/logging/layer.rs
@@ -51,7 +51,7 @@ use cbor4ii::{
     serde::{from_slice, to_writer},
 };
 use parking_lot::Mutex;
-use tracing::{Event, Subscriber, info};
+use tracing::{Event, Subscriber};
 use tracing_subscriber::layer::{Context, Layer};
 
 #[derive(Default)]


crates/amaru-observability/macros/src/define_schemas.rs
```diff
@@ -0,0 +1,1514 @@
...
+    /// Field type as string (e.g., "u64", "String")
+    ty: String,
+}
+
+/// A complete schema definition.
+#[derive(Debug, Clone)]
+pub struct Schema {
+    /// Category path components (e.g., ["ledger", "state"] for ledger::state)
+    /// Can be arbitrary depth: ["a"], ["a", "b"], ["a", "b", "c"], etc.
+    categories: Vec<String>,
```
etorreborre, line 94: How practical would it be to use a more restrictive type here instead of `String` in order to prevent an element to be named `a::b`?

crates/amaru-consensus/src/stages/fetch_block.rs
```diff
@@ -38,7 +39,7 @@ pub fn stage(
     msg: ValidateHeaderEvent,
     eff: impl ConsensusOps,
 ) -> impl Future<Output = State> {
-    let span = tracing::trace_span!(parent: msg.span(), "diffusion.fetch_block");
+    let span = tracing::trace_span!(parent: msg.span(), FETCH_BLOCK);
```
etorreborre, line 40: Should we provide our own span creation function in `amaru_observability` then ban all uses of `tracing` in all crates but `amaru_observability`?

crates/amaru/src/bin/amaru/cmd/dump_schemas.rs
```diff
@@ -0,0 +1,121 @@
...
+        serde_json::to_string(&output)?
+    } else {
+        serde_json::to_string_pretty(&output)?
+    };
+
+    eprintln!("{}", json_string);
+    Ok(())
+}
+
+fn generate_json_schema(entries: &[SchemaEntry]) -> Value {
```
etorreborre: The naming here and file name should include the word `trace` because there could be schemas for other things.

docs/TRACES.md
```diff
@@ -0,0 +1,595 @@
...
+This document lists all available spans in Amaru, auto-generated from the code.
+
+For information on how to use and filter these spans, see [monitoring/README.md](../monitoring/README.md).
+
+
+## target: `amaru::consensus::chain_sync`
+
+| name | level | description | required fields | optional fields |
+| --- | --- | --- | --- | --- |
+| `decode_header` | `TRACE` | Decode header from raw bytes |  |  |
```
etorreborre: I'm just leaving this as a note.

I think that we need to use the levels a bit better and not leave everything at the `TRACE` level.
For example, for the `amaru::consensus::chain_sync` target, the `decode_header` span should probably be at the `DEBUG` level, `pull` at the `TRACE` level (this is a low-level mechanism) and maybe `select_chain` at the `INFO` level.

This way we can use the `INFO` level to see high-level normal processing, and other levels to drill-down.
The important thing IMO is to have the ability to drill-down target by target, and for example see the possible interactions between the ledger and the consensus by selecting `amaru_ledger=debug, amaru_consensus=info`.

docs/TRACES.md
```diff
@@ -0,0 +1,595 @@
+# Available Spans
+
+This document lists all available spans in Amaru, auto-generated from the code.
```
etorreborre, line 3: I am wondering if it would be a good idea to distinguish between external/official traces and internal ones with the idea that internal ones can be useful to understand the system as it is but might change if the implementation evolves.

crates/amaru-observability/tests/trace_span.rs
```diff
@@ -0,0 +1,165 @@
...
+            }
+        }
+    }
+}
+
+#[test]
+fn test_trace_span_without_fields() {
+    // Create a span without any fields
+    let span = trace_span!(amaru::test::span_test::SpanEvent);
+    let _guard = span.enter();
```
etorreborre: Aren't all those tests just testing that the code compiles? When it compiles it is possible that the span could not be entered?

crates/amaru-consensus/src/stages/fetch_block.rs
jeluard: I thought about that unfortunately clippy runs after macro expansion so sees all the `instrumentation` usages expanded from the `trace` macro.
Also it can still be useful for tests.

So maybe if there is something more flexible than clippy?

crates/amaru/src/bin/amaru/cmd/dump_schemas.rs
jeluard: good point

docs/TRACES.md
jeluard: Akin to public API ? 
Good idea! There could be some extra data just used for schema/doc generation.
Maybe having an explicit `private`, so that everything is public by default?

crates/amaru-observability/tests/trace_span.rs
jeluard: Let me do some cleanup.

docs/TRACES.md
jeluard: I generally agree although I feel like it mostly apply to human logs, not computer traces.
For computer traces we probably want everything (and filtering some actually breaks parent/child relationships). At least for OTEL.
Now we also have an issue of too many logs generated for now. Especially when the node catches up and apply lots of blocks.

docs/TRACES.md
jeluard: cc @KtorZ

crates/amaru-observability/macros/src/define_schemas.rs
jeluard: Given that those types are not exposed I think it's sufficient to have a compile-time error when an invalid name is provided?
I'll add this.

jeluard: Done

docs/TRACES.md
jeluard: Done

crates/amaru-consensus/src/stages/fetch_block.rs
etorreborre: Ok, let's keep it like that for now.

docs/TRACES.md
etorreborre: Thanks!

PR #641, opened 2026-01-21
fix scripts/demo JSON handling

rkuhn:
`read line` removes backslashes which destroys the proper JSON emitted by our tracing subscriber

## Summary by CodeRabbit

* **Chores**
  * Reduced logging verbosity and clarified error messages; minor CI snapshot tweak.
  * Demo script: builds before run, improves output and enables controlled process termination.
  * Disabled several noisy snapshot tests.

* **Refactor**
  * Header processing and stage wiring now use an explicit state object with runtime counters and improved logging.
  * Request/queue handling moved to a pipelined/countered model; protocol mux tracing updated; related tests adjusted.

* **New Features**
  * Optional CLI color flag for logs and color-aware observability; increased runtime worker threads.

### DIFF
--- a/crates/amaru/src/bin/amaru/main.rs
+++ b/crates/amaru/src/bin/amaru/main.rs
@@ -12,7 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use amaru::{observability::setup_observability, panic::panic_handler};
+use amaru::{
+    observability::{Color, setup_observability},
+    panic::panic_handler,
+};
 
 use clap::{CommandFactory, FromArgMatches, Parser, Subcommand};
 use std::sync::LazyLock;
@@ -121,10 +124,13 @@ struct Cli {
     with_json_traces: bool,
 
     #[clap(long, action, env("AMARU_COLOR"))]
-    color: Option<bool>,
+    color: Option<Color>,
 }
 
-#[tokio::main(worker_threads = 8)]
+// TODO(rkuhn): properly measure and design the Tokio runtime setup we need.
+// (probably one runtime for network with 1-2 threads, one for CPU-bound tasks according to parallelism,
+// one for running the consensus pipeline incl. Store access with 2+ threads)
+#[tokio::main(worker_threads = 4)]
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
     panic_handler();
 
@@ -133,8 +139,11 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         .get_matches();
     let args = <Cli as FromArgMatches>::from_arg_matches(&matches)?;
 
-    let (metrics, teardown) =
-        setup_observability(args.with_open_telemetry, args.with_json_traces, args.color);
+    let (metrics, teardown) = setup_observability(
+        args.with_open_telemetry,
+        args.with_json_traces,
+        Color::is_enabled(args.color),
+    );
 
     info!(
         with_open_telemetry = args.with_open_telemetry,
--- a/crates/amaru/src/observability.rs
+++ b/crates/amaru/src/observability.rs
@@ -20,6 +20,7 @@ use std::{
     env::VarError,
     error::Error,
     io::{self, IsTerminal},
+    str::FromStr,
 };
 use tracing::{info, warn};
 use tracing_subscriber::{
@@ -109,14 +110,10 @@ impl TracingSubscriber<Registry> {
         }
     }
 
-    pub fn init(self, color: Option<bool>) {
+    pub fn init(self, color: bool) {
         let (default_filter, warning) = new_default_filter(AMARU_LOG_VAR, DEFAULT_AMARU_LOG_FILTER);
 
-        let log_format = || {
-            tracing_subscriber::fmt::format()
-                .with_ansi(color.unwrap_or(io::stderr().is_terminal()))
-                .compact()
-        };
+        let log_format = || tracing_subscriber::fmt::format().with_ansi(color).compact();
         let log_writer = || io::stderr as fn() -> io::Stderr;
         let log_events = || FmtSpan::CLOSE;
         let log_filter = || default_filter;
@@ -320,7 +317,7 @@ fn new_default_filter(var: &str, default: &str) -> (EnvFilter, DelayedWarning) {
 pub fn setup_observability(
     with_open_telemetry: bool,
     with_json_traces: bool,
-    color: Option<bool>,
+    color: bool,
 ) -> (
     Option<SdkMeterProvider>,
     Box<dyn FnOnce() -> Result<(), Box<dyn std::error::Error>>>,
@@ -347,4 +344,38 @@ pub fn setup_observability(
     }
 
     (metrics, teardown)
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Color {
+    Never,
+    Always,
+    Auto,
+}
+impl FromStr for Color {
+    type Err = &'static str;
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "never" => Ok(Color::Never),
+            "always" => Ok(Color::Always),
+            "auto" => Ok(Color::Auto),
+            _ => Err("valid color settings are 'never', 'always' or 'auto'"),
+        }
+    }
+}
+impl Color {
+    pub fn is_enabled(this: Option<Self>) -> bool {
+        match this {
+            Some(Color::Never) => false,
+            Some(Color::Always) => true,
+            Some(Color::Auto) => std::io::stderr().is_terminal(),
+            None => {
+                if std::env::var("NO_COLOR").iter().any(|s| !s.is_empty()) {
+                    false
+                } else {
+                    std::io::stderr().is_terminal()
+                }
+            }
+        }
+    }
 }
--- a/crates/amaru/src/bin/ledger/main.rs
+++ b/crates/amaru/src/bin/ledger/main.rs
@@ -12,7 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use amaru::{observability::setup_observability, panic::panic_handler};
+use amaru::{
+    observability::{Color, setup_observability},
+    panic::panic_handler,
+};
 use clap::{Parser, Subcommand};
 use tracing::info;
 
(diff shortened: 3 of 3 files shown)


jeluard:
Looks like now both e2e reach target epoch but fail with:

```rust
BlockHeight(3631987))"},"target":"amaru_consensus::consensus::stages::pull"}
2026-01-21T12:42:12.4839594Z Target epoch reached, stopping the process.
2026-01-21T12:42:12.6438191Z make: *** [Makefile:64: demo] Error 130
2026-01-21T12:42:12.6465031Z ##[error]Process completed with exit code 2.
```

rkuhn:
> Looks like now both e2e reach target epoch but fail with:

That failure is exactly what I‘d expect as the normal outcome of the `demo` script: it kills the `amaru` process with signal two, leading to an exit code of 128+2.

Looking deeper.

crates/amaru-protocols/src/chainsync/initiator.rs
```diff
@@ -222,24 +225,26 @@ impl ProtocolState<Initiator> for InitiatorState {
...
                 outcome().result(InitiatorResult::RollForward(content, tip)),
-                Idle,
+                if *n == 0 { Idle } else { CanAwait(*n - 1) },
             ),
-            (CanAwait | MustReply, Message::RollBackward(point, tip)) => (
+            (CanAwait(n) | MustReply(n), Message::RollBackward(point, tip)) => (
                 outcome().result(InitiatorResult::RollBackward(point, tip)),
-                Idle,
+                if *n == 0 { Idle } else { CanAwait(*n - 1) },
             ),
```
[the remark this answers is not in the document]
rkuhn: no, the responder won’t read the second RequestNext until its state has returned to Idle again

crates/amaru-protocols/src/chainsync/initiator.rs
```diff
@@ -256,9 +261,18 @@ impl ProtocolState<Initiator> for InitiatorState {
...
+                CanAwait(0),
+            ),
+            (CanAwait(n), InitiatorAction::RequestNext) => (
+                outcome().send(Message::RequestNext(1)).want_next(),
+                CanAwait(*n + 1),
+            ),
+            (MustReply(n), InitiatorAction::RequestNext) => (
+                outcome().send(Message::RequestNext(1)).want_next(),
+                MustReply(*n + 1),
+            ),
```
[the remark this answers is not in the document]
rkuhn: the RequestNext comes from the local node, not from a peer, and it is bounded by the design of the logic (modulo bugs)

crates/amaru-protocols/src/blockfetch/initiator.rs
```diff
@@ -95,9 +99,7 @@ impl StageState<State, Initiator> for BlockFetchInitiator {
     ) -> anyhow::Result<(Option<InitiatorAction>, Self)> {
         match input {
             BlockFetchMessage::RequestRange { from, through, cr } => {
-                let action = self
-                    .queue
-                    .is_empty()
+                let action = (self.queue.len() < 2)
                     .then_some(InitiatorAction::RequestRange { from, through });
                 self.queue.push_back((from, through, cr));
                 Ok((action, self))
```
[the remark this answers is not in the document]
rkuhn: you’re right that currently the verification using ProtoSpec::check() is incapable of understanding pipelining as specified for Ouroboros; the implementation is correct, though, because the responder won’t look at the next RequestRange in Busy state, only once the current request has been responded to in full.

.github/workflows/continuous-integration.yml
```diff
@@ -199,7 +199,7 @@ jobs:
         network:
           - name: preprod
             magic: 1
-            target_epoch: 182
+            target_epoch: 176
```
KtorZ, line 202: ?

crates/amaru-consensus/src/consensus/stages/receive_header.rs
```diff
@@ -48,23 +65,32 @@ pub fn stage(
...
                 };
 
+                state.recv_count += 1;
+                if header.point() == tip.point() {
+                    tracing::info!(%peer, point = ?header.point(), "received header");
+                } else if state.recv_count & 0xff == 0 {
+                    tracing::info!(%peer, point = ?header.point(), ?tip, recv_count = %state.recv_count, "received header (catching up)");
+                } else {
+                    tracing::debug!(%peer, point = ?header.point(), ?tip, recv_count = %state.recv_count, "received header (catching up)");
+                }
```
KtorZ:
```suggestion
                    tracing::info!(%peer, point = %header.point(), "received header");
                } else if state.recv_count & 0xff == 0 {
                    tracing::info!(%peer, point = %header.point(), %tip, recv_count = %state.recv_count, "received header (catching up)");
                } else {
                    tracing::debug!(%peer, point = %header.point(), %tip, recv_count = %state.recv_count, "received header (catching up)");
                }
```

Also, the `& 0xff` looks ... strange? Did you mean `% 256` ?

crates/amaru/src/bin/amaru/main.rs
```diff
@@ -119,9 +119,12 @@ struct Cli {
 
     #[clap(long, action, env("AMARU_WITH_JSON_TRACES"))]
     with_json_traces: bool,
+
+    #[clap(long, action, env("AMARU_COLOR"))]
+    color: Option<bool>,
```
KtorZ: What about relying on the usual `NO_COLOR` env var for this instead? It's _pretty standard_.

crates/amaru/src/bin/amaru/main.rs
```diff
@@ -119,9 +119,12 @@ struct Cli {
...
 }
 
-#[tokio::main]
+#[tokio::main(worker_threads = 8)]
```
KtorZ: Seems oddly specific? Why 8?

.github/workflows/continuous-integration.yml
rkuhn: As pointed out on discord: the current setup only gets to epoch 178 within 15min. We’ll fix that, but in the meantime we don’t want `main` to be always red on CI.

crates/amaru-consensus/src/consensus/stages/receive_header.rs
rkuhn: The bitwise and is probably due to the (lack of) colour of my beard — I thought this was the usual way to express it.

Your proposal of using Display for Point is good, but for Tip it is erasing exactly that part of the information I wanted to see, namely the slot number. Is this string representation mandated by some standard or can we add the slot number to it?

crates/amaru/src/bin/amaru/main.rs
rkuhn: I added this so I can have color when stderr is not a TTY (I like to view logs with color in `less` for easy filtering). I also haven’t yet met the NO_COLOR env var, which means that it probably isn’t _standard enough_ ;-)

crates/amaru/src/bin/amaru/main.rs
rkuhn: Ah, this should probably be replaced by a proper management of several Tokio runtimes for different purposes — it was just an attempt at ensuring that we aren’t wasting latency due to too few threads automatically allocated on CI (or using too many on a Hetzner box). Never trust an adaptive default.

crates/amaru/tests/snapshots/preprod/generated_compare_snapshot_test_cases.incl
```diff
@@ -9,12 +9,12 @@
...
-#[test_case(176)]
-#[test_case(177)]
-#[test_case(178)]
-#[test_case(179)]
+// #[test_case(174)]
+// #[test_case(175)]
+// #[test_case(176)]
+// #[test_case(177)]
+// #[test_case(178)]
+// #[test_case(179)]
```
KtorZ:
```suggestion
```

crates/amaru-consensus/src/consensus/stages/receive_header.rs
rkuhn: switching to `tip_point = %tip.point()` as discussed elsewhere

crates/amaru/src/bin/amaru/main.rs
rkuhn: A little research has shown that there is no consensus on how to switch color on, only on how to switch it off (using NO_COLORS). There is established precedent for tool-specific environment variables (LS_COLORS, GREP_COLORS) as well as wider consensus on command line option `--colors=[never|auto|always]`.

So, I’ll add the command line option and the equivalent AMARU_COLORS environment variable, and in the absence of these signals NO_COLORS can be used to switch color off, otherwise it is enabled if stderr is a TTY. Does that sound good to you @KtorZ ?

crates/amaru/tests/snapshots/preprod/generated_compare_snapshot_test_cases.incl
rkuhn: wilco

crates/amaru/src/bin/amaru/main.rs
KtorZ: Hmm. Why not simply: enable colors by default unless not a TTY (i.e. current behavior); except if NO_COLOR is set. In which case, disable colors regardless.

rkuhn: Because the only point was to switch colors on when the output is not to a TTY.

crates/amaru/src/observability.rs
```diff
@@ -347,3 +345,37 @@ pub fn setup_observability(
...
+            Some(Color::Always) => true,
+            Some(Color::Auto) => std::io::stderr().is_terminal(),
+            None => {
+                if std::env::var("NO_COLOR").iter().any(|s| !s.is_empty()) {
+                    false
+                } else {
+                    std::io::stderr().is_terminal()
+                }
+            }
+        }
```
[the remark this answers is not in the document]
rkuhn: The “standard” (presumably https://no-colorg.org) says that NO_COLOR set to anything but the empty string shall inhibit the default color output. Adding a command line option specifically to switch color on if writing to a TTY is not default color output but deliberate user choice, so it should not be overridden by NO_COLOR.

crates/amaru/src/bin/amaru/main.rs
KtorZ: Ah. I missed that. 

Why 😶 ?

rkuhn: There are many tools I’d want to pipe the logs into, which always means “not a TTY”. And when I read the resulting logs (filtered etc.) I want the colors. I’m somewhat taken aback that this needs explaining.

PR #642, opened 2026-01-21
feat: allow to customize AMARU_TRACE when running e2e tests

jeluard:
## Summary by CodeRabbit

* **Chores**
  * CI workflow: added a new optional input to configure the demo trace level for runs.
  * Demo script: applies an optional trace-level setting at invocation and removed the prior global export.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -14,6 +14,9 @@ on:
       timeout:
         description: "Override timeout for snapshot tests"
         required: false
+      demo_trace_level:
+        description: "Override AMARU_TRACE (leave empty for default)"
+        required: false
   push:
     branches: ["main"]
   pull_request:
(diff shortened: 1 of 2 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -14,6 +14,9 @@ on:
       timeout:
         description: "Override timeout for snapshot tests"
         required: false
+      demo_trace_level:
```
rkuhn, line 17: This name suggests that this input will only affect the demo script, while in fact it will be visible to all running code and potentially influence tests.

.github/workflows/continuous-integration.yml
```diff
@@ -29,7 +32,7 @@ env:
     ~/.cargo/git/db/
     target/
   INPUT_TIMEOUT: ${{ github.event.inputs.timeout }}
-  INPUT_DEMO_TARGET_EPOCH: ${{ github.event.inputs.demo_target_epoch }}
+  AMARU_TRACE: ${{ github.event.inputs.demo_trace_level }}
```
rkuhn, line 35: I haven't tried it: are you sure that the variable is unset when the input is unset? It looks to me that it might also be set to the empty string, which will lead to some head scratching.

PR #643, opened 2026-01-22
feat: finish the implementation of the block fetch protocol

etorreborre:
This PR completes (on the responder side) and tests the retrieval of blocks with the blockfetch protocol:

 - The streaming of blocks with the `StartBatch / Block / BatchDone` messages.
 - The retrieval of a range of blocks from the `ChainStore`.
 - The test the blockfetch protocol works when an initiator and a responder are connected over TCP.
 - Checks that requested blocks are either present on the best chain or on a fork that is not too old.

There is also an additional check on the client side that the fetched blocks are valid and correspond to the requested range.

## Summary by CodeRabbit

* **New Features**
  * Enhanced block streaming with range-based requests and slot-order validation.
  * Added era history tracking for improved network compatibility.

* **Bug Fixes**
  * Improved handling of missing blocks during chain synchronization.
  * Refined block validation error reporting.

* **Refactor**
  * Streamlined block retrieval to better handle unavailable blocks.
  * Reorganized block validation and streaming workflows for clarity.

### DIFF
--- a/crates/amaru-protocols/src/blockfetch/responder.rs
+++ b/crates/amaru-protocols/src/blockfetch/responder.rs
@@ -69,11 +69,13 @@ impl PointsRange {
         self,
         store: &dyn ChainStore<BlockHeader>,
     ) -> anyhow::Result<(RawBlock, Option<PointsRange>)> {
+        // points are stored from most recent to oldest, so we pop from the end
+        let (last, rest) = self.0.pop();
+        let last_hash = last.hash();
         let stored_block = store
-            .load_block(&self.0.first().hash())?
-            .ok_or_else(|| anyhow::anyhow!("block {} was pruned", self.0.first().hash()))?;
-        let points_range = self.0.pop().1.map(PointsRange);
-        Ok((stored_block, points_range))
+            .load_block(&last_hash)?
+            .ok_or_else(|| anyhow::anyhow!("block {} was pruned", last_hash))?;
+        Ok((stored_block, rest.map(PointsRange)))
     }
 
     /// Return a points range:
@@ -120,8 +122,13 @@ impl PointsRange {
             // load the header for the current hash
             if let Some(header) = store.load_header(&current_hash) {
                 result.push(header.point());
+                // if we found the from point, we're done
                 if current_hash == from.hash() {
                     break;
+                }
+                // if we reached a slot before 'from', abort
+                if header.slot() < from.slot_or_default() {
+                    return Ok(None);
                 }
                 if let Some(parent_hash) = header.parent_hash() {
                     current_hash = parent_hash
@@ -350,11 +357,11 @@ pub mod tests {
         assert_eq!(
             result,
             PointsRange::from_vec(vec![
-                headers[0].point(),
-                headers[1].point(),
-                headers[2].point(),
-                headers[3].point(),
                 headers[4].point(),
+                headers[3].point(),
+                headers[2].point(),
+                headers[1].point(),
+                headers[0].point(),
             ])
         );
     }
@@ -426,6 +433,26 @@ pub mod tests {
     }
 
     #[test]
+    fn test_request_range_slot_before_from_abort() {
+        // Create a chain with 5 headers
+        let (store, headers) = make_store_with_chain(5);
+        store_blocks(store.clone(), &headers);
+
+        // Create a 'from' point that has a slot within the chain range but with a non-existent hash.
+        // When traversing backwards from 'through', we'll pass the slot of 'from' without finding it,
+        // and then hit a block with a slot before 'from', triggering the abort condition.
+        let from_slot = headers[2].slot();
+        let non_existent_hash = run_strategy(any_fake_header()).hash();
+        let from = Point::Specific(from_slot, non_existent_hash);
+
+        let result = PointsRange::request_range(&*store, from, headers[4].point()).unwrap();
+        assert_eq!(
+            result, None,
+            "should return None when we reach a slot before 'from' without finding 'from'"
+        );
+    }
+
+    #[test]
     fn test_request_range_exactly_max_blocks() {
         // Create a chain longer than MAX_BLOCKS
         let (store, headers) = make_store_with_chain(MAX_FETCHED_BLOCKS);
@@ -444,7 +471,7 @@ pub mod tests {
     #[test]
     fn test_request_range_max_blocks_limit() {
         // Create a chain longer than MAX_BLOCKS
-        let chain_length = MAX_FETCHED_BLOCKS + 10;
+        let chain_length = MAX_FETCHED_BLOCKS + 1;
         let (store, headers) = make_store_with_chain(chain_length);
         store_blocks(store.clone(), &headers);
 
@@ -486,9 +513,9 @@ pub mod tests {
         store_blocks(store.clone(), &headers);
 
         let (block, remaining_range) = PointsRange::from_vec(vec![
-            headers[0].point(),
-            headers[1].point(),
             headers[2].point(),
+            headers[1].point(),
+            headers[0].point(),
         ])
         .unwrap()
         .next_block(&*store)
@@ -504,7 +531,7 @@ pub mod tests {
         // Should have remaining points
         assert_eq!(
             remaining_range,
-            PointsRange::from_vec(vec![headers[1].point(), headers[2].point()])
+            PointsRange::from_vec(vec![headers[2].point(), headers[1].point()])
         );
     }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -378,7 +378,6 @@ dependencies = [
  "clap",
  "futures-util",
  "hex",
- "minicbor",
  "pallas-crypto",
  "pallas-primitives",
  "parking_lot",
--- a/crates/amaru-kernel/src/data_structures/non_empty_vec.rs
+++ b/crates/amaru-kernel/src/data_structures/non_empty_vec.rs
@@ -28,6 +28,10 @@ impl<T: Eq> NonEmptyVec<T> {
         &self.0[0]
     }
 
+    pub fn last(&self) -> &T {
+        &self.0[self.0.len() - 1]
+    }
+
     /// Pop the last element from the NonEmptyVec, returning it and an Option with the remaining
     #[expect(clippy::expect_used)]
     pub fn pop(self) -> (T, Option<NonEmptyVec<T>>) {
--- a/simulation/amaru-sim/Cargo.toml
+++ b/simulation/amaru-sim/Cargo.toml
@@ -18,7 +18,6 @@ cbor4ii.workspace = true
 clap.workspace = true
 futures-util.workspace = true
 hex.workspace = true
-minicbor.workspace = true
 pallas-crypto.workspace = true
 pallas-primitives.workspace = true
 parking_lot.workspace = true


crates/amaru-protocols/src/blockfetch/responder.rs
```diff
@@ -34,34 +37,101 @@ pub fn responder() -> Miniprotocol<State, BlockFetchResponder, Responder> {
...
+            Self {
+                muxer,
+                current_range: VecDeque::default(),
+            },
+        )
+    }
+
+    /// Load the first existing block in the current range, if any
+    /// Returns None if no blocks are found.
+    /// Whether a block is found or not, the first element of the current_range is consumed.
```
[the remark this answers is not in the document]
etorreborre: Fixed.

crates/amaru-ouroboros-traits/src/stores/consensus/mod.rs
```diff
@@ -92,6 +92,29 @@ where
...
+                }
+            } else {
+                break;
+            }
+        }
+        if let Some(header) = self.load_header(&current_hash) {
+            headers.push(header.hash());
+        }
+        headers.reverse();
+        headers
```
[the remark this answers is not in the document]
etorreborre: That's fine, we return what we can.

crates/amaru-ouroboros-traits/src/stores/consensus/mod.rs
```diff
@@ -92,6 +92,29 @@ where
             Box::new(vec![*hash].into_iter())
         }
     }
+
+    /// Get the range of headers from `from_inclusive` to `to_inclusive`.
+    fn get_range(&self, from_inclusive: &HeaderHash, to_inclusive: &HeaderHash) -> Vec<HeaderHash> {
+        let mut headers = vec![];
+        let mut current_hash = *to_inclusive;
+        while current_hash != *from_inclusive {
```
rkuhn: This loop will run until genesis if from_inclusive is not an ancestor of to_inclusive — this is an attack vector. We should probably set an upper bound on how many hashes are returned at maximum. What does the Haskell node do?

crates/amaru-protocols/src/blockfetch/responder.rs
```diff
@@ -34,34 +37,101 @@ pub fn responder() -> Miniprotocol<State, BlockFetchResponder, Responder> {
...
 }
 
 impl BlockFetchResponder {
     pub fn new(muxer: StageRef<MuxMessage>) -> (State, Self) {
-        (State::Idle, Self { muxer })
+        (
+            State::Idle,
+            Self {
+                muxer,
+                current_range: VecDeque::default(),
```
rkuhn: since we only consume from one end and don’t add to the other, perhaps a Vec is enough (with parent after child, so we can pop off the end)

crates/amaru-protocols/src/blockfetch/responder.rs
```diff
@@ -34,34 +37,101 @@ pub fn responder() -> Miniprotocol<State, BlockFetchResponder, Responder> {
...
-        _eff: &Effects<Inputs<Self::LocalIn>>,
+        eff: &Effects<Inputs<Self::LocalIn>>,
     ) -> anyhow::Result<(Option<ResponderAction>, Self)> {
         match input {
-            ResponderResult::RequestRange { .. } => Ok((Some(ResponderAction::NoBlocks), self)),
+            ResponderResult::RequestRange { from, through } => {
+                let store = Store::new(eff.clone());
+                let range = store.get_range(&from.hash(), &through.hash());
+                self.current_range = VecDeque::from(range);
+                if let Some(block) = self.load_first_block(eff).await {
```
rkuhn: I think it would be cleaner and easier to read if there was a `check_block_exists` function to decide this case here and then only have a `Streaming` message without further content that is used in `local()` to send the next block or end the stream. Your proposal is correct, though, and avoids one call to the database. I’m on the fence — have you tried the other way to see which one you like better?

crates/amaru-protocols/src/tests.rs
```diff
@@ -149,6 +134,56 @@ async fn start_initiator(addr: impl Into<Option<SocketAddr>>) -> anyhow::Result<
...
+            _ = responder.clone().join() => (),
+            _ = initiator.clone().join() => (),
+        }
+    })
+    .await?;
+    Ok(())
+}
+
+/// Verify that both nodes have the same state: same best chain, same blocks, same transactions.
+#[track_caller]
```
rkuhn: this is only needed when this function is called from multiple places

crates/amaru-protocols/src/tests.rs
```diff
@@ -210,22 +245,44 @@ fn setup_logging(enable: bool) {
...
-) {
+) -> ChainSyncStageState {
     use crate::chainsync::InitiatorResult::*;
     match msg.msg {
         Initialize => {
             tracing::info!(peer = %msg.peer,"initializing chainsync");
         }
         IntersectFound(point, tip) => {
-            tracing::info!(peer = %msg.peer, ?point, ?tip, "intersect found");
+            tracing::info!(peer = %msg.peer, %point, %tip, "intersect found");
```
rkuhn:
```suggestion
            tracing::info!(peer = %msg.peer, %point, tip_point = %tip.point(), "intersect found");
```

The issue with `%tip` is that it doesn’t show the slot number, which makes it harder to see the relation to `%point`.

crates/amaru-protocols/src/tests.rs
```diff
@@ -235,20 +292,79 @@ async fn test_chainsync_stage(
...
+                RESPONDER_BLOCKS_NB - 1 - state.total_requested_blocks;
+
+            // If the last batch isn't full but would allow us to complete the retrieval, we fetch it as well.
+            if state.blocks_to_fetch.len() == 3
+                || state.blocks_to_fetch.len() == remaining_number_of_blocks_to_retrieve
+            {
+                let from = *state.blocks_to_fetch.first().unwrap();
+                let through = *state.blocks_to_fetch.last().unwrap();
+                let blocks = eff
+                    .call(&state.manager, Duration::from_secs(200), move |cr| {
```
rkuhn: 200s seems a bit long :-)

crates/amaru-protocols/src/tests.rs
```diff
@@ -235,20 +292,79 @@ async fn test_chainsync_stage(
...
+                            &RawBlock::from(to_cbor(&block).as_slice()),
+                        )
+                        .unwrap();
+                }
+                state.blocks_to_fetch.clear();
+            };
+
+            if state.total_requested_blocks == RESPONDER_BLOCKS_NB - 1 {
+                tracing::info!("all blocks retrieved, done");
+                eff.terminate::<Void>().await;
```
rkuhn: this basically means that we assert that tx sub is faster than chainsync+blockfetch, which may well break on CI; allow tx sub to be slower would mean checking for completion of sending all tx there and sending a completion signal to a common stage (as well as from here) which only terminates once both are received

crates/amaru-protocols/src/tests.rs
```diff
@@ -308,18 +427,70 @@ fn initialize_chain_store(
...
     headers.insert(0, root_header);
 
-    for header in headers.iter() {
+    for (i, header) in headers.iter().enumerate() {
         chain_store.store_header(header)?;
         chain_store.roll_forward_chain(&header.point())?;
         chain_store.set_best_chain_hash(&header.hash())?;
+
+        // Add a block for each header
+        // We skip one block to test that the initiator can try to fetch missing blocks
```
rkuhn: this is unclear to me: it was my understanding that a range can only be returned if all blocks are present, otherwise NoBlocks. did I read wrong?

crates/amaru-protocols/src/tests.rs
```diff
@@ -308,18 +427,70 @@ fn initialize_chain_store(
...
+        } else {
+            current_hash = None;
+        }
+    }
+    blocks.reverse();
+    blocks
+}
+
+fn get_block_headers(store: Arc<dyn ChainStore<BlockHeader>>) -> Vec<HeaderHash> {
+    get_blocks(store)
```
rkuhn: why not use `retrieveBestChain` on the store?

crates/amaru-ouroboros-traits/src/stores/consensus/mod.rs
etorreborre: This one is a naive strategy indeed. The Haskell node returns an iterator on blocks when they are on the best chain, only if `from` starts before the chain anchor (before `k` blocks). Then, there's no limitation that I can see. Otherwise we are on a fork and the amount of streamed blocks is limited.

I propose to tackle this in a follow-up PR if that's ok with you.

crates/amaru-protocols/src/blockfetch/responder.rs
etorreborre: This might go away entirely in a follow-up PR.

crates/amaru-protocols/src/blockfetch/responder.rs
etorreborre: I haven't tried both, but I indeed wanted to avoid one call.

crates/amaru-protocols/src/tests.rs
etorreborre: Why don't we change the `Tip` `Display` instance to include the slot?

crates/amaru-protocols/src/tests.rs
etorreborre: I was not expecting to hit that timeout during the test since there's an overall timeout for the test :-).

crates/amaru-protocols/src/tests.rs
etorreborre: It was my impression that we could start streaming from several peers at the same time, where some peers might not yet have all the blocks on the required chain. So we respond in a best-effort manner. This is definitely something that we could contribute to the blueprint because the specification is not clear-cut on the subject.

crates/amaru-protocols/src/tests.rs
etorreborre: Because I forgot that this function existed :-).

crates/amaru-protocols/src/tests.rs
etorreborre: You're right, I'll fix that.

crates/amaru-protocols/src/tests.rs
etorreborre: Fixed.

crates/amaru-protocols/src/tests.rs
etorreborre: Removed.

crates/amaru-protocols/src/tests.rs
etorreborre: Done.

crates/amaru-protocols/src/tests.rs
```diff
@@ -146,44 +140,109 @@ async fn start_initiator(addr: impl Into<Option<SocketAddr>>) -> anyhow::Result<
...
+/// Timeout after a few seconds if that's not the case
+async fn wait_for_termination(
+    responder_done: Arc<Notify>,
+    initiator_done: Arc<Notify>,
+) -> anyhow::Result<()> {
+    let _ = tokio::time::timeout(Duration::from_secs(5), async {
+        tokio::join!(responder_done.notified(), initiator_done.notified());
+    })
+    .await;
     Ok(())
```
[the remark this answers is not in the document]
etorreborre: Adjusted the comment to explain why.

crates/amaru-protocols/src/tests.rs
rkuhn: I asked @KtorZ the same and we agreed to just log the Point instead. Ideally, he said, we'd log structured data, not a string.

crates/amaru-ouroboros-traits/src/stores/consensus/mod.rs
etorreborre: I added a commit to change this. We either stream blocks from the best chain or from a limited fork.

crates/amaru-protocols/src/blockfetch/responder.rs
etorreborre: Done in 2c963600

crates/amaru-protocols/src/tests.rs
```diff
@@ -146,44 +141,124 @@ async fn start_initiator(addr: impl Into<Option<SocketAddr>>) -> anyhow::Result<
...
+async fn wait_for_termination(
+    responder_done: Arc<Notify>,
+    initiator_done: Arc<Notify>,
+) -> anyhow::Result<()> {
+    let _ = tokio::time::timeout(Duration::from_secs(5000), async {
+        tokio::join!(responder_done.notified(), initiator_done.notified());
+    })
+    .await;
     Ok(())
 }
```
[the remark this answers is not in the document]
etorreborre: Done.

crates/amaru-protocols/src/blockfetch/responder.rs
```diff
@@ -36,32 +38,167 @@ pub struct BlockFetchResponder {
...
+                    let range = PointsRange::BestChain { from, through };
+                    Some(range)
+                }
+            } else {
+                tracing::debug!(%from, %through, "both points must be on the best chain");
+                None
+            }
+        } else {
+            // Otherwise the range must on a fork
+            tracing::debug!(%from, "requested 'from' point is not on the best chain");
```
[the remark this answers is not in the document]
etorreborre: done.

crates/amaru-protocols/src/tests.rs
etorreborre: The latest version will only stream block present on the best chain or a current fork (this might not actually be the case, they way we process headers at the moment. We need to work on that).

crates/amaru-protocols/src/blockfetch/responder.rs
```diff
@@ -36,32 +38,167 @@ pub struct BlockFetchResponder {
...
+                            Some((block, self))
+                        }
+                        Err(StoreError::NotFound { .. }) => None,
+                        Err(other) => panic!("{other:?}"),
+                    }
+                } else {
+                    None
+                }
+            }
+            PointsRange::Empty => None,
```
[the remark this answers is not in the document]
etorreborre: Good point. Fixed.

crates/amaru-protocols/src/blockfetch/initiator.rs
```diff
@@ -222,4 +318,172 @@ pub mod tests {
...
+        let header2 = make_header(2, 101, Some(wrong_parent_hash));
+        let block_header2 = BlockHeader::from(header2.clone());
+        let point2 = Point::Specific(101u64.into(), HeaderHash::from([2u8; 32]));
+
+        let blocks = vec![make_block(&block_header1), make_block(&block_header2)];
+
+        assert!(!BlockFetchInitiator::is_valid_block_range(
+            &blocks, point1, point2
+        ));
+    }
```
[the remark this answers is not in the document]
etorreborre: Done.

crates/amaru-ouroboros-traits/src/stores/consensus/mod.rs
```diff
@@ -92,6 +92,27 @@ where
...
+                }
+                if h.hash() == anchor {
+                    return None;
+                }
+            }
+            Some(result)
+        } else {
+            None
+        }
+    }
```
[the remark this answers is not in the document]
etorreborre: Done.

crates/amaru-protocols/src/tests.rs
```diff
@@ -308,18 +466,49 @@ fn initialize_chain_store(
...
+        .retrieve_best_chain()
+        .iter()
+        .filter_map(|h| {
+            store
+                .load_block(h)
+                .ok()
+                .map(|b| (*h, cbor::decode(b.deref()).unwrap()))
+        })
+        .collect()
+}
```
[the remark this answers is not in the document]
etorreborre: Done.

crates/amaru-stores/src/rocksdb/consensus/mod.rs
```diff
@@ -826,6 +826,70 @@ pub mod test {
...
+        with_db(|store| {
+            let chain = populate_db(store.clone());
+            store.set_anchor_hash(&chain[3].hash()).unwrap();
+            let from = chain[1].point();
+            let to = chain[5].point();
+            let result = store.ancestors_points(&from, &to);
+
+            // Should return None because the ancestry walk breaks before reaching from_inclusive
+            assert_eq!(result, None);
+        });
```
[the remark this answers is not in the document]
etorreborre: Done.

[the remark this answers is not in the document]
etorreborre: Done.

crates/amaru-protocols/src/tests.rs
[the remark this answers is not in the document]
etorreborre: Fixed

crates/amaru-kernel/src/is_header/block_header.rs
```diff
@@ -0,0 +1,285 @@
...
+            self.hash(),
+            self.parent_hash()
+                .map(|p| format!(" ({p})"))
+                .unwrap_or_default()
+        ))?;
+        Ok(())
+    }
+}
+
+/// We serialize both the hash and the header, but we use serde's flattening
```
rkuhn, line 52: Why serialize the hash, which is strictly redundant?

crates/amaru-kernel/src/is_header/block_header.rs
```diff
@@ -0,0 +1,285 @@
...
+use std::{
+    cmp::Ordering,
+    fmt::{self, Debug, Display, Formatter},
+};
+
+/// This header type encapsulates a header and its hash to avoid recomputing
+#[derive(PartialEq, Eq, Clone)]
+pub struct BlockHeader {
+    pub(super) header: Header,
+    pub(super) hash: HeaderHash,
```
rkuhn: If we want to avoid hashing then this should be a lazy property. Arguably, this doesn’t even need a lock because the result of hashing is deterministic. But we’d need to remove the `pub(super)` to make it easier to see that the `Header` can never be modified. I might look into it if this train ride is long enough.

crates/amaru-kernel/src/is_header/block_header.rs
```diff
@@ -0,0 +1,285 @@
...
+    /// For blocks in stored format (without era tag), use `extract_header_from_stored_block` instead.
+    pub fn extract_header(block_bytes: &[u8]) -> Result<BlockHeader, cbor::decode::Error> {
+        let mut decoder = Decoder::new(block_bytes);
+
+        // Network format: [era_tag, MintedBlock]
+        // Read the outer array (should be length 2)
+        decoder.array()?;
+
+        // Skip the era tag (u16)
+        decoder.u16()?;
```
rkuhn: shouldn’t we validate that we can actually decode this era?

crates/amaru-kernel/src/is_header/block_header.rs
```diff
@@ -0,0 +1,285 @@
...
+
+impl PartialOrd for BlockHeader {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for BlockHeader {
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.hash.cmp(&other.hash())
```
rkuhn: while this is probabilistically correct, I’d like to ask what the intended usage is — perhaps it makes more sense to compare slot numbers first

crates/amaru-protocols/src/blockfetch/initiator.rs
```diff
@@ -86,6 +86,91 @@ impl BlockFetchInitiator {
...
+    #[expect(clippy::expect_used)]
+    pub fn is_valid_block_range(blocks: &[Vec<u8>], from: Point, through: Point) -> bool {
+        assert!(
+            !blocks.is_empty(),
+            "some blocks should have been fetched from {from} to {through}"
+        );
+
+        // Extract headers from all blocks
+        let mut headers = Vec::with_capacity(blocks.len());
+        for (idx, block_bytes) in blocks.iter().enumerate() {
```
rkuhn: should check for max block size here

crates/amaru-protocols/src/blockfetch/initiator.rs
```diff
@@ -126,9 +211,18 @@ impl StageState<State, Initiator> for BlockFetchInitiator {
...
-                eff.send(&cr, Blocks { blocks }).await;
+                if BlockFetchInitiator::is_valid_block_range(&blocks, from, through) {
+                    eff.send(&cr, Blocks { blocks }).await;
+                } else {
+                    tracing::warn!(
+                        ?from,
+                        ?through,
+                        "received blocks do not form a valid range; terminating the connection"
+                    );
+                    eff.terminate::<Void>().await;
```
rkuhn: the idiom is `return eff.terminate().await` because that tells the compiler that this line diverges (and fixes the type)

crates/amaru-protocols/src/blockfetch/initiator.rs
```diff
@@ -86,6 +86,91 @@ impl BlockFetchInitiator {
...
     }
+
+    /// Return true if the provided blocks form a valid chain from `from` to `through`.
+    /// This includes checks that:
+    /// - The first block matches the `from` point.
+    /// - The last block matches the `through` point.
+    /// - Each block's slot is strictly greater than the previous block's slot.
+    /// - Each block's parent hash matches the hash of the previous block.
+    #[expect(clippy::expect_used)]
+    pub fn is_valid_block_range(blocks: &[Vec<u8>], from: Point, through: Point) -> bool {
```
rkuhn: why not return an error and let the caller worry about printing the error?

crates/amaru-protocols/src/blockfetch/responder.rs
```diff
@@ -36,32 +38,201 @@ pub struct BlockFetchResponder {
...
 }
 
+/// This data type represents a range of points to fetch blocks for.
+/// It can represent either a range along the best chain (from `from` to `through` inclusive),
+/// or a list of points representing a fork. The main difference is that the points in a fork cannot
+/// go beyond the current best chain anchor.
+#[derive(Debug, PartialEq, Eq, Clone, serde::Serialize, serde::Deserialize)]
+pub enum PointsRange {
+    BestChain { from: Point, through: Point },
+    Fork(Vec<Point>),
```
rkuhn: not sure I understand: why would we ever fetch blocks for headers which are not on the current best chain candidate? and if these are to be fetched, then shouldn’t they be characterised as a range as well?

crates/amaru-protocols/src/blockfetch/responder.rs
```diff
@@ -36,32 +38,201 @@ pub struct BlockFetchResponder {
...
+    thread_local! {
+        static BUFFER: RefCell<Vec<u8>> = const { RefCell::new(Vec::new()) };
+    }
+
+    BUFFER.with_borrow_mut(|buffer| {
+        let mut encoder = cbor::Encoder::new(&mut *buffer);
+        // Encode just the wrapper: [era_tag, ...]
+        encoder.array(2).expect("failed to encode outer array");
+        encoder
+            .u16(amaru_kernel::CONWAY_ERA_TAG)
```
rkuhn: this should probably be done via slot arithmetic so that we’ll have an easier time updating it after relevant hard forks

crates/amaru-protocols/src/blockfetch/responder.rs
```diff
@@ -36,32 +38,201 @@ pub struct BlockFetchResponder {
...
+    BestChain { from: Point, through: Point },
+    Fork(Vec<Point>),
+    Empty,
+}
+
+/// Wrap a stored format block in network format by adding the era tag wrapper.
+/// Network format: [era_tag: u16, stored_block]
+/// This converts blocks from database format to the format expected by the block fetch protocol.
+#[expect(clippy::expect_used)]
+fn wrap_in_network_format(stored_block: &RawBlock) -> RawBlock {
```
rkuhn: seeing this function signature creeps me out: one of these should not be a RawBlock but a different type, because the data format is different (unless we have methods on RawBlock to see this dynamically, but still, the design looks like we distinguish this statically anyway)

crates/amaru-protocols/src/blockfetch/responder.rs
```diff
@@ -36,32 +38,201 @@ pub struct BlockFetchResponder {
...
+    /// Each time we attempt to fetch a block we pop its point from the current_range.
+    #[expect(clippy::panic)]
+    async fn next_block<T>(mut self, effects: Effects<T>) -> Option<(RawBlock, PointsRange)> {
+        let store = Store::new(effects);
+        match self {
+            PointsRange::BestChain { from, through } => match store.load_block(&from.hash()) {
+                Ok(block) => {
+                    if from == through {
+                        self = PointsRange::Empty;
+                    } else if let Some(next) = store.next_best_chain(&from) {
```
rkuhn: This won’t work: the stored best chain may well change while we are serving these blocks. I think PointsRange should collapse into a `Vec<HeaderHash>`.

crates/amaru-protocols/src/blockfetch/responder.rs
```diff
@@ -36,32 +38,201 @@ pub struct BlockFetchResponder {
...
+                }
+            } else {
+                tracing::debug!(%from, %through, "both points must be on the best chain");
+                None
+            }
+        } else {
+            // Otherwise the range must on a fork
+            tracing::debug!(%from, "requested 'from' point is not on the best chain. It might be on a fork");
+
+            if let Some(points) = store.ancestors_points(&from, &through) {
```
rkuhn: `ancestors_points` yields a sequence from newest to oldest while the blocks are shipped beginning to end in `PointsRange::Fork()`; isn’t that the wrong order?

crates/amaru-protocols/src/blockfetch/responder.rs
```diff
@@ -36,32 +38,201 @@ pub struct BlockFetchResponder {
...
+                        eff.me_ref(),
+                        Inputs::Local(BlockStreaming::NextBlock(points_range)),
+                    )
+                    .await;
+                    Ok((Some(ResponderAction::Block(block.to_vec())), self))
+                } else {
+                    // No more blocks to send, finish the batch.
+                    eff.send(eff.me_ref(), Inputs::Local(BlockStreaming::Done))
+                        .await;
+                    Ok((None, self))
```
rkuhn: why the extra loop instead of sending BatchDone from here?

crates/amaru-kernel/src/is_header/block_header.rs
etorreborre: That's because those fields get eventually embedded in traces and that makes it easier for the visualizations to pick up the hash of messages being tracked.

crates/amaru-kernel/src/is_header/block_header.rs
etorreborre: We always want to hash but the previous version was rehashing all the time. That's why this was introduced.
It was also to help with building test data IIRC.

crates/amaru-kernel/src/is_header/block_header.rs
etorreborre: Yes, done.

crates/amaru-kernel/src/is_header/block_header.rs
etorreborre: Yes, we should be comparing points instead. Done.

crates/amaru-protocols/src/blockfetch/initiator.rs
etorreborre: I missed that, thanks.

crates/amaru-protocols/src/blockfetch/responder.rs
etorreborre: When we are fetching a block we don't know yet if it is on the best chain of an upstream peer.
The other thing is that, in the database, we only maintain one forward chain, for the best chain. 
For that chain, we can stream all the points (and then all the blocks) without having to materialize the full range of headers. For forks, we can do this and we just materialize the headers up to the current anchor.

crates/amaru-protocols/src/blockfetch/initiator.rs
etorreborre: I didn't see the point of returning an error since that function is just used once to make a decision. I removed the `pub` now, since it's really a local function.

crates/amaru-protocols/src/blockfetch/responder.rs
etorreborre: Yes, I was thinking of leaving that for later but I'll do it now along with the next comment.

crates/amaru-protocols/src/blockfetch/responder.rs
etorreborre: Yes, I agree, I don't feel good about that too. I'm going to change it.

crates/amaru-protocols/src/blockfetch/responder.rs
etorreborre: Right, I propose to stream point before the anchor and materialize the rest after the anchor.

crates/amaru-protocols/src/blockfetch/responder.rs
etorreborre: Oops, totally. I should have added a test for this.

crates/amaru-protocols/src/blockfetch/responder.rs
etorreborre: Indeed.

crates/amaru-protocols/src/blockfetch/responder.rs
etorreborre: Done. I made a distinction between `NetworkBlock` containing an `era_tag` + a `RawBlock`, and the `RawBlock` that is just a `Vec<u8>`.

crates/amaru-protocols/src/blockfetch/responder.rs
etorreborre: Done.

crates/amaru-kernel/src/is_header/block_header.rs
rkuhn: Ah, then this reasoning should be included in the doc comment. And the deserialization should then probably also check that the hash is correct and raise an Err() if not.

crates/amaru-kernel/src/is_header/block_header.rs
rkuhn: Okay, then laziness wouldn’t buy us anything. I still think that `pub(super)` should be removed to make it easy to ensure that `header` is never mutated.

crates/amaru-protocols/src/blockfetch/responder.rs
rkuhn: :+1:

crates/amaru-protocols/src/blockfetch/responder.rs
rkuhn: It was my understanding that there is no data older than the anchor within the chain store. If we do have older data then the anchor is just an older block. There shouldn’t be an artificial distinction that leads to two code paths — we have only one chain store for everything anyway.

crates/amaru-protocols/src/blockfetch/responder.rs
```diff
@@ -34,34 +42,170 @@ pub fn responder() -> Miniprotocol<State, BlockFetchResponder, Responder> {
...
+        loop {
+            if result.len() > MAX_BLOCKS {
+                tracing::debug!(
+                    %from,
+                    %through,
+                    max_blocks = MAX_BLOCKS,
+                    "requested range exceeds maximum allowed blocks"
+                );
+                return None;
+            }
```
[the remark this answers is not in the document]
etorreborre: Done.

crates/amaru-protocols/src/blockfetch/responder.rs
etorreborre: Reverted to a simpler choice of loading a limited list of headers.

crates/amaru-protocols/src/blockfetch/responder.rs
etorreborre: This is simplified now.

crates/amaru-ouroboros-traits/src/stores/consensus/mod.rs
```diff
@@ -92,6 +98,24 @@ where
...
+                result.push(h.point());
+                if &h.point() == from_inclusive {
+                    return Some(result);
+                }
+            }
+            None
+        } else {
+            None
+        }
+    }
```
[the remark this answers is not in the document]
etorreborre: Removed that code.

crates/amaru-protocols/src/blockfetch/responder.rs
```diff
@@ -34,34 +42,170 @@ pub fn responder() -> Miniprotocol<State, BlockFetchResponder, Responder> {
...
+            if let Some(parent_hash) = header.parent_hash() {
+                current_hash = parent_hash
+            } else {
+                return None;
+            }
+        }
+        result.reverse();
+        Some(PointsRange(result))
+    }
```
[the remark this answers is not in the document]
etorreborre: Good point, done.

crates/amaru-protocols/src/blockfetch/initiator.rs
```diff
@@ -117,18 +241,45 @@ impl StageState<State, Initiator> for BlockFetchInitiator {
...
+                    } else {
+                        tracing::warn!(
+                            "the responder sent more {MAX_FETCHED_BLOCKS} blocks; terminating the connection"
+                        );
+                        return eff.terminate().await;
+                    }
+                } else {
+                    tracing::warn!("received invalid block CBOR; terminating the connection");
+                    return eff.terminate().await;
+                }
```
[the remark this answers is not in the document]
etorreborre: Done.

crates/amaru/src/stages/consensus/forward_chain/client_protocol.rs
```diff
@@ -344,10 +344,13 @@ async fn block_fetch<H: IsHeader>(
...
-        server.send_start_batch().await?;
-        server.send_block(block.to_vec()).await?;
-        server.send_batch_done().await?;
+        if let Some(block) = store.load_block(&Hash::from(&from_network_point(&lb_point)))? {
+            server.send_start_batch().await?;
+            server.send_block(block.to_vec()).await?;
+            server.send_batch_done().await?;
+        } else {
+            return Err(ClientError::CannotServeRange(lb_point, ub_point).into());
+        }
```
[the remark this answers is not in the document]
etorreborre: The forwarding will be reworked soon.

crates/amaru-ouroboros-traits/src/stores/consensus/mod.rs
```diff
@@ -62,14 +62,20 @@ where
...
         let anchor = self.get_anchor_hash();
+        let anchor_point = match self.load_header(&anchor) {
+            Some(header) => header.point(),
+            None => panic!("Anchor header not found in chain store"),
+        };
+
         Box::new(successors(Some(start), move |h| {
-            if h.hash() == anchor {
+            if h.slot() <= anchor_point.slot_or_default() {
                 None
```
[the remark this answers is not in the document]
etorreborre: Fixed by returning `Point::Origin`

crates/amaru-protocols/src/blockfetch/initiator.rs
```diff
@@ -224,4 +377,249 @@ pub mod tests {
...
+        let mut incomplete_bytes = Vec::new();
+        let mut encoder = cbor::Encoder::new(&mut incomplete_bytes);
+        encoder.array(2).expect("failed to encode array");
+        encoder.u16(1).expect("failed to encode tag");
+        encoder.array(5).expect("failed to encode inner array");
+        let raw_block = RawBlock::from(incomplete_bytes.as_slice());
+
+        // from raw block should work since the era tag is present
+        NetworkBlock::try_from(raw_block).unwrap()
+    }
```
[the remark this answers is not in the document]
etorreborre: Done.

crates/amaru-protocols/src/blockfetch/responder.rs
```diff
@@ -36,32 +39,176 @@ pub struct BlockFetchResponder {
...
+            tracing::debug!(%from, %through, "requested range is invalid: from > through");
+            return Ok(None);
+        };
+
+        if from == through {
+            return if store.load_block(&from.hash())?.is_some() {
+                Ok(Some(PointsRange::singleton(from)))
+            } else {
+                Ok(None)
+            };
```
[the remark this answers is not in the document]
etorreborre: That's ok.

crates/amaru-protocols/src/blockfetch/mod.rs
```diff
@@ -79,13 +79,14 @@ pub async fn register_blockfetch_initiator<M>(
...
     conn_id: ConnectionId,
+    era_history: Arc<amaru_slot_arithmetic::EraHistory>,
     eff: &Effects<M>,
 ) -> StageRef<BlockFetchMessage> {
     use crate::protocol::PROTO_N2N_BLOCK_FETCH;
     let blockfetch = eff
         .wire_up(
             eff.stage("blockfetch", initiator()).await,
-            BlockFetchInitiator::new(muxer.clone(), peer, conn_id),
+            BlockFetchInitiator::new(muxer.clone(), peer, conn_id, era_history.clone()),
```
rkuhn: unnecessary clone?

crates/amaru-protocols/src/blockfetch/responder.rs
```diff
@@ -36,32 +39,176 @@ pub struct BlockFetchResponder {
     muxer: StageRef<MuxMessage>,
 }
 
+/// This data type represents a range of points to fetch blocks for.
+/// The points are ordered from the oldest to the most recent and at least one point is present
+#[derive(Debug, PartialEq, Eq, Clone, serde::Serialize, serde::Deserialize)]
+pub struct PointsRange {
+    first: Point,
+    rest: Vec<Point>,
```
rkuhn: why not use NonEmptyVec?

crates/amaru-protocols/src/blockfetch/responder.rs
```diff
@@ -36,32 +39,176 @@ pub struct BlockFetchResponder {
...
+        self,
+        store: &dyn ChainStore<BlockHeader>,
+    ) -> anyhow::Result<(RawBlock, Option<PointsRange>)> {
+        let stored_block = store
+            .load_block(&self.first.hash())?
+            .ok_or_else(|| anyhow::anyhow!("block {} was pruned", self.first.hash()))?;
+        let points_range = self.rest.split_first().map(|(first, rest)| PointsRange {
+            first: *first,
+            rest: rest.to_vec(),
+        });
```
rkuhn: it would be vastly more efficient to store in the reversed order and pop instead of moving everything for every operation

crates/amaru-protocols/src/blockfetch/responder.rs
```diff
@@ -36,32 +39,176 @@ pub struct BlockFetchResponder {
...
+            if store.load_block(&current_hash)?.is_none() {
+                return Ok(None);
+            }
+
+            // load the header for the current hash
+            if let Some(header) = store.load_header(&current_hash) {
+                result.push(header.point());
+                if current_hash == from.hash() {
+                    break;
+                }
```
rkuhn, line 128: should bail if the parent slot number is <= the from slot number

crates/amaru-protocols/src/blockfetch/responder.rs
```diff
@@ -140,16 +290,265 @@ pub enum ResponderResult {
...
+        )
+        .unwrap();
+
+        assert_eq!(result.unwrap().points().len(), MAX_FETCHED_BLOCKS);
+    }
+
+    #[test]
+    fn test_request_range_max_blocks_limit() {
+        // Create a chain longer than MAX_BLOCKS
+        let chain_length = MAX_FETCHED_BLOCKS + 10;
```
rkuhn: I'd use +1 here to test exact overflow detection

crates/amaru-protocols/src/connection.rs
```diff
@@ -56,6 +59,7 @@ impl Connection {
                 role,
                 magic,
                 pipeline,
+                era_history: era_history.clone(),
```
rkuhn: again: unnecessary clone?

crates/amaru-protocols/src/keepalive/tests.rs
```diff
@@ -46,6 +47,7 @@ fn test_keepalive_with_node() {
         .resources()
         .put::<ConnectionResource>(Arc::new(conn));
 
+    let era_history: &EraHistory = NetworkName::Preprod.into();
```
rkuhn, line 50: For these I strongly prefer something like `EraHistory::for_network(Preprod)` (and I'd create that associated function if it doesn't yet exist).

crates/amaru-protocols/src/manager.rs
```diff
@@ -61,11 +63,16 @@ enum ConnectionState {
...
+    pub fn new(
+        magic: NetworkMagic,
+        chain_sync: StageRef<ChainSyncInitiatorMsg>,
+        era_history: Arc<amaru_slot_arithmetic::EraHistory>,
+    ) -> Self {
         Self {
             peers: BTreeMap::new(),
             magic,
             chain_sync,
+            era_history: era_history.clone(),
```
rkuhn: unnecessary clone?

crates/amaru-stores/src/rocksdb/consensus/mod.rs
```diff
@@ -221,14 +221,13 @@ macro_rules! impl_ReadOnlyChainStore {
...
-                self.db
+            fn load_block(&self, hash: &HeaderHash) -> Result<Option<RawBlock>, StoreError> {
+                Ok(self.db
                     .get_pinned([&BLOCK_PREFIX[..], &hash[..]].concat())
                     .map_err(|e| StoreError::ReadError {
                         error: e.to_string(),
                     })?
-                    .ok_or(StoreError::NotFound { hash: *hash })
-                    .map(|bytes| bytes.as_ref().into())
+                    .map(|bytes| bytes.as_ref().into()))
```
rkuhn, line 230: I agree with separating errors from nominal results. soon we will have more than two success cases, namely the new "block exists but was invalid". Perhaps while you're at it you could introduce a BlockStat enum (needs better name) with Known, NotKnown, Invalid variants.

crates/amaru-protocols/src/blockfetch/responder.rs
etorreborre: I didn't see that we had one. It's missing some operations. I added them to use it.

crates/amaru-protocols/src/blockfetch/responder.rs
etorreborre: Done with the `NonEmptyVec` refactor.

crates/amaru-protocols/src/keepalive/tests.rs
etorreborre: That wouldn't work with the crates dependencies because `amaru-kernel` depends on `amaru-slot-arithmetic`.

crates/amaru-protocols/src/blockfetch/responder.rs
etorreborre: That's a good idea.

crates/amaru-stores/src/rocksdb/consensus/mod.rs
etorreborre: I'd rather introduce it when we need it, but also I'm not sure about what you're referring to. What is `NotKnown`?

crates/amaru-protocols/src/blockfetch/responder.rs
```diff
@@ -36,32 +39,171 @@ pub struct BlockFetchResponder {
...
+                if let Some(parent_hash) = header.parent_hash() {
+                    current_hash = parent_hash
+                } else {
+                    return Ok(None);
+                }
+            } else {
+                return Ok(None);
+            }
+        }
+        Ok(PointsRange::from_vec(result))
```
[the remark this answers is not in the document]
etorreborre: That's ok because when the PointsRange is used, its points are popped in the reverse order.

crates/amaru-protocols/src/keepalive/tests.rs
rkuhn: Ah,  but then we can make it `Preprod.as_era_history()`, which conveniently returns the reference. That’s much more discoverable.

crates/amaru-stores/src/rocksdb/consensus/mod.rs
rkuhn: In light of today’s meeting we’ll probably need four states:

1. NotKnown — we have no knowledge of such a block
2. Unvalidated — we fetched it but it hasn’t been sent to the ledger yet, e.g. because it’s on a losing fork
3. Invalid — we checked it and didn’t like it
4. Valid — can  be used downstream with a clear conscience

Variants 2–4 may as well contain the bytes because we fetched them from disk already.

rkuhn:
Looking good. The two remaining things can also be done later.

PR #653, opened 2026-01-26
test: add a test showing that an initiator can eventually connect when the responder is started

etorreborre:
The PR makes sure that we can connect to an upstream node if:

 - It is not yet started when we start.
 - It is restarted while we are connected.

In order to do this:

 - The reading/writing stages in the `Muxer` are supervised.
 - The `Muxer` itself is supervised by the `Connection` which closes the loop since the `Connection` is supervised by the `Manager`.
 - `FetchBlock` errors are not considered as fatal anymore since they can happen on disconnection.
 - Similarly a rollback can fail on reconnection if the corresponding block was not yet applied to the ledger
    -> This seems to be a fundamental issue with how we maintain consistency between the ledger and the chain store. This will be addressed in another PR.
 
 In addition:

 - Two tests were added to support the 2 disconnection scenarii presented earlier.
 - The tests were refactored since the setup and the assertions contains more code now.
 - The `Muxer` has a `role` field that is used to annotate traces, which is useful when reading log messages.
 - The connection timeout that was hard-coded in the `Manager` is now part of its configuration.

## Summary by CodeRabbit

* **New Features**
  * Role-aware networking: connection role is now propagated and shown in logs/UI.
  * Network blocks now include encoded payloads for improved decoding.

* **Improvements**
  * Manager connection and reconnect timings are configurable via a new ManagerConfig.
  * Consensus validation and rollback handling refined to distinguish disconnect vs. investigatory validation errors.

* **Tests**
  * Large rework: new modular end-to-end and unit test harness and utilities; previous monolithic tests replaced.

### DIFF
--- a/crates/amaru-protocols/src/connection.rs
+++ b/crates/amaru-protocols/src/connection.rs
@@ -164,7 +164,6 @@ async fn do_initialize(
     eff: Effects<ConnectionMessage>,
 ) -> State {
     let muxer = eff.stage("mux", mux::stage).await;
-    let muxer = eff.supervise(muxer, ConnectionMessage::Disconnect);
     let muxer = eff
         .wire_up(
             muxer,
--- a/crates/amaru-protocols/src/manager.rs
+++ b/crates/amaru-protocols/src/manager.rs
@@ -87,6 +87,18 @@ pub struct ManagerConfig {
     pub reconnect_delay: Duration,
 }
 
+impl ManagerConfig {
+    pub fn with_reconnect_delay(mut self, reconnect_delay: Duration) -> Self {
+        self.reconnect_delay = reconnect_delay;
+        self
+    }
+
+    pub fn with_connection_timeout(mut self, connection_timeout: Duration) -> Self {
+        self.connection_timeout = connection_timeout;
+        self
+    }
+}
+
 impl Default for ManagerConfig {
     fn default() -> Self {
         Self {
--- a/crates/amaru-protocols/src/tests/configuration.rs
+++ b/crates/amaru-protocols/src/tests/configuration.rs
@@ -36,7 +36,7 @@ pub(super) struct Configuration {
     pub(super) chain_store: Arc<dyn ChainStore<BlockHeader>>,
     pub(super) mempool: Arc<InMemoryMempool<Transaction>>,
     pub(super) addr: SocketAddr,
-    pub(super) connection_timeout: Duration,
+    pub(super) reconnect_delay: Duration,
     pub(super) processing_wait: Option<Duration>,
     pub(super) chain_length: usize,
     pub(super) slow_manager: bool,
@@ -49,7 +49,7 @@ impl Configuration {
             chain_length: 0,
             mempool: Arc::new(InMemoryMempool::default()),
             addr: SocketAddr::from(([127, 0, 0, 1], 3000)),
-            connection_timeout: Duration::from_secs(1),
+            reconnect_delay: Duration::from_secs(1),
             processing_wait: None,
             slow_manager: false,
         };
@@ -64,7 +64,7 @@ impl Configuration {
             chain_length: 0,
             mempool: Arc::new(InMemoryMempool::default()),
             addr: SocketAddr::from(([127, 0, 0, 1], 0)),
-            connection_timeout: Duration::from_secs(1),
+            reconnect_delay: Duration::from_secs(1),
             processing_wait: None,
             slow_manager: false,
         };
@@ -109,8 +109,8 @@ impl Configuration {
         self
     }
 
-    pub(super) fn with_connection_timeout(mut self, timeout: Duration) -> Self {
-        self.connection_timeout = timeout;
+    pub(super) fn with_reconnect_delay(mut self, timeout: Duration) -> Self {
+        self.reconnect_delay = timeout;
         self
     }
 
--- a/crates/amaru-protocols/src/tests/test_cases.rs
+++ b/crates/amaru-protocols/src/tests/test_cases.rs
@@ -54,7 +54,7 @@ async fn test_connect_initiator_reconnection() -> anyhow::Result<()> {
     let (initiator, initiator_done) = start_initiator_with_configuration(
         Configuration::initiator()
             .with_addr(addr)
-            .with_connection_timeout(Duration::from_millis(500)),
+            .with_reconnect_delay(Duration::from_millis(500)),
     )
     .await?;
     tokio::time::sleep(Duration::from_secs(2)).await;
@@ -77,7 +77,7 @@ async fn test_connect_initiator_reconnection_on_responder_restart() -> anyhow::R
     let (initiator, initiator_done) = start_initiator_with_configuration(
         Configuration::initiator()
             .with_addr(addr)
-            .with_connection_timeout(Duration::from_millis(500))
+            .with_reconnect_delay(Duration::from_millis(500))
             .with_processing_wait(Duration::from_millis(100)),
     )
     .await?;
@@ -201,10 +201,12 @@ async fn start_initiator_with_configuration(
         ),
     );
 
+    let manager_config =
+        ManagerConfig::default().with_reconnect_delay(configuration.reconnect_delay);
     let era_history: &EraHistory = NetworkName::Preprod.into();
     let initiator_manager = Manager::new(
         NetworkMagic::PREPROD,
-        ManagerConfig::default(),
+        manager_config,
         chainsync_stage.without_state(),
         Arc::new(era_history.clone()),
     );


crates/amaru-consensus/src/errors.rs
```diff
@@ -76,9 +76,11 @@ pub enum ConsensusError {
     #[error("{0}")]
     InvalidHeaderParent(Box<InvalidHeaderParentData>),
     #[error("Failed to roll forward chain from {0}: {1}")]
-    RollForwardChainFailed(amaru_kernel::Hash<32>, amaru_ouroboros::StoreError),
+    RollForwardChainFailed(amaru_kernel::Hash<32>, StoreError),
     #[error("Failed to rollback chain at {0}: {1}")]
-    RollbackChainFailed(Point, amaru_ouroboros::StoreError),
+    RollbackChainFailed(Point, StoreError),
+    #[error("Failed to rollback block at {0}: {1}")]
+    RollbackBlockFailed(Point, BlockValidationError),
```
rkuhn, line 80: This needs some more explanation: my understanding was that a rollback can only target a Point that has been locally validated before (i.e. chains back to origin, all blocks valid, etc.). How could we end up processing a rollback to a point we didn’t previously validate?

crates/amaru-protocols/src/tests/setup.rs
```diff
@@ -0,0 +1,78 @@
...
+    };
+    let _ = tracing_subscriber::fmt()
+        .with_env_filter(EnvFilter::from_default_env())
+        .with_test_writer()
+        .try_init();
+}
+
+/// Get an ephemeral localhost address by binding to port 0
+/// Be careful, this does not guarantee that the port will always remain free!
+pub(super) fn ephemeral_localhost_addr() -> anyhow::Result<SocketAddr> {
```
rkuhn, line 45: Yes, this is a pattern that is used in many places, but I try to avoid it whenever possible due to the inherent reliability issues. Can we not bind just once and communicate the result to the connecting entity?

crates/amaru-protocols/src/tests/test_cases.rs
```diff
@@ -0,0 +1,229 @@
...
+
+#[tokio::test]
+async fn test_connect_initiator_reconnection() -> anyhow::Result<()> {
+    setup_logging(true);
+    let addr = ephemeral_localhost_addr()?;
+    tracing::info!("starting test at address {}", addr);
+    let (initiator, initiator_done) = start_initiator_at_with_configuration(
+        Configuration::initiator()
+            .with_addr(addr)
+            .with_connection_timeout(Duration::from_millis(500)),
```
rkuhn: This test case aims to produce an error before the Manager gets the accepted connection message. Using the `ephemeral_localhost_addr` usually achieves this goal. I’m wondering whether we should assert even more control by using [netsim-embed](https://crates.io/crates/netsim-embed), we used that quite extensively to test actyx and had David on the team (these two facts are of course highly correlated). This would allow us to drop packets for a while to prevent a connection, then allow it and observe the sync. We could also allow the TCP handshake and then drop all following packets.

This PR is already quite useful, my above suggestions are for future extensions. One caveat of netsim is that is doesn’t work on Windows (it does work in WSL, though) because it uses the same container infrastructure as Docker and friends.

crates/amaru-protocols/src/tests/test_cases.rs
```diff
@@ -0,0 +1,229 @@
...
+    wait_for_termination(responder_done, initiator_done).await?;
+    check_state(initiator, responder)?;
+
+    Ok(())
+}
+
+#[tokio::test]
+async fn test_connect_initiator_reconnection_on_responder_restart() -> anyhow::Result<()> {
+    setup_logging(true);
+    let addr = ephemeral_localhost_addr()?;
```
rkuhn: In this test, I’d just get the address from the first responder bind.

crates/amaru-protocols/src/tests/test_cases.rs
```diff
@@ -0,0 +1,229 @@
...
+) -> anyhow::Result<(TokioRunning, Arc<Notify>)> {
+    let addr = addr
+        .into()
+        .unwrap_or_else(|| SocketAddr::from(([127, 0, 0, 1], 3000)));
+    start_initiator_at_with_configuration(Configuration::initiator().with_addr(addr)).await
+}
+
+/// Create and start an initiator node that connects to the given port.
+/// ChainSync events are sent to a stage that stores them in the in-memory store without further processing.
+async fn start_initiator_at_with_configuration(
```
rkuhn: The `_at_` sounds misplaced to me, how about start_initiator_with_configuration?

crates/amaru-protocols/src/mux.rs
```diff
@@ -373,16 +387,26 @@ impl Header {
 #[expect(clippy::disallowed_types)]
 type Protocols = HashMap<ProtocolId<Erased>, PerProto>;
 
-#[derive(Debug, Default, PartialEq, serde::Serialize, serde::Deserialize)]
+#[derive(Debug, PartialEq, serde::Serialize, serde::Deserialize)]
 pub struct Muxer {
     protocols: Protocols,
     outgoing: Vec<ProtocolId<Erased>>,
     next_out: usize,
+    role: Role,
```
rkuhn, line 395: The role seems to be used only for log messages, correct? I didn’t put it in because the muxer doesn’t care about roles much, and the same muxer will also be used with both initiator and responder roles for different protocols (once we support full duplex connections). So what is the intended use of this information?

etorreborre: It's indeed used to make sense of the logs. Maybe once we go full duplex we will just see "full duplex" in there, but still see `initiator` and `responder` in those tests, which is still useful?

crates/amaru-protocols/src/tests/test_cases.rs
etorreborre: Yes, that was a remnant of passing parameters separately

crates/amaru-consensus/src/errors.rs
etorreborre: This happened because of the bug fixed in #656. Once merged, I'll rebase and make it a processing error.

crates/amaru-protocols/src/tests/setup.rs
etorreborre: The only case where it is used now is when the responder is not yet available at a specific address. I suppose that the next best thing is what you suggest below.

crates/amaru-protocols/src/tests/test_cases.rs
etorreborre: Being able to simulate very specific network condition is very nice indeed. I wonder how much of that can already be simulated simply via our simulation tests when we put all of the network stack under test?

rkuhn: Indeed, that’s also possible, good point! The main challenge I see here is to come up with a usable API for describing the test scenario within the simulation testing framework. So far, I saw the simulation testing as a random exploration of how the system unfolds in a static world, but we can of course also add changing world conditions (i.e. dropped or delayed messages).

crates/amaru-protocols/src/mux.rs
rkuhn: Ah, now I understand. With the current tests this split into two is sufficient, but going forward we’ll have to split the test participants’ logs or tag them somehow, otherwise we will indeed not be able to make sense of them. Good point!

crates/amaru-protocols/src/tests/setup.rs
rkuhn: yeah, good enough for now

crates/amaru/src/stages/build_stage_graph.rs
```diff
@@ -56,12 +56,37 @@ pub fn build_stage_graph(
         with_consensus_effects(forward_chain::stage),
     );
 
-    // TODO: currently only validate_header errors, will need to grow into all error handling
     let validation_errors_stage =
         network.stage("validation_errors", async move |manager, error, eff| {
             let ValidationFailed { peer, error } = error;
-            tracing::error!(%peer, %error, "peer error");
-            eff.send(&manager, ManagerMessage::RemovePeer(peer)).await;
+            use super::ConsensusError::*;
```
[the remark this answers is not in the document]
etorreborre: This compiles

crates/amaru-protocols/src/tests/assertions.rs
```diff
@@ -0,0 +1,115 @@
...
+/// We timeout after a bit to avoid hanging tests. In that case the test will fail later when checking the state.
+pub(super) async fn wait_for_termination(
+    responder_done: Arc<Notify>,
+    initiator_done: Arc<Notify>,
+) -> anyhow::Result<()> {
+    let _ = tokio::time::timeout(Duration::from_secs(3), async {
+        tokio::join!(responder_done.notified(), initiator_done.notified());
+    })
+    .await;
+    Ok(())
```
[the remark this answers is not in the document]
etorreborre: That's on purpose

crates/amaru-protocols/src/tests/configuration.rs
```diff
@@ -0,0 +1,165 @@
...
+) -> anyhow::Result<()> {
+    // Use the same root header for both initiator and responder
+    let origin_hash: HeaderHash = amaru_kernel::Hash::from_str(
+        "4df4505d862586f9e2c533c5fbb659f04402664db1b095aba969728abfb77301",
+    )?;
+    let root_header = BlockHeader::from(make_header(100_000_000, 100_000_000, Some(origin_hash)));
+    chain_store.set_anchor_hash(&root_header.hash())?;
+    let mut headers = run_strategy(any_headers_chain_with_root(
+        chain_length - 1, // -1 since we already have the root header
+        root_header.point(),
```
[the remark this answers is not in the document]
etorreborre: That's ok

crates/amaru-protocols/src/connection.rs
```diff
@@ -164,10 +164,11 @@ async fn do_initialize(
     eff: Effects<ConnectionMessage>,
 ) -> State {
     let muxer = eff.stage("mux", mux::stage).await;
+    let muxer = eff.supervise(muxer, ConnectionMessage::Disconnect);
```
rkuhn: This shouldn’t change anything: if the muxer terminates without this line, then the connection will terminate, which is exactly what it does in response to a Disconnect message.

crates/amaru-protocols/src/manager.rs
```diff
@@ -129,13 +132,13 @@ pub async fn stage(
...
-                .connect(addr, Duration::from_secs(10))
+                .connect(addr, manager.connection_timeout)
                 .await
             {
                 Ok(conn_id) => conn_id,
                 Err(err) => {
-                    tracing::error!(?err, %peer, "failed to connect to peer");
-                    eff.schedule_after(ManagerMessage::Connect(peer), Duration::from_secs(10))
+                    tracing::error!(?err, %peer, reconnecting_in=?manager.connection_timeout, "failed to connect to peer. Scheduling reconnect");
+                    eff.schedule_after(ManagerMessage::Connect(peer), manager.connection_timeout)
```
rkuhn: the reconnect delay will usually be different from the connection timeout; we’ll have more such config parameters, so it makes sense to create a config struct that is passed to all places that need it (should be small and `Copy`)

crates/amaru-protocols/src/mux.rs
```diff
@@ -279,32 +288,37 @@ async fn handle_msg(
             }
             Ok(())
         }
+        MuxMessage::Terminate => {
+            tracing::error!(role=%muxer.role(), "terminating muxer due to read/write error");
```
rkuhn: this error log is the only difference to the unsupervised case, and I don’t think we should log an error upon peer disconnect — this is a normal part of the node lifecycle

crates/amaru/src/stages/build_stage_graph.rs
```diff
@@ -56,12 +56,37 @@ pub fn build_stage_graph(
...
+                FetchBlockFailed(_)
+                | StoreHeaderFailed(_, _)
+                | RemoveHeaderFailed(_, _)
+                | SetAnchorHashFailed(_, _)
+                | SetBestChainHashFailed(_, _)
+                | UpdateBestChainFailed(_, _, _)
+                | StoreBlockFailed(_, _)
+                | RollbackBlockFailed(_, _) // this can failed if the block was not downloaded in the first place
+                | UnknownPeer(_) => {
+                    tracing::error!(%peer, %error, "validation error, this needs to be investigated");
```
rkuhn, line 87: these are not validation errors, most of them are internal errors; I would even go so far as to declare them failures (if we cannot store a block on disk then we should probably panic)

crates/amaru-consensus/src/effects/ledger_effects.rs
```diff
@@ -47,7 +47,7 @@ pub trait LedgerOps: Send + Sync {
         peer: &Peer,
         point: &Point,
         ctx: opentelemetry::Context,
-    ) -> BoxFuture<'_, anyhow::Result<(), ProcessingFailed>>;
+    ) -> BoxFuture<'_, anyhow::Result<(), ValidationFailed>>;
```
rkuhn, line 50: This naming seems to imply that failures won’t be returned, i.e. the process panics upon disk I/O errors or similar. Is this correct?

(ValidationFailed is valid English, albeit slightly misleading because it always is an error, not a [failure](https://www.reactivemanifesto.org/glossary#Failure))

etorreborre: > This naming seems to imply that failures won’t be returned, i.e. the process panics upon disk I/O errors or similar. Is this correct?

Right now `ProcessingFailed` means "log and terminate the node" whereas `ValidationFailed` means "log and continue processing messages".

> ValidationFailed is valid English, albeit slightly misleading because it always is an error, not a [failure](https://www.reactivemanifesto.org/glossary#Failure)

I didn't realize that the reactive manifesto take on failures / errors was kind of opposite to the testing world where failures are expected and errors mean that the test itself is broken.

I'm happy to review that whole naming scheme so that expectations are clear.

crates/amaru-protocols/src/manager.rs
etorreborre: Done.

crates/amaru-protocols/src/connection.rs
etorreborre: You're right. I've been overzealous with supervision. I removed the line and the automated tests + manual tests pass ok.

etorreborre:
@rkuhn I've made the 2 changes about the reconnection delay and the supervision, then left the rework on failures/errors for later.

crates/amaru-protocols/src/tests/configuration.rs
```diff
@@ -0,0 +1,165 @@
...
+
+    pub(super) fn with_connection_timeout(mut self, timeout: Duration) -> Self {
+        self.reconnect_delay = timeout;
+        self
+    }
+
+    pub(super) fn with_processing_wait(mut self, wait: Duration) -> Self {
+        self.processing_wait = Some(wait);
+        self
+    }
```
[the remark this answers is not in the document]
etorreborre: fixed

crates/amaru-protocols/src/tests/test_cases.rs
```diff
@@ -0,0 +1,230 @@
...
+        ),
+    );
+
+    let era_history: &EraHistory = NetworkName::Preprod.into();
+    let initiator_manager = Manager::new(
+        NetworkMagic::PREPROD,
+        ManagerConfig::default(),
+        chainsync_stage.without_state(),
+        Arc::new(era_history.clone()),
+    );
```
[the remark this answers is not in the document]
etorreborre: Done.

PR #656, opened 2026-01-28
fix: allow the ledger to rollback a state that has just been rolled forward on the volatile state

etorreborre:
This issue has been found when trying the reconnection behavior of the node where we happen to roll forward some blocks, and then try to rollback to the last block of the volatile sequence.

The previous function was not making the distinction between finding the last point of the sequence as the rollback point and not finding it at all.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved rollback behavior to validate target positions, handle out-of-range requests, rebuild in-memory state up to the target, and avoid incorrect error/no-op outcomes.

* **Logging**
  * Normalized a previously logged condition to use error-level logging.

* **Tests**
  * Added comprehensive rollback tests and helpers covering before-first, exact-last, middle, after-last, and between-element scenarios.

### DIFF
--- a/crates/amaru-ledger/src/state/volatile_db.rs
+++ b/crates/amaru-ledger/src/state/volatile_db.rs
@@ -28,7 +28,6 @@ use amaru_kernel::{
 };
 use amaru_slot_arithmetic::Epoch;
 use std::collections::{BTreeMap, BTreeSet, VecDeque};
-use tracing::error;
 
 pub const EVENT_TARGET: &str = "amaru::ledger::state::volatile_db";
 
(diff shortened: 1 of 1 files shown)


crates/amaru-ledger/src/state/volatile_db.rs
```diff
@@ -96,25 +96,33 @@ impl VolatileDB {
...
         on_unknown_point: impl Fn(&Point) -> E,
     ) -> Result<(), E> {
+        let target_slot = point.slot_or_default();
+
+        // Check upfront if the target point is beyond the sequence
+        if let Some(last) = self.sequence.back()
+            && last.anchor.0.slot_or_default() < target_slot
+        {
+            return Err(on_unknown_point(point));
+        }
```
KtorZ: Not sure I agree with the condition here; it seems backward. 

What I would expect: 

- We fail if asked to rollback for more than the maximum rollback length; that is, if `target_slot < self.sequence.front().anchor.0.slot_or_default()` 

Here, we're failing if we are asked to rollback to a point we haven't seen yet (the target slot is more recent than the tip of the volatile db). 

I don't think the latter is an error per se; In the same way that dropping 5 elements from an empty list may be deem as okay. Although it's a sign that something went off, because we shouldn't be asked to rollback to a point we have not seen (roll forward / backward events are expected to be properly ordered). 

So I would suggest to simply exit immediately in this case, with no error, but a warning log that indicates something's off and we should've seen the point before. And more importantly, the function should fail if asked to rollback too far!

crates/amaru-ledger/src/state/volatile_db.rs
```diff
@@ -396,3 +404,113 @@ fn add_proposals(
...
+
+        // Try to rollback to slot 40 (after the sequence)
+        let rollback_point = Point::Specific(Slot::from(40), Hash::new([0u8; 32]));
+
+        let result = db.rollback_to(&rollback_point, |_| "Point not found");
+
+        // This should fail
+        assert!(
+            result.is_err(),
+            "Rolling back to a point after the sequence should fail"
```
KtorZ: As per previous comment; this should succeed, but raise a warning.

crates/amaru-ledger/src/state/volatile_db.rs
```diff
@@ -396,3 +404,113 @@ fn add_proposals(
...
+
+        // Rollback to slot 5 (before the first element at slot 10)
+        // This represents rolling back to a point in the stable DB
+        let rollback_point = Point::Specific(Slot::from(5), Hash::new([0u8; 32]));
+
+        let result = db.rollback_to(&rollback_point, |_| "Point not found");
+
+        // This should succeed and clear all volatile changes
+        // (rolling back to stable DB discards all volatile state)
+        assert!(result.is_ok());
```
KtorZ: This should fail; we are not allowed to rollback beyond what's in the volatile. So here 10 is the oldest point we're allowed to rollback to.

crates/amaru-ledger/src/state/volatile_db.rs
etorreborre: Makes sense. I've updated the code and the tests.

crates/amaru-ledger/src/state/volatile_db.rs
etorreborre: Done.

crates/amaru-ledger/src/state/volatile_db.rs
etorreborre: Done.

Issue #658, opened 2026-01-30
move upstream peer tracking and block validation out of `select_chain` stage


rkuhn:
### Abstract

We bring the `select_chain` stage back within the complexity budget for a single stage by factoring out parts of its behaviour that are not intrinsically required for selecting chains.

### Why?

Currently, `select_chain` does four things:

1. it tracks upstream peer progress and asserts correct peer behaviour
2. it selects a best chain candidate whenever new information becomes available
3. it uses the ledger to validate blocks
4. it extends the chain for downstream peers upon successful validation

This is the most complex stage in the current setup and it needs to be simplified before we can make substantial improvements. The complexity needs to be factored into multiple orthogonal parts that can then be improved independently.

### How?

The upstream peer tracking will be moved into the per-peer part of the setup, i.e. before header validation step. In effect, this replaces the `pull` stage with one stage per peer. Header validation can then be skipped if the header is already stored, avoid duplication of work. Incorrect peer behaviour is recognised closer to the network stack and will trigger disconnections.

Block validation will be moved into a new part of the pipeline which will also fetch blocks with a strategy that allows batching to reduce the impact of network link latency — our current setup requires one RTT plus bandwidth delay for each block, and it does this in 1:1 correspondence with header processing, thereby slowing the whole pipeline down. The main difficulty will be to correctly feed back block validation errors into `select_chain` to trigger the switch to a different fork (this is not unclear, it only will be the most subtle part of the logic).

### Testing Strategy / Acceptance Criteria

Currently, e2e testing on CI has been reduced to run only up to epoch 176, with these changes it should easily run until 182 again; it is reasonable to expect that it can run even further within 15min.

Tests will be added for the disconnection of misbehaving upstream peers.

### Discussion points

This will likely be implemented via multiple PRs, starting with factoring out upstream peer tracking.

### Checklist

- [x] I understand that feature requests and unrefined work item should be open as [GitHub Discussions](https://github.com/pragma-org/amaru/discussions) instead.
- [x] I have assigned this item to an existing milestone from the roadmap
- [x] I have added a label capturing the impact of this item (i.e. value for users/stakeholders if successful)
- [x] I have added a label capturing the delivery risk of this item (i.e. how likely is it that this task will succeed as planned)
- [x] I have added a label capturing the effort of this item (i.e. how large is the task?)

PR #671, opened 2026-02-07
add new track_peers stage

rkuhn:
This stage will replace pull+receive_header+validate_header stages as well as incorporate the upstream peer tracking currently performed in the HeadersTree that is employed by select_chain. This is a sketch of the envisioned consensus stages, whereof this PR is the first step.
EE60C7DC-EBBF-440D-8027-88D2387BF137_1_102_a
This is also the first step of #658.

## Summary by CodeRabbit

* **New Features**
  * Added peer tracking functionality for chain synchronization operations with comprehensive error handling
  * Enhanced block header validation with detailed error classification

* **Improvements**
  * Strengthened era type system with improved naming conventions
  * Refined error reporting for invalid header heights, points, and variants
  * Improved network stage logging for better observability

* **Refactor**
  * Reorganized module structure for better maintainability


crates/amaru-consensus/src/stages/track_peers.rs
```diff
@@ -0,0 +1,1289 @@
...
+        Ok(())
+    }
+}
+
+#[tracing::instrument(
+    level = tracing::Level::TRACE,
+    skip_all,
+    name = "chain_sync.decode_header",
+)]
+pub fn decode_header(raw_header: HeaderContent) -> Result<BlockHeader, ConsensusError> {
```
etorreborre: Shouldn't we pass the `EraHistory` here to avoid hardcoding `Conway`?

crates/amaru-consensus/src/stages/track_peers.rs
```diff
@@ -0,0 +1,1289 @@
...
+                        PerPeer {
+                            current,
+                            highest: tip,
+                        },
+                    );
+                }
+                IntersectNotFound(tip) => {
+                    tracing::info!(%peer, highest = %tip.point(), reason = "intersect not found", "stopping chainsync");
+                    eff.send(&handler, chainsync::InitiatorMessage::Done).await;
+                    state.upstream.remove(&peer);
```
etorreborre: Shouldn't we send a message to the `Manager` to remove the peer as well? And, in that case refactor with  L218-L219 + L233-L234 + L255-L256.

crates/amaru-consensus/src/stages/track_peers.rs
```diff
@@ -0,0 +1,1289 @@
...
+                },
+            )));
+        }
+        if header.block_height() != per_peer.current.block_height() + 1 {
+            return Err(ConsensusError::InvalidHeaderHeight {
+                actual: header.block_height(),
+                expected: per_peer.current.block_height() + 1,
+            });
+        }
+        let highest = per_peer.highest.point().max(tip.point());
```
etorreborre: When would `per_peer.highest.point() > tip.point()`? Shouldn't we validate separately that:
 
- `header.point <= tip.point()` -> the new header and new tip are consistent
- `per_peer.highest.point() <= tip.point()` -> the new tip is further along the previous tip

And maybe make 3 different errors by adding the case `header.point() < per_peer.current.point()`?

crates/amaru-consensus/src/stages/track_peers.rs
```diff
@@ -0,0 +1,1289 @@
...
+    chainsync::{self, ChainSyncInitiatorMsg, HeaderContent},
+    manager::ManagerMessage,
+    store_effects::Store,
+};
+use pure_stage::{Effects, StageRef};
+use std::collections::BTreeMap;
+use tracing::Span;
+use tracing_opentelemetry::OpenTelemetrySpanExt;
+
+#[derive(Debug, Clone, PartialEq, serde::Serialize, serde::Deserialize)]
```
etorreborre: It would be nice to have a few comments on this data type to explain what is tracked and why we need to track it.

crates/amaru-consensus/src/stages/track_peers.rs
```diff
@@ -0,0 +1,1289 @@
...
+            manager,
+            downstream,
+        }
+    }
+
+    async fn validate_header(
+        &self,
+        peer: &Peer,
+        raw_header: HeaderContent,
+        tip: Tip,
```
etorreborre: In terms of comments it could be good to have a comment reminding the reader what is this `Tip` here.

crates/amaru-consensus/src/stages/track_peers.rs
```diff
@@ -0,0 +1,1289 @@
...
+        let trace = tb
+            .iter_entries()
+            .filter_map(|(_, e)| (!matches!(e, TraceEntry::Resume { .. })).then_some(e))
+            .collect::<Vec<_>>();
+        tb.clear();
+        pretty_assertions::assert_eq!(trace, expected);
+    }
+
+    #[test]
+    fn test_new_peer() {
```
etorreborre: That's a personal preference but I'd rather see the tests coming up first in a `tests` module and then I can scroll to helper functions if needed to understand a given test.

crates/amaru-consensus/src/stages/track_peers.rs
```diff
@@ -0,0 +1,1289 @@
...
+                TraceEntry::input("tp-1", Box::new(msg)),
+                TraceEntry::state("tp-1", Box::new(state)),
+            ],
+        );
+        logs.assert_log(|l| l.contains("initializing chainsync"));
+        logs.assert_no_log(|l| l.contains("WARN") || l.contains("ERROR"));
+    }
+
+    #[test]
+    fn test_initialize_existing_peer() {
```
etorreborre: Maybe this function should be named `test_initialize_existing_peer_is_ok` to make it clear that this is authorized. But that raises the question: why would initialize be called twice with the same peer? Should there be a warning there? If not it would be good to leave a note to explain why this can happen.

crates/amaru-consensus/src/stages/track_peers.rs
```diff
@@ -0,0 +1,1289 @@
...
+            StageRef::named_for_tests("downstream"),
+        );
+        let msg = TrackPeersMsg::FromUpstream(ChainSyncInitiatorMsg {
+            peer: Peer::new("peer1"),
+            conn_id: ConnectionId::initial(),
+            handler: StageRef::named_for_tests("handler"),
+            msg: chainsync::InitiatorResult::Initialize,
+        });
+
+        let rt = Builder::new_current_thread().build().unwrap();
```
etorreborre: It seems that we could cut a bit on test verbosity by inlining the `rt` creation into `setup`.
Maybe also creating an `assert_logs` function like this:
```
assert_logs(logs).contain(&["initializing chainsync"]).dont_contain(&["WARN", "ERROR"]);

// or this, but which does not signal that the first list means "included" and the second one "excluded"
check_logs(logs, &["initializing chainsync"], &["WARN", "ERROR"]);
```

crates/amaru-consensus/src/stages/track_peers.rs
```diff
@@ -0,0 +1,1289 @@
...
+                TraceEntry::input("tp-1", Box::new(msg)),
+                TraceEntry::state("tp-1", Box::new(state)),
+            ],
+        );
+        logs.assert_log(|l| l.contains("initializing chainsync"));
+        logs.assert_no_log(|l| l.contains("WARN") || l.contains("ERROR"));
+    }
+
+    #[test]
+    fn test_intersect_found_missing_header_sends_done() {
```
etorreborre: It's nice to have all these tests and I think it would be even better to refactor the setup with some additional function so that we can better see what's changing in the initial conditions for each test. 
For example we could hide a bit more the initial state by making it a function called in `setup` and in `assert_trace`, also abstract over `handler` and `conn_id`.

crates/amaru-consensus/src/stages/track_peers.rs
```diff
@@ -0,0 +1,1289 @@
...
+                send("tp-1", &handler, chainsync::InitiatorMessage::Done),
+                TraceEntry::state("tp-1", Box::new(expected)),
+            ],
+        );
+        logs.assert_log(|l| l.contains("INFO") && l.contains("intersect not found"));
+        logs.assert_no_log(|l| l.contains("WARN") || l.contains("ERROR"));
+    }
+
+    #[test]
+    fn test_roll_forward_unknown_peer_removes() {
```
etorreborre: `test_roll_forward_unknown_peer_removes_peer` to stay consistent

crates/amaru-consensus/src/stages/track_peers.rs
```diff
@@ -0,0 +1,1289 @@
...
+                TraceEntry::state("tp-1", Box::new(state)),
+                TraceEntry::input("tp-1", Box::new(msg)),
+                send("tp-1", &handler, chainsync::InitiatorMessage::RequestNext),
+                validate_header_effect("tp-1", header.clone()),
+                has_header_effect("tp-1", header.hash()),
+                TraceEntry::state("tp-1", Box::new(expected)),
+            ],
+        );
+        logs.assert_log(|l| {
+            l.contains("DEBUG") && l.contains("roll forward") && l.contains("already stored")
```
etorreborre: We could add `header already stored` to make it clear that this is about the header, not the peer.

crates/amaru-consensus/src/stages/track_peers.rs
```diff
@@ -0,0 +1,1289 @@
...
+                load_header_effect("tp-1", current.hash()),
+                TraceEntry::state("tp-1", Box::new(expected)),
+            ],
+        );
+        logs.assert_log(|l| l.contains("INFO") && l.contains("roll backward"));
+        logs.assert_no_log(|l| l.contains("WARN") || l.contains("ERROR"));
+    }
+
+    #[test]
+    fn test_roll_backward_unknown_peer_removes() {
```
etorreborre: `test_roll_backward_unknown_peer_removes_peer`

crates/amaru-consensus/src/stages/track_peers.rs
```diff
@@ -0,0 +1,1289 @@
...
+        );
+        logs.assert_log(|l| {
+            l.contains("ERROR")
+                && l.contains("chain_sync.roll_backward.failed")
+                && l.contains("Unknown peer")
+        });
+    }
+
+    #[test]
+    fn test_roll_backward_unknown_point_removes() {
```
etorreborre: `test_roll_backward_unknown_point_removes_peer`

crates/amaru-consensus/src/stages/track_peers.rs
rkuhn: This is the validation step, where the new incoming tip from the RollForward hasn't yet been duly stored.

crates/amaru-consensus/src/stages/track_peers.rs
rkuhn: no, because we can only decode Conway headers. Once we add more variants, this function will need to be updated.

crates/amaru-consensus/src/stages/track_peers.rs
rkuhn: Here,  we primarily care about chainsync only. The current connection setup will terminate the connection when reaching the Done state, but in the future we may just demote from hot to warm.

crates/amaru-consensus/src/stages/track_peers.rs
rkuhn: My preference is the other way around, but perhaps the scaffolding is so large in this case that it should be moved into a separate file.

crates/amaru-consensus/src/stages/track_peers.rs
rkuhn: I'd like to have better log assertions that "acknowledge" certain errors/warnings and then state that there were no others. Will have to be a subsequent PR.

crates/amaru-consensus/src/stages/track_peers.rs
rkuhn: I didn't see a reason to forbid it, and this stage is not the one deciding the ground truth of which peer is currently connected. The trouble with these systems is that disconnects and reconnects may happen very quickly.

crates/amaru-consensus/src/stages/track_peers.rs
rkuhn: yes, will push in that direction

crates/amaru-consensus/src/stages/track_peers.rs
rkuhn: oops, did it in this PR already :-)

PR #681, opened 2026-02-22
feat: add new select_chain stage

rkuhn:
currently not wired up and under the name select_chain_new because rewiring will require new block_fetch and validate stages as well

## Summary by CodeRabbit

* **New Features**
  * Multi-fork chain selection stage with dynamic best-tip selection and validation-aware behavior.
  * Stores now track per-block validity and provide validity-aware header retrieval and mutators.
  * Testable store wrapper allowing per-method overrides for rich simulation.

* **Tests**
  * Extensive select-chain test suite with shared test scaffolding and trace-driven verification.
  * Centralized in-memory log-capture and assertion utilities.

* **Documentation**
  * Selected re-exports now appear in generated docs.

* **Chores**
  * Added parking_lot to workspace dependencies.

### DIFF
--- a/crates/amaru-consensus/src/stages/select_chain_new/mod.rs
+++ b/crates/amaru-consensus/src/stages/select_chain_new/mod.rs
@@ -84,30 +84,26 @@ impl SelectChain {
             chain.push(tip.hash());
             self.tips.insert(tip.hash(), chain);
         } else {
-            // the message type allows that we get a Tip that was already sent earlier, the track_peers stage
-            // shouldn’t send this, so we don’t defend against it here
+            // since track_peers will only send newly stored tips, this is the case where
+            // a new fork is detected; while the new fork can only be one header long, it
+            // may still require multiple block validations to reach a valid chain
             let mut valid = true;
-            let mut extending = false;
             let mut ancestors = store
                 .ancestors_with_validity(parent.hash())
-                .take_while(|(h, v)| {
+                .take_while(|(_h, v)| {
                     if *v == Some(false) {
                         valid = false;
                         false
                     } else {
-                        if self.tips.remove(&h.hash()).is_some() {
-                            extending = true;
-                        };
                         v.is_none()
                     }
                 })
                 .map(|(h, _)| h.hash())
                 .collect::<Vec<_>>();
             if valid {
-                let action = if extending { "extending" } else { "new" };
-                tracing::debug!(%parent, %tip, "{action} chain");
+                tracing::debug!(%parent, %tip, "new chain");
                 ancestors.reverse();
-                ancestors.push(tip.hash());
+                ancestors.push(tip.hash()); // new block must be validated by definition
                 self.tips.insert(tip.hash(), ancestors);
             } else {
                 tracing::info!(%parent, %tip, "upstream tip depends on invalid block");
--- a/crates/amaru-consensus/src/stages/select_chain_new/test_setup.rs
+++ b/crates/amaru-consensus/src/stages/select_chain_new/test_setup.rs
@@ -21,9 +21,9 @@ use amaru_protocols::store_effects::{
     ResourceHeaderStore, SetBlockValidEffect,
 };
 use pure_stage::{
-    DeserializerGuards, Effect, Name, StageGraph, StageRef,
+    DeserializerGuards, Effect, Name, StageGraph, StageRef, TerminationReason,
     simulation::{SimulationBuilder, SimulationRunning},
-    trace_buffer::{TerminationReason, TraceBuffer, TraceEntry},
+    trace_buffer::{TraceBuffer, TraceEntry},
 };
 use tokio::runtime::{Builder, Runtime};
 use tracing::Level;
--- a/crates/amaru-stores/src/rocksdb/consensus/mod.rs
+++ b/crates/amaru-stores/src/rocksdb/consensus/mod.rs
@@ -27,8 +27,8 @@ use rocksdb::{
 use crate::rocksdb::{
     RocksDbConfig,
     consensus::util::{
-        ANCHOR_PREFIX, BEST_CHAIN_PREFIX, BLOCK_PREFIX, CHAIN_PREFIX, CHILD_PREFIX, CONSENSUS_PREFIX_LEN,
-        HEADER_PREFIX, NONCES_PREFIX, open_db, open_or_create_db,
+        ANCHOR_PREFIX, BEST_CHAIN_PREFIX, BLOCK_PREFIX, CHAIN_DB_VERSION, CHAIN_PREFIX, CHILD_PREFIX,
+        CONSENSUS_PREFIX_LEN, HEADER_PREFIX, NONCES_PREFIX, open_db, open_or_create_db,
     },
 };
 
@@ -121,7 +121,7 @@ impl RocksDBStore<OptimisticTransactionDB> {
         }
 
         let (_, db) = open_or_create_db(&config)?;
-        set_version(&db)?;
+        set_version(&db, CHAIN_DB_VERSION)?;
 
         Ok(Self { db, basedir })
     }
@@ -877,7 +877,7 @@ pub mod test {
         let result = migrate_db_path(target).expect("Migration should succeed");
 
         let db = RocksDBStore::open(&config).expect("DB should successfully be opened as it's been migrated");
-        assert_eq!((1, 2), result);
+        assert_eq!((1, 3), result);
         let header: Option<HeaderHash> = <RocksDBStore as ReadOnlyChainStore<BlockHeader>>::load_from_best_chain(
             &db,
             &Point::Specific(5.into(), Hash::from_str(SAMPLE_HASH).unwrap()),
--- a/crates/amaru-consensus/src/stages/track_peers/mod.rs
+++ b/crates/amaru-consensus/src/stages/track_peers/mod.rs
@@ -98,7 +98,7 @@ impl TrackPeers {
         // this is the point up to which the upstream peer has validated its best chain, which
         // can be less advanced than the currently transmitted header
         let highest = tip.point();
-        if header.point() < per_peer.current.point() {
+        if header.slot() <= per_peer.current.slot() {
             return Err(ConsensusError::InvalidHeaderPoint(Box::new(InvalidHeaderPoint {
                 actual: header.point(),
                 parent: per_peer.current.point(),
@@ -221,7 +221,6 @@ impl TrackPeers {
                     tracing::error!(%error, %peer, "chain_sync.roll_backward.failed");
                     self.upstream.remove(&peer);
                     eff.send(&self.manager, ManagerMessage::RemovePeer(peer)).await;
-                    return;
                 }
             }
         }
--- a/crates/amaru-consensus/src/stages/track_peers/test_setup.rs
+++ b/crates/amaru-consensus/src/stages/track_peers/test_setup.rs
@@ -86,24 +86,24 @@ pub fn make_block_header(block_number: u64, slot: u64, parent: Option<HeaderHash
     BlockHeader::from(make_header(block_number, slot, parent))
 }
 
-pub fn validate_header_effect(at_stage: &str, header: BlockHeader) -> TraceEntry {
+pub fn te_validate_header(at_stage: &str, header: BlockHeader) -> TraceEntry {
     TraceEntry::suspend(Effect::external(at_stage, Box::new(ValidateHeaderEffect::new(&header, Context::new()))))
 }
 
-pub fn load_header_effect(at_stage: &str, hash: HeaderHash) -> TraceEntry {
+pub fn te_load_header(at_stage: &str, hash: HeaderHash) -> TraceEntry {
     TraceEntry::suspend(Effect::external(at_stage, Box::new(LoadHeaderEffect::new(hash))))
 }
 
-pub fn has_header_effect(at_stage: &str, hash: HeaderHash) -> TraceEntry {
+pub fn te_has_header(at_stage: &str, hash: HeaderHash) -> TraceEntry {
     TraceEntry::suspend(Effect::external(at_stage, Box::new(HasHeaderEffect::new(hash))))
 }
 
-pub fn store_header_effect(at_stage: &str, header: BlockHeader) -> TraceEntry {
+pub fn te_store_header(at_stage: &str, header: BlockHeader) -> TraceEntry {
     TraceEntry::suspend(Effect::external(at_stage, Box::new(StoreHeaderEffect::new(header))))
 }
 
-pub fn send(from: impl AsRef<str>, to: impl AsRef<str>, msg: impl SendData) -> TraceEntry {
-    TraceEntry::suspend(Effect::send(from, to, Box::new(msg)))
+pub fn te_send(from: impl AsRef<str>, to: impl AsRef<str>, msg: impl SendData) -> TraceEntry {
+    TraceEntry::suspend(pure_stage::Effect::send(from, to, Box::new(msg)))
 }
 
 fn register_guards() -> DeserializerGuards {
--- a/crates/amaru-consensus/src/stages/track_peers/tests.rs
+++ b/crates/amaru-consensus/src/stages/track_peers/tests.rs
@@ -25,8 +25,8 @@ use tracing::Level;
 use crate::stages::track_peers::{
     TrackPeersMsg,
     test_setup::{
-        FailingHeaderValidation, assert_trace, build_store, has_header_effect, load_header_effect, make_block_header,
-        send, setup, setup_with_validation, store_header_effect, test_prep, validate_header_effect,
+        FailingHeaderValidation, assert_trace, build_store, make_block_header, setup, setup_with_validation,
+        te_has_header, te_load_header, te_send, te_store_header, te_validate_header, test_prep,
     },
 };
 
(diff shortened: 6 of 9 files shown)


crates/amaru-stores/src/rocksdb/consensus/mod.rs
```diff
@@ -123,143 +161,135 @@ impl RocksDBStore {
...
+    }
+}
+
 pub(crate) fn store_chain_point(db: &OptimisticTransactionDB, point: &Point) -> Result<(), StoreError> {
     let slot = u64::from(point.slot_or_default()).to_be_bytes();
     db.put([&CHAIN_PREFIX[..], &slot[..]].concat(), point.hash().as_ref())
         .map_err(|e| StoreError::WriteError { error: e.to_string() })
 }
 
-macro_rules! impl_ReadOnlyChainStore {
```
rkuhn, line 132: using macros to implement traits is really bad because it makes it basically impossible to edit the implementation in a sane fashion and it hides methods from finding all references, so I went on a side-quest and figured out how to abstract over mutability in RocksDB — the truth is ugly and that crate’s API sucks but I find this approach more maintainable

crates/amaru-consensus/src/stages/select_chain_new/mod.rs
```diff
@@ -0,0 +1,187 @@
...
+    }
+}
+
+impl SelectChain {
+    async fn handle_tip_from_upstream(mut self, tip: Tip, eff: Effects<SelectChainMsg>) -> SelectChain {
+        let store = Store::new(eff.clone());
+
+        let (Some(header), valid) = store.load_header_with_validity(&tip.hash()) else {
+            tracing::error!(%tip, "tip not found");
+            return self;
```
etorreborre: Shouldn't we terminate at that stage since the `track_peers` stage should have been executed and should have stored the header?

crates/amaru-consensus/src/stages/select_chain_new/mod.rs
```diff
@@ -0,0 +1,187 @@
...
+        };
+
+        if let Some(valid) = valid {
```
etorreborre, line 69: I wonder if we should rather deal with this condition in  the `track_peers` stage when we receive a new `RollForward` and simply not propagate a `NewTip` if there's already a valid block, meaning that the peer did not send us new information (and possibly flag it for that reason). Then consider in `select_chain` that we only deal with unvalidated tips.

crates/amaru-consensus/src/stages/select_chain_new/mod.rs
```diff
@@ -0,0 +1,187 @@
...
+            tracing::warn!(%tip, %valid, "got tip from upstream that was already validated");
+            return self;
+        } else {
+            tracing::debug!(%tip, "got new tip from upstream");
+        }
+
+        if let Some(parent) = header.parent_hash() {
```
etorreborre: What if `TipFromUpstream` already contained the parent hash, since we have already loaded the header in `track_peers`?

crates/amaru-consensus/src/stages/select_chain_new/mod.rs
```diff
@@ -0,0 +1,187 @@
...
+            // if parent is in tips, extend that chain; otherwise check store for fragment
+            if let Some(mut chain) = self.tips.remove(&parent) {
+                tracing::debug!(%parent, %tip, "extending chain");
+                chain.push(tip.hash());
+                self.tips.insert(tip.hash(), chain);
+            } else {
+                // the message type allows that we get a Tip that was already sent earlier, the track_peers
```
etorreborre:
```suggestion
                // the message type allows that we get a Tip that was already sent earlier, the track_peers stage
```

crates/amaru-consensus/src/stages/select_chain_new/mod.rs
```diff
@@ -0,0 +1,187 @@
...
+                            if self.tips.remove(&h.hash()).is_some() {
+                                extending = true;
+                            };
+                            v.is_none()
+                        }
+                    })
+                    .map(|(h, _)| h.hash())
+                    .collect::<Vec<_>>();
+                if valid {
+                    if extending {
```
etorreborre: I guess this could be
```
let action = if extending { "extending" } else { "new "};
tracing::debug!(%parent, %tip, format!("{} chain", action))
```
with or without inlining `action`.

crates/amaru-consensus/src/stages/select_chain_new/test_setup.rs
```diff
@@ -0,0 +1,205 @@
...
+
+use super::*;
+use crate::stages::test_utils::{BufferWriter, Logs};
+
+pub fn make_block_header(block_number: u64, slot: u64, parent: Option<HeaderHash>) -> BlockHeader {
+    BlockHeader::from(make_header(block_number, slot, parent))
+}
+
+/// Header tree for testing block invalidation and chain selection:
+/// - h0: genesis (block 1, slot 1, no parent)
```
etorreborre, line 40: An ascii tree representation would be nice here.

crates/amaru-consensus/src/stages/select_chain_new/mod.rs
```diff
@@ -0,0 +1,187 @@
...
+            let removed = prev_tips - self.tips.len();
+
+            if !self.tips.contains_key(&self.best_tip.hash()) {
+                tracing::info!(%removed, "best tip candidate invalidated");
+                // need to pick new best tip
+                self.best_tip = self
+                    .tips
+                    .keys()
+                    .filter_map(|h| store.load_header(h).map(|h| h.tip()))
+                    .max_by(|a, b| a.block_height().cmp(&b.block_height()).then_with(|| b.slot().cmp(&a.slot())))
```
etorreborre: As discussed we could extract this comparison function to show that it is the same as the one in L.118

crates/amaru-consensus/src/stages/select_chain_new/mod.rs
```diff
@@ -0,0 +1,187 @@
...
+use amaru_ouroboros::{ChainStore, ReadOnlyChainStore};
+use amaru_protocols::store_effects::Store;
+use pure_stage::{Effects, StageRef, TryInStage};
+
+#[derive(Clone, Debug, PartialEq, serde::Serialize, serde::Deserialize)]
+pub struct SelectChain {
+    downstream: StageRef<Tip>,
+    /// Maps all block tree tips to the list of headers whose blocks are yet to be validated
+    /// (oldest first)
+    tips: BTreeMap<HeaderHash, Vec<HeaderHash>>,
```
etorreborre, line 27: It could be good to specify here that:

- The map key is the last header hash of the map value.
- For each entry, the chain is following a parent -> child relationship (no skipping of parents, or is it possible?).
- The vec of headers cannot be longer than `k` (or start with a header older than the anchor).

crates/amaru-consensus/src/stages/select_chain_new/mod.rs
```diff
@@ -0,0 +1,187 @@
...
+                let mut valid = true;
+                let mut extending = false;
+                let mut ancestors = store
+                    .ancestors_with_validity(parent)
+                    .take_while(|(h, v)| {
+                        if *v == Some(false) {
+                            valid = false;
+                            false
+                        } else {
+                            if self.tips.remove(&h.hash()).is_some() {
```
etorreborre: This whole expression deserves a comment IMO since we are both iterating ancestors and mutating `self.tips`. Maybe what I don't understand is how we can end up with a new tip that "jumps ahead" and has ancestors extending an existing chain. Shouldn't we see one of these ancestors as a `NewTip` before?

crates/amaru-consensus/src/stages/select_chain_new/mod.rs
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    async fn handle_block_validation_result(
+        mut self,
+        point: Point,
+        valid: bool,
+        eff: Effects<SelectChainMsg>,
+    ) -> SelectChain {
+        let store = Store::new(eff.clone());
+        if !store.has_header(&point.hash()) {
```
etorreborre, line 128: Is it necessary to be defensive here? In that why not terminate? Also the message says "block not found" while we just check if the header is found.

crates/amaru-consensus/src/stages/select_chain_new/mod.rs
```diff
@@ -0,0 +1,187 @@
...
+        store
+            .set_block_valid(point.hash(), valid)
+            .or_terminate(&eff, async |error| {
+                tracing::error!(%error, %valid, "failed to store block validation result");
+            })
+            .await;
+
+        if valid {
+            self.tips.retain(|_k, v| {
+                if v.first() == Some(&point.hash()) {
```
etorreborre: We could add a comment to explain why the valid block is necessary the first element of the headers we track.

crates/amaru-consensus/src/stages/select_chain_new/mod.rs
```diff
@@ -0,0 +1,187 @@
...
+                    v.remove(0);
+                }
+                true
```
etorreborre: Why not, instead of `retain/true`
```
let h = point.hash();
self.tips.values_mut().for_each(|v| {
    if v.first() == Some(&h) { v.remove(0); }
});
```

crates/amaru-consensus/src/stages/select_chain_new/mod.rs
```diff
@@ -0,0 +1,187 @@
...
+                    .unwrap_or_else(|| {
+                        store.load_header(&store.get_best_chain_hash()).map(|h| h.tip()).unwrap_or(Tip::origin())
+                    });
+                if self.best_tip != Tip::origin() {
+                    self.tips.entry(self.best_tip.hash()).or_insert(vec![]);
+                }
+                tracing::debug!(%self.best_tip, "new best tip candidate");
+                eff.send(&self.downstream, self.best_tip).await;
+            } else if removed > 0 {
+                tracing::warn!(%removed, "chain fork removed due to invalid block");
```
etorreborre:
```suggestion
                tracing::warn!(%removed, "chain fork(s) removed due to invalid block");
```

crates/amaru-consensus/src/stages/select_chain_new/mod.rs
```diff
@@ -0,0 +1,187 @@
...
+    match msg {
+        SelectChainMsg::TipFromUpstream(tip) => state.handle_tip_from_upstream(tip, eff).await,
+        SelectChainMsg::BlockValidationResult(point, valid) => {
+            state.handle_block_validation_result(point, valid, eff).await
+        }
+    }
+}
+
+impl SelectChain {
+    async fn handle_tip_from_upstream(mut self, tip: Tip, eff: Effects<SelectChainMsg>) -> SelectChain {
```
etorreborre: We need to handle anchor management when we extends chains past `k`, and set a new anchor. This used to be done during chain selection but does not appear here anymore. 
One other aspect of anchor management is that we would stop tracking a peer that's following a fork that becomes longer than `k`.

crates/amaru-consensus/src/stages/select_chain_new/tests.rs
```diff
@@ -0,0 +1,534 @@
...
+use crate::stages::{
+    select_chain_new::test_setup::{
+        assert_trace, setup, te_get_anchor_hash, te_get_best_chain_hash, te_has_header, te_load_header, te_send,
+        te_set_block_valid, te_terminate, te_terminated, test_prep,
+    },
+    test_utils::{te_input, te_state},
+};
+
+#[test]
+fn test_tip_not_found() {
```
etorreborre, line 32: Those are useful unit tests but I think I would be more comfortable to complement them with a property test like we had before where we generate a tree of headers and send random `NewTip` messages from that tree.

etorreborre (commented):
I left various comments for you to consider. 
The part that I don't visualize properly is the one where we fetch the ancestors of a new tip. Maybe let's discuss this live with an example?

crates/amaru-consensus/src/stages/select_chain_new/mod.rs
rkuhn: hmm, I’m on the fence; since I don’t see how this could ever happen, I can’t trust any extrapolation into the case where it does actually happen — would it be appropriate to give up?

crates/amaru-consensus/src/stages/select_chain_new/mod.rs
rkuhn: a lot of the code here is more defensive than it needs to be because the “guarantees” of upstream stages cannot be expressed in Rust types

crates/amaru-consensus/src/stages/select_chain_new/mod.rs
rkuhn: good point!

crates/amaru-consensus/src/stages/select_chain_new/mod.rs
rkuhn: The first point is not true: the Vec may very well be empty

crates/amaru-consensus/src/stages/select_chain_new/mod.rs
rkuhn: yes, normally there shouldn’t be jumps, but we cannot express that in the types, so I prefer to handle it

crates/amaru-consensus/src/stages/select_chain_new/mod.rs
rkuhn: yes, will improve

crates/amaru-consensus/src/stages/select_chain_new/mod.rs
rkuhn: okay, will add a TODO

crates/amaru-consensus/src/stages/select_chain_new/tests.rs
rkuhn: I wouldn’t know how to formulate the property based on the various kinds of inputs that can reach this stage. I think I have covered all the basic cases in the unit tests, and I think we’ll cover the rest in the full node integration tests. If you have a good idea of a property-based test to add then let’s add that in a subsequent PR.

crates/amaru-consensus/src/stages/select_chain_new/mod.rs
etorreborre: The best thing IMO is to terminate and investigate what kind of fundamental assumption is broken.

crates/amaru-consensus/src/stages/select_chain_new/mod.rs
etorreborre: I'd rather terminate than continue computing for situation that should not happen. The risk is to produce data that become difficult (if not impossible without large coordination?) to disentangle later on.

crates/amaru-consensus/src/stages/select_chain_new/tests.rs
etorreborre: I think we should be able to implement something similar to the previous test for `HeadersTree`. I can try that on a separate PR.

crates/amaru-consensus/src/stages/select_chain_new/mod.rs
rkuhn: The second point logically follows from the existing comment. The third point isn’t true, we need to connect to the chain that we have already validated, so the anchor point cannot arbitrarily cut the sequence. Cardano will eventually see a network split of more than 12h and I really don’t want such an event to end the chain.

crates/amaru-consensus/src/stages/select_chain_new/mod.rs
rkuhn: This might actually happen after node restart, when track_peers doesn’t have its full state yet.

crates/amaru-consensus/src/stages/select_chain_new/mod.rs
```diff
@@ -0,0 +1,192 @@
...
+                        }
+                    })
+                    .map(|(h, _)| h.hash())
+                    .collect::<Vec<_>>();
+                if valid {
+                    let action = if extending { "extending" } else { "new" };
+                    tracing::debug!(%parent, %tip, "{action} chain");
+                    ancestors.reverse();
+                    ancestors.push(tip.hash());
+                    self.tips.insert(tip.hash(), ancestors);
```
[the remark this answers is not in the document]
rkuhn: There is no concurrency within a pure-stage. This is a design decision taken precisely to avoid these kinds of issues and thus simplify the code.

crates/amaru-consensus/src/stages/select_chain_new/mod.rs
```diff
@@ -0,0 +1,192 @@
...
+                }
+                // block validation results should arrive in order, from oldest to newest
+                debug_assert!(!v.contains(&h));
+            });
+        } else {
+            // remove all tips depending on the invalid block
+            // (if a peer sends further headers on this chain, we will ignore them)
+            let prev_tips = self.tips.len();
+            self.tips.retain(|_k, v| v.first() != Some(&point.hash()));
+            let removed = prev_tips - self.tips.len();
```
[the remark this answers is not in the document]
rkuhn: A tracked tip with an empty pending list means that it has been fully validated already and can thus not be invalidated.

jeluard:
@rkuhn is this PR still in WIP mode?

PR #684, opened 2026-02-25
fix: return intersection points further in the past

etorreborre:
When we try to connect 2 amaru nodes the downstream node will fail to find an intersection with the  node if the downstream node is too for in the past. That because we don't look far enough for intersection points.

This PR tries to find an intersection with the `ChainStore::load_from_best_chain` function which can look past the current anchor for an intersection.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved chain synchronization reliability through enhanced intersection detection logic.

* **Tests**
  * Expanded test coverage for chain synchronization scenarios, including edge cases and comprehensive chain state validations.


etorreborre:
Sorry @rkuhn I accidentally approved the wrong PR. I meant to merge another one. Please tell me if you think it's off and I'll revert or adjust.

PR #688, opened 2026-02-26
chore: prepare for amaru release

jeluard:
Some steps to prepare for a release:
* vendor in crypto related deps that used to rely on git
* add release script and integrate with GH workflow
* add some extra metadata

What's blocking the release:
* uplc-turbo being itself released

Note that vendored dependencies are apache2/BSD3, both allows to be vendored in.

## Summary by CodeRabbit

* **New Features**
  * Added a vendored Curve25519-derived cryptography crate providing Edwards/Montgomery/Ristretto support, scalar/field arithmetic, multiscalar algorithms, SIMD backends and benchmarks.

* **Chores**
  * Updated workspace manifests and crate metadata (descriptions, README links, keywords, categories) and switched several workspace dependencies to vendored packages for reproducible builds.
  * Added a release/publish workflow.

* **Documentation**
  * Added README, changelog, contributing, license and CI docs for the vendored crate.


Cargo.toml
```diff
@@ -7,10 +7,13 @@ authors = ["Amaru Maintainers <amaru@pragma.builders>"]
...
 homepage = "https://github.com/pragma-org/amaru"
 documentation = "https://docs.rs/amaru"
+keywords = ["cardano", "blockchain", "node", "ouroboros", "crypto"]
+categories = ["cryptography::cryptocurrencies"]
 rust-version = "1.93"                                    # ⚠️ Also change in rust-toolchain.toml
 
 [workspace]
-members = ["crates/*", "simulation/*"]
+members = ["crates/*", "crates/amaru-observability/macros/", "simulation/*"]
+exclude = ["crates/vendor"]
```
KtorZ, line 16: I am not sure to understand why this is necessary? Does it mean we won't be publishing the crypto crates?

scripts/publish-crates
```diff
@@ -0,0 +1,35 @@
...
+cargo publish -p amaru-observability-macros && sleep 30
+cargo publish -p amaru-kernel              && sleep 30
+cargo publish -p amaru-observability       && sleep 30
+cargo publish -p amaru-ouroboros-traits    && sleep 30
+cargo publish -p amaru-plutus              && sleep 30
+cargo publish -p amaru-ouroboros           && sleep 30
+cargo publish -p amaru-mempool             && sleep 30
+cargo publish -p pure-stage                && sleep 30
+cargo publish -p amaru-network             && sleep 30
+cargo publish -p amaru-protocols
```
KtorZ: Perhaps better to use: https://github.com/crate-ci/cargo-release ? We've been using for Aiken and it (a) does take care of publishing the crates in the right order (b) polls crates.io and awaits for the crate to be published before moving to the next one. Seems less brittle than `sleep 30` 😅

Cargo.toml
jeluard: Releasing separately was the original idea now it makes sense to have everything bundled.
Still `vendor` has to be excluded, and `vendor/*` included, because of cargo workspaces limitations.

scripts/publish-crates
jeluard: 👍

PR #726, opened 2026-03-17
test: `select_chain_new` property test

etorreborre:
This PR adds a property test to check the new `select_chain` stage.

## Summary by CodeRabbit

* **Improvements**
  * Block header display now includes block height in its printed representation
  * Test data generation updated so generated slots vary near block height while remaining strictly increasing along branches

* **Tests**
  * Added comprehensive property-based tests for chain selection and adjusted unit test expectations
  * Introduced additional test-only modules to support the new property tests

* **Chores**
  * Updated a dev dependency in the test configuration


crates/amaru-consensus/src/headers_tree/data_generation/actions.rs
```diff
@@ -525,15 +525,16 @@ impl GeneratedActionsStatistics {
...
+        let peers = peers.clone();
+        (1..u64::MAX).prop_map(move |seed| generate_random_walks(seed, &generated_tree, &peers))
     })
 }
 
 /// Generate a random list of actions, for a fixed number of peers, with a given tree of headers.
-pub fn any_select_chains_from_tree(tree: &GeneratedTree, peers: &[Peer]) -> impl Strategy<Value = GeneratedActions> {
-    (1..u64::MAX).prop_map(move |seed| generate_random_walks(seed, tree, peers))
+pub fn any_select_chains_from_tree(tree: GeneratedTree, peers: Vec<Peer>) -> impl Strategy<Value = GeneratedActions> {
+    (1..u64::MAX).prop_map(move |seed| generate_random_walks(seed, &tree, &peers))
```
rkuhn: why these changes? seems unnecessary to me

crates/amaru-consensus/src/stages/select_chain_new/property_tests.rs
```diff
@@ -0,0 +1,332 @@
...
+        invalid_blocks.insert(candidates[index]);
+    }
+
+    invalid_blocks
+}
+
+/// Build an in-memory header store from the roll-forward actions only.
+///
+/// Rollbacks do not contribute new headers, so they are ignored here.
+fn initialize_store(actions: &[Action]) -> Arc<InMemConsensusStore<BlockHeader>> {
```
rkuhn: why is this function called in so many places instead of just passing around the built store?

crates/amaru-consensus/src/stages/select_chain_new/property_tests.rs
```diff
@@ -0,0 +1,332 @@
...
+            if !sent_tips.contains(&header.hash()) {
+                let message = SelectChainMsg::TipFromUpstream(header.tip(), parent_point);
+                result.push(message);
+                sent_tips.insert(header.hash());
+            }
+
+            if !validated_headers.contains(&header.hash()) {
+                let message =
+                    SelectChainMsg::BlockValidationResult(header.point(), !invalid_blocks.contains(&header.hash()));
+                result.push(message);
```
rkuhn: shouldn’t the result only come back after the tip has been sent to fetch_blocks?

crates/amaru-consensus/src/stages/select_chain_new/property_tests.rs
```diff
@@ -0,0 +1,332 @@
...
+    if candidates.is_empty() {
+        return BTreeSet::new();
+    }
+
+    let mut rng = StdRng::seed_from_u64(invalid_seed);
+    let mut invalid_blocks = candidates.iter().copied().filter(|_| rng.random_bool(0.35)).collect::<BTreeSet<_>>();
+
+    if invalid_blocks.is_empty() {
+        let index = rng.random_range(0..candidates.len());
+        invalid_blocks.insert(candidates[index]);
```
rkuhn: I don’t think an invalid block is always required

rkuhn (commented):
So this basically asserts that select_chain selects one of the longest chains (but doesn’t care which one) while getting there through a sequence of actions coming from the HeadersTree part of the codebase, right? That’s okay for me, although it feels a bit like belt and suspenders.

crates/amaru-consensus/src/headers_tree/data_generation/actions.rs
etorreborre: Yes, I reverted it.

crates/amaru-consensus/src/stages/select_chain_new/property_tests.rs
etorreborre: I refactored it.

crates/amaru-consensus/src/stages/select_chain_new/property_tests.rs
etorreborre: Yes we simulate a bit more than regular behavior here. In a way that serves as a test that the stage is resilient to those messages when they are irrelevant.

crates/amaru-consensus/src/stages/select_chain_new/property_tests.rs
etorreborre: I've added a random boolean to guard this behavior so that we don't always insert an invalid block during a test.

crates/amaru-consensus/src/stages/select_chain_new/property_tests.rs
```diff
@@ -0,0 +1,339 @@
...
+
+        // check the actual best chain is one of the expected best chains
+        // as given by the longest valid chains in the generated tree of headers.
+        let expected = make_expected_chains(&select_messages);
+
+        let success = if expected.is_empty() {
+            actual.is_empty()
+        } else {
+            expected.contains(&actual)
+        };
```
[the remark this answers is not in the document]
etorreborre: Done

crates/amaru-consensus/src/stages/select_chain_new/property_tests.rs
```diff
@@ -0,0 +1,380 @@
...
+            if !sent_tips.contains(&header.hash()) {
+                let message = SelectChainMsg::TipFromUpstream(header.tip(), parent_point);
+                result.push(message);
+                sent_tips.insert(header.hash());
+            }
+
+            if !validated_headers.contains(&header.hash()) {
+                let message =
+                    SelectChainMsg::BlockValidationResult(header.tip(), !invalid_blocks.contains(&header.hash()));
+                result.push(message);
```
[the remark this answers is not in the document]
etorreborre: Made a better generator now.

crates/amaru-consensus/src/stages/select_chain_new/property_tests.rs
etorreborre: I changed the generation now.

crates/amaru-consensus/src/headers_tree/data_generation/actions.rs
etorreborre: Actually, that's necessary to appease the borrow checker.

etorreborre:
@rkuhn I changed the property to assert that we should always emit the best tip that we know of when a previous tip has been invalidated.

Then I made the data generation generate more interesting cases for invalidating blocks and incorporating the possibility of restarts. That led me to fix the `select_chain` code to recompute the best tip candidate from store because it might not include the `best_chain_hash` anymore.

Does that seem right to you? What would you change?

crates/amaru-consensus/src/stages/select_chain/property_tests.rs
```diff
@@ -0,0 +1,71 @@
...
+/// from the sent headers and validation results so far.
+///
+fn execute_messages(generated_tree: GeneratedTree, seed: u64) -> Option<AssertionFailure> {
+    let store = Arc::new(InMemConsensusStore::default());
+    let mut harness = Harness::new(store.clone());
+    let mut state = TestState::new(generated_tree, seed, store.clone());
+
+    while let Some(action) = state.next_action() {
+        let outputs = harness.execute_action(&action, &state);
+        state.assert_best_known_tip(&outputs)?;
```
rkuhn, line 67: This line suggests that it would be better to return `Result<(), AssertionFailure>` instead of `Option<...>`.

crates/amaru-consensus/src/stages/select_chain/mod.rs
```diff
@@ -248,6 +228,126 @@ impl SelectChain {
...
+            }
+        })
+        .map(|(header, _)| header.hash())
+        .collect::<Vec<_>>();
+
+    valid.then_some(ancestors)
+}
+
+/// Return whether `hash` lies on a chain suffix whose ancestry includes `ancestor`.
+fn depends_on_hash(store: &(impl ChainStore<BlockHeader> + ?Sized), hash: HeaderHash, ancestor: HeaderHash) -> bool {
```
rkuhn: The `store` argument should simply be `&dyn ChainStore<BlockHeader>` everywhere.

crates/amaru-consensus/src/stages/select_chain/mod.rs
```diff
@@ -248,6 +228,126 @@ impl SelectChain {
...
+    let mut current = Some(hash);
+
+    while let Some(hash) = current {
```
rkuhn: Why not something like `store.ancestors(hash).any(|h| h == ancestor)`?

crates/amaru-consensus/src/stages/select_chain/mod.rs
```diff
@@ -248,6 +228,126 @@ impl SelectChain {
...
+/// returns `(leaf, [parent, leaf])`
+///
+/// If the leaf suffix crosses an invalid block, the candidate collapses to the
+/// best ancestor above that invalid segment:
+///
+/// leaf (None)
+///  |
+/// invalid (Some(false))
+///  |
+/// ancestor (None)
```
rkuhn: note that this constellation is not possible: there cannot be validation gaps on a fragment

crates/amaru-consensus/src/stages/select_chain/mod.rs
```diff
@@ -248,6 +228,126 @@ impl SelectChain {
...
+///  |
+/// validated (Some(true))
+///
+/// returns `(ancestor, [ancestor])`
+fn best_non_invalidated_candidate_from_leaf(
+    chain_store: &(impl ChainStore<BlockHeader> + ?Sized),
+    leaf_hash: HeaderHash,
+) -> Option<(BlockHeader, Vec<HeaderHash>)> {
```
rkuhn: actual algorithm should be:
- traverse upwards
- return None when hitting invalid
- return leaf + accumulator when hitting valid or anchor

crates/amaru-consensus/src/stages/select_chain/mod.rs
```diff
@@ -160,51 +147,44 @@ impl SelectChain {
             // remove all tips depending on the invalid block
             // (if a peer sends further headers on this chain, we will ignore them)
             let prev_tips = self.tips.len();
-            self.tips.retain(|_k, v| v.first() != Some(&tip.hash()));
+            self.tips.retain(|tip_hash, v| !v.contains(&tip.hash()) && !depends_on_hash(&store, *tip_hash, tip.hash()));
```
rkuhn: The `self.tips` values are sorted in order of validation and validation occurs strictly in chain sequence, so by definition any tip depending on this invalid block must have that hash first in the list.

crates/amaru-consensus/src/stages/select_chain/mod.rs
```diff
@@ -160,51 +147,44 @@ impl SelectChain {
...
+                                        "failed to load parent {:?} of best tip candidate {:?}",
+                                        parent,
+                                        new_best_tip
+                                    );
+                                })
+                                .await
+                                .point()
+                        } else {
+                            Point::Origin
+                        };
```
rkuhn: with the rest nicely cleaned up, this should also be moved into a separate function

crates/amaru-consensus/src/stages/select_chain/mod.rs
```diff
@@ -248,6 +228,126 @@ impl SelectChain {
...
+/// - Filter out tips that have invalid blocks.
+/// - Take the best one using the cmp function.
+/// - Also return the list of still non-validated block hashes for that tip.
+///
+pub fn best_tip_candidate_from_store(
+    store: &(impl ChainStore<BlockHeader> + ?Sized),
+) -> Option<(BlockHeader, Vec<HeaderHash>)> {
+    store
+        .child_tips(&store.get_anchor_hash())
+        .filter_map(|tip| best_non_invalidated_candidate_from_leaf(store, tip.hash()))
```
rkuhn: This is quadratic complexity because it keeps re-scanning the branch we’re walking on. It also is not needed because child_tips doesn’t walk invalid nodes. I’d suggest picking the best candidate header and then figuring out the list of hashes to be validated.

crates/amaru-consensus/src/stages/select_chain/mod.rs
```diff
@@ -160,51 +147,44 @@ impl SelectChain {
...
                 tracing::info!(%removed, "best tip candidate invalidated");
-                // need to pick new best tip
-                let (parent, new_best_tip) = self
-                    .tips
-                    .keys()
-                    .filter_map(|h| store.load_header(h).map(|h| (h.parent_hash(), Some(h))))
-                    .max_by(|a, b| cmp_tip(a.1.as_ref(), b.1.as_ref()))
-                    .unwrap_or_else(|| {
-                        store
-                            .load_header(&store.get_best_chain_hash())
```
rkuhn, line 178: I understand now what you meant by going further back than required, good point.

rkuhn (changes requested):
I reviewed the stage logic changes and a bit more, but I find the `property_tests` directory way too complex and don’t really get it. The point of a test should be that it is more obviously correct than the code it tests. At least at this hour I fail to see that this is the case here.

crates/amaru-consensus/src/stages/select_chain/property_tests.rs
etorreborre: Isn't that strictly equivalent? Also, `AssertionFailure` is too large and would need to be boxed.

crates/amaru-consensus/src/stages/select_chain/mod.rs
etorreborre: Done.

crates/amaru-consensus/src/stages/select_chain/mod.rs
etorreborre: Because `ancestors` takes a header as an input and iterates on block headers not header hashes. Anyway, this is not used anymore.

crates/amaru-consensus/src/stages/select_chain/mod.rs
etorreborre: Removed.

crates/amaru-consensus/src/stages/select_chain/mod.rs
etorreborre: This is not used anymore.

crates/amaru-consensus/src/stages/select_chain/mod.rs
etorreborre: I see, reverted.

crates/amaru-consensus/src/stages/select_chain/mod.rs
etorreborre: Done.

crates/amaru-consensus/src/stages/select_chain/mod.rs
etorreborre: I changed the fallback implementation

etorreborre:
@rkuhn I improved the description of the property we are testing. It is, IMHO, not so complicated:
```rust
/// Assert that the latest non-invalidated tip emitted downstream by `select_chain`
/// is among the best non-invalidated tips sent from upstream.
///
///
/// (a non-invalidated tip is a tip whose chain does not contain an invalidated block)
```

Here is the code:
```rust
pub fn assert_best_known_tip(&self, outputs: &[(Tip, Point)]) -> Option<AssertionFailure> {
    let latest_tip = self.latest_non_invalidated_tip_sent_downstream(outputs);

    // Some tips might be ex-aequo, this is why we have a Vec of tips here.
    let best_generated_tips = self.best_non_invalidated_generated_tips();

    // The latest tip sent downstream must be in that best tips list
    let ok = best_generated_tips.is_empty() && latest_tip.is_none()
        || latest_tip.as_ref().is_some_and(|tip| best_generated_tips.contains(tip));

    self.make_assertion_failure(!ok, outputs, latest_tip, best_generated_tips)
}
```

PR #728, opened 2026-03-19
test: check the tx submission protocol in the simulation

etorreborre:
This PR extends the current simulation test with the generation of transactions in downstream nodes and the verification that those transactions eventually end-up in upstream nodes.

## Summary by CodeRabbit

* **New Features**
  * Generate and inject transactions into network simulations; nodes accept injected transactions and track them in mempools
  * Simulation tests now validate transaction propagation across nodes
  * Persist generated transactions alongside actions in simulation outputs

* **UX / Configuration**
  * Test configs accept event lists (including transaction injections) and a transaction count
  * New CLI/env option --number-of-transactions to control generated transaction volume


simulation/amaru-sim/src/simulator/run_tests.rs
```diff
@@ -83,7 +86,9 @@ fn run_test_nb(run_config: &RunConfig, test_run_dir: &Path, test_number: u32) ->
     let actions = generate_actions(run_config);
     display_actions_statistics(&actions);
 
-    let test_result = run_test(run_config, &actions);
+    let transactions = generate_transactions(run_config);
+
+    let test_result = run_test(run_config, &actions, &transactions);
```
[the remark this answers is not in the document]
etorreborre: Good idea, done.

crates/amaru-consensus/src/headers_tree/data_generation/shrink.rs
```diff
@@ -97,6 +97,20 @@ impl<A: Debug + Clone> Shrinkable for Vec<A> {
...
+        Self: Sized,
+    {
+        let (a, b) = self;
+        (a.complement(from, to), b.complement(from, to))
+    }
+
+    fn len(&self) -> usize {
+        self.0.len() + self.1.len()
+    }
+}
```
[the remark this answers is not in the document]
etorreborre: Done and added tests.

crates/amaru-protocols/src/tx_submission/initiator.rs
```diff
@@ -466,6 +473,57 @@ mod tests {
...
+    #[tokio::test]
+    async fn blocking_request_waits_for_the_first_transaction() -> anyhow::Result<()> {
+        let mempool = Arc::new(SizedMempool::with_capacity(4));
+        let txs = create_transactions(1);
+        let mut initiator = TxSubmissionInitiator::new(StageRef::named_for_tests("muxer")).1;
+
+        let mempool_for_insert = mempool.clone();
+        let tx = txs[0].clone();
+        let insert_handle = tokio::spawn(async move {
+            tokio::task::yield_now().await;
```
rkuhn: this doesn’t do anything

crates/amaru-protocols/src/tx_submission/initiator.rs
```diff
@@ -466,6 +473,57 @@ mod tests {
...
+            mempool_for_insert.add(tx)
+        });
+
+        let action = step(&mut initiator, request_tx_ids(0, 2, Blocking::Yes), mempool.as_ref()).await?;
+        insert_handle.await??;
```
rkuhn: this also doesn’t do anything

crates/amaru-protocols/src/tx_submission/initiator.rs
```diff
@@ -466,6 +473,57 @@ mod tests {
...
+        let mut initiator = TxSubmissionInitiator::new(StageRef::named_for_tests("muxer")).1;
+
+        let mempool_for_insert = mempool.clone();
+        let tx = txs[0].clone();
+        let insert_handle = tokio::spawn(async move {
+            tokio::task::yield_now().await;
+            mempool_for_insert.add(tx)
+        });
+
+        let action = step(&mut initiator, request_tx_ids(0, 2, Blocking::Yes), mempool.as_ref()).await?;
```
rkuhn: what are you trying to achieve here? If I’m right, then you’d want to `tokio::spawn` this one, remove the insert_handle, wait a couple 100ms here, assert that no action was generated, then add the txs and assert that within a couple 100ms the action is generated.

rkuhn (commented):
Looks good to me except for that one test

crates/amaru-protocols/src/tx_submission/initiator.rs
```diff
@@ -730,4 +793,61 @@ mod tests {
...
+        fn tx_ids_since(&self, from_seq: MempoolSeqNo, limit: u16) -> Vec<(TxId, u32, MempoolSeqNo)> {
+            self.inner.tx_ids_since(from_seq, limit)
+        }
+
+        fn wait_for_at_least(&self, seq_no: MempoolSeqNo) -> Pin<Box<dyn Future<Output = bool> + Send + '_>> {
+            let wait_started = self.wait_started.clone();
+            let wait = self.inner.wait_for_at_least(seq_no);
+            Box::pin(async move {
+                wait_started.notify_one();
+                wait.await
```
rkuhn, line 827: given that the return value captures the `&self` lifetime, shouldn’t this also work?
```suggestion
            Box::pin(async move {
                self.wait_started.notify_one();
                self.inner.wait_for_at_least(seq_no).await
```

crates/amaru-protocols/src/tx_submission/initiator.rs
```diff
@@ -466,6 +476,59 @@ mod tests {
...
+        let mempool_for_step = mempool.clone();
+        let mempool_for_insert = mempool.clone();
+        let tx = txs[0].clone();
+        let step_handle = tokio::spawn(async move {
+            step(&mut initiator, request_tx_ids(0, 2, Blocking::Yes), mempool_for_step.as_ref()).await
+        });
+
+        wait_started.notified().await;
+        mempool_for_insert.add(tx)?;
+        let action = step_handle.await??;
```
rkuhn, line 526: This now tests the following:

- `step()` is getting to the point where it wants to wait for the txn and starts waiting
- we insert the txn to unblock `step()`

Do I understand correctly that this excludes `step()` from already having returned something because the `tx` cannot be known to the `initiator` before we insert it? In this case, I’d recommend adding a comment.

rkuhn (approved):
LGTM apart from two nitpicks.

PR #742, opened 2026-04-03
chore: add encoding benchmarks

rkuhn:
in anticipation of wanting the TraceBuffer activated in production, which will need some thinking

## Summary by CodeRabbit

* **Tests**
  * Added Criterion performance benchmarks measuring CBOR serialization/encoding for consensus stage messages and kernel types, covering validation, chain selection, adoption, block fetching, and peer-tracking scenarios.
* **Chores**
  * Enabled benchmarking in the dev environment and added configuration for new benchmark targets, gated to run only when the test-utils feature is enabled.

### DIFF
--- a/crates/amaru-consensus/benches/stage_msgs.rs
+++ b/crates/amaru-consensus/benches/stage_msgs.rs
@@ -1,0 +1,112 @@
+// Copyright 2026 PRAGMA
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::{hint::black_box, time::Duration};
+
+use amaru_consensus::stages::{
+    adopt_chain::AdoptChainMsg,
+    fetch_blocks::FetchBlocksMsg,
+    select_chain::SelectChainMsg,
+    track_peers::{DeferReqNextMsg, TrackPeersMsg},
+    validate_block::ValidateBlockMsg,
+};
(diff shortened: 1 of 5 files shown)


crates/amaru-consensus/benches/stage_msgs.rs
```diff
@@ -0,0 +1,154 @@
...
+use amaru_ouroboros::ConnectionId;
+use amaru_protocols::chainsync::{ChainSyncInitiatorMsg, HeaderContent, InitiatorMessage, InitiatorResult};
+use criterion::{Criterion, criterion_group, criterion_main};
+use pure_stage::{SendData, serde::to_cbor};
+
+fn stage_msgs(c: &mut Criterion) {
+    let mut group = c.benchmark_group("Stage Messages");
+    group.measurement_time(Duration::from_secs(5));
+
+    let tip = Tip::new(Point::Origin, BlockHeight::from(0));
```
etorreborre: There's `Tip::origin()` for this.

crates/amaru-consensus/benches/stage_msgs.rs
```diff
@@ -0,0 +1,154 @@
...
+    let point = Point::Origin;
+    let bh = BlockHeight::from(100);
+
+    group.bench_function("ValidateBlockMsg", |b| {
```
etorreborre: Maybe we could refactor a bit with `add_bench_function("data type", <boxed data>)`?

etorreborre (approved):

crates/amaru-consensus/benches/stage_msgs.rs
rkuhn: Ah right, of course. But it would probably be better to use some numbers that require 32bit CBOR because that is where we are today. Will fix.

crates/amaru-consensus/benches/stage_msgs.rs
rkuhn: I’m a big fan of keeping things as obvious as possible. These closures are really small and simple already, adding a layer of abstraction makes them one step more difficult to read.

rkuhn: I optimised the formatting a bit further so that each benchmark is a single line, with test values created before.

PR #744, opened 2026-04-07
Fix Required Vkey Witnesses

yHSJ:
Currently, we're missing 2 places that vkey witnesses are required: collateral inputs and stake pool *owners*. Stake pool owners are the set of keys that provide the pledge. It could be the same key as the stake pool operator.

Neither the conformance tests nor unit tests provide coverage for the stake pool owner case right now, but this behavior now matches the [Haskell node](https://github.com/IntersectMBO/cardano-ledger/blob/master/eras/shelley/impl/src/Cardano/Ledger/Shelley/UTxO.hs#L217-L226)

## Release Notes

* **Bug Fixes**
  * Enhanced witness validation for collateral inputs based on address credentials (vkey or bootstrap witnesses)
  * Strengthened pool registration by requiring witness validation for all pool owners


etorreborre (approved):
@yHSJ looks all good!

PR #747, opened 2026-04-09
refactor: remove sync effects

etorreborre:
This PR removes "external sync effects" so that only async external effects can be executed in `pure-stage` stages. This will allow the simulation to fully simulate the passing of time by suspending all these effects and simulating the time they take to execute.

All of those effects are related to accesses to the chain store. Some of this code was doing repeated accesses on iterators retrieved from the database, with data that could potentially mutate while being traversed. 
This kind of processing has now been encapsulated in specific effects that take a snapshot of the chain store in order to work on consistent data.

## Summary by CodeRabbit

* **New Features**
  * Read-only chain-store snapshots for consistent queries
  * Explicit missing-blocks detection to optimize block fetching
  * Added non-empty vector constructor for convenience
  * Trace/schema: rollback event renamed to SWITCH_TO_FORK

* **Refactor**
  * Migrated ledger, store, mempool, and protocol plumbing to async-first APIs
  * Reworked fork switching (switch_to_fork) and simplified reorg handling
  * Unified termination helper (OrTerminateWith) and streamlined error propagation


etorreborre:
@coderabbitai review

crates/amaru-consensus/src/effects/ledger_effects.rs
```diff
@@ -33,79 +33,78 @@ pub trait LedgerOps: Send + Sync {
...
-        self.0.external_sync(ContainsPointEffect::new(point))
+    fn contains_point(&self, point: &Point) -> BoxFuture<'static, bool> {
+        self.effects.external(ContainsPointEffect::new(point))
     }
 
-    fn tip(&self) -> Tip {
-        self.0.external_sync(TipEffect)
+    fn tip(&self) -> BoxFuture<'static, Tip> {
+        self.effects.external(TipEffect)
     }
```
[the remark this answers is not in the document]
etorreborre: Fixed

crates/amaru-consensus/src/stages/fetch_blocks/mod.rs
```diff
@@ -71,67 +74,61 @@ impl FetchBlocks {
...
-        assert!(self.missing.is_empty(), "missing blocks should be empty when starting a new tip: {:?}", self.missing);
-        let initial = store
-            .load_header(&tip.hash())
-            .or_terminate(store.eff(), async |_| {
+        let store = Store::new(eff.clone());
+        assert!(
+            self.missing.is_none(),
+            "there shouldn't be any missing blocks when starting a new tip: {:?}",
+            self.missing
+        );
```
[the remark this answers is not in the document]
etorreborre: That's ok

crates/amaru-stores/src/rocksdb/consensus/mod.rs
```diff
@@ -429,59 +571,85 @@ impl<H: IsHeader + Clone + Debug + for<'d> cbor::Decode<'d, ()>> ChainStore<H> f
...
+                tx.put([&CHAIN_PREFIX[..], &slot[..]].concat(), point.hash().as_ref())
+                    .map_err(|e| StoreError::WriteError { error: e.to_string() })?;
+            }
+
+            tx.put(BEST_CHAIN_PREFIX, forward_points.last().hash().as_ref())
+                .map_err(|e| StoreError::WriteError { error: e.to_string() })?;
+
+            Ok(())
+        })
+    }
```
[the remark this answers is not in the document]
etorreborre: Done.

crates/amaru/src/stages/build_node.rs
```diff
@@ -95,41 +95,9 @@ pub fn build_node(
...
-            } else {
-                break;
-            }
-        }
-        best_missing.reverse();
-    }
+    let best_candidate = startup_best_candidate(chain_store.as_ref());
 
-    let tip = best_candidate.as_ref().map(|h| h.point()).unwrap_or(Point::Origin);
+    let tip = best_candidate.as_ref().map(|(h, _)| h.point()).unwrap_or(Point::Origin);
```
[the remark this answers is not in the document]
etorreborre: Done.

crates/amaru-ouroboros-traits/src/stores/consensus/missing_blocks.rs
```diff
@@ -0,0 +1,59 @@
...
+
+    pub fn nb_missing_blocks(&self) -> usize {
+        self.missing.len()
+    }
+
+    /// This method is called when the first missing block has been fetched.
+    /// It is then removed from the list of missing blocks and becomes the boundary.
+    pub fn shift_one_block(&mut self) {
+        self.boundary = self.missing.remove(0);
+    }
```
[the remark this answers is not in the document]
etorreborre: Done.

crates/amaru-ouroboros-traits/src/stores/consensus/mod.rs
```diff
@@ -208,11 +210,309 @@ impl<H: IsHeader> ReadOnlyChainStore<H> for Box<dyn ChainStore<H>> {
...
+            return Ok(NextBestChainHeader::AtTip);
+        };
+        let Some(header) = snapshot.load_header(&point.hash()) else {
+            return Ok(NextBestChainHeader::MissingHeader { point });
+        };
+        if header.parent().unwrap_or(ORIGIN_HASH) != pointer.hash() {
+            return Ok(NextBestChainHeader::NeedRollback);
+        }
+        Ok(NextBestChainHeader::RollForward { point, header })
+    }
```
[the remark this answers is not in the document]
etorreborre: done.

crates/amaru-ouroboros-traits/src/stores/consensus/mod.rs
```diff
@@ -208,11 +210,309 @@ impl<H: IsHeader> ReadOnlyChainStore<H> for Box<dyn ChainStore<H>> {
...
+            if index + 1 == spacing {
+                points.push(last);
+                spacing *= 2;
+            }
+        }
+        if points.last() != Some(&last) {
+            points.push(last);
+        }
+        points
+    }
```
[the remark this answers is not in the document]
etorreborre: That's the previous behavior, not changing it in this PR.

crates/amaru-ouroboros-traits/src/stores/consensus/overriding_consensus_store.rs
```diff
@@ -327,6 +328,10 @@ impl<H: IsHeader + Send + Sync + 'static> ReadOnlyChainStore<H> for OverridingCh
 }
 
 impl<H: IsHeader + Send + Sync + 'static> ChainStore<H> for OverridingChainStore<H> {
+    fn snapshot(&self) -> Box<dyn ReadOnlyChainStore<H> + '_> {
+        self.inner.snapshot()
+    }
```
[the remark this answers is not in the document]
etorreborre: Valid point. Fixed.

crates/amaru-protocols/src/connection.rs
```diff
@@ -279,11 +279,11 @@ async fn do_handshake(
...
     } else {
         let store = Store::new(eff.clone());
-        let upstream = store.get_best_chain_hash();
+        let upstream = store.get_best_chain_hash().await;
         let upstream = if upstream == ORIGIN_HASH {
             Tip::new(Point::Origin, 0.into())
         } else {
-            let header = store.load_header(&upstream).expect("best chain hash not found");
+            let header = store.load_header(&upstream).await.expect("best chain hash not found");
             header.tip()
```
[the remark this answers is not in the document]
etorreborre: Added a `get_best_chain_tip` function to avoid this 2-steps dance.

crates/amaru-consensus/src/stages/validate_block/mod.rs
```diff
@@ -144,70 +147,99 @@ fn validation_failed_when_contains_claimed_rollback(rollback_target: Point, vf:
...
-        Err(ValidationFailed::new(
-            // TODO: figure out which peer to blame
-            &Peer::new("unknown"),
-            ConsensusError::RollbackBlockFailed(
-                parent,
-                anyhow::anyhow!("rollback point not found in volatile db").into(),
-            ),
-        ))
+        current_hash = parent_hash;
     }
```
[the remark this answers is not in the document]
etorreborre: I'm running the search in a snapshot now.

crates/amaru-stores/src/rocksdb/consensus/mod.rs
```diff
@@ -177,6 +192,124 @@ pub(crate) fn store_chain_point(db: &OptimisticTransactionDB, point: &Point) ->
...
+                None
+            }
+        })
+    }
+
+    fn next_best_chain(&self, point: &Point) -> Option<Point> {
+        let mut readopts = self.read_options();
+        readopts.set_iterate_range(PrefixRange(CHAIN_PREFIX));
+        let prefix = [&CHAIN_PREFIX[..], &(u64::from(point.slot_or_default()) + 1).to_be_bytes()].concat();
+        let mut iter = self.db.iterator_opt(IteratorMode::From(&prefix, rocksdb::Direction::Forward), readopts);
```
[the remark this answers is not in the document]
etorreborre: Done and refactored the code that was similar

crates/amaru-consensus/src/stages/adopt_chain/mod.rs
```diff
@@ -69,35 +68,50 @@ pub async fn stage(mut state: AdoptChain, msg: AdoptChainMsg, eff: Effects<Adopt
...
-    let current_best = store
-        .load_header(&state.current_best_tip.hash())
-        .or_terminate(store.eff(), async |_| {
-            tracing::warn!("failed to load current best");
-        })
-        .await;
-
-    if cmp_tip(Some(&incoming_header), Some(&current_best)) != Ordering::Greater {
+    let current_best = if state.current_best_tip.point() == Point::Origin {
+        None
```
rkuhn, line 90: This has irked me since the beginning: There’s this one Point we can’t ask the store about. Wouldn’t it be better to return “the root header” when calling `.load_header(&ORIGIN_HASH)`? It has hash zero, no body, no parent.

We probably shouldn’t do it this week in particular, but please leave a TODO in the store trait.

crates/amaru-consensus/src/stages/adopt_chain/mod.rs
```diff
@@ -108,82 +122,34 @@ pub async fn stage(mut state: AdoptChain, msg: AdoptChainMsg, eff: Effects<Adopt
...
-        }
-        forward_points.push(point);
-    }
-    Err(StoreError::ReadError { error: "rollback point not found: new tip has no ancestor on best chain".to_string() })
+    let Some((fork_point, forward_points)) = store.find_fork_point(incoming_header.hash()).await else {
+        return Err(StoreError::ReadError {
+            error: "rollback point not found: new tip has no ancestor on best chain".to_string(),
+        });
+    };
+    store.switch_to_fork(&fork_point, &forward_points).await
```
rkuhn: why not compute the fork_point inside the `switch_to_fork` effect as well?

crates/amaru-consensus/src/stages/adopt_chain/mod.rs
```diff
@@ -108,82 +122,34 @@ pub async fn stage(mut state: AdoptChain, msg: AdoptChainMsg, eff: Effects<Adopt
...
-
-    if target_anchor_height <= current_anchor_tip.block_height() {
-        return Ok(());
+/// blocks behind the adopted tip. The walk along the best chain runs in a single
+/// store effect against a consistent snapshot; only the resulting anchor write
+/// is a separate effect. Only moves the anchor forward, never backward.
+async fn drag_anchor_forward(store: &Store, tip: &Tip, consensus_security_param: u64) -> Result<(), StoreError> {
+    let target_height = tip.block_height() - consensus_security_param;
+    if let Some(new_anchor) = store.find_anchor_at_height(target_height).await {
+        store.set_anchor_hash(&new_anchor).await?;
```
rkuhn, line 178: This should really also be `set_anchor_tip` to avoid having to `.load_header` when using it.

Again a TODO for now.

crates/amaru-consensus/src/stages/fetch_blocks/mod.rs
```diff
@@ -71,67 +74,61 @@ impl FetchBlocks {
...
+
+        match store.find_missing_blocks(tip.hash(), MAX_MISSING_BLOCKS_PER_BATCH).await {
+            Ok(None) => {
```
rkuhn: This seems wrong: when failing to load a header that should be there, the result should be `Err` not `Ok`.

With this implied signature for `find_missing_blocks`, I’d think that `Ok(None)` means that we’re at the tip already, nothing is missing. Perhaps that would be better than returning a MissingBlocks that can then indicate that nothing was actually missing — which seems a bit surprising.

crates/amaru-consensus/src/stages/fetch_blocks/mod.rs
```diff
@@ -71,67 +74,61 @@ impl FetchBlocks {
...
+                self.missing = None;
+                tracing::info!(tip = %tip.point(), parent = %parent, "no blocks to fetch");
+                eff.send(&self.upstream, SelectChainMsg::FetchNextFrom(tip.point())).await;
+            }
+            Some((from, to)) => {
+                tracing::debug!(%from, %to, length = missing_blocks.nb_missing_blocks(), "requesting blocks");
+                self.req_id += 1;
+                eff.send(
+                    &self.manager,
+                    ManagerMessage::FetchBlocks2 {
```
rkuhn, line 120: ah, we can now retire the old FetchBlocks variant that I left in place while transitioning to the new API — made a new issue #778

crates/amaru-consensus/src/stages/fetch_blocks/mod.rs
```diff
@@ -148,46 +145,56 @@ impl FetchBlocks {
...
         }
-        // check that parent is as expected
-        if header.parent_hash() != self.missing.first().map(|p| p.hash()) {
-            tracing::warn!(expected = ?self.missing.first().map(|p| p.hash()), actual = ?header.parent_hash(), "block parent hash mismatch");
+        let Some(missing_blocks) = self.missing.as_ref() else {
+            tracing::warn!("received block with no outstanding missing blocks");
+            return;
+        };
+        if header.parent_hash() != Some(missing_blocks.boundary().hash()) {
+            tracing::warn!(expected = ?Some(missing_blocks.boundary().hash()), actual = ?header.parent_hash(), "block parent hash mismatch");
```
rkuhn:
```suggestion
            tracing::warn!(expected = %missing_blocks.boundary().hash(), actual = %header.parent_hash().unwrap_or(ORIGIN_HASH), "block parent hash mismatch");
```
We shouldn’t log Option’s Debug impl, especially not when it is known `Some(…)`.

crates/amaru-consensus/src/stages/fetch_blocks/mod.rs
```diff
@@ -148,46 +145,56 @@ impl FetchBlocks {
...
             return;
         }
-        // check that block's point is as expected
-        if Some(&point) != self.missing.get(1) {
-            tracing::warn!(expected = ?self.missing.get(1), actual = ?point, "block point mismatch");
+        if Some(point) != missing_blocks.first() {
+            tracing::warn!(expected = ?missing_blocks.first(), actual = ?point, "block point mismatch");
```
rkuhn: same here

crates/amaru-consensus/src/stages/fetch_blocks/mod.rs
```diff
@@ -148,46 +145,56 @@ impl FetchBlocks {
             tracing::warn!(expected = %header.header().header_body.block_body_hash, actual = %block.body_hash(), "block body hash mismatch");
             return;
         }
-        // check that parent is as expected
-        if header.parent_hash() != self.missing.first().map(|p| p.hash()) {
-            tracing::warn!(expected = ?self.missing.first().map(|p| p.hash()), actual = ?header.parent_hash(), "block parent hash mismatch");
+        let Some(missing_blocks) = self.missing.as_ref() else {
```
rkuhn: just use `.as_mut()` here and you have a `&mut MissingBlocks` for the rest of the function

crates/amaru-consensus/src/stages/fetch_blocks/mod.rs
```diff
@@ -148,46 +145,56 @@ impl FetchBlocks {
...
-        // check if we are done
-        if self.missing.len() < 2 {
-            self.missing.clear();
+        let done = {
+            #[expect(clippy::expect_used)]
+            let missing = self.missing.as_mut().expect("checked above that this exists");
+            missing.shift_one_block();
+            missing.is_empty()
+        };
+        if done {
```
rkuhn:
```suggestion
        missing.shift_one_block();
        if missing.is_empty() {
```
(with the above `.as_mut()` suggestion)

crates/amaru-consensus/src/stages/fetch_blocks/mod.rs
```diff
@@ -148,46 +145,56 @@ impl FetchBlocks {
             tracing::warn!(expected = %header.header().header_body.block_body_hash, actual = %block.body_hash(), "block body hash mismatch");
             return;
         }
-        // check that parent is as expected
-        if header.parent_hash() != self.missing.first().map(|p| p.hash()) {
-            tracing::warn!(expected = ?self.missing.first().map(|p| p.hash()), actual = ?header.parent_hash(), "block parent hash mismatch");
+        let Some(missing_blocks) = self.missing.as_ref() else {
+            tracing::warn!("received block with no outstanding missing blocks");
```
rkuhn, line 155: We should eventually accept blocks that we want to have but didn’t get within the timeout, which will require checking some invariants (like that the parent block must already be stored and no parent is invalid). If you agree, could you add a TODO here?

crates/amaru-consensus/src/stages/fetch_blocks/tests.rs
```diff
@@ -117,15 +113,7 @@ fn test_new_tip_blocks_to_fetch() {
...
-            te_load_header("fb-1", tip.hash()),
-            te_get_anchor_hash("fb-1"),
-            te_get_anchor_hash("fb-1"),
-            te_load_header("fb-1", prep.headers.h0.hash()),
-            te_load_header("fb-1", prep.headers.h1.hash()),
-            te_load_block("fb-1", prep.headers.h2.hash()),
-            te_load_header("fb-1", prep.headers.h0.hash()),
-            te_load_block("fb-1", prep.headers.h1.hash()),
-            te_load_block("fb-1", prep.headers.h0.hash()),
+            te_find_missing_blocks("fb-1", tip.hash(), 10),
```
rkuhn: these changes feel very good

crates/amaru-consensus/src/stages/validate_block/mod.rs
```diff
@@ -144,70 +148,54 @@ fn validation_failed_when_contains_claimed_rollback(rollback_target: Point, vf:
...
                 anyhow::anyhow!("rollback point not found in volatile db").into(),
             ),
-        ))
+        )),
     }
 }
 
+async fn find_rollback_point(ledger: &Ledger, store: &Store, parent: Point) -> RollbackPointSearchResult {
+    let ledger_tip = ledger.tip().await;
+    store.find_rollback_point(parent.hash(), ledger_tip.point()).await
```
rkuhn: This is not correct: the search must stop as soon as we hit a point that is part of the volatile DB

crates/amaru-ouroboros-traits/src/stores/consensus/missing_blocks.rs
```diff
@@ -0,0 +1,63 @@
...
+    pub fn new(boundary: Point, missing: Vec<Point>) -> Self {
+        Self { boundary, missing: VecDeque::from(missing) }
+    }
+
+    pub fn boundary(&self) -> Point {
+        self.boundary
+    }
+
+    pub fn first(&self) -> Option<Point> {
+        self.missing.front().cloned()
```
rkuhn: please use `.copied()` for these to assert bitwise copy

crates/amaru-ouroboros-traits/src/stores/consensus/missing_blocks.rs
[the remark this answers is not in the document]
rkuhn: nah, this is fine; we could add a `debug_assert!()`

crates/amaru-ouroboros-traits/src/stores/consensus/mod.rs
```diff
@@ -58,6 +60,12 @@ where
         self.load_header(hash).map(|h| h.tip())
     }
 
+    #[expect(clippy::expect_used)]
+    fn get_best_chain_tip(&self) -> Tip {
```
rkuhn, line 64: This is good, please put in a TODO to optimize this by actually storing the tip.

crates/amaru-ouroboros-traits/src/stores/consensus/mod.rs
```diff
@@ -208,11 +216,365 @@ impl<H: IsHeader> ReadOnlyChainStore<H> for Box<dyn ChainStore<H>> {
...
+            return Ok(NextBestChainHeader::NeedRollback);
+        }
+        let Some(point) = snapshot.next_best_chain(pointer) else {
+            return Ok(NextBestChainHeader::AtTip);
+        };
+        let Some(header) = snapshot.load_header(&point.hash()) else {
+            return Ok(NextBestChainHeader::MissingHeader { point });
+        };
+        if header.parent().unwrap_or(ORIGIN_HASH) != pointer.hash() {
+            return Ok(NextBestChainHeader::NeedRollback);
```
rkuhn: This can no longer happen due to the snapshot, so this would be a panic in my book.

rkuhn (commented):
got up to the start of the ChainStore trait, will continue in a few hours

crates/amaru-ouroboros-traits/src/stores/consensus/mod.rs
```diff
@@ -208,11 +216,365 @@ impl<H: IsHeader> ReadOnlyChainStore<H> for Box<dyn ChainStore<H>> {
...
+    ///
+    /// Returns `(C, [F, G])`.
+    ///
+    /// Returns None if the start point is already on the best chain.
+    fn find_fork_point(&self, start: HeaderHash) -> Option<(Point, NonEmptyVec<Point>)>
+    where
+        H: 'static,
+    {
+        let snapshot = self.snapshot();
+        let header = snapshot.load_header(&start)?;
```
rkuhn: This should be an error condition. We are currently not entirely consistent on these between `panic!` and terminating the stage after logging an error. My preference is to return a Result.

crates/amaru-ouroboros-traits/src/stores/consensus/mod.rs
```diff
@@ -208,11 +216,365 @@ impl<H: IsHeader> ReadOnlyChainStore<H> for Box<dyn ChainStore<H>> {
...
+    /// O--A--B--C
+    ///          \
+    ///           F--G
+    ///              ^
+    ///            start = new best tip
+    ///
+    /// Returns `(C, [F, G])`.
+    ///
+    /// Returns None if the start point is already on the best chain.
+    fn find_fork_point(&self, start: HeaderHash) -> Option<(Point, NonEmptyVec<Point>)>
```
rkuhn:
```suggestion
    fn find_ancestor_on_best_chain(&self, start: HeaderHash) -> Option<(Point, NonEmptyVec<Point>)>
```

crates/amaru-ouroboros-traits/src/stores/consensus/mod.rs
```diff
@@ -208,11 +216,365 @@ impl<H: IsHeader> ReadOnlyChainStore<H> for Box<dyn ChainStore<H>> {
...
+    where
+        H: 'static,
+    {
+        let snapshot = self.snapshot();
+        let header = snapshot.load_header(&start)?;
+        let mut forward_points = None;
```
rkuhn: No need for the `Option`, I’d just start with `Vec::new` here — the `.pop()?` further down ensures the `None` return as appropriate.

crates/amaru-ouroboros-traits/src/stores/consensus/mod.rs
```diff
@@ -208,11 +216,365 @@ impl<H: IsHeader> ReadOnlyChainStore<H> for Box<dyn ChainStore<H>> {
...
+            if snapshot.load_from_best_chain(&point).is_some() {
+                // Both `?` here implement the "start was already on best chain" case from the
+                // contract: if the first ancestor we visit is already on the best chain, we
+                // never populated `forward_points` (outer `?`), or we only pushed `start`
+                // itself and then popped it (inner `?` leaves nothing behind). Either way we
+                // return `None`.
+                let mut forward_points: Vec<Point> = forward_points?;
+                let first = forward_points.pop()?;
+                forward_points.reverse();
+                return Some((point, NonEmptyVec::new(first, forward_points)));
```
rkuhn: so we already had the right vector but reconstruct it again in the NonEmptyVec constructor — I’d prefer to just use `NonEmptyVec::try_from(forward_points).ok()?`

crates/amaru-ouroboros-traits/src/stores/consensus/mod.rs
```diff
@@ -208,11 +216,365 @@ impl<H: IsHeader> ReadOnlyChainStore<H> for Box<dyn ChainStore<H>> {
...
+                    break;
+                }
+            }
+            if let Some(a_point) = chain1.peek() {
+                if *a_point == point {
+                    return Some(point);
+                }
+            } else {
+                break;
+            }
```
rkuhn:
```suggestion
        'outer: for point in snapshot.ancestors(header2).map(|h| h.point()) {
            while let Some(a_point) = chain1.peek() {
                if a_point > point {
                    chain1.next();
                } else if a_point == point {
                    return Some(point);
                } else {
                    continue 'outer;
                }
            }
            break;
```

crates/amaru-ouroboros-traits/src/stores/consensus/mod.rs
```diff
@@ -208,11 +216,365 @@ impl<H: IsHeader> ReadOnlyChainStore<H> for Box<dyn ChainStore<H>> {
...
+
+    /// Find the first point, in the list of points, that intersects with the best chain.
+    ///
+    /// Return None if none of the points is on the best chain
+    fn find_intersect_point(&self, mut points: Vec<Point>) -> Option<Point>
+    where
+        H: 'static,
+    {
+        let snapshot = self.snapshot();
+        let best_chain_is_non_empty = snapshot.get_best_chain_hash() != ORIGIN_HASH;
```
rkuhn: If I remember correctly, Point::Origin is always a valid intersection, even when the best chain is empty. OTOH we typically won’t have the blocks, but that is another topic.

crates/amaru-ouroboros-traits/src/stores/consensus/mod.rs
```diff
@@ -208,11 +216,365 @@ impl<H: IsHeader> ReadOnlyChainStore<H> for Box<dyn ChainStore<H>> {
...
+            (point == Point::Origin && best_chain_is_non_empty) || snapshot.load_from_best_chain(&point).is_some()
+        })
+    }
+
+    /// Find the closest rollback point when walking ancestors from `parent_hash`.
+    ///
+    /// The search runs against a single store snapshot. It returns the ledger tip
+    /// itself if encountered, otherwise it returns the closest ancestor that is both
+    /// on the best chain and marked valid. Invalid ancestors stop the search, and
+    /// ancestors below `ledger_tip` are rejected as below the rollback boundary.
```
rkuhn: I don’t understand this explanation: everything on the best chain has been validated with positive result, by definition. So this sounds like it is `find_ancestor_on_best_chain` with a cutoff point, right? In that case I’d fold both methods into one.

crates/amaru-ouroboros-traits/src/stores/consensus/mod.rs
```diff
@@ -208,11 +216,365 @@ impl<H: IsHeader> ReadOnlyChainStore<H> for Box<dyn ChainStore<H>> {
...
+    fn sample_ancestor_points(&self) -> Vec<Point>
+    where
+        H: 'static,
+    {
+        let snapshot = self.snapshot();
+        let best = snapshot.get_best_chain_hash();
+        if best == ORIGIN_HASH {
+            return vec![Point::Origin];
+        }
+        let Some(best) = snapshot.load_header(&best) else {
```
rkuhn, line 413: again, this is a failure and should be treated as such

crates/amaru-ouroboros-traits/src/stores/consensus/mod.rs
```diff
@@ -208,11 +216,365 @@ impl<H: IsHeader> ReadOnlyChainStore<H> for Box<dyn ChainStore<H>> {
...
+            return vec![Point::Origin];
+        };
+        let best_point = best.tip().point();
```
rkuhn:
```suggestion
        let best_point = best.point();
```
doesn’t this work?

crates/amaru-stores/src/rocksdb/consensus/mod.rs
```diff
@@ -177,6 +192,107 @@ pub(crate) fn store_chain_point(db: &OptimisticTransactionDB, point: &Point) ->
...
+            }
+        })
+    }
+
+    fn next_best_chain(&self, point: &Point) -> Option<Point> {
+        get_next_best_chain_point(self.db, point, self.read_options())
+    }
+}
+
 impl<H, T: DbOps> ReadOnlyChainStore<H> for RocksDBStore<T>
```
rkuhn, line 216: You could also reuse this impl for the snapshot by putting the `SnapshotWithThreadMode` into the `RocksDBStore<>` and then only implementing DbOps.

crates/amaru/src/stages/build_node.rs
```diff
@@ -251,3 +216,110 @@ fn make_validate_header(
...
+) -> Option<(BlockHeader, Vec<amaru_kernel::HeaderHash>)> {
+    let best_chain_hash = chain_store.get_best_chain_hash();
+    let anchor = chain_store.get_anchor_hash();
+    chain_store
+        .child_tips(&best_chain_hash)
+        .filter(|tip| tip.hash() != best_chain_hash)
+        .filter_map(|tip| {
+            let candidate = chain_store.load_header(&tip.hash())?;
+            let candidate_hash = candidate.hash();
+            let (hashes, valid) = chain_store.unvalidated_ancestor_hashes(candidate_hash);
```
rkuhn: This is a very expensive check and I don’t think it should be needed: `.child_tips` should not return invalid points (I don’t think it does).

crates/amaru/src/stages/build_node.rs
```diff
@@ -251,3 +216,110 @@ fn make_validate_header(
...
+/// a known invalid block. This prevents node restart from restoring a branch that was already
+/// determined to be invalid before shutdown.
+fn startup_best_candidate(
+    chain_store: &(impl ChainStore<BlockHeader> + ?Sized),
+) -> Option<(BlockHeader, Vec<amaru_kernel::HeaderHash>)> {
+    let best_chain_hash = chain_store.get_best_chain_hash();
+    let anchor = chain_store.get_anchor_hash();
+    chain_store
+        .child_tips(&best_chain_hash)
+        .filter(|tip| tip.hash() != best_chain_hash)
```
rkuhn: Why filter this out? If the best_chain_hash is the best starting point then we should use that — ledger_tip is 2160 blocks in the past.

crates/amaru/src/stages/build_node.rs
```diff
@@ -251,3 +216,110 @@ fn make_validate_header(
...
+        .filter_map(|tip| {
+            let candidate = chain_store.load_header(&tip.hash())?;
+            let candidate_hash = candidate.hash();
+            let (hashes, valid) = chain_store.unvalidated_ancestor_hashes(candidate_hash);
+            if valid {
+                let missing = hashes.into_iter().filter(|h| *h != anchor).collect();
+                Some((candidate, missing))
```
rkuhn: If the `missing` are needed (which I don’t yet understand) then please only obtain them for the one point we return, not for every visited tip (which is not only the tips but all tree nodes).

crates/amaru/src/stages/build_node.rs
```diff
@@ -125,41 +124,9 @@ pub fn build_node(
...
-            } else {
-                break;
-            }
-        }
-        best_missing.reverse();
-    }
+    let best_candidate = startup_best_candidate(chain_store.as_ref());
 
-    let tip = best_candidate.as_ref().map(|h| h.point()).unwrap_or(Point::Origin);
+    let tip = best_candidate.as_ref().map(|(h, _)| h.point()).unwrap_or(ledger_tip.point());
```
rkuhn: `ledger_tip` should not be here, this is only about the chain store.

rkuhn (commented):
Great work! Some things to fix, but we’re taking steps towards a better ChainStore :-)

crates/amaru-consensus/src/stages/adopt_chain/mod.rs
etorreborre: Done.

crates/amaru-consensus/src/stages/adopt_chain/mod.rs
etorreborre: To avoid pushing  too much logic into the store.

crates/amaru-consensus/src/stages/adopt_chain/mod.rs
etorreborre: Done.

crates/amaru-consensus/src/stages/fetch_blocks/mod.rs
etorreborre: From the point of view of the `Store` we only return errors that are generic to the store access: IO, retries, etc... All the rest is considered Ok from the point of view of the store. I can be clearer by using an enum that explicitly flags "missing start header" and "no missing blocks". See what I did in #[4ffe635](https://github.com/pragma-org/amaru/pull/747/commits/4ffe63560bc6ef5740121bd147b1af2e2c0e100e)

crates/amaru-consensus/src/stages/fetch_blocks/mod.rs
etorreborre: Done

crates/amaru-consensus/src/stages/fetch_blocks/mod.rs
etorreborre: Done

crates/amaru-consensus/src/stages/fetch_blocks/mod.rs
etorreborre: Done.

crates/amaru-consensus/src/stages/fetch_blocks/mod.rs
etorreborre: Done.

crates/amaru-consensus/src/stages/fetch_blocks/mod.rs
etorreborre: Makes sense, I added the TODO.

crates/amaru-ouroboros-traits/src/stores/consensus/missing_blocks.rs
etorreborre: Done.

crates/amaru-ouroboros-traits/src/stores/consensus/mod.rs
etorreborre: Done.

crates/amaru-ouroboros-traits/src/stores/consensus/mod.rs
etorreborre: Right, I'm removing the test entirely.

etorreborre:
Thanks a lot @rkuhn for the careful review! I pushed a few commits and will tackle part 2 next. When starting this I did not realize how much work was necessary to make the store tight :-).

crates/amaru-consensus/src/stages/adopt_chain/mod.rs
rkuhn: Yes, after reading the rest I concur. This does assume only one actor modifying the best chain because the read happens on a potentially different database version. AFAICS this is currently okay, but it should be documented.

crates/amaru-ouroboros-traits/src/stores/consensus/mod.rs
etorreborre: Yes, fixed.

crates/amaru-ouroboros-traits/src/stores/consensus/mod.rs
etorreborre: Done, by introducing an enum.

crates/amaru-ouroboros-traits/src/stores/consensus/mod.rs
etorreborre: Done.

crates/amaru-ouroboros-traits/src/stores/consensus/mod.rs
etorreborre: Good point on the whole implementation, I fixed it.

crates/amaru-ouroboros-traits/src/stores/consensus/mod.rs
etorreborre: 👍

crates/amaru-ouroboros-traits/src/stores/consensus/mod.rs
etorreborre: Done.

crates/amaru-ouroboros-traits/src/stores/consensus/mod.rs
etorreborre: Done.

crates/amaru-ouroboros-traits/src/stores/consensus/mod.rs
etorreborre: Yes, I simplified that.

crates/amaru-stores/src/rocksdb/consensus/mod.rs
etorreborre: Good idea, done.

crates/amaru/src/stages/build_node.rs
etorreborre: I see. I misinterpreted the original logic because `child_tips` made me think of actual leaves on the tree of nodes.

That being said we are already exploring descendants in `child_tips`, then crawling back to get points with missing blocks, so we might as well collect the non-validated points at the same time. I'll do this.

etorreborre: I pushed a consolidated version.

crates/amaru/src/stages/build_node.rs
etorreborre: 👍 see the new version and its comments

crates/amaru/src/stages/build_node.rs
etorreborre: This has been reworked now.

crates/amaru/src/stages/build_node.rs
etorreborre: Done.

crates/amaru-consensus/src/stages/validate_block/mod.rs
etorreborre: Right, what I wanted to do was both wrong and unnecessary. Moving back.

etorreborre:
@rkuhn I've now addressed your comments and rebased on main:

- You can check in particular: #8a025ff where I simplified the code for fetching the best candidate on startup.
- There are definitely still some improvements to be made on the store, like:
   - Storing tips directly.
   - Have uniform results, where each store function returns `Result<Something, StoreError>`.

rkuhn:
One remark up front: for this complex PR with long commit history — and in particular when referencing commit hashes from comments — it is much preferred to merge `main` into the branch instead of rebasing. Do I assume correction that [dde0611](https://github.com/pragma-org/amaru/pull/747/commits/dde06110cddab38bac43f43eec4f7050fc29d182) is the commit you intended to point out?

etorreborre:
@rkuhn yes, that's the one. Note that if we merge https://github.com/pragma-org/amaru/pull/726, we can use #726 code instead

etorreborre:
@rkuhn I made the changes we discussed, including some test support to unit test functions that are used inside `pure-stage` without having to test the full stage.

crates/pure-stage/src/simulation/simulation_builder.rs
```diff
@@ -378,3 +379,58 @@ fn deliver_message(
...
+
+/// Run a function using store effects only inside a stage and get back the result as the stage state.
+pub fn run_function_with_resource<S, R, F, Fut>(r: R, function: F) -> S
+where
+    F: FnMut(Effects<()>) -> Fut + 'static + Send,
+    Fut: Future<Output = S> + 'static + Send,
+    S: SendData + PartialEq + serde::de::DeserializeOwned + serde::Serialize + 'static + Clone,
+    R: Send + Sync + 'static + Clone,
+{
+    run_function(|resources| resources.put::<R>(r.clone()), function)
```
rkuhn: This should be just `resources.put(r)`, which will require `run_tests` to take `FS: FnOnce(&Resources)`.

crates/amaru-protocols/src/blockfetch/responder.rs
```diff
@@ -529,20 +528,24 @@ pub mod tests {
...
-
-        let stage = network.stage("points_range_test", points_range_test_stage);
-        let stage = network.wire_up(stage, PointsRangeTestState::default());
-        network.preload(&stage, [msg]).unwrap();
-
-        let mut running = network.run();
-        running.run_until_blocked_incl_effects(runtime.handle()).assert_idle();
-
-        running.get_state(&stage).unwrap().result.clone().unwrap()
+        let store: Arc<dyn ChainStore<BlockHeader>> = store.clone();
```
rkuhn: This line should be removed: the `Arc<InMemStore>` can be coerced into an `Arc<dyn ChainStore>` without further gymnastics.

crates/amaru-consensus/src/stages/select_chain/tests.rs
```diff
@@ -557,3 +577,109 @@ fn test_startup_with_non_empty_store() {
...
+        let store: Arc<dyn ChainStore<BlockHeader>> = Arc::new(InMemConsensusStore::default());
+        for header in [&a, &b, &c, &d] {
+            store.store_header(header).unwrap();
+        }
+        store.set_anchor_hash(&a.hash()).unwrap();
+        for header in [&a, &b] {
+            store.roll_forward_chain(&header.point()).unwrap();
+        }
+        store.set_block_valid(&b.hash(), true).unwrap();
+        let d_hash = d.hash();
```
rkuhn, line 633: why for d_hash but not for c_hash? (I’d just remove this)

crates/amaru-consensus/src/stages/select_chain/mod.rs
```diff
@@ -240,6 +222,91 @@ impl SelectChain {
...
+///                     \
+///                      C(?)--D(?)         candidate 1
+///                      \
+///                       E(?)--F(invalid)--G(?)  candidate 2, rejected
+///                       \
+///                       H(?)                candidate 3
+///
+/// The candidates are the reachable leaves below `best_chain_hash`: `D`, `G`, and `H`.
+/// The branch ending at `G` is rejected because it contains `F(invalid)`.
+/// Between `D` and `H`, the highest tip wins according to `cmp_tip`.
```
rkuhn: E is also an option

crates/amaru-consensus/src/stages/select_chain/mod.rs
```diff
@@ -240,6 +222,91 @@ impl SelectChain {
...
+/// Return the highest tip candidate that extends the current best chain, together with
+/// the unknown suffix after the last validated ancestor.
+///
+/// Example:
+///
+/// O--A(valid)--B(valid)  best_chain_hash
+///                     \
+///                      C(?)--D(?)         candidate 1
+///                      \
+///                       E(?)--F(invalid)--G(?)  candidate 2, rejected
```
rkuhn: F cannot be invalid while E is unknown

crates/amaru-consensus/src/stages/select_chain/mod.rs
```diff
@@ -240,6 +222,91 @@ impl SelectChain {
...
+///
+/// The candidates are the reachable leaves below `best_chain_hash`: `D`, `G`, and `H`.
+/// The branch ending at `G` is rejected because it contains `F(invalid)`.
+/// Between `D` and `H`, the highest tip wins according to `cmp_tip`.
+///
+/// If `D` wins, this returns:
+/// `(D, [C, D])` since `C` and `D` have not-yet-validated blocks.
+pub async fn best_tip_from_store(store: &Store) -> anyhow::Result<Option<(BlockHeader, Vec<HeaderHash>)>> {
+    let best_chain_hash = store.get_best_chain_hash().await;
+    if best_chain_hash != ORIGIN_HASH && store.load_header_with_validity(&best_chain_hash).await.is_none() {
```
rkuhn: should bind that loaded header to a variable because it is needed below

crates/amaru-consensus/src/stages/select_chain/mod.rs
```diff
@@ -240,6 +222,91 @@ impl SelectChain {
...
+        // Don't use this header if its block is invalid
+        if validity == Some(false) {
+            continue;
+        };
+
+        // Continue the traversal with all direct descendants of the current header.
+        let children = store.get_children(&hash).await;
+
+        if children.is_empty()
+            && best_candidate.as_ref().is_none_or(|current| cmp_tip(Some(&header), Some(current)).is_gt())
```
rkuhn: simply use `cmp_tip(Some(&header), best_candidate.as_ref()).is_gt()`

crates/amaru-consensus/src/stages/select_chain/mod.rs
```diff
@@ -240,6 +222,91 @@ impl SelectChain {
...
+        {
+            best_candidate = Some(header);
+        } else {
+            // Propagate the updated suffix state to each child branch.
+            to_visit.extend(children);
+        }
+    }
+
+    if let Some(best_candidate) = best_candidate {
+        let (best_missing, is_valid) = store.unvalidated_ancestor_hashes(best_candidate.hash()).await;
```
rkuhn: `is_valid` cannot be `false` here, that would be an internal invariant violation

crates/amaru-consensus/src/stages/select_chain/mod.rs
```diff
@@ -240,6 +222,91 @@ impl SelectChain {
...
+    }
+
+    let mut best_candidate = None;
+
+    // ORIGIN_HASH cannot have a block, so we start from its direct children
+    let mut to_visit = if best_chain_hash == ORIGIN_HASH {
+        store.get_children(&best_chain_hash).await.into_iter().collect()
+    } else {
+        // If the best chain hash is not origin, we have a first best candidate.
+        best_candidate = store.load_header(&best_chain_hash).await;
```
rkuhn: In order to find the best candidate from scratch, this will need to start at the anchor, not at the best hash: there might be a better branch that we haven’t been able to validate yet.

rkuhn (commented):
going in the right direction, but I found some more issues

crates/amaru-consensus/src/stages/select_chain/mod.rs
etorreborre: I made a better diagram, I had the wrong mental model.

crates/amaru-consensus/src/stages/select_chain/tests.rs
etorreborre: That's because `d` is used twice on line 632.

crates/amaru-protocols/src/blockfetch/responder.rs
etorreborre: Indeed.

crates/pure-stage/src/simulation/simulation_builder.rs
etorreborre: Done.

crates/amaru-consensus/src/stages/select_chain/mod.rs
etorreborre: Oh, yes, right, updated

crates/amaru-consensus/src/stages/select_chain/mod.rs
etorreborre: Done.

crates/amaru-consensus/src/stages/select_chain/mod.rs
etorreborre: Much better

crates/amaru-consensus/src/stages/select_chain/mod.rs
etorreborre: Yes.

crates/amaru-consensus/src/stages/select_chain/mod.rs
etorreborre: Done.

etorreborre:
@rkuhn I made all the changes in the last commit, thanks for your patience!

crates/amaru-protocols/src/store_effects.rs
```diff
@@ -53,7 +53,7 @@ impl Store {
         self.effects.external(GetAnchorHashEffect::new())
     }
 
-    pub fn get_best_chain_hash(&self) -> BoxFuture<'static, HeaderHash> {
+    pub fn get_anchor(&self) -> BoxFuture<'static, HeaderHash> {
         self.effects.external(GetBestChainHashEffect::new())
```
rkuhn, line 57: ouch, this is quite an Easter egg — how can we make sure that such a change doesn’t slip through while I’m out of capacity for such reviews?

crates/amaru-consensus/src/stages/select_chain/mod.rs
```diff
@@ -222,40 +222,40 @@ impl SelectChain {
...
 /// Example:
 ///
-/// O--A(valid)--B(valid)  best_chain_hash
+/// O--A(valid)--B(valid)  anchor_hash
 ///                     \
-///                      C(?)--D(?)         candidate 1
+///                      C(valid)--D(?)--E(?) candidate 1
 ///                      \
-///                       E(?)--F(invalid)--G(?)  candidate 2, rejected
+///                       H(valid)--I(valid)--J(invalid)  candidate 2, rejected
```
rkuhn, line 234: please add a test where `I` is the best tip to show that the algorithm also works in this case (I don’t think it will due to the `children.is_empty()` prerequisite for emitting)

crates/amaru-consensus/src/stages/select_chain/tests.rs
```diff
@@ -389,7 +389,6 @@ fn test_block_validation_result_invalid_best_tip_invalidated() {
             te_has_header("sc-1", tip.hash()),
             te_set_block_valid("sc-1", tip.hash(), false),
             te_get_best_chain_hash("sc-1"),
```
rkuhn: This must be “get anchor” and not “get best chain hash”.

rkuhn (changes requested):

crates/amaru-protocols/src/store_effects.rs
etorreborre: Whaaaat? I don't understand how that could happen.

crates/amaru-consensus/src/stages/select_chain/tests.rs
etorreborre: Fixed

crates/amaru-consensus/src/stages/select_chain/mod.rs
etorreborre: I think it's covered by `test_block_validation_result_invalid_best_tip_invalidated` (and you will notice that I removed the check on `children.is_empty`.

crates/amaru-protocols/src/store_effects.rs
etorreborre: See the forced commit.

etorreborre:
@rkuhn In my latest commit I generalized the parent-child relationship so that we can query children with the `ORIGIN_HASH`. This is helpful for the tests.

rkuhn (approved):
Yes, you’re right, the test should cover this case.

PR #750, opened 2026-04-10
Fix Metadata Predicates

yHSJ:
This PR implements the  `validateScriptsWellFormed` predicate and adds the script validation logic that exists in `validateMetadata`. 

The `valdiateMetadata` conformance test is expected to fail, though it is now failing with a decoding error. This is actually the expected state, but the test harness expects the transactions to decode successfully. This would be a more significant change that falls outside the scope of this PR.

## Release Notes

* **New Features**
  - Added comprehensive Plutus script validation for witness scripts, reference input scripts, and auxiliary data scripts.

* **Bug Fixes**
  - Metadata text and byte string fields now validated to not exceed 64 bytes.
  - Improved protocol-version-aware script decoding with stricter validation for newer protocol versions.

* **Chores**
  - Updated dependencies.


crates/amaru-ledger/src/rules/transaction/phase_one/mod.rs
```diff
@@ -103,7 +103,7 @@ where
 {
     let transaction_id = transaction_body.id();
 
-    metadata::execute(&transaction_body, transaction_auxiliary_data)?;
+    metadata::execute(&transaction_body, transaction_auxiliary_data, protocol_parameters.protocol_version.major())?;
```
KtorZ: Would be more readable to pass the protocol version here, we don't gain much by passing a naked u32 but we loose context.

crates/amaru-ledger/src/rules/transaction/phase_one/metadata.rs
```diff
@@ -28,25 +31,37 @@ pub enum InvalidTransactionMetadata {
...
-                // the validateMetadata logic is not implemented here (unlike the Haskell code), but instead during deserialization (TODO)
+                data.plutus_v1_scripts()
+                    .iter()
+                    .try_for_each(|s| validate_plutus_script(s, PlutusVersion::V1, protocol_version_major))?;
+                data.plutus_v2_scripts()
+                    .iter()
+                    .try_for_each(|s| validate_plutus_script(s, PlutusVersion::V2, protocol_version_major))?;
+                data.plutus_v3_scripts()
+                    .iter()
+                    .try_for_each(|s| validate_plutus_script(s, PlutusVersion::V3, protocol_version_major))?;
```
KtorZ: Seems like a common behavior to attach to `data` here.

crates/amaru-ledger/src/context/assert.rs
```diff
@@ -132,8 +135,13 @@ impl UtxoSlice for AssertValidationContext {
     }
 
     fn produce(&mut self, input: TransactionInput, output: MemoizedTransactionOutput) {
+        self.produced.insert(input.clone());
         self.utxo.insert(input, output);
     }
+
+    fn produced_inputs(&self) -> Vec<&TransactionInput> {
```
KtorZ: That's not right. We precisely want to avoid this kind of thing in the context and during rules execution. This only exists to be called in a single place followed by `context.lookup` to extract scripts from UTxOs? 

The scripts to validate should rather be collected while validating inputs, stashed for later.

crates/amaru-ledger/src/rules/transaction/phase_one/scripts.rs
```diff
@@ -323,14 +328,116 @@ fn fail_on_missing_datums(missing: BTreeSet<u32>) -> Result<(), InvalidScripts>
     Ok(())
 }
 
+pub(crate) fn validate_plutus_script<const V: usize>(
+    script: &PlutusScript<V>,
+    plutus_version: PlutusVersion,
+    protocol_version_major: u32,
+) -> Result<(), FlatDecodeError> {
+    let flat_bytes = script.flat_bytes().map_err(|e| FlatDecodeError::Message(format!("{e}")))?;
+    let arena = Arena::new();
```
KtorZ: x_x

crates/amaru-ledger/src/rules/transaction/phase_one/scripts.rs
```diff
@@ -323,14 +328,116 @@ fn fail_on_missing_datums(missing: BTreeSet<u32>) -> Result<(), InvalidScripts>
...
+        );
+    }
+    if let Some(scripts) = witness_set.plutus_v3_script.as_ref() {
+        malformed.extend(
+            scripts
+                .iter()
+                .filter(|s| validate_plutus_script(s, PlutusVersion::V3, protocol_version_major).is_err())
+                .map(|s| s.script_hash()),
+        );
+    }
```
KtorZ: That's the same code repeated over, which indicates it would be better captured as a shared function.

crates/amaru-ledger/src/rules/transaction/phase_one/scripts.rs
yHSJ: I know I know, I agree. I just haven't figured out the right way to do this, do we want to use the same arena pool we use for validation?

crates/amaru-ledger/src/rules/transaction/phase_one/metadata.rs
yHSJ: Yeah, but `AuxiliaryData` lives in the kernel, which doesn't depend on `uplc-turbo` right now. We could add an extension, but really I'm not sure it is that valuable here. I can extract this to a separate function though, just to make it easier to reason about the core `execute` function

crates/amaru-ledger/src/rules/transaction/phase_one/scripts.rs
KtorZ: Well, yes. And there's no point even flat-decoding the scripts twice. So we should find ways to connect this back to phase-2 validations. If scripts have already been decoded during phase-1 validations, then they should be ready to apply in phase2. 

My only MAJOR concern here is that this gets shoved in here without any bells and whistle when it should have a big fat;

```rs
// FIXME: expensive Arena creation in the middle of validation function
// The allocation arenas should never be created on the fly but created ahead of time and re-used. 
// They loose otherwise all their benefits. 
```

That we don't address the problem _now_ isn't a big deal; that we silently introduce a potentially big issue to the code without proper warnings for us to find it back later, that's bad!

crates/amaru-ledger/src/rules/transaction/phase_one/scripts.rs
```diff
@@ -323,14 +333,111 @@ fn fail_on_missing_datums(missing: BTreeSet<u32>) -> Result<(), InvalidScripts>
...
+}
+
+pub(crate) fn validate_auxiliary_data_scripts(
+    data: &AuxiliaryData,
+    protocol_version: ProtocolVersion,
+) -> Result<(), FlatDecodeError> {
+    data.plutus_v1_scripts().iter().try_for_each(|s| validate_plutus_script(s, PlutusVersion::V1, protocol_version))?;
+    data.plutus_v2_scripts().iter().try_for_each(|s| validate_plutus_script(s, PlutusVersion::V2, protocol_version))?;
+    data.plutus_v3_scripts().iter().try_for_each(|s| validate_plutus_script(s, PlutusVersion::V3, protocol_version))?;
+    Ok(())
```
KtorZ: Note that at the *very least*, the arena could be created here before all those `.iter()` and re-used for all those intermediate validations. That makes the problem already WAY worse.

crates/amaru-kernel/src/cardano/protocol_version.rs
```diff
@@ -17,3 +17,14 @@ pub use pallas_primitives::conway::ProtocolVersion;
...
+pub trait ProtocolVersionExt {
+    /// Returns the major protocol version as a u32.
+    fn major(&self) -> u32;
+}
+
+impl ProtocolVersionExt for ProtocolVersion {
+    fn major(&self) -> u32 {
+        self.0 as u32
+    }
+}
```
KtorZ, line 27: Seems a bit overkill to access the first element of a tuple?

crates/amaru-ledger/src/rules/transaction/phase_one/outputs.rs
```diff
@@ -76,6 +76,11 @@ where
             context.allow_supplemental_datum(*hash);
         }
 
+        // TODO: avoid cloning here; requires changing produce() to not take ownership
+        if let Some(ref script) = output.script {
+            context.produce_script(script.clone());
+        }
```
KtorZ: That's going to be tricky here. Given that we later give ownership of the output to the context too when producing. 

We could/should use a similar trick to what was done for `known_scripts`; and store the location of where to find the script in the context.  I am actually a bit puzzled as for why we even need to store produced scripts now when we already have `known_scripts`?

crates/amaru-ledger/src/rules/transaction/phase_one/scripts.rs
```diff
@@ -69,13 +71,23 @@ pub enum InvalidScripts {
...
-pub fn execute<C>(context: &mut C, witness_set: &WitnessSet) -> Result<(), InvalidScripts>
+pub fn execute<C>(
+    context: &mut C,
+    witness_set: &WitnessSet,
+    protocol_version: ProtocolVersion,
+) -> Result<(), InvalidScripts>
 where
     C: UtxoSlice + WitnessSlice + fmt::Debug,
 {
+    fail_on_malformed_scripts(witness_set, protocol_version, context.produced_scripts())?;
```
KtorZ: Seems like this could be actually done _earlier_ so that we don't even end up registering as known scripts those that are invalid? Because it kind of feels like double work here: we have to re-traverse all the scripts that we have already traversed once when we could simply do it once or not at all should one be invalid (thus saving precious computing resources!).

crates/amaru-ledger/src/rules/transaction/phase_one/scripts.rs
```diff
@@ -323,14 +333,111 @@ fn fail_on_missing_datums(missing: BTreeSet<u32>) -> Result<(), InvalidScripts>
...
+fn collect_malformed<const V: usize>(
+    scripts: &[PlutusScript<V>],
+    plutus_version: PlutusVersion,
+    protocol_version: ProtocolVersion,
+    malformed: &mut BTreeSet<Hash<SCRIPT>>,
+) {
+    malformed.extend(
+        scripts
+            .iter()
+            .filter(|s| validate_plutus_script(s, plutus_version, protocol_version).is_err())
```
KtorZ: That doesn't sound right. Now we re-validate AGAIN all the scripts to figure out which one failed? Why not return the culprit during the first validation pass already?

crates/amaru-ledger/src/rules/transaction/phase_one/scripts.rs
```diff
@@ -323,14 +333,111 @@ fn fail_on_missing_datums(missing: BTreeSet<u32>) -> Result<(), InvalidScripts>
...
+        .filter_map(|script| {
+            let result = match script {
+                MemoizedScript::PlutusV1Script(s) => validate_plutus_script(s, PlutusVersion::V1, protocol_version),
+                MemoizedScript::PlutusV2Script(s) => validate_plutus_script(s, PlutusVersion::V2, protocol_version),
+                MemoizedScript::PlutusV3Script(s) => validate_plutus_script(s, PlutusVersion::V3, protocol_version),
+                MemoizedScript::NativeScript(_) => return None,
+            };
+            result.err().map(|_| script.script_hash())
+        })
+        .collect()
```
KtorZ: Same remark here; all these are duplication of the same bits of logic which should all come from the same source actually: we should not register invalid scripts to the context to begin with and bail right away with an appropriate error.

crates/amaru-kernel/src/cardano/protocol_version.rs
yHSJ: I thought it was nice to be explicit about what is being access, but yes, it is definitely overkill

crates/amaru-ledger/src/rules/transaction/phase_one/outputs.rs
yHSJ: > I am actually a bit puzzled as for why we even need to store produced scripts now when we already have known_scripts?

They're disjoint sets. `produced_scripts` are all the scripts which are newly created in outputs of this transaction, and we need to validate they are well-formed bytes. `known_scripts` is used to validate needed scripts are provided, only containing scripts in the inputs/reference inputs.

> That's going to be tricky here. Given that we later give ownership of the output to the context too when producing.

We could/should use a similar trick to what was done for known_scripts; and store the location of where to find the script in the context.

This could work, given that we have the `produced` outputs in the context. That probably does make more sense.

crates/amaru-ledger/src/rules/transaction/phase_one/scripts.rs
yHSJ: I think this probably does make the most sense. I tried to keep the "script related logic" all in one place, but perhaps this is actually "output related logic", since we need to validate the structure of a piece of an output here. That would also remove the need to have produced_outputs at all.

crates/amaru-ledger/src/rules/transaction/phase_one/scripts.rs
yHSJ: We haven't revalidated more than once here, but I've restructured the logic to make more sense.

Now, while we collect the plutus scripts in the witness set, we validate that they are wellformed. If there are any malformed scripts, we return an error, otherwise we return the `BTreeMap<Hash, ScriptKind>`. That removes the extra iteration cycle.

crates/amaru-ledger/src/rules/transaction/phase_one/scripts.rs
yHSJ: Though this is no longer relevant, as I've refactored this logic to avoid an unnecessary iteration while hopefully making it more explicit what is happening, I'll provide some clarity here.

`known_scripts` is populated only from the inputs and reference inputs. If they contain a script, we add them to our context inside of `inputs.rs`. Those scripts have necessarily already being validated, as we now have validation in `output.rs` which validates the well-formedness for newly created reference scripts.

However, any scripts in the witness set could be malformed, since they have not been validated yet. That validation is what was happening here. But, you're right that in the case of no malformed scripts (which is the happy path), we are doing two iterations: one to find malformed scripts and another to create the script map we use.

I've since refactored this code so that we find malformed scripts and create the script map in the same iteration. So that should be more clear now.

KtorZ (approved):
OK

PR #755, opened 2026-04-15
Ledger Predicate Fixes

yHSJ:
This PR introduces some fixes for some general "low hanging fruit" in terms of fixing/implementing some of the ledger predicates. While this PR groups many different predicates in a single PR, I tried to make it easy to isolate the predicates by commit, which may help with reviews.

Notes:
1) Many of these predicates are missing coverage in both the unit tests, and in the conformance tests.
2) There is an important divergence from the Haskell logic in the Byron address attributes predicate right now. This is due to the underlying Pallas type.

## Summary by CodeRabbit

* **New Features**
  * Track per-transaction byte sizes; enforce max transaction size and use sizes during validation
  * Block-level validations: body-hash and header protocol-version checks
  * Stronger network ID checks across transactions, withdrawals, pool registrations, proposals and addresses
  * Pool retirement, cost and deposit validations; stricter proposal and governance checks
  * Enforce execution-unit limits for scripts; bootstrap address-attribute size check

* **Refactor**
  * Transaction iteration and phase-one validation wired to pass transaction sizes and network info
  * Public reward-account parsing now also returns the parsed network
  * New CBOR decoding helpers exposed for item/value byte-size reporting

* **Tests**
  * Updated conformance expectations and unit tests to reflect new validations


crates/amaru-ledger/src/rules/transaction/phase_one/certificates.rs
```diff
@@ -121,13 +142,44 @@ where
...
+            let current_epoch = era_history.slot_to_epoch(pointer.slot(), pointer.slot())?;
+            let retirement_epoch = Epoch::from(epoch);
+            let max_epoch = current_epoch + protocol_parameters.stake_pool_max_retirement_epoch;
+            if retirement_epoch <= current_epoch || retirement_epoch > max_epoch {
+                return Err(InvalidCertificates::PoolRetirementWrongEpoch {
+                    epoch: retirement_epoch,
+                    current_epoch,
+                    max_epoch,
+                });
+            }
```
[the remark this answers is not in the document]
yHSJ: The note is correct. I actually think you're correct about using `slot_to_epoch_unchecked_horizon` in this case, but the comment is correct and worth including.

crates/amaru-ledger/src/rules/transaction/phase_one/proposals.rs
```diff
@@ -40,6 +77,156 @@ pub(crate) fn execute<C>(
...
+    reject_zero(ppu.collateral_percentage, "collateral_percentage".into())?;
+    reject_zero(ppu.committee_term_limit, "committee_term_limit".into())?;
+    reject_zero(ppu.governance_action_validity_period, "governance_action_validity_period".into())?;
+    reject_zero(ppu.pool_deposit, "pool_deposit".into())?;
+    reject_zero(ppu.governance_action_deposit, "governance_action_deposit".into())?;
+    reject_zero(ppu.drep_deposit, "drep_deposit".into())?;
+
+    if pv.0 != 9 {
+        reject_zero(ppu.ada_per_utxo_byte, "ada_per_utxo_byte".into())?;
+    }
```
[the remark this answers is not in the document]
yHSJ: This is wrong. They're all `u64`s.

crates/amaru-ledger/src/rules/transaction/phase_one/proposals.rs
```diff
@@ -40,6 +84,167 @@ pub(crate) fn execute<C>(
...
+            }
+            let added_keys: std::collections::BTreeSet<_> = added.iter().map(|(k, _)| k).collect();
+            let removed_keys: std::collections::BTreeSet<_> = removed.iter().collect();
+            if !added_keys.is_disjoint(&removed_keys) {
+                return Err(InvalidProposals::ConflictingCommitteeUpdate);
+            }
+
+            // NOTE: conformance tests are brittle on this check due to era_history mismatch.
+            // (see certificates.rs PoolRetirement comment for details)
+            let current = era_history.slot_to_epoch(pointer.slot, pointer.slot)?;
```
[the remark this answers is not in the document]
yHSJ: We are operating at the tip here, so it's the same thing.

crates/amaru-ledger/src/rules/block/body_hash.rs
```diff
@@ -0,0 +1,45 @@
...
+
+    use crate::rules::block::InvalidBlockDetails;
+
+    macro_rules! fixture {
+        ($number:literal) => {
+            include_cbor!(concat!("blocks/preprod/", $number, "/valid.cbor"))
+        };
+    }
+
+    #[test_case(fixture!("2667657"); "valid")]
```
KtorZ, line 41: We're missing at least one scenario where the hash is invalid.

crates/amaru-minicbor-extra/src/decode.rs
```diff
@@ -88,6 +88,81 @@ pub fn heterogeneous_array<'d, A>(
...
+                items.push(item);
+            }
+        }
+        None => {
+            while d.datatype()? != Type::Break {
+                let (item, bytes) = tee(d, |d| T::decode(d, ctx))?;
+                sizes.push(bytes.len() as u64);
+                items.push(item);
+            }
+            d.skip()?;
```
KtorZ: This is wrong in two ways:

1. It re-implements and duplicate existing logic (from heterogenous_array), now creating twice the opportunity for bugs. 
2. It re-implements it _wrong_ and forces the allocation of an intermediate `Vec<T>` instead of giving control to the caller over the decoded elements. 

In this kind of situation, it is better to amend and generalize the existing functions (`heterogenous_array`) to keep the critical part in one place, but allow different behavior on top with specialized helpers. 

In this particular occurrence, I don't think it's even needed because the place where it's called shouldn't need it. I explain:

- First, the decoding of block parts have been amended to `let ((transaction_bodies, transaction_body_sizes), transaction_bodies_bytes) = cbor::tee(d, |d| cbor::decode_array_with_item_sizes::<TransactionBody, _>(d, ctx))?;`; so we now collect sizes alongside bodies. This is also done for witnesses and aux data. 

- Then, everything is zipped together, with some padding:
   ```
    // Pre-compute the per-transaction size matching Haskell's logic.
    // Importantly, the serialization the Haskell node does is different from the on-chain serialization.
    // The logic below is from: https://github.com/IntersectMBO/cardano-ledger/blob/0cfbf861cfb456660a7b73281c6fb714a53d40f9/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Tx.hs#L351-L362
    //
    //   encodeListLen 3
    //     <> encCBOR body
    //     <> encCBOR witnesses
    //     <> encodeNullStrictMaybe encCBOR aux
    //
    // Notably, the array length is 3, NOT 4, and the isValid flag is intentionally
    // excluded for backward compatibility with pre-Alonzo eras. Aux data is encoded as
    // CBOR `null` (0xf6, 1 byte) when absent.
    const TX_FRAMING_BYTES: u64 = 1; // array(3) header = 0x83
    const NULL_AUX_DATA_BYTES: u64 = 1; // CBOR null = 0xf6
    let transaction_sizes: Vec<u64> = transaction_body_sizes
        .iter()
        .zip(transaction_witness_sizes.iter())
        .enumerate()
        .map(|(i, (body_size, witness_size))| {
            let aux_size = auxiliary_data_sizes.get(&(i as u32)).copied().unwrap_or(NULL_AUX_DATA_BYTES);
            TX_FRAMING_BYTES + body_size + witness_size + aux_size
        })
        .collect();
   ```

- This is then stashed away as a `transaction_sizes` field on blocks, and later accessed when iterating over transactions: `let tx_size = transaction_sizes.get(i as usize).copied().unwrap_or(0);`; 

- The size is added to the iterator item as an extra `u64` 

So we've spread the size a bit everywhere and added some extra minicbor helpers when all the pieces were *already* there in the iterator. Instead of all those intermediate values, I suggest we defined a `SizedTransaction` type to be returned by the iterator, that contains a transaction and size. The size can be computed on-the-fly in the iterator from storing the `original_size` of each sub-element and adding the necessary padding. 

This is better in few ways:

- Less code spread over modules
- Makes the iterator type more explicit, rather than a now-cryptic triple (u32, Transaction, u64)
- Avoid polluting the block decoder context with transaction-specific elements
- Keep the size-related matters near-located

crates/amaru-kernel/src/cardano/witness_set.rs
```diff
@@ -65,7 +69,62 @@ pub struct WitnessSet {
...
+                    0 => ws.vkeywitness = d.decode_with(ctx)?,
+                    1 => ws.native_script = d.decode_with(ctx)?,
+                    2 => ws.bootstrap_witness = d.decode_with(ctx)?,
+                    3 => ws.plutus_v1_script = d.decode_with(ctx)?,
+                    4 => ws.plutus_data = d.decode_with(ctx)?,
+                    5 => ws.redeemer = d.decode_with(ctx)?,
+                    6 => ws.plutus_v2_script = d.decode_with(ctx)?,
+                    7 => ws.plutus_v3_script = d.decode_with(ctx)?,
+                    _ => {
+                        d.skip()?;
```
KtorZ: Wow! No. Never. Forbidden. 😬

KtorZ: Do not silently ignore errors.

crates/amaru-ledger/src/rules/transaction/phase_one/outputs.rs
```diff
@@ -88,6 +94,44 @@ where
     Ok(())
 }
 
+fn validate_bootstrap_attributes(output: &MemoizedTransactionOutput) -> Result<(), InvalidOutput> {
+    if let Address::Byron(addr) = &output.address {
+        // This logic assumes the address (and thus the payload) has already been checked
+        let Some(payload) = from_cbor::<AddressPayload>(&addr.payload.0) else {
+            return Ok(());
```
KtorZ, line 116: I don't think that's the right behavior here. The comment is right, therefore, if we ever get a `None` here, that's an invariant violation and we want the whole world to know about it. So this `return Ok(())` should actually be 

```rs
unreachable!("failed to decode address payload during bootstrap attribute validation: {addr:?}");
```

**Do not silently ignore errors.**

crates/amaru-ledger/src/rules/transaction/phase_one/proposals.rs
```diff
@@ -40,6 +84,167 @@ pub(crate) fn execute<C>(
...
+                    Ok(Address::Stake(addr)) => {
+                        let actual = addr.network();
+                        if actual != network {
+                            return Err(InvalidProposals::TreasuryWithdrawalWrongNetwork { expected: network, actual });
+                        }
+                    }
+                    _ => return Err(InvalidProposals::MalformedReturnAddress),
+                }
+            }
+            if !wdrls.iter().any(|(_, coin)| *coin > 0) {
```
KtorZ, line 133: .all?

crates/amaru-ledger/src/rules/transaction/phase_one/proposals.rs
```diff
@@ -40,6 +84,167 @@ pub(crate) fn execute<C>(
...
+    }
+
+    Ok(())
+}
+
+fn pv_can_follow(current: ProtocolVersion, new: ProtocolVersion) -> bool {
+    let (cur_major, cur_minor) = current;
+    let (new_major, new_minor) = new;
+    (new_major == cur_major + 1 && new_minor == 0) || (new_major == cur_major && new_minor == cur_minor + 1)
+}
```
KtorZ, line 187: This one is worth unit testing and illustrate with various cases. I've always found it confusing as hell.

crates/amaru-ledger/src/rules/transaction/phase_one/proposals.rs
```diff
@@ -40,6 +84,167 @@ pub(crate) fn execute<C>(
...
+        && ppu.max_collateral_inputs.is_none()
+        && ppu.pool_voting_thresholds.is_none()
+        && ppu.drep_voting_thresholds.is_none()
+        && ppu.min_committee_size.is_none()
+        && ppu.committee_term_limit.is_none()
+        && ppu.governance_action_validity_period.is_none()
+        && ppu.governance_action_deposit.is_none()
+        && ppu.drep_deposit.is_none()
+        && ppu.drep_inactivity_period.is_none()
+        && ppu.minfee_refscript_cost_per_byte.is_none();
```
KtorZ, line 241: Sounds risky and not maintainable over time. If new fields are added to the protocol parameters, we get no warnings from the compiler. 

I'd suggest instead: 

- Define a `default` instance for the `ProtocolParamUpdate` (or an equivalent function if we don't own the type _yet_)
- Compare with the default. 

This way, we are forced to construct a complete type always, and the compiler can tell us where to add code on changes.

crates/amaru-ledger/src/rules/transaction/phase_one/withdrawals.rs
```diff
@@ -42,15 +47,23 @@ where
...
+                if network != &account_network {
+                    Err(InvalidWithdrawals::NetworkMismatch {
+                        expected: *network,
+                        received: account_network,
+                        context: TransactionField::Withdrawals,
+                        position,
+                    })
+                } else {
+                    Ok((credential, st))
+                }
```
KtorZ:
```suggestion
                if network != &account_network {
                    return Err(InvalidWithdrawals::NetworkMismatch {
                        expected: *network,
                        received: account_network,
                        context: TransactionField::Withdrawals,
                        position,
                    })
                } 
                
                Ok((credential, st))
```

Seems like an aesthetic suggestion but it's not. Prefer this style always (assert errors first + early return) because these are more maintainable. `else` branch like this are particularly dangerous over time, as things get added, they're easy to miss and it might cause intricate logic bugs.

crates/amaru-ledger/src/rules/transaction/phase_one/outputs.rs
yHSJ: I got lazy here, and forgot to go back and update this. Good catch

crates/amaru-ledger/src/rules/transaction/phase_one/proposals.rs
yHSJ: My original plan was a default, but I decided a default `ProtocolParamUpdate` that was empty/invalid felt strange. However, I guess empty defaults are the convetion (like the default `Vec`).

KtorZ:
(@yHSJ: I am currently rebasing / reviewing / wrapping up this one)

KtorZ (approved):

PR #764, opened 2026-04-20
fix: Redeemer map behavior to match Haskell impl

jonathanlim222:
Added remove before insert on script_table in all 6 branches of `ScriptPurpose::builder`. `BTreeMap::insert` keeps the existing key when an Ord-equal key is inserted, only replacing the value. This meant duplicate redeemers (same tag+index, different ExUnits) would silently retain the first redeemer's ExUnits — diverging from Haskell's Map.fromList which keeps the last duplicate. The remove-before-insert pattern ensures the new key (with correct ExUnits) fully replaces the old one.

## Summary by CodeRabbit

* **New Features**
  * Redeemer handling and script-context construction now use a keyed redeemers map for consistent script resolution and serialization.

* **Bug Fixes**
  * Duplicate redeemer entries deterministically keep the last occurrence; Plutus serialization and script mapping are more reliable.

* **Tests**
  * Added/updated tests validating redeemer key semantics and deterministic script-context outputs.

* **Refactor**
  * Removed legacy ordered-redeemer public type and adjusted redeemer-to-Plutus serialization and public exports.

[Review Change Stack](https://app.coderabbit.ai/change-stack/pragma-org/amaru/pull/764?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


jonathanlim222:
@coderabbitai review

crates/amaru-plutus/src/script_context/mod.rs
```diff
@@ -388,6 +389,7 @@ impl<'a> ScriptPurpose<'a> {
             RedeemerTag::Mint => mint.0.keys().nth(index).copied().and_then(|policy_id| {
                 let script = scripts.get(&policy_id);
                 script.map(|script| {
+                    script_table.remove(redeemer);
                     script_table.insert(redeemer.clone(), script.clone());
```
KtorZ: I'd be better to be explicit about the behavior here and limit the unecessary work as well by using:

- .entry()
- .and_modify()
- .or_insert

KtorZ (dismissed):
Please leverage -> https://doc.rust-lang.org/std/collections/btree_map/enum.Entry.html

crates/amaru-kernel/src/cardano/ordered_redeemer.rs
```diff
@@ -61,3 +61,113 @@ impl<'a> From<&'a Redeemer> for OrderedRedeemer<'a> {
...
+
+pub struct UpsertMap<K: Ord, V>(BTreeMap<K, V>);
+
+impl<K: Ord, V> UpsertMap<K, V> {
+    pub fn new() -> Self {
+        Self(BTreeMap::new())
+    }
+
+    pub fn upsert(&mut self, key: K, value: V) {
+        self.0.remove(&key);
```
yHSJ: @KtorZ's feedback had two main points:
1. explicitness
2. efficiency 

This change only addressed the first point. The second point has yet to be addressed. I would argue that's acceptable for now, but it's worth noting (in the PR, in a comment, both?) that hasn't been addressed.

crates/amaru-plutus/src/script_context/mod.rs
yHSJ: I went ahead and restructured our `ScriptContext` in a way that I think makes more sense. There is no need to be bound to the redeemers shape that we had previously. So the updated logic now has fewer clones and iterations in general, while still being explicit.

KtorZ (commented):
Could we introduce at least one extra test that is showcasing the problem? I believe that a transaction with a duplicate redeemer key would do?

yHSJ (approved):

PR #765, opened 2026-04-21
fix: too many `adopted tip` messages at the `INFO` level during catch-up

etorreborre:
This PR sends down the best known block height from the `select_chain` stage to the `adopt_chain` stage.

Otherwise the `max_block_height` information that the `adopt_chain` receives comes only from the max block height that the `fetch_blocks` stage has been required to fetch. Then that height will drag along closely to the adopted tip and we will then print too many `adopted_tip` messages at the `INFO` level.

@rkuhn do you agree with that diagnosis and fix?

## Summary by CodeRabbit

* **Bug Fixes**
  * Consensus pipeline now consistently propagates block-height information through tip selection and block-fetching stages, improving correctness of tip handling and fetch behavior.
* **Tests**
  * Updated test expectations and benchmarks to reflect the new block-height-aware message shape, keeping test coverage aligned with runtime behavior.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -78,10 +78,10 @@ jobs:
             target: riscv32gc-unknown-linux-gnu
             title: riscv32
             packages: -p amaru-ledger
-            extra-args: +nightly -Zbuild-std=std,panic_abort
+            extra-args: +nightly-2025-11-21 -Zbuild-std=std,panic_abort
             command: build
             setup: |
-              rustup component add rust-src --toolchain nightly
+              rustup component add rust-src --toolchain nightly-2025-11-21
               curl -L -o riscv32-glibc-llvm.tar.xz \
                 "https://github.com/riscv-collab/riscv-gnu-toolchain/releases/download/2025.11.04/riscv32-glibc-ubuntu-22.04-llvm.tar.xz"
               tar -xvf riscv32-glibc-llvm.tar.xz
--- a/crates/amaru-consensus/src/stages/select_chain/mod.rs
+++ b/crates/amaru-consensus/src/stages/select_chain/mod.rs
@@ -14,14 +14,14 @@
 
 use std::{cmp::Ordering, collections::BTreeMap};
 
-use amaru_kernel::{BlockHeader, HeaderHash, IsHeader, Point, Tip};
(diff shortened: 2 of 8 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -78,10 +78,10 @@ jobs:
             target: riscv32gc-unknown-linux-gnu
             title: riscv32
             packages: -p amaru-ledger
-            extra-args: +nightly -Zbuild-std=std,panic_abort
+            extra-args: +nightly-2025-11-21 -Zbuild-std=std,panic_abort
```
rkuhn, line 81: While I applaud fixing some date, why not choose something more recent?

crates/amaru-consensus/src/stages/select_chain/mod.rs
```diff
@@ -123,7 +126,7 @@ impl SelectChain {
             tracing::debug!(tip = %tip.point(), height = %tip.block_height(), %best_tip, "new best tip candidate");
             if self.may_fetch_blocks {
                 self.may_fetch_blocks = false;
-                eff.send(&self.downstream, (tip, parent)).await;
+                eff.send(&self.downstream, (tip, parent, header.block_height())).await;
```
rkuhn, line 129: This seems to be the core change, and I fail to understand what exactly it should be changing: the `header` here is exactly the one referred to by `tip`, so the `tip.block_height` should exactly match `header.block_height()`. Is there a bug that makes this false?

rkuhn (changes requested):
As far as I can see this doesn’t change anything. If you see fewer log lines for `adopted tip` then that is probably just the bi-stable behaviour of the system I (try to) defend against in #749. This performance aspect is not yet understood, so my proposed fix (in that PR) is to throttle the logs based on clock time — which is the desired behaviour and therefore the proper fix in any case.

.github/workflows/continuous-integration.yml
jeluard, line 81: Unrelated? Any reason to include that?

etorreborre: `main` is broken at the moment because nightly has changed and that job doesn't pass anymore.

etorreborre:
I closed this PR because when we are in sync mode, the block height sent by the `select_chain` stage is not that far from the `fetch_block` max height, so the situation would not change much.

PR #781, opened 2026-04-25
feat: get peer candidates from ledger

rkuhn:
This is only the effect implementation, the addresses aren’t used yet.

## Summary by CodeRabbit

* **New Features**
  * Nodes and ledger now expose and surface registered relay socket addresses from stake pools for network use
  * Runtime provides a stake-pools resource at startup for downstream stages

* **Tests**
  * Test harnesses updated to register and exercise the stake-pools capability

* **Chores**
  * CI snapshot workflow matrix target epoch adjusted

### DIFF
--- a/crates/amaru-ledger/src/registered_relay_addrs.rs
+++ b/crates/amaru-ledger/src/registered_relay_addrs.rs
@@ -37,15 +37,15 @@ pub fn collect_from_read_store(db: &impl ReadStore) -> Result<BTreeSet<SocketAdd
 fn relay_to_socket_addrs(relay: &Relay, set: &mut BTreeSet<SocketAddr>) {
     match relay {
         Relay::SingleHostAddr(port, ipv4, ipv6) => {
-            let Some(port) = null_to_port(port) else {
+            let Some(port) = nullable_to_port(port) else {
                 return;
             };
-            if let Some(ip) = null_ipv4_to_ip(ipv4)
+            if let Some(ip) = nullable_ipv4_to_ip(ipv4)
                 && !is_excluded_relay_ip(ip)
             {
                 set.insert(SocketAddr::new(ip, port));
             }
-            if let Some(ip) = null_ipv6_to_ip(ipv6)
+            if let Some(ip) = nullable_ipv6_to_ip(ipv6)
                 && !is_excluded_relay_ip(ip)
             {
                 set.insert(SocketAddr::new(ip, port));
@@ -57,14 +57,14 @@ fn relay_to_socket_addrs(relay: &Relay, set: &mut BTreeSet<SocketAddr>) {
     }
 }
 
-fn null_to_port(port: &Nullable<u32>) -> Option<u16> {
+fn nullable_to_port(port: &Nullable<u32>) -> Option<u16> {
     match port {
         Nullable::Some(p) => u16::try_from(*p).ok(),
         Nullable::Null | Nullable::Undefined => None,
     }
 }
 
-fn null_ipv4_to_ip(null: &Nullable<Bytes>) -> Option<IpAddr> {
+fn nullable_ipv4_to_ip(null: &Nullable<Bytes>) -> Option<IpAddr> {
     let bytes = match null {
         Nullable::Some(b) => <[u8; 4]>::try_from(b).ok()?,
         Nullable::Null | Nullable::Undefined => return None,
@@ -72,7 +72,7 @@ fn null_ipv4_to_ip(null: &Nullable<Bytes>) -> Option<IpAddr> {
     Some(IpAddr::V4(Ipv4Addr::from_octets(bytes)))
 }
 
-fn null_ipv6_to_ip(null: &Nullable<Bytes>) -> Option<IpAddr> {
+fn nullable_ipv6_to_ip(null: &Nullable<Bytes>) -> Option<IpAddr> {
     let mut bytes = match null {
         Nullable::Some(b) => <[u8; 16]>::try_from(b).ok()?,
         Nullable::Null | Nullable::Undefined => return None,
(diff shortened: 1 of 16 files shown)


crates/amaru-ledger/src/registered_relay_addrs.rs
```diff
@@ -0,0 +1,107 @@
...
+    Some(IpAddr::V4(Ipv4Addr::from_octets(bytes)))
+}
+
+fn null_ipv6_to_ip(null: &Nullable<Bytes>) -> Option<IpAddr> {
+    let mut bytes = match null {
+        Nullable::Some(b) => <[u8; 16]>::try_from(b).ok()?,
+        Nullable::Null | Nullable::Undefined => return None,
+    };
+    // manual inspection of the ledger contents has confirmed that Haskell node
+    // serialization of IPv6 addresses is in “screwed-endian” byte order
```
yHSJ, line 81: It would be nice to provide a link to the specific logic that proved this

crates/amaru-ledger/src/registered_relay_addrs.rs
```diff
@@ -0,0 +1,107 @@
...
+    for (_, row) in db.iter_pools()? {
+        for relay in &row.current_params.relays {
+            relay_to_socket_addrs(relay, &mut set);
+        }
+    }
+    span.record("addresses", set.len());
+    Ok(set)
+}
+
+fn relay_to_socket_addrs(relay: &Relay, set: &mut BTreeSet<SocketAddr>) {
```
yHSJ, line 37: NIT: I generally prefer avoiding passing mutable collections like this. I feel like it's very easy to mess up later. I'd personally prefer filtering and mapping to produce a new collection which is returned. Or providing some mapping of `Relay` to `SocketAddr`, and then adding it to the collection at the call site.

crates/amaru-ledger/src/registered_relay_addrs.rs
```diff
@@ -0,0 +1,107 @@
...
+                set.insert(SocketAddr::new(ip, port));
+            }
+        }
+        Relay::SingleHostName(..) | Relay::MultiHostName(..) => {
+            // DNS usage is very inconsistent, don't bother for now
+        }
+    }
+}
+
+fn null_to_port(port: &Nullable<u32>) -> Option<u16> {
```
yHSJ: NIT: these names are rather confusing. We're not converting `null` to a port, we're reading from a type that "optionally" (bad choice of words, since it's not an `Option`), contains a port.

crates/amaru-ledger/src/registered_relay_addrs.rs
```diff
@@ -0,0 +1,107 @@
...
+                set.insert(SocketAddr::new(ip, port));
+            }
+            if let Some(ip) = null_ipv6_to_ip(ipv6)
+                && !is_excluded_relay_ip(ip)
+            {
+                set.insert(SocketAddr::new(ip, port));
+            }
+        }
+        Relay::SingleHostName(..) | Relay::MultiHostName(..) => {
+            // DNS usage is very inconsistent, don't bother for now
```
yHSJ, line 55: This should be a TODO, which we search for semi-regularly in the ledger to identify open work to complete. Otherwise, this may be missed and not implemented in the future.

crates/amaru-ledger/src/registered_relay_addrs.rs
rkuhn: This code is on the critical path in that it blocks ledger operations, so I want to avoid useless allocations. Since the number of addresses isn’t statically known, this is the best pattern, and it is idiomatic Rust (which is not Haskell).

crates/amaru-ledger/src/registered_relay_addrs.rs
rkuhn: TBH I don’t think there is a risk of this: MultiHostName depends on CIP-0155 which is not ready to be accepted (I’d contribute improvements but that isn’t the best time spent by the Amaru team at this point). Since it will probably be us driving that CIP forward at some future point, we will most certainly then revisit our own implementation. Until then, this is not actually a TODO.

crates/amaru-ledger/src/registered_relay_addrs.rs
rkuhn: Right, will rename to `nullable_to_X`; the Nullable type does indeed not make sense in the way it is currently used, Option would work just fine, but here we are.

crates/amaru-ledger/src/registered_relay_addrs.rs
rkuhn: I was unable to find it, even with “AI”. @KtorZ confirmed that this is correct and somehow inherent to some IP library in Haskell. But manual inspection is quite unambiguous here given that the only block of public IPv6 currently allocated is `2000::/3` and the actually registered addresses on chain.

crates/amaru-ledger/src/state.rs
```diff
@@ -246,6 +247,12 @@ impl<S: Store, HS: HistoricalStores> State<S, HS> {
         self.volatile.view_back().map(|st| st.anchor.0)
     }
 
+    #[expect(clippy::unwrap_used)]
+    pub fn registered_relay_socket_addrs(&self) -> Result<BTreeSet<SocketAddr>, StateError> {
+        let db = self.stable.lock().unwrap();
+        Ok(crate::registered_relay_addrs::collect_from_read_store(&*db)?)
+    }
```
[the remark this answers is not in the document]
rkuhn: Yes, correct, but there is no other way. I’ll make the constraint clear in the doc comment.

crates/amaru-ledger/src/registered_relay_addrs.rs
yHSJ: The second option, of taking a single relay, converting it to a `SocketAddr`, and then inserting it into the collection at the call site is also an option that avoids passing the mutable collection.

You're right that it is idiomatic rust, we do it some other places. But I feel like it is more confusing, easier to mess up, and less readable. But that's just a code style difference, and not a blocking concern.

PR #789, opened 2026-04-30
General Ledger Cleanup

yHSJ:
There are ~50 `TODO` or `FIXME` comments in the `amaru-ledger` crate.

Many of them are rather trivial improvements, so I have begun attacking some of the most obvious/low-hanging fruit on this branch. The commits are scoped to specific `FIXME`/`TODO` changes, so if it is easier to break this up into smaller PRs scoped to single changes, that should not be too difficult to do.

## Summary by CodeRabbit

* **New Features**
  * Added a memoized value type that preserves decoded values and their original serialized bytes.
  * Added transaction-output original_size() and a supplemental-datum policy to control datum inclusion.
  * Introduced a dedicated iterator for unique redeemers.

* **Performance Improvements**
  * Avoided redundant CBOR re-serialization by reusing cached byte representations.
  * Refactored APIs to stream results via iterators instead of building allocated vectors.

* **API Enhancements / Errors**
  * Replaced generic string errors with specific typed error variants for invalid address/payload cases.


crates/amaru-ledger/src/rules/transaction/phase_one/outputs.rs
```diff
@@ -40,17 +40,14 @@ pub enum InvalidOutput {
...
-    #[error("uncategorized error: {0}")]
-    UncategorizedError(String),
 }
 
 pub fn execute<C>(
     context: &mut C,
     protocol_parameters: &ProtocolParameters,
     network: &Network,
     outputs: Vec<MemoizedTransactionOutput>,
+    allow_supplemental_datum: bool,
```
KtorZ: Please use an enum 🙏, avoid booleans whenever possible.

crates/amaru-kernel/src/cardano/memoized/transaction_output.rs
```diff
@@ -15,32 +15,71 @@
...
 
     pub datum: MemoizedDatum,
 
     #[serde(serialize_with = "serialize_script")]
-    #[serde(deserialize_with = "deserialize_script")]
     pub script: Option<MemoizedScript>,
 }
 
+#[derive(serde::Deserialize)]
+struct MemoizedTransactionOutputDe {
```
KtorZ, line 45: Why is this necessary?

KtorZ (approved):
Apart from the this boolean thing, agreed.

PR #790, opened 2026-05-01
feat: leaner bootstrap and snapshots creation

### LINKED ISSUE #645: Generate more recent snapshots & document (manual) bootstrap configuration & steps
### Abstract

Our snapshots are getting slightly _old_ causing spurious issues in our continuous integration due to timeouts; and, forcing long re-sync when bootstrapping from scratch. 

So, we want to produce more recent snapshots, and while at it, address gaps in documentation around _bootstrapping_; in particular answering the following questions:

- [ ] What is covered by the `bootstrap` command?
- [ ] What configuration elements come into play? 
- [ ] How can one obtain the various config pieces?
- [ ] When and why is bootstrap necessary?
- [ ] How far can Amaru "bootstrap"?

### Why?

Bootstrapping Amaru is made "easy" by providing a dedicated `bootstrap` command and pre-configured snapshots. However, there are cases where one may want to bootstrap manually (if only to create a custom network not covered by the `bootstrap` command). 

In which case, there are currently clear gaps in documentation about what exactly is _bootstrapping_, what it required, and how to obtain the various data.

Being able to produce more recent snapshots in a more straightforward manner seems like a good idea. This was only done once, and the knowledge of how this happened hasn't been properly captured.

### How?

- [ ] Pair programming session between _those-in-the-know_ (e.g. @KtorZ @jeluard) and those who want to learn about bootstrap, to create more recent snapshots.
- [ ] Start with preview or preprod, and then mainnet.
- [ ] As part of this exercise, document the steps & information given out.

- [ ] At least one more recent snapshot(s) for preview, preprod and mainnet are available (i.e. updated to cloudflare).  
- [ ] Snapshot generation steps are documented, and visible on the website (https://amaru.global)
- [ ] Get feedback from at least 2 external actors on the updated documentation

### Discussion points

_No response_

_No response_

### Checklist

jeluard:
This is a focused attempt at making it easy to create new snapshots.

Introduce a simple command to create new snapshots that can then be used to bootstrap amaru.
Older amaru snapshots are not supported anymore (a nice error message detects those).

## Usage

Make sure you have a docker instance running and access to a remote peer.

```shell
export AMARU_PEER_ADDRESS=preprod-node.play.dev.cardano.org:3001

# First generate required snapshots to bootstrap some network at some epoch
cargo run --bin amaru -- create-snapshots -f --network preprod --epoch 163

# Then bootstrap
cargo run --bin amaru -- bootstrap -f --network preprod


# To make those usable by any amaru end user
# First publish them on S2/R2

AWS_ACCESS_KEY_ID=YOUR_KEY \
AWS_SECRET_ACCESS_KEY=YOUR_SECRET \
AWS_DEFAULT_REGION=auto \
AMARU_NETWORK=preprod \
BOOTSTRAP_SNAPSHOT_EPOCH=182 \
BUCKET_NAME=cardano-ledger-snapshots \
ENDPOINT=... \
make publish-bootstrap-snapshots

# You should then commit changes to `crates/amaru/config/bootstrap/$NETWORK/snapshots.json`
# those will be bundled in amaru binary and used by `amaru bootstrap` to locate snapshots not available locally; required for end users

```

## Behavior

Necessary `headers` are fetched from live peers. `nonces` are exptracted from full `cardano_node` persistency.
Proper slot numbers are retrieved from `koios`.
`cbor` state files are still retrieved from `r2` although they are now full `cardano_node` persistency.

This implies that your node must have internet access while bootstrapping.

## Notes

Now the following files are not needed anymore:

* headers/*
* nonces.json

Currently only support `InMem` cardano-node.

It would make sense to have default peer per network, at least for the bootstrap commands

Removed `docker/testnet` as it's being moved to [cardano-node-antithesis](https://github.com/cardano-foundation/cardano-node-antithesis) (see https://github.com/cardano-foundation/cardano-node-antithesis/issues/72). It relies on now removed amaru commands.

## Next

EraHistory can now be extracted from the `cardano_node` snapshots. Is it worth persisting it so that it can evolve and we don't need to hardcode it anymore?
It would also simplify custom testnet support.

Make a similar push to simplify tests snapshots creation.

### Preprod

```shell
export AMARU_PEER_ADDRESS=preprod-node.play.dev.cardano.org:3001

# Generate snpashots into snapshots
cargo run --bin amaru -- create-snapshots --network preprod --epoch 163

# Bootstrap from those snapshots
cargo run --bin amaru -- bootstrap -f
```

Then you can perform a full diff of amaru dbs to ensure newly ones are bytes equivalent with the one generated with previous bootstrap method
(assuming you have them available as `chain.preprod.old.db` and `ledger.preprod.old.db/*`)

```shell
rocksdb-diff chain.preprod.old.db chain.preprod.db --prefix acct,comm,dlg,drep,pool,pots,prop,slot,utxo,vote

rocksdb-diff ledger.preprod.old.db/163 ledger.preprod.db/163 --prefix acct,comm,dlg,drep,pool,pots,prop,slot,utxo,vote
rocksdb-diff ledger.preprod.old.db/164 ledger.preprod.db/164 --prefix acct,comm,dlg,drep,pool,pots,prop,slot,utxo,vote
rocksdb-diff ledger.preprod.old.db/165 ledger.preprod.db/165 --prefix acct,comm,dlg,drep,pool,pots,prop,slot,utxo,vote
```

### Preview

```shell
export AMARU_PEER_ADDRESS=preview-node.play.dev.cardano.org:3001

cargo run --bin amaru -- create-snapshots --network preview --epoch 646
cargo run --bin amaru -- bootstrap --network=preview -f
```

```shell
rocksdb-diff chain.preview.old.db chain.preview.db --prefix acct,comm,dlg,drep,pool,pots,prop,slot,utxo,vote

rocksdb-diff ledger.preview.old.db/646 ledger.preview.db/646 --prefix acct,comm,dlg,drep,pool,pots,prop,slot,utxo,vote
rocksdb-diff ledger.preview.old.db/647 ledger.preview.db/647 --prefix acct,comm,dlg,drep,pool,pots,prop,slot,utxo,vote
rocksdb-diff ledger.preview.old.db/648 ledger.preview.db/648 --prefix acct,comm,dlg,drep,pool,pots,prop,slot,utxo,vote
```

### Mainnet

```shell
export AMARU_PEER_ADDRESS=backbone.cardano.iog.io:3001

cargo run --bin amaru -- create-snapshots --network mainnet --epoch 507
cargo run --bin amaru -- bootstrap --network=mainnet -f
```

```shell
rocksdb-diff chain.mainnet.old.db chain.mainnet.db --prefix acct,comm,dlg,drep,pool,pots,prop,slot,utxo,vote

rocksdb-diff ledger.mainnet.old.db/507 ledger.mainnet.db/507 --prefix acct,comm,dlg,drep,pool,pots,prop,slot,utxo,vote
rocksdb-diff ledger.mainnet.old.db/508 ledger.mainnet.db/508 --prefix acct,comm,dlg,drep,pool,pots,prop,slot,utxo,vote
rocksdb-diff ledger.mainnet.old.db/509 ledger.mainnet.db/509 --prefix acct,comm,dlg,drep,pool,pots,prop,slot,utxo,vote
```

```shell
❯ rocksdb-diff ledger.mainnet.old.db/509 ledger.mainnet.db/509 --prefix acct,comm,dlg,drep,pool,pots,prop,slot,utxo,vote

Diff summary
─────────────────────────────────────────

  [default]

    ▸ pool (3094 identical keys)
      ~ 1 replace (same key, different value)
          key=581c800cce924571de8aa9dec5ffd29a9aeae5ea49c6658761aed866092c
            old: …1832581de1e3af434a5516854f20191807cc5ea85b57b4fd0f050f3eab28af19…
            new: …1832581de0e3af434a5516854f20191807cc5ea85b57b4fd0f050f3eab28af19…
```

Note `...e1...` vs `...e0...`
This appears to be an issue with the original snaphots created with ogmios. e0 here relates to the stake address of an account, with e0 speicifcally identifying a testnet address.
In the raw `cardano_node` snapshot the value is indeed e0, while the previous ogmios based snapshot contains e1. This suggests that some normalization happended dureing those past snapshot generation losing the original (mistaken?) stake address in the process.

Fixes https://github.com/pragma-org/amaru/issues/645

## Summary by CodeRabbit

* **New Features**
  * Improved transaction-output decoding and state-snapshot parsing; support for importing UTxOs from node TVAR snapshot files.
  * Public exports now include multiasset types for richer value handling.

* **Refactor**
  * Import/bootstrap flow now uses on-disk consensus state and directory-based snapshot imports; nonces/headers are persisted to the store.
  * CLI import behavior simplified: ledger snapshot imports now expect a snapshot directory.

* **Chores**
  * Removed legacy ledger-conversion/import subcommands and testnet Docker/sample genesis/config artifacts.


jeluard:
@KtorZ Please take a look and push the relevant preprod snapshots using `make publish-bootstrap-snapshots`

KtorZ:
Quick first comments from reading the PR description:

>  Older amaru snapshots are not supported anymore.

This means we need proper warnings and errors from trying to handle old snapshots. 

> This implies that your node must have internet access while bootstrapping.

I don't like this at all; I see no good reasons to not make the snapshot procedure fully local if provided with the right data. So may we separate the steps so that we (1) fetch, then (2) bootstrap; but not interleave fetching within bootstrap?

crates/amaru-kernel/src/cardano/value.rs
```diff
@@ -12,4 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-pub use pallas_primitives::conway::Value;
+pub use pallas_primitives::{
+    NonEmptyKeyValuePairs as MultiassetKeyValuePairs,
+    conway::{Multiasset, Value},
```
KtorZ: Please make separate modules. One type per module in the kernel 🙏. Those imports should ideally not exist.

crates/amaru-mithril/src/lib.rs
KtorZ, line 1: I feel like this is going to grow; maybe worth parking in its own `amaru-mithril` and build a proper anti-corruption layer on top from the get-go? So that we own our interface to Mithril?

crates/amaru/src/cardano_node/tvar.rs
```diff
@@ -0,0 +1,235 @@
...
+        d.array()?;
+        Ok(d.map()?.map(|len| len as usize))
+    })?;
+
+    let estimated_size = size.unwrap_or(match network {
+        NetworkName::Mainnet => 11_000_000,
+        NetworkName::Preview => 1_500_000,
+        NetworkName::Preprod => 1_500_000,
+        NetworkName::Testnet(..) => 1,
+    });
```
KtorZ: Should be a method on `NetworkName` ?

crates/amaru/src/bootstrap.rs
```diff
@@ -51,105 +57,389 @@ struct Snapshot {
...
+    DownloadInvalidStatusCode(String, StatusCode),
+
+    #[error("Missing cardano-node snapshot directory {0}")]
+    MissingSnapshotDirectory(PathBuf),
+
+    #[error("Missing bootstrap parent point for snapshot {0}")]
+    MissingParentPoint(String),
+
+    #[error("{0}")]
+    SnapshotSelection(String),
```
KtorZ: May we try to do better than `String` 🙏, add context + details to errors.

crates/amaru-ledger/src/bootstrap.rs
```diff
@@ -42,38 +42,82 @@ static DEFAULT_CERTIFICATE_POINTER: LazyLock<CertificatePointer> = LazyLock::new
...
     with_progress: impl Fn(usize, &str) -> Box<dyn ProgressBar>,
-    // Assumes the presence of fully computed rewards when set.
+    decode_pool_state: for<'a> fn(
+        &mut cbor::Decoder<'a>,
+        NetworkName,
+    ) -> Result<
+        (BTreeMap<PoolId, PoolParams>, BTreeMap<PoolId, PoolParams>, BTreeMap<PoolId, Epoch>),
+        cbor::decode::Error,
+    >,
+    pool_state_tail_skips: usize,
```
KtorZ: What's that?

crates/amaru/src/bin/amaru/cmd/generate_epoch_snapshots/db_analyser.rs
```diff
@@ -0,0 +1,421 @@
...
+use super::{config::DbAnalyserBuildConfig, repo_root};
+
+const DB_ANALYSER_PROGRESS_REPORT_INTERVAL_SECS: f64 = 30.0;
+
+pub(super) fn ensure_db_analyser_image(
+    db_analyser_build_config: &DbAnalyserBuildConfig,
+) -> Result<String, Box<dyn std::error::Error>> {
+    let image_ref =
+        format!("{}:{}", db_analyser_build_config.image, sanitize_image_tag(&db_analyser_build_config.git_ref));
+    let status = ProcessCommand::new("docker").args(["image", "inspect", image_ref.as_str()]).status()?;
```
KtorZ: Hmmm. I am not sure about this 😰 ... I get the spirit but now, our CLI makes assumption about Docker being present?

What about: 

- decoupling what's generated from db-analyzer from the amaru-cli by extracting intermediate data 
- have the amaru cli accepts those data as inputs

We can still provide a little command/instruction to get this from db-analyzer, or from a pre-hosted dataset, but running docker within the command-line is not going to work out nicely.

crates/amaru/src/cardano_node/tvar.rs
jeluard: Some util fn would make sense now I feel like it's too specific to be at the type level.

jeluard:
> I don't like this at all; I see no good reasons to not make the snapshot procedure fully local if provided with the right data. So may we separate the steps so that we (1) fetch, then (2) bootstrap; but not interleave fetching within bootstrap?

We could have try to load local ad-hoc files before loading remote files (similarly to what we do for cbor archive). Now not super clear what it provides as you need at some point internet access to bootstrap and most importantly to run your node.

KtorZ:
> > I don't like this at all; I see no good reasons to not make the snapshot procedure fully local if provided with the right data. So may we separate the steps so that we (1) fetch, then (2) bootstrap; but not interleave fetching within bootstrap?
> 
> We could have try to load local ad-hoc files before loading remote files (similarly to what we do for cbor archive). Now not super clear what it provides as you need at some point internet access to bootstrap and most importantly to run your node.

To run the node, yes; but to prepare it _not necessarily_. Thinking for example, someone building a fat Docker image with Amaru pre-bootstrapped, or simply provisioning their environment on a machine where they have tight firewall rules. Not even mentioning the _trust_ aspect of it. Making fetching from online stores the default is dangerous. 

It is going to backfire to have this be fully remote; having two separate steps is crucial here. We may allow fetching this sort-of automatically with a flag (`--fetch-from-remote`), but that flat should never be on by default.

crates/amaru-ledger/src/bootstrap.rs
jeluard: Simplified

crates/amaru/src/bin/amaru/cmd/generate_epoch_snapshots/db_analyser.rs
jeluard: An alternative is to rely on the `db-analyzer` binary provided by cardano-node releases.
The bootstrap creation mechanism integrates `db-analyzer` in the middle of the process with pretty specific arguments, and calls it 3 times. So adding another set of scripts to call it outside of this flow makes everything more complex and brittle. 
Here the value is to embed everything in `amaru` binary and have the snapshots created in a single command/step, granted some binary is available. This sounds a reasonanble tradeoff to me.

jeluard:
> We may allow fetching this sort-of automatically with a flag (--fetch-from-remote), but that flat should never be on by default.

I understand this as a change of behavior from what we have now then? Currently cbor files are downloaded silently if missing locally.

So, what I get is:

```shell
# Download all necessary files locally
amaru prepare-for-bootstrap

# Do the bootstrap based on local files
amaru bootstrap

# Both steps can be combined with
amaru bootstrap --fetch-from-remote

# Then run
amaru run
```

Issue #795, opened 2026-05-01
Remove MemoryStore


yHSJ:
### Abstract

Remove the `MemoryStore`, and any dependencies on it (1 test).

### Why?

The `MemoryStore` has diverged significantly from the `RocksDBStore`, and it doesn't provide us significant value. While it's great idea to have, it would be easier to revisit it later. Since the store code is "in flux", it will be easier to rewrite from scratch, to extract shared logic, or logic that shouldn't belong to at the store layer.

### How?

Delete the `MemoryStore` and clean up dependencies. There will be at least one failing test as a result, that needs to be updated to use the `RocksDBStore`

### Testing Strategy / Acceptance Criteria

If everything compiles and tests pass after `MemoryStore` was removed, the task is completed.

### Checklist

- [x] I understand that feature requests and unrefined work item should be open as [GitHub Discussions](https://github.com/pragma-org/amaru/discussions) instead.
- [x] I have assigned this item to an existing milestone from the roadmap
- [x] I have added a label capturing the impact of this item (i.e. value for users/stakeholders if successful)
- [x] I have added a label capturing the delivery risk of this item (i.e. how likely is it that this task will succeed as planned)
- [x] I have added a label capturing the effort of this item (i.e. how large is the task?)

yHSJ:
This is more involved than I had originally suspected, and I'd like to open this up for some conversation (cc: @KtorZ @etorreborre @rkuhn)

To summarize the goal, we want to *remove* the `MemoryStore`, as it's greatly diverged from `RocksDB`. While having an in memory storage option is a great and useful idea, it's more burden than it's worth right now to maintain it (highlighted by the fact that it *hasn't* been maintained). Removing it would also simplify logic in other places.

Current places where we use `MemoryStore`:
1) State rollback test
2) `amaru_sim` test harness
3) the `nodejs` examples
4) There is lots of plumbing that may have dependencies *or* are shaped to support in memory storage. Some of these are shared between the ledger stores and consensus store.

1 and 2 are easy fixes; we simply use a `tempdir` when setting up tests. 4 probably doesn't need to change, even though it could potentially be simplified, since it still supports the in memory consensus store which doesn't have the same maintenance issues.

3 is more difficult. From my experimenting, rocksDB can't compile to WASM for a variety of issues, and it can't be simply fixed by providing the right C++ toolchain.

[...]

PR #801, opened 2026-05-05
fix: don't fail rollback inside the volatile ledger

etorreborre:
We can't currently rollback the ledger to a previous volatile state.

## Summary by CodeRabbit

* **Improvements**
  * Refined ledger state rollback behavior with enhanced error handling that distinguishes between different rollback failure scenarios, including unknown rollback points and future-dated rollback targets.

* **Tests**
  * Added new test suite for ledger rollback functionality covering success cases and various error conditions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/pragma-org/amaru/pull/801)

### DIFF
--- a/crates/amaru-ledger/Cargo.toml
+++ b/crates/amaru-ledger/Cargo.toml
@@ -50,6 +50,7 @@ toml.workspace = true
 
 # Internal dependencies ───────────────────────────────────────────────────────┐
 amaru-kernel = { workspace = true, features = ["test-utils"] }
+amaru-stores.workspace = true
 amaru-tracing-json.workspace = true
 
 [build-dependencies]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -220,6 +220,7 @@ dependencies = [
  "amaru-ouroboros-traits",
  "amaru-plutus",
  "amaru-progress-bar",
+ "amaru-stores",
  "amaru-tracing-json",
  "amaru-uplc",
  "anyhow",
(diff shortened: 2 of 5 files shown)


crates/amaru-ledger/Cargo.toml
```diff
@@ -50,6 +50,7 @@ toml.workspace = true
 
 # Internal dependencies ───────────────────────────────────────────────────────┐
 amaru-kernel = { workspace = true, features = ["test-utils"] }
+amaru-stores.workspace = true
```
jeluard, line 53: It's unfortunate that we introduce this dependency just for a test, especially given that `MemoryStore` is pretty broken and will be removed soon AFAIK.

Maybe there is a way to test this w/o it?

etorreborre: Let me check.

etorreborre: I don't see anything immediate. It also seems to me that having a reliable in-memory ledger store is a useful thing to have. One way to validate it is to do what we do for the `ChainStore` and run the same tests across the in-memory version and the RocksDB version.

KtorZ: It definitely is useful; we discussed it with Josh last week. For now though the plan is to clean up the old in memory store implementation that is obsolete.

etorreborre: @KtorZ is this PR good to merge as is?

yHSJ: Just for context: we have #795 to track removing the `MemoryStore` (which I misremembered as `InMemoryStore`)

yHSJ (approved):

crates/amaru-ledger/Cargo.toml
KtorZ: @etorreborre no. The rollback logic here used to be quite straightforward and after a few successive updates from different people, I find it very confusing. I'll try to make a quick pass on it.

KtorZ (approved):

Issue #797, opened 2026-05-01
Generic Ledger Tests


yHSJ:
### Abstract

We should ensure our ledger tests are rust-agnostic, and can be consumed by anyone.

### Why?

Many of our tests are rust-specific, or require the rust code to fully understand. These tests should be self describing language-agnostic, with the fixtures providing everything needed to run the test. This not only benefits the Cardano ecosystem, it also greatly improves our codebase. 

### How?

Every test should be:
  -  Valid transactions if the predicate under test were removed
  -  Self describing (can be by the path)
  - Self contained (all data needed is provided in the fixture)
  - Language agnostic
  - Encoded on disk in some standard JSON schema

All groups of tests should have some pseudocode to explain how to test the provided data. Generally, the flow would be:
Deserialize test fixture, run ledger on the provided transaction given the provided initial state, serialize the output, compare to the expected result.

### Testing Strategy / Acceptance Criteria

- Every test should pass if the specific predicate under test were removed
- By judgement, they should be simple to consume from another codebase in a different language.

### Checklist

- [x] I understand that feature requests and unrefined work item should be open as [GitHub Discussions](https://github.com/pragma-org/amaru/discussions) instead.
- [x] I have assigned this item to an existing milestone from the roadmap
- [x] I have added a label capturing the impact of this item (i.e. value for users/stakeholders if successful)
- [x] I have added a label capturing the delivery risk of this item (i.e. how likely is it that this task will succeed as planned)
- [x] I have added a label capturing the effort of this item (i.e. how large is the task?)

yHSJ:
This should really be broken up into several pieces of work, and not all of them need to be in the same milestone.

- [x] Define schema and harness for fixtures
- [ ] Close existing coverage gaps
- [ ] Migrate existing unit tests to the new fixture schema (in progress)
- [ ] Increase coverage for fixture schema

The last point may not necessarily be a single issue, but perhaps instead it should be a requirement of ledger work that there is a test covering the changes in each PR.

yHSJ:
I am closing this issue in favor of #885. This issue being complete marks the schema and harness implementation.

Future issues will be created for the above work that isn't covered in #885

PR #804, opened 2026-05-07
Ledger Test Harness & Some Initial Fixtures

yHSJ:
This PR is the first step to resolving #797. I am introducing a prototype of the ledger test harness and the shape of the test fixtures. While the rust code is open for review (particularly in terms of how we want to decode the JSON fixtures), the most important discussion I want to have is around the test fixtures and whether or not they meet the requirements of #797.

**Some key decisions I made:**
- I decided to follow the `cardano-cli`'s output shape for the protocol parameters, which is defined here: https://github.com/IntersectMBO/cardano-api/blob/master/cardano-api/src/Cardano/Api/ProtocolParameters.hs
- I decided to use the predicate names from the Haskell codebase as the "canonical errors". Implementations don't have to match them, but they should be able to map from their error to the error in the fixture. The predicate failures can optionally have a "minimum required data" that all implementations should provide.
- Notably, the initial state right now is NOT the complete initial state that all nodes would need, it's just the state we need to construct our validation context. As we expand the validation context, this will have to expand as well, eventually being a complete view of a starting state.
- The transaction as the "unit" of the test. The ledger generally operates in blocks, which a transaction is a piece of. But, the structure of a block is a bit more obnoxious to work with. Using a transaction felt like a reasonable trade off, as it should be simple to isolate phase one logic.

Once this work is settled, the rest of the work is mostly mechanical, moving existing test fixtures and modifying them to make sure they are all inline with the expectations of #797.

UPDATE:
I've merged in some new fixtures to introduce some basic coverage (First introduced in #807). This helps make it clear how these fixtures will be used and what they look like.

## Summary by CodeRabbit

* **Tests**
  * Added extensive phase-one JSON pass/fail fixtures and a conformance test harness that runs transactions against canonical predicates.
* **Chores**
  * Added test-only proxy/serde helpers (including $ref resolution, UTxO/CBOR decoding) and mappings from validation errors to canonical predicates to simplify fixture loading.
* **Documentation**
  * Added README and JSON Schema describing fixture layout, fields, merge semantics, and contribution steps for new test vectors.


crates/amaru-ledger/src/rules/transaction/phase_one/fixture.rs
yHSJ, line 1: I think it probably makes more sense to have the kernel types own the deserialization, with small helpers/wrappers wherever that is needed. I decided that for a prototype, it was better to have the decoding live in it's own layer though.

crates/amaru-ledger/tests/data/phase-one/fail/InvalidWitnessesUTXOW/0.json
```diff
@@ -0,0 +1,154 @@
...
+          "epochSizeSlots": 432000,
+          "slotLengthMs": 1000,
+          "eraName": "Conway"
+        }
+      }
+    ]
+  },
+  "protocolParameters": {
+    "protocolVersion": { "major": 9, "minor": 0 },
+    "txFeePerByte": 44,
```
yHSJ: The names for parameters here come from the `cardano-cli`. They differ in a few places from the Haskell node, but I found they are generally better names.

crates/amaru-ledger/tests/data/phase-one/pass/simple-transfer.json
```diff
@@ -0,0 +1,154 @@
...
+          "datum": null,
+          "script": null
+        }
+      ]
+    ],
+    "votingState": {
+      "numDormantEpochs": 0
+    }
+  },
+  "ledgerEnv": {
```
yHSJ, line 22: `ledgerEnv` comes from the Haskell; it is the name of the type that carries information about where in the chain the transaction occurs. We use a `TransactionPointer` type. I am not entirely pleased with either one of these for the fixture, since they both are specific to an implementation.

crates/amaru-ledger/tests/data/phase-one/pass/simple-transfer.json
```diff
@@ -0,0 +1,154 @@
+{
+  "title": "simple-transfer",
```
yHSJ, line 2: We do not actually consume the title and description in Amaru, but I felt that it was helpful for humans reviewing the fixtures to, at a glance, understand what they are testing.

crates/amaru-ledger/tests/data/phase-one/fail/InvalidWitnessesUTXOW/0.json
yHSJ, line 1: my end goal is to be able to package these tests in a tarball. The path to the file is part of the test being "self-describing". There could be many test cases for any predicate, so the naming of each test case is a bit challenging.

I'd also be open to removing title/description from the JSON, since the path is self describing enough.

crates/amaru-ledger/src/rules/transaction/phase_one/fixture.rs
KtorZ: Agreed.

crates/amaru-ledger/tests/data/phase-one/fail/InvalidWitnessesUTXOW/0.json
KtorZ: My personal favorites are Ogmios' names, but I am _obviously_ biased there.

crates/amaru-ledger/tests/data/phase-one/fail/InvalidWitnessesUTXOW/0.json
```diff
@@ -0,0 +1,154 @@
...
+        8942, 0, 1, 159378, 8813, 0, 1, 107490, 3298, 1, 106057, 655, 1,
+        1964219, 24520, 3
+      ]
+    }
+  },
+  "initialState": {
+    "utxo": [
+      [
+        {
+          "transaction_id": "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
```
KtorZ: POV: you're looking at how the script integrity hash is computed.

crates/amaru-ledger/tests/data/phase-one/fail/InvalidWitnessesUTXOW/0.json
```diff
@@ -0,0 +1,154 @@
...
+        "end": null,
+        "params": {
+          "epochSizeSlots": 432000,
+          "slotLengthMs": 1000,
+          "eraName": "Conway"
+        }
+      }
+    ]
+  },
+  "protocolParameters": {
```
KtorZ, line 8: Note: it might be worth putting a reference here; like it was done for the existing ledger-state tests extracted from the Haskell nodes by Ethan & al. The parameters don't change _that_ much. So I would suggest to content-address them here (by a somewhat canonical hash) and leave a JSON pointer here to the right file.

crates/amaru-ledger/tests/data/phase-one/fail/InvalidWitnessesUTXOW/0.json
```diff
@@ -0,0 +1,154 @@
...
+          "transaction_id": "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
+          "index": 0
+        },
+        {
+          "address": "6091de07fed8a2b52a92686dd68bb2b8d443d6d42089cf508259774675",
+          "value": { "Coin": 5000000 },
+          "datum": null,
+          "script": null
+        }
+      ]
```
KtorZ: It'd probably be better to store the UTxO as CBOR too. At least, the inputs and outputs. We already feel the pain in Amaru of having to support dual encoding for these things. And it's another layer of potential mistakes. So what about a JSON list of objects: `{ "input": "<hex-encoded cbor input>", "output": "<hex-encoded cbor output>" }` ? 

That will heavily simplify some of our own code, and still remain very readable at the high level for those inspecting test scenarios.

crates/amaru-ledger/tests/data/phase-one/pass/simple-transfer.json
KtorZ: The name is definitely not super insightful by itself.

crates/amaru-ledger/tests/data/phase-one/pass/simple-transfer.json
KtorZ: Isn't it redundant with the file title?

crates/amaru-ledger/tests/data/phase-one/fail/InvalidWitnessesUTXOW/0.json
yHSJ: I considered this, but I wanted to prioritize the "self contained" aspect of these fixtures. I think this is a *very* reasonable tradeoff, I just wanted to defer that trade off until later.

crates/amaru-ledger/tests/data/phase-one/fail/InvalidWitnessesUTXOW/0.json
yHSJ: You know what, let's go with the Ogmios names, since this decision is almost entirely arbitrary. You deserve it 😛

crates/amaru-ledger/tests/data/phase-one/pass/simple-transfer.json
yHSJ: I swear I responded to this comment, but GitHub has been struggling recently.

Yes, in theory it is redundant. I worry that, for more complex test cases, it can be harder to have a clear file name that is also ergonomic. Maybe the best solution is to allow these fields to be optionally included, in the case that the file name alone is not descriptive enough. It shouldn't actually impact anything in the test harness implementation, but does serve some human readability benefits.

crates/amaru-ledger/tests/data/phase-one/fail/InvalidWitnessesUTXOW/0.json
KtorZ: A JSON reference is still "self-contained"; just over two files 🤓

yHSJ:
I've moved a lot of the deserialization logic to at least be *colocated* with the relevant kernel types. We have a few different patterns established for this kind of work, with private helpers in deserialization functions and `HasProxy` implementations as the most notable. I decided the `HasProxy` was the most readable and easiest to understand, but I am not really married to it, and it could definitely be a misuse of the proxy concept.

I also gated on the `test-utils` feature, to prevent hurting compilation times. This may add some complexity and can definitely be "cleaned up".

All that being said, I think I can to some very reasonable decisions and it allows me to continue moving forward here. Improving this code can be done in a follow up ticket, if we decide it is necessary.

PR #805, opened 2026-05-08
fix: validate downloaded but still unvalidated blocks in fetch_blocks

etorreborre:
This PR fixes the case where the `fetch_block` stage might have downloaded blocks that were not yet validated before. Without it, we could have an inconsistent state on restart.

It also sets the proper chain store values for the anchor and best hash when we start from a bootstrapped state.

## Release Notes

* **New Features**
  * Added block recovery during consensus initialization to resubmit stored but unvalidated blocks before resuming synchronization.

* **Improvements**
  * Optimized block existence validation checks for better performance.
  * Enhanced chain store initialization and anchor management.

[Review Change Stack](https://app.coderabbit.ai/change-stack/pragma-org/amaru/pull/805)

### DIFF
--- a/crates/amaru-consensus/src/stages/fetch_blocks/mod.rs
+++ b/crates/amaru-consensus/src/stages/fetch_blocks/mod.rs
@@ -113,20 +113,25 @@ impl FetchBlocks {
         let tip = best_tip.tip();
         tracing::debug!(tip = %tip.point(), "recovering stored blocks");
 
+        let mut parent: Option<Point> = None;
         for hash in unvalidated {
             let Some(header) = store.load_header(&hash).await else {
                 tracing::error!(%hash, "failed to load candidate header");
                 return eff.terminate().await;
             };
             let tip = header.tip();
-            let parent = load_parent_point(&eff, store.clone(), &header).await;
+            let block_parent = match parent {
+                Some(p) => p,
+                None => load_parent_point(&eff, store.clone(), &header).await,
+            };
             match store.has_block(&hash).await {
                 Ok(true) => {
                     tracing::debug!(point = %tip.point(), "validating stored block");
-                    eff.send(&self.downstream, (tip, parent, self.block_height)).await;
+                    eff.send(&self.downstream, (tip, block_parent, self.block_height)).await;
+                    parent = Some(tip.point());
                 }
                 Ok(false) => {
-                    self.request_missing_blocks(tip, parent, eff).await;
+                    self.request_missing_blocks(tip, block_parent, eff).await;
                     return;
                 }
                 Err(error) => {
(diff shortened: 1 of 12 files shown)


etorreborre:
@coderabbit review

crates/amaru-consensus/src/stages/fetch_blocks/tests.rs
```diff
@@ -82,6 +83,51 @@ fn test_new_tip_no_blocks_to_fetch() {
...
+            te_has_block("fb-1", prep.headers.h2.hash()),
+            te_send("fb-1", "downstream", (prep.headers.h2.tip(), prep.headers.h1.point(), BlockHeight::from(3))),
+            te_send("fb-1", "upstream", SelectChainMsg::FetchNextFrom(prep.headers.h2.point())),
+            te_state("fb-1", &expected),
+        ],
+    );
+    logs.assert_and_remove(Level::DEBUG, &["recovering stored blocks"])
+        .assert_and_remove(Level::DEBUG, &["validating stored block"])
+        .assert_and_remove(Level::INFO, &["no blocks to fetch"])
+        .assert_no_remaining_at([Level::INFO, Level::WARN, Level::ERROR]);
```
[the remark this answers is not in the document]
etorreborre: Fixed

crates/amaru-consensus/src/stages/fetch_blocks/mod.rs
```diff
@@ -78,13 +78,70 @@ impl FetchBlocks {
...
+        let tip = best_tip.tip();
+        tracing::debug!(tip = %tip.point(), "recovering stored blocks");
+
+        for hash in unvalidated {
+            let Some(header) = store.load_header(&hash).await else {
+                tracing::error!(%hash, "failed to load candidate header");
+                return eff.terminate().await;
+            };
+            let tip = header.tip();
+            let parent = load_parent_point(&eff, store.clone(), &header).await;
```
rkuhn: shouldn’t ask the store for information that we already have: the parent point is available from the previous loop iteration or from `best_tip`, and its hash must match the `header.parent()`

rkuhn (approved):
Yes, looking good, just one nitpick.

crates/amaru-consensus/src/stages/fetch_blocks/mod.rs
etorreborre: Good point, fixed.

crates/amaru-consensus/src/stages/fetch_blocks/mod.rs
```diff
@@ -113,17 +113,21 @@ impl FetchBlocks {
         let tip = best_tip.tip();
         tracing::debug!(tip = %tip.point(), "recovering stored blocks");
 
+        let mut parent = Point::Origin;
```
rkuhn: If the `unvalidated` are sorted by slot ascending, then this should be the `best_tip.point()`, shouldn’t it? If it is not, then the loop below now attributes the wrong parent.

rkuhn (commented):
.

crates/amaru-consensus/src/stages/fetch_blocks/mod.rs
etorreborre: This is just used as a sentinel value to know if we are in the first iteration of the loop. 
Back to an `Option` which is clearer.

etorreborre:
@rkuhn I merged as is to make to progress on all the opened PRs. Happy to open a new PR for refinements if necessary.

PR #810, opened 2026-05-11
fix: validation parameters

etorreborre:
## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced block validation to correctly use protocol parameters based on the block's epoch, improving reliability when processing blocks from different time periods.

[Review Change Stack](https://app.coderabbit.ai/change-stack/pragma-org/amaru/pull/810)

### DIFF
--- a/crates/amaru-ledger/src/state.rs
+++ b/crates/amaru-ledger/src/state.rs
@@ -129,6 +129,10 @@ where
     /// Updatable protocol parameters.
     protocol_parameters: ProtocolParameters,
 
+    /// Protocol parameters previewed for validating volatile blocks in an epoch that the stable DB
+    /// has not reached yet.
+    validation_protocol_parameters: Option<(Epoch, ProtocolParameters)>,
+
     /// Track the number of dormant epochs (i.e. epochs that start without any available
     /// proposals).
     governance_activity: GovernanceActivity,
@@ -201,6 +205,8 @@ impl<S: Store, HS: HistoricalStores> State<S, HS> {
             global_parameters: Arc::new(global_parameters),
 
             protocol_parameters,
+
+            validation_protocol_parameters: None,
 
             governance_activity,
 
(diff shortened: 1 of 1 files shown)


yHSJ:
I was struggling to put into words why I don't like this solution, and the short answer is it's a band-aid solution to a bigger issue of how we handle epoch transitions and the protocol parameters. Below is an *attempt* to put it into words:

The code smell that was bothering me is the `protocol_parameters_for_validation` method. When validating a block at the tip, the *active* PPs must be in the stable DB. There is a "slot of no return" which is the point at which the protocol parameters are set for some epoch. The SONR is `boundary - 6f/k`, so the protocol parameters are finalized (in the stable DB) by the time the next epoch starts.

So I think this PR is band-aiding an issue with our epoch transition handling, and it's not even correctly fixing that issue. The one wrinkle I am not sure about is that Eric mentioned he found this issue when importing a Mithril snapshot; I honestly don't have a great understanding of how we handle Mithril imports. I assume this issue may be that since we discard the `futurePParams` field in the snapshot, there are `k` blocks that will write to the volatileDB, and those `k` blocks could cross an epoch boundary that doesn't trigger the transition (since the epoch transition logic only runs when we promote a block from the volatileDB to the stableDB).

Thoughts @KtorZ?

I think to provide more insight, I'd have to spend more time looking into it.  Maybe it's best to file an issue around this so we can put in the effort to *really* fix the underlying issue, which should fix this.

I guess the biggest question is for @etorreborre: how blocked are you? Is this something you could leave a big fat `FIXME` somewhere and continue?

KtorZ (changes requested):
That should not be necessary; but @yHSJ has been pointing out, this is symptomatic of the problem I've been describing for a little while now: the epoch transition doesn't happen at the right moment. 

@etorreborre out of curiosity, in which scenario did you see that being a problem (seems like a good test case for what I am currently working on 😅).

etorreborre:
I created an issue to track this problem: https://github.com/pragma-org/amaru/issues/813.

@KtorZ I downloaded the latest mithril snapshot and tried to import it with `amaru-ledger`. Unfortunately I don't have the console logs anymore.

etorreborre:
I'm closing this PR for now.

PR #812, opened 2026-05-12
fix: include the era tag in the transactions serialization

etorreborre:
The serialization of transactions was missing the era tag. I did not notice that before because I was only testing `amaru` with `amaru` :-)

## Summary by CodeRabbit

* **New Features**
  * Transaction submission is now era-aware: txs and tx-ids carry the current Cardano era tag for correct era-scoped handling.

* **Protocol / Wire**
  * CBOR wire format updated to encode per-item era metadata; decoders validate era prefixes and CBOR-wrapped tx bytes.

* **Tests**
  * Tests expanded to cover era-tagged message encoding/decoding, CBOR roundtrips, and protocol behavior with tagged payloads.

[Review Change Stack](https://app.coderabbit.ai/change-stack/pragma-org/amaru/pull/812)

### DIFF
--- a/crates/amaru-protocols/src/tx_submission/messages.rs
+++ b/crates/amaru-protocols/src/tx_submission/messages.rs
@@ -14,20 +14,38 @@
 
 use std::fmt::Display;
 
-use amaru_kernel::{NonEmptyBytes, Transaction, TransactionId, cbor, to_cbor};
+use amaru_kernel::{EraName, NonEmptyBytes, Transaction, TransactionId, cbor, to_cbor};
 
 use crate::tx_submission::Blocking;
 
+/// A transaction id paired with the era it was created in. Encoded on the wire as
+/// `[era_index, tx_id]`, matching the Haskell hard-fork combinator's `GenTxId` shape.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, serde::Serialize, serde::Deserialize)]
+pub struct EraTaggedTxId {
+    pub era: EraName,
+    pub id: TransactionId,
+}
+
+/// A transaction paired with the era it was created in. Encoded on the wire as
+/// `[era_index, cbor tx]`.
+#[derive(Debug, Clone, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
+pub struct EraTaggedTx {
+    pub era: EraName,
+    pub tx: Transaction,
+}
+
 /// Messages for the txsubmission mini-protocol.
+///
+/// Each item carries its own era tag via [`EraTaggedTxId`] / [`EraTaggedTx`].
 #[derive(Debug, PartialEq, Eq, Clone, serde::Serialize, serde::Deserialize)]
 #[repr(u8)]
 pub enum Message {
     Init,
     RequestTxIdsBlocking(u16, u16),
     RequestTxIdsNonBlocking(u16, u16),
-    RequestTxs(Vec<TransactionId>),
-    ReplyTxIds(Vec<(TransactionId, u32)>),
-    ReplyTxs(Vec<Transaction>),
+    RequestTxs(Vec<EraTaggedTxId>),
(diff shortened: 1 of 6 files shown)


etorreborre:
@coderabbit review

crates/amaru-protocols/src/tx_submission/messages.rs
```diff
@@ -104,27 +108,39 @@ impl cbor::Encode<()> for Message {
...
-                for tx in txs {
-                    e.encode(tx)?;
+                e.begin_array()?;
+                if !txs.is_empty() {
+                    let era = era_for_non_empty::<W>(era, "ReplyTxs")?;
+                    for tx in txs {
+                        encode_tx_with_era_tag(e, era, tx)?;
+                    }
                 }
+                e.end()?;
```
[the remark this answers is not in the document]
etorreborre: Good idea

crates/amaru-protocols/src/tx_submission/messages.rs
```diff
@@ -107,24 +125,28 @@ impl cbor::Encode<()> for Message {
...
             }
             Message::RequestTxs(ids) => {
                 e.array(2)?.u16(2)?;
-                e.array(ids.len() as u64)?;
+                e.begin_array()?;
                 for id in ids {
                     e.encode(id)?;
                 }
+                e.end()?;
             }
```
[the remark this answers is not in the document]
etorreborre: The Haskell node only accepts indefinite length, so we need to keep it that way.

etorreborre:
@KtorZ I checked if it was possible to remove the `PartialEq`, `Eq` instances on the `EraTagged` types but those instances are required by the protocol machinery. That being said it's not too hard to convince oneself that those types are just used in one place then the relevant `TxId` and `Tx` are unwrapped for the rest of the processing.

crates/amaru-protocols/src/tx_submission/responder.rs
```diff
@@ -113,10 +114,14 @@ impl ProtocolState<Responder> for State {
         let _guard = _span.enter();
         Ok(match (self, input) {
             (State::Init, Message::Init) => (outcome().result(ResponderResult::Init), State::Idle),
-            (State::TxIdsBlocking | State::TxIdsNonBlocking, Message::ReplyTxIds(tx_ids)) => {
+            (State::TxIdsBlocking | State::TxIdsNonBlocking, Message::ReplyTxIds(tagged_ids)) => {
+                let tx_ids = tagged_ids.into_iter().map(|(t, size)| (t.id, size)).collect();
```
rkuhn, line 118: Shouldn’t we check the received era tag here?

crates/amaru-protocols/src/tx_submission/responder.rs
```diff
@@ -113,10 +114,14 @@ impl ProtocolState<Responder> for State {
...
                 (outcome().result(ResponderResult::ReplyTxIds(tx_ids)), State::Idle)
             }
-            (State::Txs, Message::ReplyTxs(txs)) => (outcome().result(ResponderResult::ReplyTxs(txs)), State::Idle),
+            (State::Txs, Message::ReplyTxs(tagged_txs)) => {
+                let txs = tagged_txs.into_iter().map(|t| t.tx).collect();
```
rkuhn, line 122: and here?

crates/amaru-protocols/src/tx_submission/responder.rs
etorreborre: I wouldn't know what it should be checked against, once we've decoded tx ids and txs, I think we're good.

rkuhn: It is understanding that the decoding is supposed to be guided by the tag. If we get a tag older than Conway we should not try to decode the txn. And once the next era has been added we will probably have separate decoders per era.

etorreborre: I added a check to make sure we only decode `Conway` transactions for now.

PR #820, opened 2026-05-13
Reference Script Size & Minimum Fee Calculations

### LINKED ISSUE #887: Ledger predicate: BodyRefScriptsSizeTooBig
### Abstract

Implement the `BodyRefScriptsSizeTooBig` predicate

### Why?

Conformance

### How?

[Reference Implementation](https://github.com/IntersectMBO/cardano-ledger/blob/0cfbf861cfb456660a7b73281c6fb714a53d40f9/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Bbody.hs#L332)

Introduce fixtures for coverage

### Discussion points

_No response_

_No response_

### Checklist

### LINKED ISSUE #889: Ledger predicate: ConwayTxRefScriptsSizeTooBig
### Abstract

Implement `ConwayTxRefScriptsSizeTooBig`

### Why?

Conformance

### How?

[Reference Implementation](https://github.com/IntersectMBO/cardano-ledger/blob/0cfbf861cfb456660a7b73281c6fb714a53d40f9/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ledger.hs#L475)

### Discussion points

_No response_

_No response_

### Checklist

### LINKED ISSUE #891: Ledger Predicate: FeeTooSmallUTxO
### Abstract

Implement the predicate `FeeTooSmallUTxO`

### Why?

Conformance

### How?

[Reference Implementation](https://github.com/IntersectMBO/cardano-ledger/blob/0cfbf861cfb456660a7b73281c6fb714a53d40f9/eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxo.hs#L211-L219)

### Discussion points

_No response_

_No response_

### Checklist

yHSJ:
Closes: #887, #889, #891

Introduce more missing predicates, this time around reference script size limits. I acknowledge there is duplicate work here, where we iterate to find the script sizes twice; once at the block level before we call phase one validation per transaction, and once during phase one validation. I decided that this is an acceptable situation, since it is not a particularly expensive, and it is much simpler to reason about than threading the calculation through from either phase one back to the block level, or vice versa.

There are some other options though, we could:
1. While we validate at the block level, we allocate a `Vec<usize>` to track reference script sizes of each transaction, and pass that to phase one validation. The problem with this is that we still double the work of iterating over inputs and looking up outputs, since there are other input validations in phase one.
2. We could keep a running sum during phase one validation, returned to the block validation after phase one completes. This *does* deduplicate the logic, but it means that a block which should fail would require that we do phase one validation for *each* transaction within it, before we can fail. That could expose us to potential attack vectors that aren't viable for other node implementations.
3. We could do a hybrid strategy, where we validate *all* reference script size predicates in block validation. The concern here is that we're mixing responsibilities in a confusing way.

## Summary by CodeRabbit

* **New Features**
  * Added phase-one enforcement of the minimum transaction fee, including Plutus execution-unit and reference-script cost components.
  * Enforced protocol limits for total reference-script size per transaction and per block.
  * Improved execution-unit accounting consistency during validation.
* **Bug Fixes**
  * Transactions now fail with clearer, more specific errors when fees are too low or reference-script size limits are exceeded.
* **Tests**
  * Added new pass/fail fixtures for min-fee and reference-script size boundary/overflow scenarios, including spent-input coverage.


yHSJ:
Actually, with the changes in #821, where we already *need* to track the per-tx sizes, it probably makes sense to go with option 2. That will, at the very least, reduce the times we iterate over transactions and read from our context. Though it is worth considering the *potential* for adverse impact.

crates/amaru-ledger/src/rules/transaction/phase_one/fees.rs
```diff
@@ -60,9 +83,110 @@ where
...
+    let acc = rational_add(acc, term);
+
+    let (mult_num, mult_den) = mult;
+    let next_num = rate_num.saturating_mul(mult_num);
+    let next_den = rate_den.saturating_mul(mult_den);
+    let g = gcd(next_num, next_den);
+    let next_rate = if g > 1 { (next_num / g, next_den / g) } else { (next_num, next_den) };
+
+    sum_tiers(remaining - stride, stride, next_rate, mult, acc)
+}
```
KtorZ: I really dislike the `saturating_` bits here. This is dangerous. Either we consider (and enforce as invariant) that these calculations cannot overflow; or they might and then we resort to using big integers. But the saturating primitives are a red-herring here. 

Also, the tuples of `u64` really scream for using a ratio type (there are good ones in Rust!).

crates/amaru-ledger/tests/evaluate_ledger_states.rs
```diff
@@ -202,6 +202,13 @@ pub mod tests {
                 transaction_index: ix,
             };
 
+            // NOTE: Transaction Size Calculation:
+            //
+            // As noted in block.rs, the transaction size is calcualted from an encoded 3-element list, not including the is_valid byte.
+            // We don't have that here since the fixtures encode individual transactions, instead of blocks.
+            // While the exact bytes aren't the same (the header should be 0x83 instead of 0x84), the is_valid boolean is exactly one byte.
+            // So, by subtracting one, we get the expected value.
+            let tx_size = (tx_bytes.len() - 1) as u64;
```
KtorZ, line 324: Now I just want to have a Mary-encoded transaction in these tests just to make this line fail.

crates/amaru-kernel/src/traits/has_ex_units.rs
```diff
@@ -26,3 +30,12 @@ impl HasExUnits for Block {
             .flat_map(|redeemers| redeemers.iter_unique().map(|(_, ex_units, _)| ex_units))
     }
```
KtorZ, line 31: This could have been rewritten to use `.ex_units()` now available on WitnessSet.

crates/amaru-ledger/src/rules/transaction/phase_one/fees.rs
```diff
@@ -102,18 +200,52 @@ mod tests {
...
                 )
             },
             expected_traces,
         )
     }
+
+    #[test_case(1_000, 25_600, 15, 1, 12, 10 => 15_000; "tier: within first tier (1000 bytes * 15)")]
+    #[test_case(250, 100, 10, 1, 2, 1 => 5_000; "tier: spans 3 tiers (100*10 + 100*20 + 50*40)")]
+    #[test_case(5, 5, 1, 3, 1, 1 => 1; "tier: final floor (floor(5/3) = 1)")]
+    #[test_case(10, 5, 1, 3, 1, 1 => 3; "tier: rational accumulates across tiers (floor(5/3 + 5/3) = 3, not 2+2)")]
```
KtorZ, line 219: Verified those independently against the Haskell code and added a few more ones extracted from their test suite.

KtorZ (approved):

PR #822, opened 2026-05-13
fix: use 'cargo run' for generating and validating trace schemas + revert sccache

KtorZ:
## Summary by CodeRabbit

* **Chores**
  * Streamlined CI and release pipelines by removing redundant steps and simplifying artifact handling.
  * Simplified trace schema generation and validation to run directly without separate prebuilds.
  * Revised caching behavior and defaults to improve build consistency.
  * Adjusted snapshot test flow to gate a timing step off main and relax a strict failure mode.
  * Runner cleanup now focuses on Docker images and unused files.

[Review Change Stack](https://app.coderabbit.ai/change-stack/pragma-org/amaru/pull/822)

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -111,9 +111,6 @@ jobs:
 
           cargo $EXTRA_ARGS $COMMAND $SCOPE --locked --target ${{ matrix.environments.target }}
 
-          # Run doc tests separately as --doc can't be used with --all-targets
-          cargo test --doc
-
       - name: Upload test artifacts on failure
         if: failure()
         uses: actions/upload-artifact@v7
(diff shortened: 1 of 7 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -111,9 +111,6 @@ jobs:
 
           cargo $EXTRA_ARGS $COMMAND $SCOPE --locked --target ${{ matrix.environments.target }}
 
-          # Run doc tests separately as --doc can't be used with --all-targets
-          cargo test --doc
```
jeluard, line 115: Why remove? This is about testing code in doc, not tested by default.

jeluard:
@KtorZ DCO blocking?

.github/workflows/continuous-integration.yml
KtorZ: Doc tests are executed by default 😬 

image

Though I forgot to also modify the testing command to not use `--all-targets` indeed. We run benchmarks separately _anyway_.

KtorZ:
> @KtorZ DCO blocking?

Yes, apparently git pre-commit hooks don't apply to `git revert` 🙄

PR #823, opened 2026-05-14
fix: get the tx_id from the transaction body

etorreborre:
The main fix of this PR is the retrieval of the transaction id used in the mempool and tx submission protocol from the transaction body and not from the hashing of the full transaction CBOR.

Then, in order to make the access to the transaction id more consistent and encapsulated, I defined a `tx_id()` method on `Transaction`.

Notes: 

 - I removed the `dummy` mempool in this PR instead of having to maintain it.
 - I know that #822 is coming but I just committed a small fix to make sure CI passes on this PR.

## Summary by CodeRabbit

* **Refactor**
  * Consolidated transaction identifier handling across the system for improved consistency.
  * Removed unused dummy mempool implementation.
  * Unified transaction ID derivation methods throughout the codebase to use a consistent interface.


crates/amaru-kernel/src/cardano/transaction.rs
```diff
@@ -34,3 +37,55 @@ pub struct Transaction {
...
+    pub fn tx_id(&self) -> TxId {
+        TxId(self.body.id())
+    }
+}
+
+/// Identifier for a transaction. This is the hash of the transaction body bytes.
+/// It encapsulates the TransactionId to be completely opaque to the structure of the
+/// transaction identifier.
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash, PartialOrd, Ord, serde::Serialize, serde::Deserialize)]
+pub struct TxId(TransactionId);
```
KtorZ: Please make a new kernel module 🙏, one module per type. 

Also, let's try to avoid acronyms and short names as much as we can; 👉 `TransactionId`

crates/amaru-kernel/src/cardano/transaction.rs
```diff
@@ -34,3 +37,55 @@ pub struct Transaction {
...
+impl<'b> cbor::Decode<'b, ()> for TxId {
+    fn decode(d: &mut cbor::Decoder<'b>, _ctx: &mut ()) -> Result<Self, cbor::decode::Error> {
+        let hash = Hash::<{ TRANSACTION_BODY }>::decode(d, _ctx)?;
+        Ok(TxId(hash))
+    }
+}
+
+impl Display for TxId {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", hex::encode(self.0.as_slice()))
```
KtorZ: Might want to limit to the first 10-12 chars.

crates/amaru-kernel/src/cardano/transaction.rs
```diff
@@ -34,3 +37,55 @@ pub struct Transaction {
...
+
+pub trait HasTxId {
+    fn tx_id(&self) -> TxId;
+}
+
+impl HasTxId for Transaction {
+    fn tx_id(&self) -> TxId {
+        self.tx_id()
+    }
+}
```
KtorZ: Please move to a separate module under `kernel/src/traits` + rename to `HasTransactionId`

rkuhn (commented):
It looks like shuffling around rather similarly named types in some places, but I fail to see an actual change of computational rules. I’m a bit confused by the intended difference between TxId and TransactionId, which sounds like they should be the same thing to begin with.

crates/amaru-kernel/src/cardano/transaction.rs
etorreborre: There's already a `TransactionId` used as a type alias for the hash. Should I remove that type alias then?

etorreborre:
@rkuhn sorry I should have pointed it. The fix is [here](https://github.com/pragma-org/amaru/pull/823/changes#diff-f89dc72c2e7a70056f87c5199b276dcb1643f36d3316be39db6f9c9791944fd2R43).

crates/amaru-kernel/src/cardano/transaction.rs
etorreborre: @KtorZ ☝️

etorreborre: I pushed a new commit doing that renaming and then renaming the original `TransactionId` to `TransactionIdHash`. However it's not consistent with `PoolId` which is also a hash. Are you ok to rename it to `PoolIdHash` or do you have another naming scheme that you'd prefer?

KtorZ: > TransactionIdHash

😬 ... transaction id is already a hash. So; yeah, we should NOT use a type-alias if we think this deserves to be a proper type with its own set of methods. 

It doesn't make sense to me to have `TransactionId` and `TransactionIdHash` (same for pool).

KtorZ: Looking at the code here though; I don't really see what's specific to `TransactionId` though; everything could be generally handled by the `Hash` (though we don't yet control this type, so can't really extend it).

crates/amaru-kernel/src/cardano/transaction.rs
etorreborre: Done.

rkuhn:
Hmm, I still don't fully understand, but that is probably not a blocking issue.

I agree with removing type aliases to bare Hash types from other crates, we should have a proper type to avoid accidentally mixing up IDs because the hashes happen to have the same length.

PR #825, opened 2026-05-14
feat: defer the validation of a header when the stake distribution is not available yet

etorreborre:
This PR fixes a header validation issue that appeared during testing:
```

┌────────────────────────────── ┬───────────────┬───────────────────────────────────────────────────────────────────────────────────────┐
│           State.              │     Value     │                               Description                                             │
├────────────────────────────── ┼───────────────┼───────────────────────────────────────────────────────────────────────────────────────┤
│ Most recent snapshot on disk  │ 286           │ Taken when the ledger crossed 286 -> 287                                              │
├────────────────────────────── ┼───────────────┼───────────────────────────────────────────────────────────────────────────────────────┤
│ Snapshots on disk             │ 284, 285, 286 │ MIN_LEDGER_SNAPSHOTS = 3 keeps the latest three                                       │
├────────────────────────────── ┼───────────────┼───────────────────────────────────────────────────────────────────────────────────────┤
│ Ledger stable tip.            │ 122_767_427   │ Inside epoch 287. ~ 7000 slots before the 287 / 288 boundary at slot 122_774_400      │
├────────────────────────────── ┼───────────────┼───────────────────────────────────────────────────────────────────────────────────────┤
│ In-memory stake distributions │ 285, 284      │                                                                                       |
├────────────────────────────── ┼───────────────┼───────────────────────────────────────────────────────────────────────────────────────┤
│ Peer's tip                    │ 122_899_069   │ Inside epoch 288. It streamed headers cross the 287 / 288 boundary almost immediately │
└────────────────────────────── ┴───────────────┴───────────────────────────────────────────────────────────────────────────────────────┘
```

In that situation the upstream peer sent us a header that could not be validated and the peer was unduly removed.

This PR fixes this situation by:

 1. Tracking validation errors due to missing stake distributions. The error includes the epoch where the stake distribution will be available.
 2. Keeping peers as pending until the ledger catches up.
 3. Receiving `BlockApplied` events in the `track_peers` stage to know if we have reached an epoch that allows us to validate pending headers.

## Summary by CodeRabbit

* **New Features**
  * Adopt-chain now notifies peer-tracking when a new tip is applied, enabling automatic retries of previously deferred header validations after new blocks arrive.
  * Peer-tracking gains a transient retry queue so deferred validations are re-attempted when applicable.

* **Bug Fixes**
  * Transient header-validation failures no longer cause peer removals; peers are kept and retried, improving chain-sync stability and recovery.

[Review Change Stack](https://app.coderabbit.ai/change-stack/pragma-org/amaru/pull/825?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/crates/amaru-consensus/src/stages/track_peers/mod.rs
+++ b/crates/amaru-consensus/src/stages/track_peers/mod.rs
@@ -17,7 +17,8 @@ mod defer_req_next;
 use std::{collections::BTreeMap, time::Duration};
 
 use amaru_kernel::{
-    BlockHeader, BlockHeight, EraHistory, EraName, IsHeader, ORIGIN_HASH, Peer, Point, Tip, from_cbor_no_leftovers,
+    BlockHeader, BlockHeight, Epoch, EraHistory, EraName, IsHeader, ORIGIN_HASH, Peer, Point, Tip,
+    from_cbor_no_leftovers,
 };
 use amaru_observability::trace_span;
 use amaru_protocols::{
@@ -61,6 +62,10 @@ pub struct TrackPeers {
     /// Lazily populated via [`Effects::stage`](pure_stage::Effects::stage) on first deferred `RequestNext`.
     defer_req_next: StageRef<DeferReqNextMsg>,
     defer_req_next_poll_ms: u64,
+    /// Headers whose validation failed because the ledger had not yet computed the required
+    /// stake distribution. Drained on every `BlockApplied` and re-validated against the
+    /// now-advanced ledger state.
+    pending_validations: Vec<PendingValidation>,
     ledger_applied_block_height: BlockHeight,
     ledger_last_checked_at: Instant,
 }
@@ -71,9 +76,23 @@ struct PerPeer {
     highest: Tip,
 }
 
+/// A header whose validation failed transiently and must be retried once the ledger has
+/// caught up far enough to compute the required stake distribution.
+#[derive(Debug, Clone, PartialEq, serde::Serialize, serde::Deserialize)]
+struct PendingValidation {
+    peer: Peer,
+    handler: StageRef<chainsync::InitiatorMessage>,
+    variant: EraName,
+    header: BlockHeader,
+    tip: Tip,
+    missing_epoch: Epoch,
+}
+
 #[derive(Debug, Clone, PartialEq, serde::Serialize, serde::Deserialize)]
 pub enum TrackPeersMsg {
     FromUpstream(ChainSyncInitiatorMsg),
+    /// Sent by `adopt_chain` after a block has been adopted. This is used to re-validate pending headers.
+    BlockApplied(Tip),
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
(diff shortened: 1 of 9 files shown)


etorreborre:
@coderabbit review

rkuhn:
We should probably discuss this with @KtorZ before deciding how to proceed.

1. There is a mechanism (defer_req_next) in TrackPeers that should prevent fetching headers more than 2160 blocks ahead of the adopted tip. Given the numbers you quote, this mechanism clearly has been broken (it was working when I tried it two weeks ago).

   Is this a reasonable mechanism or does the proper solution need to be aware of epochs?

2. The new proposal tries to validate headers that we don't really want to have, Which I consider a resource usage hazard. We only need to keep chain candidate selection to be a couple hundred blocks ahead of fetching and validation to get decent initial sync speed.

crates/amaru-consensus/src/stages/track_peers/tests.rs
```diff
@@ -514,14 +510,47 @@ fn test_roll_forward_header_validation_failure_removes_peer() {
...
+    let (_running, _guards, mut logs) = setup_with_validation(
+        &prep.rt_handle(),
+        state.clone(),
+        msg,
+        build_store(&[]),
+        Arc::new(TransientHeaderValidation { epoch: amaru_kernel::Epoch::from(286) }),
+    );
+
+    assert!(state.upstream.contains_key(&peer));
+
```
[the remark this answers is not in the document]
etorreborre: Done.

KtorZ:
@rkuhn 

1. Why 2160 here? The number seems arbitrarily chosen to be `k` but that I can't seem to find the motivation for it? 
    
    There's an underlying question of "how far" can we realistically fetch? The answer depends on where we are in the epoch but the *current rule* (though it's about to change with the epoch transition work) i that we calculate the stake distribution of the previous epoch once we are more than `k` deep in the current epoch (i.e. once it's stable to do so). The calculation still happens synchronously at the moment so it's "immediately" available".

    The stake distribution computed in epoch `e` is used for the leader schedule in epoch `e + 1` (so that is the stake of epoch `e - 1` that is effectively used). Soooo.... If we are less than `k` deep in an epoch, we can fetch up to the last block of the next epoch. And if we are more than `k` deep, we can extend that to the one that follows. So that's between 17*k and 27*k from the tip (given that epochs are 10*k long).

3. That seems reasonable and matches my previous observations in other contexts; pipelining more than ~100 requests does not usually improve the situation (sometimes even worsen it).

crates/amaru-consensus/src/stages/track_peers/mod.rs
```diff
@@ -125,6 +164,46 @@ impl TrackPeers {
...
+            mode,
+            eff,
+        )
+        .await;
+    }
+
+    /// Decide whether to pipeline the next `RequestNext` or defer it, based on how far the
+    /// ledger applied tip is behind the header being processed. Refreshes the cached
+    /// `ledger_applied_block_height` if it might be stale and the comparison is close.
+    async fn compute_rollforward_mode(
```
KtorZ, line 185: Why not simply always pipeline, but keep the pipeline bounded?

crates/amaru-ouroboros-traits/src/validators/headers/can_validate_headers.rs
```diff
@@ -59,25 +65,28 @@ impl Serialize for HeaderValidationError {
...
+        Ok(Self::Other(anyhow::anyhow!(s)))
     }
 }
 
 impl PartialEq for HeaderValidationError {
     fn eq(&self, other: &Self) -> bool {
-        self.0.to_string() == other.0.to_string()
+        match (self, other) {
+            (Self::MissingStakeDistribution(a), Self::MissingStakeDistribution(b)) => a == b,
+            (Self::Other(a), Self::Other(b)) => a.to_string() == b.to_string(),
```
KtorZ, line 88: Unrelated but, this implementation looks particularly fishy 😶

KtorZ (commented):
I don't necessarily see problems with the approach of handling ledger errors but, as @rkuhn pointed out; it seems that it's perhaps more reasonable to simply avoid enqueuing requests for blocks that are too far ahead? 

As I pointed out, we can likely go as far as 17*k blocks as far as the stake distribution is concerned; but that's probably already a lot more than the practical limit we would want? 

I believe that the most important element here is that the pipeline is always busy doing something; so having always _a few_ blocks request pipelined seems sufficient given that the ledger generally consumes them at a slower pace than they're fetched? Even just a pipeline of 10 might suffice?

etorreborre:
> Even just a pipeline of 10 might suffice?

That would be the simplest solution. What do you think @rkuhn?

crates/amaru-consensus/src/stages/track_peers/mod.rs
rkuhn: text isn’t perfectly clear: yes, we always pipeline, the decision here is whether there is already space in that pipeline, yet

PR #834, opened 2026-05-18
feat: improved monitoring basic

jeluard:
Make sure `monitoring/basic` reflect latest state:

* forward metrics, spans and logs
* introduced optionals (via profile) `prometheus`, `grafana` and `tempo` support
* introduce optional (via extra conf) forward support (allows to chain  other processes like [otel-ui](https://github.com/jeluard/otel-ui))

## Summary by CodeRabbit

* **New Features**
  * Basic monitoring stack with OTLP gRPC+HTTP endpoints and an additional gRPC port
  * Optional Prometheus metrics collection (prometheus profile)
  * Forwarding of traces/logs to external OTLP endpoints with configurable forward endpoint
  * Local debug export of traces/logs for easier inspection

* **Documentation**
  * Expanded monitoring docs with a “Basic” example and step‑by‑step setup for running and forwarding the stack

[Review Change Stack](https://app.coderabbit.ai/change-stack/pragma-org/amaru/pull/834?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


monitoring/README.md
```diff
@@ -55,9 +56,40 @@ For example `amaru[find_intersection]=trace` will filter all `spans` and `events
...
-A trace is a collection of spans logically connected together. A span represents a unit of work or operation in the system.
+Now prometheus UI is also accessible at `http://localhost:9090`.
+
+### Forwarding traces and logs to another OTLP backend
+
+By default, traces and logs are only printed to the collector's console output. To bridge them to an external OTLP-compatible backend (e.g. Jaeger, Grafana Tempo, a remote collector), use the `docker-compose.forward.yml` override:
+
+```bash
+# default downstream endpoint (localhost:4319)
+docker compose -f docker-compose.yml -f docker-compose.forward.yml up
```
KtorZ, line 200: Seems annoying to have both options here; it's supposed to be a `basic` configuration 😅 ... Why not just make forwarding the default? I don't see in what situation we don't want traces and logs forwarded.

I admit I am almost a little confused about why is this *now* needed when it was working fine before?

jeluard: `basic` doesn't work when traces are emitted (so with `AMARU_WITH_OPEN_TELEMETRY=true`)

Also forwarding only works if there is the right process listening on the other side. Or it isn't happy. 

It looks to me that it would make more sense to have just one set of monitoring conf, that is actually supported and working. Like having prometheus and others available as profiles.

Right now it's more a collection of barely working examples.

KtorZ: I agree. That one set is meant to be the "basic" configuration. 

So if we need forwarding and the right listening services on the other ends, lets configure it in this way. 

I am a bit less attached to the Grafana / Tempo setup, because people who deal with Grafana don't usually need help with these sort of configurations; it would still be nice to ensure it works properly, or remove it from the examples if we can't. There's no point of keeping non working stuff around.

jeluard: Some ideas about having simpler and more resilient monitoring conf:
* just a single folder, `monitoring`
* `docker compose up` is `basic` but with spans and logs support
* some profile can be activated (e.g. `prometheus`, `tempo`)

Additionally I would suggest that we settle on one tool for one task (e.g. `tempo` or `jaeger`) and provide a basic set of metric dashboard. I understand most of us have our own things, would make sense to consolidate, make sure it works and offer some nice default for the curious end user.

cc @etorreborre

KtorZ: I agree with your first points / suggestions.

Regarding:

> Additionally I would suggest that we settle on one tool for one task (e.g. tempo or jaeger)

Both actually serve a different purpose. Grafana/Tempo is what we expect most people will use in production. And it's a must have config example regardless. Jaeger is practical for testing; less heavy than Grafana. Although, if we can get Grafana to works nicely, in particular regarding spans visualisation; then we can possibly ditch Jaeger indeed.

etorreborre (approved):

PR #838, opened 2026-05-19
fix: transaction rollback

### LINKED ISSUE #836: Panic at epoch transition 607 → 608: RocksDB already has an ongoing transaction
### What revision are you using?

Amaru 0.1.0 (0075d34)

### What operating system are you using?

Linux

### What architecture are you using?

x86_64/amd64

### What problem are you having? What should be the expected behaviour?

when running the Amaru-pi app on the transition of epoch 607-608 the process dies, have seen mentions of similar issues but not sure they are exactly the same 

### If possible, provide reproductions steps.

_No response_

### If any, provide logs / error outputs from the system when the problem occurs.

{"timestamp":"2026-05-19T09:28:26.815173Z","level":"INFO","fields":{"message":"enter"},"target":"amaru_ledger::state","span":{"from":"607","into":"608","name":"epoch_transition"}}
amaru::fatal::error
thread 'tokio-runtime-worker' (5706) panicked at crates/amaru-stores/src/rocksdb/mod.rs:315:13:
RocksDB already has an ongoing transaction

### Please confirm that you've already searched issues, CHANGELOG and available documentation.

- [x] I searched, but no luck.

etorreborre:
This PR fixes #836 by:

 - Adding a `Drop` implementation that rolls back an ongoing transaction.
 - Adding a `with_transaction` helper to run code inside a transaction and make sure that we don't forget to commit.

### DIFF
--- a/crates/amaru-stores/src/rocksdb/mod.rs
+++ b/crates/amaru-stores/src/rocksdb/mod.rs
@@ -1148,14 +1148,27 @@ mod tests {
 
     #[test]
     fn dropping_transaction_does_not_leak_ongoing_flag() {
-        let tmp_dir = TempDir::new().expect("failed to create temp dir");
-        let store = RocksDB::empty(&RocksDbConfig::new(tmp_dir.path().into())).expect("failed to open store");
+        let tmp_dir = TempDir::new().unwrap();
+        let store = RocksDB::empty(&RocksDbConfig::new(tmp_dir.path().into())).unwrap();
 
         // Open a transaction and drop it without committing or rolling back.
         drop(store.create_transaction());
 
         // Since the previous transaction is dropped and has been properly cleared, we should be able
         // to create a new one.
+        let _ = store.create_transaction();
+    }
+
+    /// `with_transaction` rolls back the transaction on `Err`.
+    #[test]
+    fn with_transaction_rolls_back_on_err() {
+        let tmp_dir = TempDir::new().unwrap();
+        let store = RocksDB::empty(&RocksDbConfig::new(tmp_dir.path().into())).unwrap();
+
+        let outcome: Result<(), StoreError> = store.with_transaction(|_tx| Err(StoreError::Send));
+        assert!(matches!(outcome, Err(StoreError::Send)));
+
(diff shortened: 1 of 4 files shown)


etorreborre:
@coderabbit review

jeluard:
@etorreborre so there was some missing commit?

etorreborre:
@jeluard rather, in case of an error, there was a missing rollback.

jeluard:
So #836 must have some root error triggering this, right?

yHSJ (approved):
I like this pattern in general, it's a nicer API to work with. I am not 100% confident it will "fix" #836, I think it will just reveal the "more accurate" issue. But that's progress!

crates/amaru-stores/src/rocksdb/mod.rs
```diff
@@ -534,16 +534,28 @@ pub struct RocksDBTransactionalContext<'a> {
     db: Transaction<'a, OptimisticTransactionDB>,
 }
 
+impl Drop for RocksDBTransactionalContext<'_> {
+    fn drop(&mut self) {
```
jeluard, line 538: How does this play with https://github.com/pragma-org/amaru/blob/main/crates/amaru-stores/src/rocksdb/transaction.rs#L51 ?

KtorZ: Had the same question, since this was precisely our approach initially. But looking at the code, I now realize we never drop the OngoingTransaction object actually.... instead of being re-created with every transactions, it is created once at the beginning when the store handle is created! That completely defies the purpose of it. 

So yes; this is likely where we want this Drop instance, instead of the OngoingTransaction object.

etorreborre: @KtorZ do I keep the PR as it is, and merge it, or do you want to do that cleanup too?

yHSJ: I am not Matthias, but I think we should merge this now, since it's an improvement, and we can do some future work to clean up the store and related logic.

PR #844, opened 2026-05-20
Remove VolatileCache

yHSJ:
The `VolatileCache` was a premature optimization. As we are expanding to support more state, we will create nicer abstractions that apply to all the necessary objects.

In the mean time, we should just remove it. This makes the future work easier, and simplifies the current logic.

## Summary by CodeRabbit

* **Refactor**
  * Updated internal ledger state storage and UTxO lookup implementation to reduce architectural complexity.

[Review Change Stack](https://app.coderabbit.ai/change-stack/pragma-org/amaru/pull/844?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/crates/amaru-ledger/src/state/volatile_db.rs
+++ b/crates/amaru-ledger/src/state/volatile_db.rs
@@ -35,18 +35,8 @@ pub const EVENT_TARGET: &str = "amaru::ledger::state::volatile_db";
 // VolatileDB
 // ----------------------------------------------------------------------------
 
-// TODO: Currently, the cache owns data that are also available in the sequence. We could
-// potentially avoid cloning and re-allocation altogether by sharing an allocator and having them
-// both reference from within that allocator (e.g. an arena allocator like bumpalo)
-//
-// Ideally, we would just have the struct be self-referenced, but that isn't possible in Rust and
-// we cannot introduce a lifetime to the VolatileDB (which would bubble up to the State).
-//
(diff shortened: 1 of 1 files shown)


crates/amaru-ledger/src/state/volatile_db.rs
```diff
@@ -68,34 +58,30 @@ impl VolatileDB {
...
+            if let Some(output) = state.state.utxo.produced.get(input) {
+                return Some(output);
+            }
+        }
+        None
     }
 
     pub fn has_consumed_input(&self, input: &TransactionInput) -> bool {
-        self.cache.utxo.consumed.contains(input)
+        self.sequence.iter().any(|state| state.state.utxo.consumed.contains(input))
```
KtorZ, line 73: [binary_search_by](https://doc.rust-lang.org/std/collections/struct.VecDeque.html#method.binary_search_by) ?

yHSJ: I don't think this is applicable? The sequence is ordered by slot, which has no relation on whether or not an input was consumed in that slot (and each state has many inputs, so which do we use as the key?)

KtorZ: Maybe.

KtorZ: Yeah... we don't really have a way to correlate slots to inputs here. My bad; wrong suggestion.

PR #845, opened 2026-05-21
fix: on startup reapply downloaded and valid blocks

etorreborre:
On startup we could have up to `k` blocks that were downloaded and validated but still have a ledger that's lagging behind because it volatile part has been dropped from memory.

This PR reapplies those blocks.

## Summary by CodeRabbit

* **New Features**
  * Improved startup initialization to rebuild the ledger’s volatile state by re-applying stored blocks and ensuring missing blocks are requested or revalidated up to the best tip.

* **Tests**
  * Updated and expanded test harnesses for deterministic ledger-tip scenarios; added tip-aware validation mocks and adjusted tests to reflect the new initialization flow and logging.

* **Chores**
  * Adjusted test wiring and mocks, and updated integration tests to tolerate permission-denied setup errors as no-ops.

[Review Change Stack](https://app.coderabbit.ai/change-stack/pragma-org/amaru/pull/845?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/crates/amaru-consensus/benches/stage_msgs.rs
+++ b/crates/amaru-consensus/benches/stage_msgs.rs
@@ -69,9 +69,9 @@ fn stage_msgs(c: &mut Criterion) {
     let msg: Box<dyn SendData> = Box::new(msg);
     group.bench_function("FetchBlocksMsg::NewTip", |b| b.iter(|| black_box(to_cbor(black_box(&msg)))));
 
-    let msg = FetchBlocksMsg::RecoverStoredBlocks(point.hash());
+    let msg = FetchBlocksMsg::Initialize(point.hash());
     let msg: Box<dyn SendData> = Box::new(msg);
-    group.bench_function("FetchBlocksMsg::RecoverStoredBlocks", |b| b.iter(|| black_box(to_cbor(black_box(&msg)))));
+    group.bench_function("FetchBlocksMsg::Initialize", |b| b.iter(|| black_box(to_cbor(black_box(&msg)))));
 
     let msg = FetchBlocksMsg::Timeout(1000);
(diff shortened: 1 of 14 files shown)


etorreborre:
@coderabbit review

etorreborre:
@rkuhn thanks for having a look to see if that's still relevant.

rkuhn:
Sorry for the delay: I don’t think it is. Upon start, getting the unvalidated header hashes should always be sufficient, I don’t see why we should send the already validated blocks for validation again. They will be applied by `validate_block` anyway as soon as the first fresh block arrives, at least that is the intention, and it has worked in my tests. I’d rather think that re-sending already validated blocks is harmful because the `validate_block` stage should (in the future?) just panic if during replay of already validated blocks it finds one to be invalid (possibly overridable for the SPO if this happens as part of an unintentional hard fork mitigation).

etorreborre:
> They will be applied by validate_block anyway as soon as the first fresh block arrives.

One issue I see with that is that we have to wait for a fresh block to arrive in order to submit transactions. That leaves open a window of time where we were submitting transactions that were consistent with the ledger and its volatile state, then the node restarts to the ledger in its immutable state and now transactions might fail to be submitted until the first block is received.

rkuhn:
I see what you mean. In that case the proper fix is to send a “roll forward” message to the validate_block stage at the end of fetch_blocks initialization, instead of sending the blocks again.

etorreborre:
@rkuhn I actually wonder if we should instead synchronously set the system in a proper state before sending any message asynchronously. This way we can be sure that we start from a really consistent and up to date state.

PR #846, opened 2026-05-21
fix: order the proposals

etorreborre:
This PR fixes an issue found when syncing the `preview` network, where proposals were not returned in the correct order.

## Release Notes

* **Improvements**
  * Governance proposal processing now sorts proposals by their submission sequence.

* **Tests**
  * Added unit tests for proposal processing behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/pragma-org/amaru/pull/846?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/crates/amaru-ledger/src/governance/ratification/proposals_forest.rs
+++ b/crates/amaru-ledger/src/governance/ratification/proposals_forest.rs
@@ -107,6 +107,8 @@ impl ProposalsForest {
     ) -> Result<Self, ProposalsInsertError<ComparableProposalId>> {
         let current_epoch = self.current_epoch;
 
+        proposals.sort_by_key(|(_, row)| row.proposed_in);
+
         proposals.drain(..).try_fold::<_, _, Result<_, ProposalsInsertError<_>>>(&mut self, |forest, (id, row)| {
             // There shouldn't be any invalid proposals left at this point.
             assert!(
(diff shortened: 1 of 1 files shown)


yHSJ (approved):
Thanks for fixing this. This is one of the few places that objects are not ordered lexicographically, it would seem 😅 

I think it may be worth having a `TODO` here to provide a data structure that enforces the ordering on insertion, which may be more efficient than this (depending on the size of the vec and the frequency of calls to `drain`)

KtorZ:
> I think it may be worth having a TODO here to provide a data structure that enforces the ordering on insertion, which may be more efficient than this (depending on the size of the vec and the frequency of calls to drain)

That's more or less the role of the forest; though it states ordering of proposals as a pre-condition: 

https://github.com/pragma-org/amaru/blob/593bdc75e6a98f5959f5032cd6f5c106910fcf6d/crates/amaru-ledger/src/governance/ratification/proposals_forest.rs#L131-L132

Though that is weird here because, this should be the natural order that the database gives us already; the ordering should be unnecessary 🤔. @etorreborre do you have the epoch at which this occurs on Preview?

etorreborre:
@KtorZ unfortunately I haven't saved it.

etorreborre:
@KtorZ I found it!
```
Error processing block at point Specific(
    109036833,
    dd192e49d4dc9d672a9ba545e0a9b42a36b8c82192415405d516a6d45b36d58d
  ):
  BlockValidationError(error when ratifying proposals:
    invalid operation while creating the proposals forest:
    proposal ... transaction_id 2585fa3772a1866ae8e79bc4f19cf094c7815f4e7963e363078c8f573bd6268a, action_index 0
    has an unknown parent
    ... transaction_id 4ac138ad9ffc3021bb6168d042d86becb595ec25db02ae0f142fbc23d10c08da, action_index 0
  )
```

PR #855, opened 2026-05-27
feat: get rid of protocol v9

jeluard:
Amaru doesn't bootstrap from genesis but from well known snapshots. Now that snapshots can be more reliably be created it's been decided to drop v9 support and provide fresher snapshots (current ones are more than a year old).

## Summary by CodeRabbit

* **Features**
  * Protocol advanced to version 10; added preprod snapshots for epochs 270–272.
  * Local snapshot discovery and improved snapshot publishing/manifest merging.

* **Bug Fixes**
  * Removed legacy delegation-tracking; simplified DRep registration/removal flows.
  * Unified certificate handling across protocol versions.

* **Refactor**
  * Simplified governance voting/tally and proposal validation.
  * Stake-distribution and rewards computation now snapshot-driven.

* **Docs/Observability**
  * Removed two legacy trace schema/span entries.

[Review Change Stack](https://app.coderabbit.ai/change-stack/pragma-org/amaru/pull/855?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


crates/amaru-ledger/src/summary/governance.rs
```diff
@@ -96,10 +94,7 @@ impl GovernanceSummary {
 
         let mut dreps = db
             .iter_dreps()?
-            .filter(|(_, dreps::Row { registered_at, previous_deregistration, .. })| {
-                Some(registered_at) > previous_deregistration.as_ref()
-            })
```
KtorZ, line 101: This may still be required actually. It depends how we handle the case in bootstrap. I'll check.

crates/amaru-ledger/src/summary/stake_distribution.rs
```diff
@@ -124,31 +124,9 @@ impl StakeDistribution {
...
-                                //
-                                // 'if since > registered_at'
-                                //
-                                // It's also not correct to gate this condition by protocol version
-                                // because invalid such delegation may pre-exist in the database,
-                                // even when under PROTOCOL_VERSION_10.
-                                //
-                                // So we fallback to checking that no de-registration happened
-                                // post-delegation.
-                                if &Some(since) > previous_deregistration { Some(drep) } else { None }
```
KtorZ, line 150: Same as the previous comment. I am not 100% sure we can fully get rid of it.

PR #857, opened 2026-05-27
Migrate Phase Two Types to Kernel

### LINKED ISSUE #796: Relocate Phase 2 Types to the Kernel
### Abstract

Any types that are not "intermediate representation" should be moved to a phase 2 module with 1 type per file in the kernel.

### Why?

Follows the existing pattern, moving types out of the execution logic. This change also allows us to improve the APIs.

### How?

This work is two fold:
1. Introduce a phase 2 module in the `amaru-kernel`, and create a new file for each type, moving them from the current phase 2 execution logic.
2. Assess/improve each API, update docs, and ensure quality test coverage.

Importantly, this may create some duplication of types (phase 2 specific versions), but that is OK for now. Deduplicating this will be a separate task.

- `cargo doc` must produce docs and an API that you would be happy to use if you ran into it the wild
- type specific logic (not validation logic) should be co-located with the type itself

### Discussion points

_No response_

_No response_

### Checklist

yHSJ:
Fixes #796.

Moving these types to the kernel follows the design of the rest of our codebase (types belong in the kernel, logic operates on imported types). It also allows us to identify duplicate types that can be either merged or removed in favor of the non "phase-two" variants. This PR also adds better documentation to these types, which can be checked with `cargo doc -p amaru-kernel --no-deps --open`. 

## Summary by CodeRabbit

* **New Features**
  * Introduced enhanced transaction context representation with new `TxInfo` and `ScriptContext` types for Plutus script validation.
  * Added support for time range validity intervals and improved UTxO resolution.
  * Expanded public transaction information types including redeemers, withdrawals, datums, and votes with canonical ordering guarantees.

* **Chores**
  * Reorganized public API exports and type hierarchies across kernel modules.
  * Updated redeemer tagging semantics for improved script purpose handling.


crates/amaru-kernel/src/cardano/phase_two/certificate.rs
```diff
@@ -0,0 +1,38 @@
...
+/// Wraps a borrowed [`PallasCertificate`] and [`Deref`]s to it, so it behaves like the
+/// underlying certificate everywhere except that it also carries a [`ProtocolVersion`].
+/// That version is needed because a certificate's `ToPlutusData` encoding is
+/// version-dependent: protocol 9 omits the deposit on the newer registration
+/// certificates, a quirk that protocol 10 fixes.
+#[derive(Debug, Clone)]
+pub struct Certificate<'a> {
+    pub protocol_version: ProtocolVersion,
+    pub certificate: &'a PallasCertificate,
+}
```
KtorZ: I haven't fully made up my mind on lifetimes around kernel types. The three approaches I see being:

- We keep the borrowed types completely separate (as you are just proposing), named after their owned version but behind a different namespace.
- We keep the borrowed types close to their original types, in the same module, but use a naming convention (e.g. BorrowedCertificate<'a>)
- We try to unify both using `Cow`; possibly exporting an aliased version of the type which uses `'static` for the owned version. 

The later seems to be the most "elegant" in the long run; and the one that may reduce the most duplication. But it's a heavy undertaking.

crates/amaru-kernel/src/cardano.rs
```diff
@@ -68,6 +68,7 @@ pub mod network_name;
 pub mod non_zero_int;
 pub mod nonce;
 pub mod peer;
+pub mod phase_two;
```
KtorZ: Not sure about _that_ namespace; given that it's a mixed of types we only find during phase-2 validations, but also simply existing types that are borrowed. Cf my other comment; I think for now I'd more inclined to keep the hierarchy flat and have:

- Completely new type just be introduced as a new type in the top-level kernel (e.g. ScriptContext / ScriptInfo)
- Types borrowed from other simply be re-exported from the same module as their original, with a naming convention "BorrowedX"; so that we keep the definitions close and in-sync.

crates/amaru-kernel/src/cardano/phase_two/certificate.rs
yHSJ: I agree that unifying both using `Cow` is *probably* the nicest solution, beyond just implementing the phase 2 traits on the "original" types, and passing around references of those types (and then `ScriptInfo` and `ScriptContext` have lifetimes, but I think that's acceptable and they could live in a phase two specific module)

crates/amaru-kernel/src/cardano.rs
yHSJ: The reason I went with phase 2 is because they *aren't* simply `BorrowedX`, they're a slightly different representation of the same object in many cases. But maybe it could be `PlutusX` with some note somewhere to clarify what that type is.

crates/amaru-kernel/src/cardano/phase_two/certificate.rs
yHSJ: I'm actually going to take a stab at implementing the traits on the original types, with a timebox.

KtorZ (approved):

PR #858, opened 2026-05-27
chore: general bootstrap improvements

jeluard:
Based on feedback from @KtorZ , improves recent bootstrap changes.

## Summary by CodeRabbit

* **New Features**
  * Optional snapshot-dir to customize snapshot output location.
  * CLI help now documents defaulting behavior for key args.

* **Improvements**
  * More robust resume/chunk inference for downloads.
  * Deterministic epoch ordering and upfront metadata during snapshot builds.
  * db-analyser resolved via PATH for simpler discovery.

* **Behavior Changes**
  * Initial snapshot bootstrap no longer imports DRep delegations.

* **Documentation**
  * Expanded snapshot creation guide with prerequisites and step-by-step workflow.

[Review Change Stack](https://app.coderabbit.ai/change-stack/pragma-org/amaru/pull/858?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


crates/amaru/src/bin/amaru/cmd/bootstrap.rs
KtorZ, line 31: There's a `show_default` and `default` macro fields for clap; So not sure how applicable we can have it here since the actual default is dependent on environment variables.

jeluard: Yes they are used when possible. Here I don't think we can, and we follow the same pattern as used in `run` command.

crates/amaru-ledger/src/bootstrap.rs
```diff
@@ -487,10 +474,6 @@ fn import_dreps(
         iter::empty(),
     )?;
 
-    info!(size = delegations.len(), "dreps delegations");
-
-    transaction.add_drep_delegations(delegations)?;
```
KtorZ, line 492: What's the rationale here? Is this addressing the pre-v10 / post-v10 case for drep delegations. I don't quite remember where the delegations are stored but I recall they're stored in two places and one is redundant indeed.

PR #886, opened 2026-06-05
feat: custom/short-epoch testnet bootstrap & run support

paolino:
Draft — enables bootstrapping and running amaru against custom, short-epoch testnets (e.g. generated db-synthesizer / Antithesis chains) without Koios/Mithril and without a built-in network profile. Surfaced while building lambdasistemi's amaru-bootstrap harness on top of current main.

### create-snapshots — offline / testnet mode
- `--targets-file <json>`: explicit per-epoch targets (epoch/slot/hash/parent_point), bypassing Koios (which only serves mainnet/preprod/preview).
- `--cardano-db-dir <dir>`: use an existing local cardano-node db, bypassing the Mithril download.
Both default to None; public-network behaviour is unchanged.

### run — runtime network parameters
- `--era-history-file` / `--global-parameters-file`: override the built-in per-network era history / global parameters at runtime (required for custom testnets whose epoch length differs from the built-ins).

### bootstrap — testnet tvar snapshots
- Honor the `history.<slot>.<hash>.json` era-history sidecar when importing a node (tvar) snapshot for `Testnet(_)`. Without it, `parse_state_snapshot` hardcodes the current-era epoch size to the network default (86400) and bootstrap fails with `expected epoch 0 ... got N`.

### ledger — short-epoch robustness
- Saturating epoch subtraction in stake-distribution lookups (avoid underflow when epoch < 2).
- Testnet-only stake-distribution forecast in `get_pool` (serve next-leader-schedule lookups before the snapshot materializes).
- Guards: avoid zero-active-stake panic in header validation; guard `ln(1-f)` when active_slot_coeff >= 1.

### Verification
Full downstream harness build-gate green, and a live epoch-120 short-epoch follow cluster (cardano-node block producer + amaru relay) passes: amaru bootstraps from the snapshot and follows across epochs 3→9 (slot 1088) with zero fatal errors.

Happy to split into smaller PRs (create-snapshots / run flags / bootstrap sidecar / ledger) if preferred.

## Summary by CodeRabbit

* **New Features**
  * New CLI flags/env vars: --cardano-node-db (CARDANO_NODE_DB), --snapshot (repeatable), --epoch target semantics, ERA_HISTORY_FILE, GLOBAL_PARAMETERS_FILE, TARGETS_FILE. create-snapshots supports manual snapshot points and may skip downloads if local DB covers targets.

* **Bug Fixes**
  * Prevent epoch underflow and warn when historical epochs are missing.
  * Fix leader/stake checks for zero-active-stake and related edge cases.
  * Improved messaging when stake distributions are unavailable (slot/epoch hints).

* **Refactor**
  * Era history and global parameters centralized in config; bootstrap/import honors local era-history sidecars on Testnet.


crates/amaru-ledger/src/state.rs
```diff
@@ -883,6 +887,28 @@ pub fn initial_stake_distributions(
...
 }
 
+/// Resolve the stake distribution to use for a pool access at `epoch`.
+///
+/// Normally this is an exact match. Generated private testnets, however, can
+/// serve block headers from the next leader schedule before the stable ledger
+/// has materialized that epoch's snapshot; for those (`allow_future_epoch_forecast`)
+/// we fall back to the latest cached distribution when the requested epoch is
+/// strictly newer. Public networks keep exact lookups only.
+fn stake_distribution_for_pool_access(
```
etorreborre: Why not make `stake_distribution_for_pool_access` a function directly on `StakeDistributionObserver`? This way it is fully encapsulated and we can just pass the desired `epoch`.

etorreborre (approved):
That looks good to me, with a suggestion for refactoring

crates/amaru-ledger/src/state.rs
KtorZ: I don't think this is a problem which needs solving _in the ledger_. The consensus / network layer should not be fetching blocks that are too far ahead; I believe this was fixed already in another PR?

KtorZ (approved):

PR #934, opened 2026-06-05
refactor: address mempool review comments

etorreborre:
@rkuhn I had a look at the review comments left on the `mempool` stage. The most relevant was connected to the use of a `MempoolError` that was never triggered (that was premature engineering), so I removed it.

I don't think that the other concerns are valid but I'm happy to discuss if you think that this is not the case.

## Summary by CodeRabbit

* **Refactor**
  * Simplified mempool error handling architecture by removing layered error wrapping in transaction insertion and removal operations.
  * Mempool operations now return results directly, improving code clarity and consistency across the transaction submission pipeline.
  * Enhanced system resilience by preventing mempool failures from prematurely terminating stage operations.

### DIFF
--- a/crates/amaru-consensus/src/stages/mempool/stage.rs
+++ b/crates/amaru-consensus/src/stages/mempool/stage.rs
@@ -15,9 +15,9 @@
 use std::time::Instant;
 
 use amaru_kernel::{Tip, Transaction};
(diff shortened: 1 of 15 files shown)


rkuhn:
Looks good to me, I didn’t check the LLM suggestions in detail: I got also mostly false positives for the other stages as well. It felt like a useful exercise, though, because it triggered me to check a few things more closely, and I thus found some things to improve.

PR #938, opened 2026-06-06
update CLI EDR and add trace span EDR

### LINKED ISSUE #936: Persist tracing decisions from Porto workshop as EDR
### Abstract

We discussed how to structure the emitted OpenTelemetry spans. The resulting flipchart photographs now need to be persisted as text.

### Why?

so that we know what to do

### How?

by writing text

check that the EDR is there

### Discussion points

_No response_

_No response_

### Checklist

### LINKED ISSUE #937: Update CLI EDR with Porto workshop results
### Abstract

We refined the CLI design and consistency rules, so we need to persist these refinements.

### Why?

so that the EDR stays up to date

### How?

by editing the text

review approvals

### Discussion points

_No response_

_No response_

### Checklist

rkuhn:
- **update EDR on CLI design**
- **add EDR on tracing span design**

## Summary by CodeRabbit

* **Documentation**
  * Clarified CLI guidelines: explicit top-level command structure, consistent naming and option conventions, environment-variable fallbacks, standardized option display and documentation, logging/output rules (stdout vs stderr), progress/ shutdown semantics, and a reference for observability env vars.
  * Added tracing span design: guaranteed tracing schema, span naming/level conventions, trace-id propagation rules, and predefined performance and consensus span patterns for system observability.


rkuhn:
fixes #936 #937

engineering-decision-records/019-guidelines-for-writing-cli.md
```diff
@@ -17,37 +15,61 @@ This EDRs defines guidelines to follow when building CLI in the context of the A
 
 ## Decision
 
-<!-- Explain the decision with sufficient details to be self-explanatory. -->
+The top-level CLI command structure is as follows:
 
-<details><summary><strong>1. Options shall have associated environment variables as fallback.</strong></summary>
+- `amaru`: invoking without options or arguments will always display the short help (as for `amaru -h`)
+- `amaru -V`: displays short version information, i.e. only the `<MAX>.<MIN>.<DATE>` version string
```
etorreborre, line 21: Why not `amaru -v`

engineering-decision-records/019-guidelines-for-writing-cli.md
```diff
@@ -17,37 +15,61 @@ This EDRs defines guidelines to follow when building CLI in the context of the A
...
+- `amaru --version`: displays long version information including git commit etc.
+- `amaru -q ...`: raise the minimum log level from `INFO` to `WARN`
```
etorreborre: I thought we said we would avoid short options as much as possible (with `help` and `version` as exceptions). I think that `amaru --quiet` is fine in this case.

engineering-decision-records/019-guidelines-for-writing-cli.md
```diff
@@ -17,37 +15,61 @@ This EDRs defines guidelines to follow when building CLI in the context of the A
...
+
+Many users enjoy using env vars for configuring their deployments.
+The difficulty in documenting environment variables is mitigated by the 1:1 correspondence
+with options and the diligent documentation of those, meaning that `clap` will document the
+environment variables for us.
+
+Each such variable name starts with `AMARU_`. Besides these, industry standards like
+[OpenTelemetry variables][otel-vars] are respected.
+
+> FIXME RK: with this reasoning, `AMARU_LOG` corresponds to a `--log` option while we disrespect the `RUST_LOG` industry standard; do we really want this?
```
etorreborre: Why not treat `RUST_LOG` as `AMARU_LOG` and use `AMARU_LOG` with the highest priority?
Like this:

- People who want to be very consistent with variable names can use `AMARU_LOG` everywhere.
- People who are used to the `RUST_LOG` convention can still use it. 

`NO_COLOR` is also an exception to that rule that we might want to respect.

etorreborre (commented):
@rkuhn I made a first pass on the CLI guidelines.

engineering-decision-records/019-guidelines-for-writing-cli.md
rkuhn: Because `-V` is the established option for getting the version, `-v` is established for verbose mode.

engineering-decision-records/019-guidelines-for-writing-cli.md
rkuhn: `-q` is quite established, but as we won’t have `-v` (because switching on ALL debug logging is typically not what you want), I agree with your assessment, will fix.

engineering-decision-records/019-guidelines-for-writing-cli.md
rkuhn: I fully agree with also respecting `RUST_LOG`, I tried to make this point when @abailly was still here, perhaps your proposal gets us the best of both worlds. @KtorZ what do you think?

`NO_COLOR` is not really established, when I recently researched these settings there was no consensus. Our rules imply that people can use `AMARU_COLOR` instead.

KtorZ: I like Eric's proposal 👍.

rkuhn: okay, changing accordingly

engineering-decision-records/026-tracing-span-design.md
```diff
@@ -0,0 +1,119 @@
+---
+type: design
```
KtorZ:
```suggestion
type: architecture
```

We only have two types: `process` and `architecture`

engineering-decision-records/026-tracing-span-design.md
```diff
@@ -0,0 +1,119 @@
...
+
+> **IMPORTANT NOTICE ON `tracing` LEVELS**
+>
+> The `tracing` infrastructure we use to emit OpenTelemetry spans is built around log levels `TRACE`, `DEBUG`, `INFO`, `WARN`, `ERROR`.
+> Log messages (including spans) from `INFO` upwards are printed to `stderr` by default and are part of the CLI / TUI design; these must not occur with high frequency and be relevant to a human observer.
+> `TRACE` level is only intended for debug builds when actually debugging a specific code module, it is not considered viable to switch on all `TRACE` logging, and it is not considered viable to switch on ANY `TRACE` logging during normal operation; `TRACE` level is removed from release binaries by compile-time options.
+>
+> **For these reasons all nominally emitted trace spans intended for production use MUST be emitted at `DEBUG` level!**
+>
+> Since OpenTelemetry treats logs (i.e. tracing events a.k.a. log messages) as individually important, all debug logging that may occur at high rate MUST be emitted at `TRACE` level.
```
KtorZ, line 50: 👍

engineering-decision-records/026-tracing-span-design.md
```diff
@@ -0,0 +1,119 @@
...
+
+> **TARGET:** `amaru::LEDGER`
+
+All ledger operations (validating headers and blocks, performing reward calculations, etc.) are traced in this category.
+Those activities that are related to a specific block will automatically be associated with the right tracing context via the parent propagation added to `pure-stage` effect handlers.
+Activities spawned from such invocations that are not related to that block or header (like asynchronous rewards calculations) are NOT associated with the calling trace context and instead create new trace IDs.
+
+### Transactions
+
+> **TARGET:** `amaru::TRANSACTION`
```
KtorZ:
```suggestion
> **TARGET:** `amaru::MEMPOOL`
```

?

engineering-decision-records/026-tracing-span-design.md
```diff
@@ -0,0 +1,119 @@
...
+> `TRACE` level is only intended for debug builds when actually debugging a specific code module, it is not considered viable to switch on all `TRACE` logging, and it is not considered viable to switch on ANY `TRACE` logging during normal operation; `TRACE` level is removed from release binaries by compile-time options.
+>
+> **For these reasons all nominally emitted trace spans intended for production use MUST be emitted at `DEBUG` level!**
+>
+> Since OpenTelemetry treats logs (i.e. tracing events a.k.a. log messages) as individually important, all debug logging that may occur at high rate MUST be emitted at `TRACE` level.
+
+### Blockchain
+
+The unit of processing for the chain is one block, which is logically also the unit for tracing.
+We therefore generate a trace ID for each header received from upstream and associate the header hash with that trace ID as an OpenTelemetry property.
```
KtorZ, line 55: Are you suggesting we roll out our own ID generation instead of relying on Tokio's tracing one?

rkuhn: The text only defines when we generate a trace ID and what we use it for, it doesn’t define how we generate the trace ID. And I’m not sure whether the trace ID is generated by `tracing`, as I understand it that ID lives in the OpenTelemetry part of the stack.

engineering-decision-records/026-tracing-span-design.md
rkuhn: My opinion isn’t very strong. My reasoning is that this is about transaction handling and the mempool is an implementation detail. If our audience sees that differently then we should respect the audience where it makes sense. The times we measure are those spent outside the mempool, though.

We’ll very likely want to publish metrics on mempool usage and performance, which I’d see more clearly as `MEMPOOL`.

engineering-decision-records/026-tracing-span-design.md
```diff
@@ -0,0 +1,119 @@
...
+It is important to note that OpenTelemetry traces are not meant to be long-running while peers are long-lived entities.
+We therefore do not associate a single trace ID with each peer but rather generate a new trace ID per peer management activity.
+
+The precise design of spans and their names will be done at a later time.
+
+## Consequences
+
+## Discussion Points
+
+- shall we make the span name globally unique, enforce that in the schema, and select the target automatically based on the name?
```
KtorZ, line 128: > we make the span name globally unique

Yes and no; it depends how we look at it. I find it useful to have cross-targets names (e.g. a name "block-validation" would make sense in both the consensus and ledger targets; and re-using the same name reinforce the connection). 

For names, I would suggest the following scheme:

- three parts, separated by `::`
- 1st field: a category (setup, io, store, ...); the point here is not to use domain nouns, but keep the list small and focused on _the type of work being performed_.
- 2nd field: a function / logical unit (e.g. epoch_transition)
- 3rd field: an operation, as a verb (e.g. create)

Given that spans can be nested; we can consider every logical unit to be nested inside some namespace. Fundamentally, that should more or less follow our function splitting. 

In terms of categories, I have in mind the following (initial) "taxonomy":

- `setup`: process startup, one-time wiring, bootstrap, and background initialization work performed to make a subsystem ready.
- `store`: reads, writes, migrations, snapshots, and other operations against persistent storage backends.
- `io`: interactions with external systems other than the storage layer, such as network traffic, filesystem access, subprocesses, and remote fetches.
- `state`: logical state management, including applying changes, rolling back, long-running calculations, advancing, reconciling, or bookkeeping for internal state.
- `check`: checks that verify correctness, conformance, integrity, or authorization of inputs, state transitions, or protocol artifacts.
- `cli`: for command messages that don't really fit anywhere else 

 Few examples to make this concrete:

```
setup.initial_stake_distribution.compute
setup.ledger_store.connect
setup.tip.resolve

store.epoch_transition.flush
store.utxo.put
store.chain.migrate

io.tcp_connection.accept
io.koios.query
io.snapshot.read

state.volatile.push
state.peers.register
state.chain.select

check.header.validate
check.vrf_proof.compare
check.block.parse

cli.bootstrap.begin
cli.command.summarize
cli.epoch.reset
```

We can extend the list if needed; but I'd like us to try to remain succinct and also, "stable" in the names here. The _message_, however, should be free form and more descriptive although we should avoid as much as possible interpolations in the message and prefer structured fields.

engineering-decision-records/026-tracing-span-design.md
KtorZ: It's just that given transactions are _everywhere_, I don't find the target very useful as-is. Especially given that what's described below seems to pertain to the mempool. 

Also, I don't think the mempool is an implementation detail; it is a well-known component. What's left to implementation is how the mempool works and make choices; but fundamentally, it is as much of a component as the ledger.

engineering-decision-records/026-tracing-span-design.md
rkuhn: That is more structured than I expected, but it doesn’t sound too bad. I liked your proposal of using “`.`” to separate the name components, and I corrected your examples to match that (I think we shouldn’t use “`::`” because it has other connotations).

EDIT: I just noticed that you had changed the dot to double-colon in an edit that I had somehow not seen / mistaken. Sorry about that, in that case I wouldn’t have edited your post. I still like the dot better, though.

Shall we try this on for size or shall I immediately update the EDR?

engineering-decision-records/026-tracing-span-design.md
rkuhn: okay, then that’s settled

engineering-decision-records/026-tracing-span-design.md
rkuhn: @KtorZ while trying to figure out the necessary changes I see that the structure you propose doesn’t apply to the spans at the abstraction / granularity level named in the EDR. For each consensus stage invocation we need a span, and the only useful name of that span is the stage name, none of the categories apply. Similarly, the two spans defined specifically to cover the Cardano network health monitoring are not of the structure or nature you describe. It seems to me that your proposal only applies to more detailed code-focused tracing.

KtorZ: > I just noticed that you had changed the dot to double-colon in an edit that I had somehow not seen / mistaken. Sorry about that, in that case I wouldn’t have edited your post. I still like the dot better, though.

Well, that's my bad; I originally posted with the `.` indeed; then realized that this would be annoying for people filtering due to the discrepancy with the target. It's just going to be annoying to have `::` and `.`; overall I think I do prefer the `.` (it's less rust-tainted, more usually seen in the wild, feels like field access etc...). 

So we can settle on the dot, but only if we also use a dot for targets. Consistency shall prevail IMO.

> Shall we try this on for size or shall I immediately update the EDR?

If we like it, then I suggest to add it to the EDR and see how it goes?

> and the only useful name of that span is the stage name, none of the categories apply.

I'd argue that what you call "name" here, should really be a "target"? In my original proposal on Discord, this is what I suggested; each stage to be a target. I believe it is the right split?

> the two spans defined specifically to cover the Cardano network health monitoring are not of the structure or nature you describe.

We might need a new category 🤓! Maybe `monitoring` or `observability` (it's getting meta now!). Which ones are you referring to exactly? The "keep alive" ones?

KtorZ:
@rkuhn I forgot one rule for the command-line building related to named arguments: no "value type" prefix/suffix/infix whatsoever; we have the value name for these already. So, 

don't:

```
--ledger-dir <STRING>
--config-file <STRING>
--timeout-seconds <UINT>
```
do:

```
--ledger <DIR>
--config <FILEPATH>
--timeout <SECONDS>
```

rkuhn:
@KtorZ I updated with the changes you suggested, with slight adaptations (e.g. a `perf` category for trace names). Good to go now?

KtorZ:
> @KtorZ I updated with the changes you suggested, with slight adaptations (e.g. a `perf` category for trace names). Good to go now?

Do we keep the targets are colon-separated or do we align this with dot like for the name? I really sense that having a discrepancy there will be annoying when filtering. 

At the same time; given that they are semantically different fields, not being consistent may help readability overall. So i am not sure. 

I just want to confirm it's not an oversight and that it's a conscious choice to go with `::` for targets?

rkuhn:
Yes, it is a conscious choice to stick with established practice: the TARGET has Rust path syntax by default, which is absent from all other fields to avoid visual overlap and ease reading.

PR #940, opened 2026-06-08
feat: use backbone peer as default peer

### LINKED ISSUE #766: Add an public bootstrap peer option to help onboarding
### Abstract

The top-level onboarding flow in `README.md` currently assumes the user already has access to a synced local `cardano-node`. Add an alternative path for users who only want to try Amaru quickly by documenting a public bootstrap peer or relay endpoint, together with an explicit disclaimer about trust, availability, and performance assumptions.

### Why?

The current getting-started flow is optimized for contributors who already run Cardano infrastructure. That is a poor default for evaluators, new users, and developers who want to validate the project with minimal setup. Requiring a locally synced node adds hours of waiting and extra operational work before the first successful run.

### How?

Update the top-level `README.md` "Running" section to present two paths:
- A "local synced cardano-node" path for production-like testing (with a section explaining how to use Mithril to bootstrap faster).
- A "quick experiment" path that points to a public bootstrap peer or relay.

Document the exact environment variable to set (`AMARU_PEER_ADDRESS`) and explain the trade-offs of using a public peer. 

A fresh user can follow the documented path from the top-level README without reading internal scripts first.

### Discussion points

- We agreed that the right approach to this was to add meaningful default values to the CLI that can be overridden if necessary. 

_No response_

### Checklist

jeluard:
fixes https://github.com/pragma-org/amaru/issues/766 by using a network specific backbone peer as default peer.

## Summary by CodeRabbit

* **New Features**
  * The `run` command now automatically uses network-specific default bootstrap peers when no peer addresses are provided.
  * Added built-in defaults for Mainnet, Preprod, and Preview networks.

* **Documentation**
  * Updated CLI help text to reflect that `--peer-address` is optional and will default per-network when omitted.

### DIFF
--- a/crates/amaru/src/bin/amaru/cmd/run.rs
+++ b/crates/amaru/src/bin/amaru/cmd/run.rs
@@ -19,8 +19,8 @@ use std::{
 };
 
 use amaru::{
-    DEFAULT_DOWNSTREAM_PEERS, DEFAULT_LISTEN_ADDRESS, DEFAULT_NETWORK, DEFAULT_PEER_ADDRESS, DEFAULT_UPSTREAM_PEERS,
-    default_chain_dir, default_ledger_dir,
+    DEFAULT_DOWNSTREAM_PEERS, DEFAULT_LISTEN_ADDRESS, DEFAULT_NETWORK, DEFAULT_UPSTREAM_PEERS, default_chain_dir,
+    default_ledger_dir, default_peer_for_network,
     metrics::track_system_metrics,
     stages::{
         build_node::build_and_run_node,
@@ -138,12 +138,11 @@ pub struct Args {
     ///
     /// This option can be specified multiple times to connect to multiple peers.
     ///
-    /// At least one peer address must be specified.
+    /// If not specified, defaults to the network-specific bootstrap peer.
     #[arg(
         long,
         value_name = amaru::value_names::ENDPOINT,
         env = amaru::env_vars::PEER_ADDRESS,
-        default_value = DEFAULT_PEER_ADDRESS,
         action = ArgAction::Append,
         value_delimiter = ',',
         num_args(0..),
@@ -299,6 +298,13 @@ fn parse_args(args: Args) -> Result<Config, Box<dyn std::error::Error>> {
 
     let chain_dir = args.chain_dir.unwrap_or_else(|| default_chain_dir(network).into());
 
+    // Use network-specific default peer if no peer-address was provided
+    let peer_address = if args.peer_address.is_empty() {
+        vec![default_peer_for_network(network).to_string()]
+    } else {
+        args.peer_address
+    };
+
     let (trace_buffer_min_entries, trace_buffer_max_size) = match args.trace_buffer.as_deref() {
         None => (0usize, 0usize),
         Some(s) => parse_trace_buffer_limits(s)?,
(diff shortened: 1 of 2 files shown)


crates/amaru/src/lib.rs
```diff
@@ -192,6 +212,9 @@ pub mod env_vars {
     /// --peer-address
     pub const PEER_ADDRESS: &str = "AMARU_PEER_ADDRESS";
 
+    /// --append-default-peers
```
etorreborre: ~Shouldn't that be singular? `--append-default-peer`.~ I guess that's because the option can be repeated? Still that's inconsistent with `--peer-address` that is singular.

That being said, why not have a one default peer be the default value for `--peer-address`? When do we want both a localhost peer and a default one?

etorreborre (approved):
☮️

crates/amaru/src/bin/amaru/cmd/run.rs
```diff
@@ -138,12 +138,11 @@ pub struct Args {
     ///
     /// This option can be specified multiple times to connect to multiple peers.
     ///
-    /// At least one peer address must be specified.
+    /// If not specified, defaults to the network-specific bootstrap peer.
     #[arg(
         long,
         value_name = amaru::value_names::ENDPOINT,
         env = amaru::env_vars::PEER_ADDRESS,
-        default_value = DEFAULT_PEER_ADDRESS,
```
rkuhn, line 146: This is going in the wrong direction: we should document all the default values in `clap` so that they are visible to users. I thought we had written that down in the Porto Accords, but so far it is missing and I’ll amend the EDR.

crates/amaru/src/lib.rs
```diff
@@ -43,6 +43,15 @@ pub const DEFAULT_NETWORK: NetworkName = NetworkName::Preprod;
 
 pub const DEFAULT_PEER_ADDRESS: &str = "127.0.0.1:3001";
 
+/// Default public bootstrap peer for mainnet.
+pub const MAINNET_DEFAULT_PEER_ADDRESS: &str = "backbone.cardano.iog.io:3001";
```
rkuhn, line 47: @KtorZ This does not match the defaults put into the systemd files in #945, which one(s) should we choose?

crates/amaru/src/lib.rs
```diff
@@ -43,6 +43,15 @@ pub const DEFAULT_NETWORK: NetworkName = NetworkName::Preprod;
...
+
+/// Default public bootstrap peer for preprod.
+pub const PREPROD_DEFAULT_PEER_ADDRESS: &str = "preprod-node.play.dev.cardano.org:3001";
+
+/// Default public bootstrap peer for preview.
+pub const PREVIEW_DEFAULT_PEER_ADDRESS: &str = "preview-node.play.dev.cardano.org:3001";
+
 /// Default address to listen on for incoming connections.
 pub const DEFAULT_LISTEN_ADDRESS: &str = "0.0.0.0:3000";
```
rkuhn, line 56: Why do we listen on `3000` when the default nodes are all on `3001`?

crates/amaru/src/bin/amaru/cmd/run.rs
jeluard: I don't think this can be achieved when it depends on another argument?

rkuhn: I see, then the default should be documented in the long help text.

crates/amaru/src/lib.rs
KtorZ: Well, I avoid depending on things that we don't pay for; just out of politeness :) 

Any are valid actually. I rather pick the CF backbone since I know who I can reach out to if something goes down there.

crates/amaru/src/lib.rs
KtorZ: cardano-node does not have defaults actually; and we see both use of 3000 and 3001 in the wild but the convention seems to have settled on 3001 indeed:

- https://github.com/IntersectMBO/cardano-node/blob/390273284c76cc182cdd22b142e8a4df63edbcd3/nix/docker/README.md?plain=1#L432-L439
- https://github.com/IntersectMBO/cardano-node/blob/390273284c76cc182cdd22b142e8a4df63edbcd3/nix/nixos/cardano-node-service.nix#L539-L545
- https://github.com/IntersectMBO/cardano-node/blob/390273284c76cc182cdd22b142e8a4df63edbcd3/configuration/cardano/mainnet-topology.json#L2-L15
- https://developers.cardano.org/docs/get-started/infrastructure/node/running-cardano/#running-the-node

rkuhn: Then let’s change that.

PR #943, opened 2026-06-09
fix: remove optimistic transactions

etorreborre:
This PR removes the need for dealing with optimistic transactions inside RockDB.

The consensus stages are designed so that concurrent writes the same type of data cannot occur so there is no need to retry a failed optimistic transaction. However this PR used batched writes to make sure that related writes to the database are still made atomically.

There is also a small fix included to use a snapshot for `find_best_candidate`.

## Summary by CodeRabbit

* **Refactor**
  * Consensus DB now uses a unified store wrapper and defaults to a simpler DB backend; transaction APIs replaced by atomic batch writes for more predictable updates.

* **New Features**
  * Migration tooling and CLI updated to run migrations through the store wrapper, improving upgrade reliability and version handling.

* **Bug Fixes**
  * Snapshot/read and fork/traversal logic adjusted to reduce edge-case inconsistencies when reading, migrating, and selecting best candidates.

### DIFF
--- a/crates/amaru-stores/src/rocksdb/consensus/migration.rs
+++ b/crates/amaru-stores/src/rocksdb/consensus/migration.rs
@@ -89,6 +89,7 @@ fn migrate_to_v2(store: &RocksDBStore<DB>) -> Result<(), StoreError> {
     set_version(store, 2)
 }
 
+#[expect(clippy::expect_used)]
 fn migrate_to_v3(store: &RocksDBStore<DB>) -> Result<(), StoreError> {
     // the reason is that v3 stores the block validation result, which cannot be derived from the v2 DB without
     // running the consensus algorithm and ledger validation. previously, blocks were stored before validation,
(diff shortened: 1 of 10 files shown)


etorreborre:
@coderabbit review

crates/amaru-stores/src/rocksdb/consensus/migration.rs
```diff
@@ -42,121 +45,70 @@ static MIGRATIONS: [fn(&OptimisticTransactionDB) -> Result<(), StoreError>; CHAI
...
-    let bytes = db.get_pinned(BEST_CHAIN_PREFIX).ok().flatten()?;
-    if bytes.len() == hash::size::HEADER { Some(Hash::from(bytes.as_ref())) } else { None }
-}
-
-fn get_anchor_hash(db: &OptimisticTransactionDB) -> Option<HeaderHash> {
-    let bytes = db.get_pinned(ANCHOR_PREFIX).ok().flatten()?;
-    if bytes.len() == hash::size::HEADER { Some(Hash::from(bytes.as_ref())) } else { None }
-}
+    let original_best_chain_point = store.get_best_chain_tip().point();
+    let anchor_point = store.load_header(&store.get_anchor_hash()).map(|h| h.point()).unwrap_or(Point::Origin);
```
rkuhn: This should also be stored as anchor_hash.

rkuhn (approved):
Looking good! (I hope I didn’t overlook a detail … )

crates/amaru-stores/src/rocksdb/consensus/migration.rs
etorreborre: I didn't see the point. We load a header from its hash, retrieved from `get_anchor_hash`, then get the hash again to store it as `anchor_hash`. How can they be different? And if they are, doesn't that mean that the db is seriously corrupted?

rkuhn: It was indeed an anti-corruption measure. Perhaps an even better response is to print a big fat error if the header for the anchor hash cannot be loaded.

etorreborre: Done

PR #946, opened 2026-06-10
fix: process the case of a registered pool that hasn't produced any block yet

etorreborre:
When checking the operational certificate, if a pool has not produced any block yet, we consider that its `latest_opcert_sequence_number`, which leads us to expect either `0` or `1` as the first opcert sequence number.

Note: 

[the corresponding Haskell code](https://github.com/IntersectMBO/ouroboros-consensus/blob/main/ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos.hs#L644-L645) mixes up 2 concerns. It considers the possibility of the pool to be missing altogether when checking the certificate (the `Nothing` case [here](https://github.com/IntersectMBO/ouroboros-consensus/blob/main/ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos.hs#L663)).

This cannot happen with `amaru` since we perform a check for the pool presence _before_ checking for its certificate (and return `AssertHeaderError::UnknownPool` if that's the case).

## Tests

I took the opportunity of that PR to extend the existing header test cases in order to support the issue above + some additional cases for:

 - `AssertHeaderError::VrfProof` (5 sub-cases).
 - `AssertHeaderError::UnknownPool`.
 - `AssertHeaderError::PoolError`.

## Miscellaneous

The file `mainnet_blockheader_10817298.cbor` seems to be a left-over from some other testing activity so I removed it.

## Summary by CodeRabbit

* **Bug Fixes / Refactor**
  * Operational-certificate sequence validation now treats missing “latest operational certificate sequence number” as zero and applies direct numeric range checks for more consistent outcomes.

* **Tests**
  * Reworked header validation tests to use a new fixture-driven harness with updated expected-result handling and deserialization logic.

* **Chores**
  * Reduced unused development dependencies and expanded ignore rules for generated snapshot and backup artifacts.


etorreborre:
@KtorZ @yHSJ  @jonathanlim222 I actually had to relax the implementation temporarily because the end to end tests were failing.
That's because we don't track opcerts as we should. I created a follow-up issue for this piece of work: https://github.com/pragma-org/amaru/issues/950.
Please tell me if I misunderstood anything.

crates/amaru-ouroboros/src/praos/header.rs
```diff
@@ -407,20 +408,16 @@ impl AssertOperationalCertificateError {
...
-                // to have no opcert available?
-            }
+        if declared_sequence_number < latest_sequence_number {
+            return Err(Self::SequenceNumberTooSmall { declared_sequence_number, latest_sequence_number });
+        }
+
+        // FIXME: `latest_sequence_number == 0` currently means "no data tracked" in Amaru, since
+        // ocert_counters isn't populated from chain history yet.
+        // Enforcing an `n <= m + 1` upper bound would reject any pool with a real counter > 1.
+        // For now we skip the upper-bound check in that case
```
KtorZ: This is probably somewhere in the snapshots we get from the Haskell node. Right now we don't track operational certificates at all; `.latest_opcert_sequence_number` always return `0` I believe. 

Note that the Haskell uses `0` as a default for pools that haven't produced blocks yet:

https://github.com/IntersectMBO/cardano-ledger/blob/cd8b7fab8365c5376c4ca2929cfb455d31f4fde6/libs/cardano-protocol-tpraos/src/Cardano/Protocol/TPraos/OCert.hs#L76

KtorZ:
@etorreborre I believe this PR should go the extra mile of:

1. implementing the `latest_opcert_sequence_number` method correctly (meaning, keeping track of the op certs somewhere)
2. resolving the initial opcerts from the snapshots

So that we can fully close the topic.

etorreborre:
@KtorZ I was planning on stacking another PR on top but I'll just add some commits then

etorreborre:
Closed in favor of #965 and #966

PR #951, opened 2026-06-11
feat: workflow to create and publish bootstrap snapshots

### LINKED ISSUE #872: Automate the creation of amaru snapshots
### Abstract

We need to have a command and an environment to create amaru snapshots on demand then upload them on Cloudflare

### Why?

In order to be able to quickly synchronize an `amaru` node from snapshots.

### How?

There is already a command downloading snapshots from mithril and bootstrapping the database from those snapshots. We need to have an integrated command that will also upload the snapshots to Cloudflare for all environments: `preview`, `preprod` and `mainnet`. `mainnet` requires a large amount of disk space and @rkuhn graciously offered to run this process on demand on a Hezner machine when required.

It should be possible to download a recent snapshot to bootstrap an `amaru` node.

### Discussion points

_No response_

_No response_

### Checklist

mmahut:
Adds a manually dispatched workflow that creates the three-epoch bootstrap snapshot set, uploads it to R2, verifies it by running a full amaru bootstrap against the published archives, and opens a PR updating snapshots.json.

Fixes https://github.com/pragma-org/amaru/issues/872

Screenshot 2026-06-11 at 15 03 52

Before merging, one-time setup needed in this repo:

 - Setting up [secret variables](https://github.com/mmahut/amaru/blob/mmahut/snaps/docs/PUBLISHING_SNAPSHOTS.md#secrets-and-variables) for the Github actions
  - Settings → Actions → General → enable "Allow GitHub Actions to create and approve pull requests"
  - A [self-hosted runner with the snapshots label](https://github.com/mmahut/amaru/blob/mmahut/snaps/docs/PUBLISHING_SNAPSHOTS.md#runner-setup)
  
As a demo, this has ran on my fork:

- The Github [action run itself](https://github.com/mmahut/amaru/actions/runs/27345520522)
- The [PR it has opened](https://github.com/mmahut/amaru/pull/2)

Also:

- Fixed chunk_for_slot to derive the immutable chunk size from the network's security parameter instead of the hardcoded mainnet value (21600 slots), which made create-snapshots skip needed Mithril downloads on preview (4320 slots per chunk).
- Fixed the publish script's auto-epoch selection to use the third-latest epoch metadata file, given the latest one is the end of a snapshot set, not its start.

## Summary by CodeRabbit

* **New Features**
  * Added a GitHub Actions workflow to generate, optionally smoke-test, verify, and publish bootstrap snapshot archives to R2/S3 with manual epoch selection and automatic PR creation when `snapshots.json` changes.
* **Improvements**
  * Snapshot chunking/resume calculations are now network-aware for consistent immutable chunk mapping.
  * Snapshot publishing now selects a 3-epoch boundary from available epoch metadata.
  * Public download links support `PUBLIC_URL_BASE` override with URL validation.
* **Documentation**
  * Added an end-to-end guide for publishing bootstrap snapshots, including required runner setup and manual execution steps.

### DIFF
--- a/.github/workflows/publish-bootstrap-snapshots.yml
+++ b/.github/workflows/publish-bootstrap-snapshots.yml
@@ -1,0 +1,225 @@
+name: Publish Bootstrap Snapshots
+
+# See docs/PUBLISHING_SNAPSHOTS.md for runner provisioning and operations.
+
+on:
+  workflow_dispatch:
+    inputs:
+      network:
+        description: "Network to publish snapshots for"
+        required: true
+        type: choice
+        options:
+          - preprod
+          - preview
+          - mainnet
+          - all
+        default: preprod
+      epoch:
+        description: "Start epoch of the three-epoch snapshot set (defaults to the last three completed epochs)"
+        required: false
+        type: string
+        default: ""
+      skip_verification:
+        description: "Skip the bootstrap smoke test against the published snapshots"
+        required: false
+        type: boolean
+        default: false
+      cardano_node_version:
+        description: "cardano-node release tag to download db-analyser from (must be a published release tag, e.g. 11.0.1)"
+        required: false
+        type: string
+        default: "11.0.1"
+
+permissions:
+  contents: write
+  pull-requests: write
+  actions: write
+
+env:
+  CARGO_TERM_COLOR: always
+  RUSTUP_LOG: error
+  CARDANO_NODE_VERSION: ${{ inputs.cardano_node_version }}
+
+jobs:
+  publish:
+    name: Create & publish ${{ matrix.network }} bootstrap snapshots
+    runs-on: [self-hosted, snapshots]
+    timeout-minutes: 2880
+    strategy:
+      fail-fast: false
+      max-parallel: 1
+      matrix:
+        network: ${{ inputs.network == 'all' && fromJSON('["preview", "preprod", "mainnet"]') || fromJSON(format('["{0}"]', inputs.network)) }}
+    concurrency:
+      group: publish-snapshots-${{ matrix.network }}
+      cancel-in-progress: false
+    env:
+      AMARU_NETWORK: ${{ matrix.network }}
+      INPUT_EPOCH: ${{ inputs.epoch }}
+      MANIFEST: crates/amaru/config/bootstrap/${{ matrix.network }}/snapshots.json
+
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v6.0.2
+        with:
+          persist-credentials: false
+
+      - name: Preflight checks
+        shell: bash
+        run: |
+          set -euo pipefail
+          if [ -n "$INPUT_EPOCH" ]; then
+            if ! [[ "$INPUT_EPOCH" =~ ^[0-9]+$ ]] || [ "$INPUT_EPOCH" -eq 0 ]; then
+              echo "::error::epoch must be a positive integer (non-zero), got '$INPUT_EPOCH'" >&2
+              exit 1
+            fi
+          fi
+
+      - name: Resolve persistent cache directories
+        shell: bash
+        run: |
+          set -euo pipefail
+          CACHE_ROOT="${{ vars.SNAPSHOTS_CACHE_DIR || '' }}"
+          if [ -z "$CACHE_ROOT" ]; then
+            CACHE_ROOT="$HOME/amaru-snapshots"
+          fi
+          mkdir -p "$CACHE_ROOT/$AMARU_NETWORK"
+          echo "CACHE_ROOT=$CACHE_ROOT" >> "$GITHUB_ENV"
+          echo "AMARU_DIST_DIR=$CACHE_ROOT/$AMARU_NETWORK/dist" >> "$GITHUB_ENV"
+          echo "AMARU_SNAPSHOTS_DIR=$CACHE_ROOT/$AMARU_NETWORK/snapshots" >> "$GITHUB_ENV"
+
+      - name: Install db-analyser
+        shell: bash
+        run: |
+          set -euo pipefail
+          case "$(uname -s)" in
+            Linux) os=linux ;;
+            Darwin) os=macos ;;
+            *) echo "::error::unsupported operating system: $(uname -s)" >&2; exit 1 ;;
+          esac
+          case "$(uname -m)" in
+            x86_64 | amd64) arch=amd64 ;;
+            arm64 | aarch64) arch=arm64 ;;
+            *) echo "::error::unsupported architecture: $(uname -m)" >&2; exit 1 ;;
+          esac
+
+          asset="cardano-node-$CARDANO_NODE_VERSION-$os-$arch.tar.gz"
+          base_url="https://github.com/IntersectMBO/cardano-node/releases/download/$CARDANO_NODE_VERSION"
+          tmp="$(mktemp -d)"
+          echo "DB_ANALYSER_TMP=$tmp" >> "$GITHUB_ENV"
+          echo "downloading $base_url/$asset"
+          curl -fsSL -o "$tmp/$asset" "$base_url/$asset"
+          curl -fsSL -o "$tmp/sums.txt" "$base_url/cardano-node-$CARDANO_NODE_VERSION-sha256sums.txt"
+          (cd "$tmp" && grep " $asset\$" sums.txt | (sha256sum -c - 2>/dev/null || shasum -a 256 -c -))
+          tar -xzf "$tmp/$asset" -C "$tmp" --strip-components 1
+          rm "$tmp/$asset" "$tmp/sums.txt"
+          echo "$tmp/bin" >> "$GITHUB_PATH"
+
+      - name: Build amaru
+        shell: bash
+        run: |
+          cargo build --release --bin amaru
+
+      - name: Create snapshots
+        shell: bash
+        run: |
+          set -euo pipefail
+          ./target/release/amaru create-snapshots \
+            --network "$AMARU_NETWORK" \
+            --force \
+            ${INPUT_EPOCH:+--epoch "$INPUT_EPOCH"}
+
+      - name: Publish snapshots to R2 and update manifest
+        shell: bash
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.R2_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.R2_SECRET_ACCESS_KEY }}
+          AWS_DEFAULT_REGION: auto
+          ENDPOINT: ${{ secrets.S3_ENDPOINT }}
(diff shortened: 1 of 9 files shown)


crates/amaru-mithril/src/lib.rs
```diff
@@ -377,12 +377,16 @@ pub fn first_missing_immutable_chunk(immutable_dir: &Path) -> Result<u64, io::Er
     }
 }
 
-pub fn chunk_for_slot(slot: u64) -> u64 {
-    slot / 21_600
+pub fn chunk_for_slot(network: NetworkName, slot: u64) -> u64 {
+    // Immutable chunks span one Byron epoch, i.e. 10 * k slots: 21600 on
```
jeluard, line 381: Did you hit some issue that made you consider this fix?

.github/workflows/publish-bootstrap-snapshots.yml
```diff
@@ -0,0 +1,242 @@
...
+    concurrency:
+      group: publish-snapshots-${{ matrix.network }}
+      cancel-in-progress: false
+    env:
+      AMARU_NETWORK: ${{ matrix.network }}
+      INPUT_EPOCH: ${{ inputs.epoch }}
+      MANIFEST: crates/amaru/config/bootstrap/${{ matrix.network }}/snapshots.json
+
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v6.0.2
```
jeluard, line 63: ?

.github/workflows/publish-bootstrap-snapshots.yml
```diff
@@ -0,0 +1,242 @@
...
+        with:
+          persist-credentials: false
+
+      - name: Preflight checks
+        shell: bash
+        run: |
+          set -euo pipefail
+          for tool in cargo aws jq curl gh git; do
```
jeluard: Not needed?

.github/workflows/publish-bootstrap-snapshots.yml
```diff
@@ -0,0 +1,242 @@
...
+          mkdir -p "$CACHE_ROOT/$AMARU_NETWORK"
+          echo "CACHE_ROOT=$CACHE_ROOT" >> "$GITHUB_ENV"
+          echo "AMARU_DIST_DIR=$CACHE_ROOT/$AMARU_NETWORK/dist" >> "$GITHUB_ENV"
+          echo "AMARU_SNAPSHOTS_DIR=$CACHE_ROOT/$AMARU_NETWORK/snapshots" >> "$GITHUB_ENV"
+
+      - name: Install db-analyser if missing
+        shell: bash
+        run: |
+          set -euo pipefail
+          if command -v db-analyser >/dev/null 2>&1; then
```
jeluard: How could it be not missing?

.github/workflows/publish-bootstrap-snapshots.yml
```diff
@@ -0,0 +1,242 @@
...
+          EPOCHS="$(jq -r '[.[-3:][].epoch | tostring] | join(", ")' "$MANIFEST")"
+          TITLE="chore: update $AMARU_NETWORK bootstrap snapshots (epochs $EPOCHS)"
+          BRANCH="snapshots/$AMARU_NETWORK-$(date -u +%Y%m%d%H%M%S)"
+          HEAD_SHA="$(git rev-parse HEAD)"
+
+          gh api "repos/$GITHUB_REPOSITORY/git/refs" \
+            -f ref="refs/heads/$BRANCH" \
+            -f sha="$HEAD_SHA"
+
+          gh api graphql \
```
jeluard, line 194: I would expect there are simpler way to create PRs?

scripts/publish-bootstrap-snapshots
jeluard, line 28: What are we fixing here?

jeluard (commented):
I would be surprised if this can run on public github action runner. Could you validate that?

mmahut:
> I would be surprised if this can run on public github action runner. Could you validate that?

Definitely not, needs a dedicated self-hosted runner (cc @rkuhn).

.github/workflows/publish-bootstrap-snapshots.yml
mmahut: [The rabbit said](https://github.com/pragma-org/amaru/pull/951#discussion_r3396260767) it is better. Probably a nit-pick.

crates/amaru-mithril/src/lib.rs
mmahut: Yes, on preview the second run with a reused work dir skipped the Mithril download because chunk_for_slot assumes 21600 slots per chunk while preview chunks span 4320 (k=432), so db-analyser ran out of blocks and failed.

Sadly, I deleted the action run already, but I can try to reproduce.

.github/workflows/publish-bootstrap-snapshots.yml
mmahut: Depends on your runner, mine was missing aws for sure. But we can just fail probably.

.github/workflows/publish-bootstrap-snapshots.yml
mmahut: Because it might be already installed on the self-hosted runner.

.github/workflows/publish-bootstrap-snapshots.yml
mmahut: To be honest, I did not find any but you might be right - let me find out.

scripts/publish-bootstrap-snapshots
mmahut: When no epoch is given, the script used the latest metadata file as the set's start epoch, but that's actually the end of the set (the loop then looks for epoch+1 and epoch+2 which don't exist), so it now picks the third-latest.

.github/workflows/publish-bootstrap-snapshots.yml
mmahut: Looks like there is `peter-evans/create-pull-request` but I would rather skip adding a new dependency here.

crates/amaru-mithril/src/lib.rs
jeluard: Makes sense

.github/workflows/publish-bootstrap-snapshots.yml
jeluard: We want to control db-analyser version so I would suggest that we always install and add first in PATH the new one.

mmahut: @jeluard that makes sense, let me also add an overwrite to the version number in that case.

.github/workflows/publish-bootstrap-snapshots.yml
rkuhn: yeah, in general any non-hash reference is a supply chain attack waiting to happen

rkuhn (commented):
@mmahut Not sure what exactly your ping signifies: are you referring to my not yet publicly admitted experience in operating actions runners?

mmahut:
@rkuhn your experience has been doxxed in the https://github.com/pragma-org/amaru/issues/872 ;)

if you want, I can make the script a little bit leaner (like not checking for missing tools, which will only fail on the first tun) too

.github/workflows/publish-bootstrap-snapshots.yml
KtorZ: Note: we already depend on it; although my plan was to fork it under `pragma-org` to at least avoid random uncontrolled changes there.

mmahut:
I have made the script leaner (no checks on the self-hosted runner, as we expect that to be setup ahead) and also added an option to define your own version of db-analyser. Also rebased against main.

.github/workflows/publish-bootstrap-snapshots.yml
```diff
@@ -0,0 +1,222 @@
...
+          asset="cardano-node-$CARDANO_NODE_VERSION-$os-$arch.tar.gz"
+          base_url="https://github.com/IntersectMBO/cardano-node/releases/download/$CARDANO_NODE_VERSION"
+          tmp="$(mktemp -d)"
+          echo "downloading $base_url/$asset"
+          curl -fsSL -o "$tmp/$asset" "$base_url/$asset"
+          curl -fsSL -o "$tmp/sums.txt" "$base_url/cardano-node-$CARDANO_NODE_VERSION-sha256sums.txt"
+          (cd "$tmp" && grep " $asset\$" sums.txt | (sha256sum -c - 2>/dev/null || shasum -a 256 -c -))
+          tar -xzf "$tmp/$asset" -C "$tmp" --strip-components 1
+          rm "$tmp/$asset" "$tmp/sums.txt"
+          echo "$tmp/bin" >> "$GITHUB_PATH"
```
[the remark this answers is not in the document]
mmahut: Fair.

.github/workflows/publish-bootstrap-snapshots.yml
```diff
@@ -0,0 +1,225 @@
...
+    inputs:
+      network:
+        description: "Network to publish snapshots for"
+        required: true
+        type: choice
+        options:
+          - preprod
+          - preview
+          - mainnet
+          - all
```
KtorZ, line 16:
```suggestion
```

I am about 100% sure that these cannot lead to successful run. We need custom runners for Mainnet; but there's still value in producing snapshots for the two others.

.github/workflows/publish-bootstrap-snapshots.yml
```diff
@@ -0,0 +1,225 @@
...
+        required: false
+        type: string
+        default: ""
+      skip_verification:
+        description: "Skip the bootstrap smoke test against the published snapshots"
+        required: false
+        type: boolean
+        default: false
+      cardano_node_version:
+        description: "cardano-node release tag to download db-analyser from (must be a published release tag, e.g. 11.0.1)"
```
KtorZ, line 29: Is this controlling the version of db-analyzer or cardano-node? Both evolve rather independently.

docs/PUBLISHING_SNAPSHOTS.md
```diff
@@ -0,0 +1,112 @@
...
+Running the workflow again for the same epoch is safe: nothing is uploaded
+twice and no PR is opened if nothing changed.
+
+If a run fails with "db-analyser did not create snapshot directory", the
+latest epoch is not yet available from Mithril. This can happen shortly
+after an epoch boundary, mostly on preview. Try again a few hours later.
+
+## Runner setup
+
+The workflow needs a self-hosted runner with the labels `self-hosted` and
```
KtorZ, line 48: Okay. You can ignore my previous comment. Even on testnet actually, the required disk space is just too large so we do need self hosted runners _anyway_.

KtorZ (approved):

.github/workflows/publish-bootstrap-snapshots.yml
mmahut: This is the binary from the cardano-node release itself. Sadly looks like ouroboros-consensus is not shipping any binaries.

PR #956, opened 2026-06-12
chore: add ci for amaru-uplc

jonathanlim222:
This adds benchmark workflows for the amaru-uplc code. The CI test workflows already exist and are run automatically as part of the linux, macOS, and windows jobs.

`cargo test --workspace` runs the conformance and conformance flat tests.

## Summary by CodeRabbit

* **Chores**
  * Added automated performance benchmarking with a daily schedule and manual trigger.
  * Caches benchmark artifacts to speed runs and reduce repeated downloads.
  * Fetches and prepares benchmark data before executing the Turbo benchmark suite for consistent results.


.github/workflows/uplc-benchmarks.yml
```diff
@@ -0,0 +1,84 @@
...
+      should_run: ${{ steps.check.outputs.should_run }}
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+
+      - name: Check for changes since last run
+        id: check
+        run: |
+          LAST_RUN=$(gh run list --workflow=uplc-benchmarks.yml --branch=main --status=success --limit=1 --json createdAt --jq '.[0].createdAt' 2>/dev/null || echo "")
```
yHSJ: I know this comes straight from the other repo, but why aren't we comparing commit hashes?

```bash
LAST_SHA=$(gh run list --workflow=uplc-benchmarks.yml --branch=main --status=success --limit=1 --json headSha --jq '.[0].headSha')
```

We can then use:
```bash
git diff --quiet "$LAST_SHA" HEAD -- crates/amaru-uplc/
```

To see changes to this crate specifically. I think the wall-clock stuff is complicated and easy to break.

yHSJ (dismissed):
There are a few things:

1) Can we make sure that the `Makefile` is updated so that users can easily run benchmarks, and whatever combination they choose?
2) Where are the conformance test suites (flat and plutus)? Those need to be migrated as well
3) Do we want to update the other Amaru workflow files to *not* run on changes related to the uplc crate? Maybe not, since UPLC changes could break syncing/building, but worth considering.

KtorZ (approved):

PR #960, opened 2026-06-15
feat: new test snapshots

jeluard:
## Release Notes

* **New Features**
  * Added per-epoch conformance snapshot archive generation during snapshot creation, including deterministic packaging of extracted conformance data.

* **Tests**
  * Automated snapshot test case generation now triggers before end-to-end tests run, downloading and regenerating missing conformance archives as needed.

* **Chores**
  * Improved snapshot publishing flexibility via configurable snapshot output directory.
  * Updated the conformance extractor build and conformance dependency version to keep snapshots compatible.

### DIFF
--- a/conformance-tests/haskell-node-extractor/src/Data/PoolParameters.hs
+++ b/conformance-tests/haskell-node-extractor/src/Data/PoolParameters.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE DataKinds #-}
 {-# LANGUAGE NamedFieldPuns #-}
 
 module Data.PoolParameters
@@ -6,25 +7,33 @@ module Data.PoolParameters
 
 import Relude
 
+import Cardano.Ledger.Address
+    ( AccountAddress (..)
+    , AccountId
+    )
 import Cardano.Ledger.BaseTypes
     ( BoundedRational (unboundRational)
+    , Network (Testnet)
     )
 import Cardano.Ledger.Hashes
-    ( VRFVerKeyHash (unVRFVerKeyHash)
+    ( KeyHash
+    , VRFVerKeyHash (unVRFVerKeyHash)
     )
-import Cardano.Ledger.PoolParams
+import Cardano.Ledger.Keys
+    ( KeyRole (..)
+    )
+import Cardano.Ledger.State
     ( PoolMetadata
-    , PoolParams
-        ( PoolParams
-        , ppCost
-        , ppId
-        , ppMargin
-        , ppMetadata
-        , ppOwners
-        , ppPledge
-        , ppRelays
-        , ppRewardAccount
-        , ppVrf
+    , StakePoolState
+        ( StakePoolState
+        , spsCost
+        , spsMargin
+        , spsMetadata
+        , spsOwners
+        , spsRelays
+        , spsPledge
+        , spsVrf
+        , spsAccountId
         )
     )
 import Data.Aeson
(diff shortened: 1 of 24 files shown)


conformance-tests/haskell-node-extractor/src/Data/PoolParameters.hs
```diff
@@ -63,22 +72,22 @@ import Data.RewardAccount
...
-            , "reward_account" .= JsonRewardAccount ppRewardAccount
-            , "owners" .= Set.toAscList ppOwners
-            , "relays" .= fmap JsonPoolRelay (toList ppRelays)
+            [ "id" .= JsonPoolId poolId
+            , "vrf" .= unVRFVerKeyHash spsVrf
+            , "pledge" .= JsonCoin spsPledge
+            , "cost" .= JsonCoin spsCost
+            , "margin" .= JsonRational (unboundRational spsMargin)
+            , "reward_account" .= JsonRewardAccount (AccountAddress Testnet spsAccountId)
+            , "owners" .= Set.toAscList spsOwners
```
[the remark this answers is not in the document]
KtorZ: Rabbit is right here, and to avoid issues altogether; we could simply encode the account credential, without network whatsoever. It doesn't fundamentally changes the data.

scripts/publish-bootstrap-snapshots
KtorZ, line 1: Not fan of conflating the two things here; even if both the bootstrapping snapshots and the test snapshots are stuff we publish to cloudflare; they are fundamentally different artifacts. We should probably not even publish them in the same bucket because they have different "SLA".

crates/amaru/src/bin/amaru/cmd/create_snapshots.rs
```diff
@@ -296,15 +303,17 @@ pub async fn run(args: Args) -> Result<(), Box<dyn std::error::Error>> {
...
 
     if force {
-        remove_target_outputs(&snapshot_output_dir, &targets)?;
+        remove_target_outputs(&snapshot_output_dir, network, &targets)?;
     }
 
     let existing_snapshots = existing_snapshot_paths(&snapshot_output_dir, &targets);
     let existing_archives = existing_archive_paths(&snapshot_output_dir, &targets);
-    if !existing_snapshots.is_empty() || !existing_archives.is_empty() {
+    let existing_conformance = existing_conformance_archive_paths(&snapshot_output_dir, network, &targets);
```
KtorZ, line 311: I don't think we should mix the conformance tests with the snapshot creation; those are separate concern that deserve separate commands.

scripts/generate-test-snapshots
KtorZ, line 1: I am not sure to understand why we have both a scripts/ and a Rust command? Shouldn't we have either and not both?

crates/amaru/src/bin/amaru/cmd/create_snapshots.rs
jeluard: They both require cardano-node persistent data so that was the reasoning here. 
Separating them might trigger extra db-analyser processing. Are we fine with that?

KtorZ: >  Separating them might trigger extra db-analyser processing. Are we fine with that?

It shouldn't though? The db-analyser looks for ledger directories with a `_db-analyser` suffix, so provided one is available, it should use them.

jeluard: Right it should, but won't be guaranteed anymore if both steps are separate. It's not like those folders are smallish :D 
Now I guess we can always address that when it's an issue.

PR #961, opened 2026-06-15
Make DiffBind::Append Infallible

yHSJ:
On preview, the epoch transition from 1098 -> 1099 currently crashes, with the following error (first reported by @Cerkoryn on [Discord](https://discord.com/channels/1202416088776712253/1514660741490278460/1514747308041044030)):
```
2026-06-11T21:37:27.917392Z ERROR amaru_consensus::stages::validate_block: failed to validate block error=BlockValidationError: inconsistent or invalid volatile states; failed to create an aggregated volatile view point=94953610.bc237c0372978edb4de333a43ab35f53739051c22ca4563f28694b7e27080128
2026-06-11T21:37:27.917397Z  WARN pure_stage::tokio: stage `validate_block-7` terminated
2026-06-11T21:37:27.917452Z ERROR amaru::cmd::run: Consensus died, this should not happen! Please report this incl. preceding logs to the Amaru team.
```

That error is due to transaction`c4f6162a3dd4f1136952b7a46f5b76d92c904d1a7702f317eda8af21223ac33f` in which a user deregisters a stake credential and then immediately re-registers it. This caused a `MergeError::AlreadyRegistered` when constructing the `VolatileView`.

Validation in `DiffBind` happens during construction, and `append` is only ever going to be merging two already valid `DiffBind`s, so we do not need this extra check. This generally simplifies our logic as well.

Future work here should look at how we can simplify `DiffBind` and whether or not it's still needed, following the 2-sequence volatile DB changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved ledger state stability by making volatile view creation non-failing and streamlining volatile state transitions.
  * Updated state binding merging so re-registration cleanly supersedes prior bindings, and subsequent updates preserve existing registration state.

* **Tests**
  * Added unit tests covering re-registration precedence and preservation of registration state during binding updates.

* **Chores**
  * Removed and stopped exporting obsolete public error types related to volatile view creation and merge-time registration handling.

### DIFF
--- a/crates/amaru-ledger/src/state/diff_bind.rs
+++ b/crates/amaru-ledger/src/state/diff_bind.rs
@@ -110,19 +110,6 @@ pub enum RegisterError<K> {
 }
 
 #[derive(thiserror::Error, Debug)]
-pub enum MergeError<K> {
-    #[error("key is already registered")]
-    AlreadyRegistered(K),
-}
-
-impl<K: ToOwned<Owned = K>> MergeError<&K> {
-    pub fn to_owned(self) -> MergeError<K> {
-        let Self::AlreadyRegistered(k) = self;
-        MergeError::AlreadyRegistered(k.to_owned())
-    }
-}
-
-#[derive(thiserror::Error, Debug)]
 pub enum BindError<K> {
     #[error("key is already unregistered")]
     AlreadyUnregistered(K),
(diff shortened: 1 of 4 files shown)


crates/amaru-ledger/src/state/diff_bind.rs
```diff
@@ -137,16 +124,19 @@ impl<K: Ord, L, R, V> DiffBind<K, L, R, V> {
     }
 
     /// Merge two states together, assuming that the other is a more recent update.
-    pub fn append(&mut self, most_recent: Self) -> Result<&mut Self, MergeError<K>> {
+    ///
+    /// Importantly, this composes two already-validated `DiffBind`s, it does not re-validate them.
+    /// This composes two already-validated diffs; it does not re-validate them.
+    ///
```
KtorZ, line 130:
```suggestion
```

KtorZ (approved):
Yes. That sounds reasonable and I've been probably _too defensive_ here. And assuming DiffBinds are "correct" is the right thing to do; we need not to revalidate them. And it's even demonstrated here; because each individual DiffBind is effectively an aggregate of multiple events, we can't reason about them being individual transactions. 

Fundamentally, we have to treat them as pending database updates; and "blindly" apply them to the rolling state. Soooo., okay,

PR #963, opened 2026-06-15
Dual-series volatile DB

### LINKED ISSUE #863: Explore Improved Data Models for Block Validation
### Abstract

Explore creating a clear separation of the view of on-disk state and the view of state in our volatile DB. This would simplify state management during validation and could make our storage (and rollback) more time and space efficient.

### Why?

This work arose after @KtorZ's https://github.com/pragma-org/amaru/pull/832, which allowed further work on https://github.com/pragma-org/amaru/pull/831 (#832 included changes to storage, which made it, in theory, easier to construct the `ValidationContext` with the right state). However, when doing that work, it became clear that it was very difficult to be sure we're constructing it with the *correct* state, and the data was all tied together.

In parallel, we were also discussing how to make data lookups more efficient than linear time on the volatileDB, and considering rollbacks. Having a better data model for the ledger would address all the concerns.

### How?

The current plan (names are all TBD):
- `VolatileFragment` would contain the current set of diffs it has, along with an `AggregateVolatileView` object
-`AggregateVolatileView` would have an indexed view of the entire `VolatileDB`, making lookups fast (bascial
- `DiskView` would have an indexed view of the data on disk we need to validate the specific block
- `ValidationContext` becomes the previous block's the tuple `[AggregateVolatileView, DiskView, DestructiveChanges]` where `DestructiveChanges` is the set of changes which removed objects from state (so we can prevent double spends).

This is a very abstract representation, some of these objects could be merged and may not actually have struct definitions in the code.

If we're able to implement all the current slices on the new context, we should be able to drop in replace it, and all tests should pass.

### Discussion points

_No response_

_No response_

### Checklist

yHSJ:
This PR introduces a new design for the volatileDB, in which we end the `VolatileSeries` at the epoch boundary, and start a new one. This change should make it easier to reason about the state, simplify querying state, and even potentially simplify some of our `DiffX` types that we use to enforce invariants about ledger state.

This is the first step of resolving #863. The next step will be finishing up the `ValidationContext` changes, exposing the other ledger state we don't currently include. This PR is scoped to just moving the existing logic to the new design.

## Summary by CodeRabbit

* **Refactor**
  * Improved ledger volatile state handling with safer sealing and rollback consistency checks.
  * Redesigned volatile storage to support front/back views and correct cross-window lookups.
  * Simplified state operation merging and validation wiring.
* **New Features**
  * Added support for decoding “new epoch” snapshot state, including pool and authorization data.
  * Expanded validation context to support ledger predicates using additional state.
* **Tests**
  * Added/updated coverage for volatile storage behavior and new epoch state decoding.


crates/amaru-ledger/src/state/volatile/view/iter_pools.rs
```diff
@@ -48,15 +48,15 @@ use crate::{
...
+    registrations: BTreeMap<PoolId, Registrations<&'volatile Arc<(PoolParams, CertificatePointer)>>>,
     retirements: BTreeMap<PoolId, Epoch>,
 }
 
 impl<'volatile, DBIter: Iterator<Item = (PoolId, Pool)>> IterPools<'volatile, DBIter> {
     pub fn new(
         epoch: Epoch,
         db_iterator: DBIter,
-        pools: &mut DiffEpochReg<PoolId, &'volatile (PoolParams, CertificatePointer)>,
+        pools: &mut DiffEpochReg<PoolId, &'volatile Arc<(PoolParams, CertificatePointer)>>,
```
KtorZ: Suspicious (mutable) reference of an `Arc` here; the iterator operates on references whose ownership is checked at runtime here, the `Arc` should be superfluous.

crates/amaru-ledger/src/state/volatile/series.rs
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    fn contains(&self, point: &Point) -> bool {
+        self.sequence.binary_search_by_key(point, |anchored| anchored.point()).is_ok()
+    }
+
+    fn pop_front(&mut self) -> Option<AnchoredVolatileFragment> {
+        let popped = self.sequence.pop_front();
+        if popped.is_some() {
+            self.recompute_aggregate();
```
KtorZ:
```suggestion
            // TODO: consider introducing "undo" fragments? 
            //
            // This recompute is 'expensive', in that it will happen for *every block* with only a little change 
            // since the vast majority of the fragment sequence stay the same (only one change). So we should
            // find a way to avoid re-computing this entirely every time. One option could be to introduce 
            // 'undo' operations so that we can simply revert a single fragment while keeping the others.   
            self.recompute_aggregate();
```

crates/amaru-ledger/src/state/volatile.rs
```diff
@@ -13,10 +13,104 @@
...
+
+    pub(crate) const VOLATILE_WINDOW: usize = 6;
+
+    prop_compose! {
+        /// A window of [`DiffSet`]s where each tagged UTxO has a unique lifecycle: produced once and
+        /// optionally consumed at a strictly later index. Mirrors UTxO uniqueness, so a newest-first
+        /// walk has a well-defined answer for every key.
+        pub(crate) fn unique_lifecycle_diffs()(
+            plan in prop::collection::btree_map(
+                0u8..16,
```
KtorZ: Kind of odd/convoluted to resort to a btree_map here; I believe we could do with a vector and then use `.enumerate` when iterating to get a unique index? Since this seems to be the goal.

crates/amaru-ledger/src/state/volatile.rs
```diff
@@ -13,10 +13,104 @@
...
+                (0usize..VOLATILE_WINDOW, prop::option::of(0usize..VOLATILE_WINDOW)),
```
KtorZ: Note: you can have proptest strategies take arguments, instead of resorting to a constant here. This allows to test with different volatile window sizes more easily. So for example:

```rs
        pub(crate) fn unique_lifecycle_diffs(volatile_window: usize)(
            plan in prop::collection::btree_map(...) 
        )
```

crates/amaru-ledger/src/state.rs
```diff
@@ -813,6 +813,12 @@ impl<S: Store, HS: HistoricalStores> State<S, HS> {
                     self.overlay.rollback();
                 }
 
+                debug_assert_eq!(
```
```suggestion
                assert_eq!(
```

Yolo.

crates/amaru-ledger/src/state/volatile/fragment.rs
```diff
@@ -52,12 +55,42 @@ impl VolatileFragment {
...
     }
+
+    /// Fold `more_recent` into this fragment, treating it as applied *after* `self`.
+    /// This maintains the running aggregate of a [`crate::state::volatile::VolatileSeries`].
+    ///
+    /// `accounts`, `dreps` (with `dreps_deregistrations`) and `committee` are
+    /// deliberately not folded here: they are resolved against the stable store and
+    /// will be maintained as materialized state rather than a composed diff. The
+    /// exhaustive destructuring forces this decision to be revisited if a column is
+    /// added.
```
KtorZ: Not sure to follow this comment 🤔? Why aren't we doing that for accounts, dreps, etc.. ?

crates/amaru-ledger/src/state/volatile/fragment.rs
```diff
@@ -52,12 +55,42 @@ impl VolatileFragment {
...
+            dreps_deregistrations: _,
+            committee: _,
+        } = more_recent;
+
+        self.utxo.merge(utxo.clone());
+        self.votes.merge(votes.clone());
+        self.pools.append(pools.clone());
+        self.withdrawals.extend(withdrawals.iter().cloned());
+        self.proposals.extend(proposals.iter().map(|(id, value)| (id.clone(), value.clone())));
+        self.fees += *fees;
```
KtorZ: Yeah, so that's where things get a little ugly. `.merge` on `BTreeMap` isn't particularly adequate here since it forces to clone the whole structure whereas we only want to clone the individual elements (which .iter().cloned() does nicely for the `BTreeSet`).

KtorZ (approved):
LGTM, modulo the remaining pending points.

crates/amaru-ledger/src/state/volatile/fragment.rs
yHSJ: This is old and won't be there after #975 is merged in. This was just a temporary note, I was only folding in the state that was previously tracked.

crates/amaru-ledger/src/state/volatile/fragment.rs
yHSJ: No, it's quite ugly. This is one obvious improvement place that we need to focus on in the future.

crates/amaru-ledger/src/state.rs
```diff
@@ -409,7 +472,9 @@ impl<S: Store, HS: HistoricalStores> State<S, HS> {
 
             drop(db); // Dropping the *mutable reference*, not the *actual database* :)
 
-            self.overlay.transition(effective_rewards, pools_updates, governance_updates);
+            self.volatile.overlay_mut().transition(effective_rewards, pools_updates, governance_updates);
+
+            self.volatile.transition();
```
KtorZ, line 479: Seems off; I would avoid exposing the `overlay` outside of the volatile in this new design. If it's self-managed; then let's simply interact with the volatile and never surface the overlay to external callers.

Conceptually, the "volatile" becomes the owner of _anything that's computed but not yet applied_.

crates/amaru-ledger/src/state.rs
```diff
@@ -540,6 +606,353 @@ impl<S: Store, HS: HistoricalStores> State<S, HS> {
         Ok(result)
     }
 
+    /// Resolves pools, confirming the existence of the provided `pool_ids`.
```
KtorZ: NOTE: haven't looked at any of the implementation below _yet_; but setting a reminder.

crates/amaru-ledger/src/state/overlay.rs
```diff
@@ -206,60 +203,79 @@ impl StateOverlay {
...
+    }
+
+    /// The account's pending reward credit at the not-yet-flushed epoch boundary: its effective
+    /// reward plus any pool-deposit refund. `0` outside the straddle window.
+    pub fn pending_reward_credit(&self, credential: &StakeCredential) -> Lovelace {
+        let reward = match &self.rewards {
+            RewardsState::Effective(effective) => effective.accounts().get(credential).copied().unwrap_or(0),
+            RewardsState::NotReady | RewardsState::Computed(..) => 0,
+        };
+        let refund = self.pools_updates.as_ref().map(|updates| updates.refund(credential)).unwrap_or(0);
```
KtorZ, line 278: Missing governance proposal refunds?

crates/amaru-ledger/src/context/default/validation.rs
```diff
@@ -274,8 +358,18 @@ impl CommitteeSlice for DefaultValidationContext {
 }
 
 impl ProposalsSlice for DefaultValidationContext {
+    fn lookup(&self, id: &ComparableProposalId) -> Option<&ProposalState> {
+        // TODO: also fold proposals acknowledged earlier in this block, once the voting rule needs
+        // it. Their expiry is the current epoch's horizon, which that rule already has.
```
KtorZ:
```suggestion
        // FIXME: also fold proposals discovered in the block during validation
```

crates/amaru-ledger/src/state.rs
yHSJ: Agreed, this will be done in a follow up PR, both for scope and to ensure that anyone waiting on this PR is unblocked (or doesn't have to manage significant merge conflicts)

PR #964, opened 2026-06-15
chore: add protocol version to cost model determination

jonathanlim222:
This change adds cost model determination using protocol version. This change is needed for the upcoming protocol version 11 fork.

## Summary by CodeRabbit

* **New Features**
  * Added cost model support for Ripemd-160, ExpModInteger, DropList, LengthOfArray, ListToArray, and IndexArray.
  * Exposed a default V3 cost model for use by evaluations and tooling.

* **Bug Fixes**
  * Corrected the `blake2b_224` memory cost configuration to align with the expected parameter source.

* **Refactor**
  * Protocol/version handling is more explicit: program evaluation and parsing now require an explicit protocol version.
  * Updated `constr`/`case` availability gating to match protocol compatibility rules.


crates/amaru-uplc/src/machine/cost_model/mod.rs
```diff
@@ -57,6 +61,28 @@ impl<B: BuiltinCostModel + Default> Default for CostModel<B> {
     }
 }
 
+/// Default V3 cost model values (251 base + 46 PLOMIN = 297 entries).
+/// These match the Cardano mainnet default cost parameters for PlutusV3.
```
yHSJ, line 65: Technically, these match the Cardano mainnet Pv11 cost model for PlutusV3.

yHSJ (approved):
Looks acceptable to me. We need these changes for Pv11, so we should just get them in.

Issue #882, opened 2026-06-05
Fully process an AssertOperationalCertificateError


etorreborre:
### Abstract

In `praos/header.rs` there is a `FIXME`:
```
FIXME: Double-check whether we mustn't fail in this case or if it is acceptable
to have no opcert available?
```
That requires reading the `cardano-node` code to check what needs to be done

### Why?

For conformance

### How?

By implementing the correct behaviour.

### Testing Strategy / Acceptance Criteria

Unit test

### Checklist

- [x] I understand that feature requests and unrefined work item should be open as [GitHub Discussions](https://github.com/pragma-org/amaru/discussions) instead.
- [x] I have assigned this item to an existing milestone from the roadmap
- [x] I have added a label capturing the impact of this item (i.e. value for users/stakeholders if successful)
- [x] I have added a label capturing the delivery risk of this item (i.e. how likely is it that this task will succeed as planned)
- [x] I have added a label capturing the effort of this item (i.e. how large is the task?)

PR #965, opened 2026-06-16
refactor: move header validation

etorreborre:
This PR refactors the code base around header validation in order to support the upcoming fix for #882.
The main changes are:

 - The move of all the header validation code under an effect, eventually calling `HeaderValidator`, like we have `BlockValidator` already:
    - This removes the need for a `PraosChainStore` and the `Praos` trait in `amaru-ouroboros-traits`.
    - There's only one mock left for `BlockValidator`.
    - The header verification uses a trait `HasPools` implemented directly by the ledger `State`, instead of having an intermediate `struct`.
 - Use a specific trait `HasPeersData` to retrieve registered relays from the ledger.
    - Even if it is also implemented by the ledger `State`, I preferred to separate this concern from `HasPools` since it is quite different from retrieving the `PoolSummary`.
 - Some data types were moved to the `kernel` and/or better encapsulated:
    - `PoolId` (this will help with the implementation of #882)
    - `Nonce` (with functions to make or evolve a nonce from a header)
    - `vrf` 
    - `kes`
    - The `randomness_stability_window` can now be accessed directly on `EraHistory`
    - `Bytes`
  - As a result the `amaru-ouroboros` crate is not necessary anymore and has been removed.  
  - A `FragmentAnchor` now keeps track of its `issuer` (needed later for #882). This data type is added, it used to be a pair.
  - The `LedgerConfig` is a subset of the overall `Config`, which helps for instantiating the `Ledger` in `amaru` and simplify the initial creation of the `BlockValidator` and the `HeaderValidator` at the top-level (some helpers were removed thanks to this).
  - This also helps with the implementation of `sync` which just uses a `BlockValidator` and a `HeaderValidator` now with high-level methods.

 

## Summary by CodeRabbit

* **New Features**
  * Added header validation components and wired them into the ledger flow.
  * Added ledger-backed peer data to improve relay socket address discovery.
  * Introduced operational certificate sequence number tracking and exposed it via new pool summary APIs.
  * Extended randomness stability window computations (epoch/slot boundary awareness).
* **Improvements**
  * Refactored node/ledger initialization around a unified ledger configuration.
  * Enhanced pool and nonce handling with new accessors and updated serialization.
  * Improved transaction mempool insert reporting with a structured insert result type.


crates/amaru-kernel/src/cardano/nonce.rs
```diff
@@ -12,9 +12,81 @@
...
-use crate::{Hash, size::NONCE};
+use pallas_crypto::hash::Hasher;
 
-pub type Nonce = Hash<NONCE>;
+use crate::{Hash, IsHeader, cbor, size::NONCE};
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, serde::Serialize, serde::Deserialize)]
+#[repr(transparent)]
+pub struct Nonce(Hash<NONCE>);
+
```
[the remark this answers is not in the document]
etorreborre: Done.

crates/amaru-ledger/src/header_validator/assertions.rs
```diff
@@ -167,6 +110,51 @@ pub fn assert_all<'a>(
...
+            (Self::KnownLeaderVrf(l0), Self::KnownLeaderVrf(r0)) => l0 == r0,
+            (Self::VrfProof(l0), Self::VrfProof(r0)) => l0 == r0,
+            (Self::LeaderStake(l0), Self::LeaderStake(r0)) => l0 == r0,
+            (Self::KesSignature(l0), Self::KesSignature(r0)) => l0 == r0,
+            (Self::OperationalCertificate(l0), Self::OperationalCertificate(r0)) => l0 == r0,
+            (Self::TryFromSliceError, Self::TryFromSliceError) => true,
+            (Self::UnknownPool { pool: l_pool }, Self::UnknownPool { pool: r_pool }) => l_pool == r_pool,
+            _ => false,
+        }
+    }
```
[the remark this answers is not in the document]
etorreborre: Fixed

crates/amaru/src/stages/build_stage_graph.rs
```diff
@@ -47,12 +47,12 @@ use crate::stages::config::Config;
...
     config: &Config,
-    era_history: &EraHistory,
-    global_parameters: &GlobalParameters,
     ledger_tip: Tip,
     best_hash: HeaderHash,
     stage_graph: &mut impl StageGraph,
 ) -> NodeStages {
+    let era_history = config.era_history();
+    let global_parameters = config.global_parameters();
     let manager = stage_graph.stage("manager", manager::stage);
```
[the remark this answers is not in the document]
etorreborre: Done.

crates/amaru/src/stages/ledger.rs
```diff
@@ -12,60 +12,106 @@
...
-        Ok(Arc::new(state.view_stake_distribution()))
+    #[expect(clippy::panic)]
+    fn make_state(config: &LedgerConfig) -> anyhow::Result<state::State<RocksDB, RocksDBHistoricalStores>> {
+        let era_history: &EraHistory =
+            config.network.as_era_history().unwrap_or_else(|| panic!("missing default EraHistory for network"));
+        let global_parameters: &GlobalParameters = config
+            .network
+            .as_global_parameters()
+            .unwrap_or_else(|| panic!("missing default GlobalParameters for network"));
+
```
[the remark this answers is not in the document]
etorreborre: Done.

crates/amaru/src/stages/ledger.rs
```diff
@@ -12,60 +12,106 @@
...
+        if anchor_hash == ORIGIN_HASH {
+            tracing::info!(anchor = %ledger_tip, "first initialization - setting anchor and best chain");
+            self.chain_store.set_anchor_hash(&ledger_tip.hash())?;
+            self.chain_store.set_block_valid(&ledger_tip.hash(), true)?;
+            self.chain_store.roll_forward_chain(&ledger_tip)?;
+        };
+
+        // Check that the ledger tip can be retrieved from a stored header and return it
+        self.chain_store.load_tip(&ledger_tip.hash()).ok_or(anyhow!("ledger tip header not found"))
     }
```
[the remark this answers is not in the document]
etorreborre: Done.

crates/amaru-ouroboros-traits/src/validators/blocks/can_validate_blocks_mock.rs
```diff
@@ -12,47 +12,117 @@
...
     }
 
-    fn rollback_block(&self, _to: &Point) -> Result<(), BlockValidationError> {
+    fn rollback_block(&self, to: &Point) -> Result<(), BlockValidationError> {
+        let mut inner = self.inner.lock();
+        if inner.rollback_fails {
+            return Err(BlockValidationError::new(anyhow::anyhow!("mock rollback failed")));
+        }
+        inner.tip = *to;
+        inner.contains.retain(|p| p.slot_or_default() <= to.slot_or_default());
```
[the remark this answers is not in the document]
etorreborre: Done.

crates/amaru/src/stages/config.rs
```diff
@@ -85,26 +74,34 @@ impl Config {
...
     }
+
+    pub fn network(&self) -> NetworkName {
+        self.ledger_config.network
+    }
+
+    #[expect(clippy::panic)]
+    pub fn era_history(&self) -> &EraHistory {
+        self.ledger_config.network.as_era_history().unwrap_or_else(|| panic!("missing default EraHistory for network"))
+    }
```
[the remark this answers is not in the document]
etorreborre: Done.

crates/amaru-ledger/src/header_validator/validator.rs
```diff
@@ -0,0 +1,365 @@
...
+
+/// Evolve the given nonce by combining it in an arbitrary way with other data. When
+/// `within_stability_window` is false, this also modifies the candidate nonce for the next
+/// epoch.
+///
+/// Once the stability window has been reached, the candidate is fixed for the epoch and will
+/// be used once crossing the epoch boundary to produce the next epoch nonce.
+fn evolve_nonce(
+    consensus_parameters: &ConsensusParameters,
+    store: Arc<dyn ChainStore>,
```
rkuhn:
```suggestion
    store: &dyn ChainStore,
```

crates/amaru-ledger/src/peers_data.rs
```diff
@@ -0,0 +1,162 @@
...
+    state,
+    state::State,
+    store::{HistoricalStores, ReadStore, Store, StoreError as LedgerStoreError},
+};
+
+/// This data type accesses the ledger state to extract peers data.
+pub struct PeersData<S, HS>
+where
+    S: Store + Send,
+    HS: HistoricalStores + Send,
```
rkuhn, line 36: These constraints should only be on impls unless they are required for the pure declaration of this type itself.

crates/amaru-ledger/src/peers_data.rs
```diff
@@ -0,0 +1,162 @@
...
+{
+    pub state: Arc<Mutex<state::State<S, HS>>>,
```
rkuhn, line 38: Why is it better to have this wrapper instead of just implementing `registered_relay_socket_addrs` on the state?

crates/amaru-ledger/src/state.rs
```diff
@@ -253,20 +253,33 @@ impl<S: Store, HS: HistoricalStores> State<S, HS> {
...
+    pub fn get_pool_summary(&self, slot: Slot, pool: &PoolId) -> Result<Option<PoolSummary>, GetPoolError> {
+        let epoch = self
+            .era_history
+            // NOTE: This function is called by the consensus when validating block headers. So in
+            // theory, the slot is either within the current epoch or the next since blocks must
+            // form a chain. Either the previous block is well within the current epoch, or it was
+            // the last block of the previous epoch.
+            //
+            // Either way, we do know at this point how to forecast this slot.
+            .slot_to_epoch_unchecked_horizon(slot)
```
rkuhn: I hope “unchecked” means “it will panic if this assumption turns out to be false”

crates/amaru/src/stages/build_node.rs
```diff
@@ -100,48 +98,24 @@ impl NodeRunning {
...
     stage_builder: &mut impl StageGraph,
 ) -> anyhow::Result<NodeStages> {
-    let era_history = &config.era_history;
-
-    // Make the ledger and get its tip
-    let ledger = Ledger::new(config, era_history.clone(), global_parameters.clone())
+    // Create the chain store
+    let chain_store = create_chain_store(config)?;
+    // Make the ledger
+    let ledger = Ledger::new(&config.ledger_config, chain_store.clone())
```
rkuhn, line 107: seems a little weird to see the ledger needing a chain store, is this because of the nonces? should the nonces be moved into the ledger db instead? they always felt a bit misplaced to me /cc @KtorZ

crates/amaru/src/stages/build_node.rs
```diff
@@ -100,48 +98,24 @@ impl NodeRunning {
...
-        "build_ledger"
-    );
-
-    // Make the chain store, either from the network resources if already set
-    // or from the configuration.
-    // This also makes sure that the chain store tip and anchors are exactly aligned to the
-    // ledger tip.
-    let chain_store = initialize_chain_store(config, ledger_tip)?;
-    let ledger_tip = chain_store.load_tip(&ledger_tip.hash()).ok_or(anyhow!("ledger tip header not found"))?;
+    let ledger_tip = ledger.initialize_chain_store()?;
```
rkuhn, line 110: this is also weird — we need to sit down and do a clear design on these DBs and their relationship

crates/amaru/src/stages/ledger.rs
```diff
@@ -12,60 +12,104 @@
...
 use amaru_plutus::arena_pool::ArenaPool;
 use amaru_stores::rocksdb::{RocksDB, RocksDBHistoricalStores};
+use anyhow::anyhow;
 
-use crate::stages::config::Config;
+use crate::stages::config::LedgerConfig;
 
 /// Representation of the ledger as used by the consensus stages.
-pub struct Ledger(BlockValidator<RocksDB, RocksDBHistoricalStores>);
+pub struct Ledger {
```
rkuhn, line 27: This is also diverging: the `stages` module should be used  by stages, and this is not a stage.

rkuhn (commented):
Nice cleanup! There are a couple of points to discuss, though.

crates/amaru-ledger/src/header_validator/validator.rs
etorreborre: Done.

crates/amaru-ledger/src/peers_data.rs
etorreborre: `State` itself requires those constraints.

crates/amaru-ledger/src/peers_data.rs
etorreborre: Because we need to lock the `State`. Do you see another way?

crates/amaru-ledger/src/state.rs
etorreborre: Here `unchecked` means that we select the epoch if it doesn't have a defined `end` (this was done some time ago now: https://github.com/pragma-org/amaru/pull/326)

crates/amaru/src/stages/build_node.rs
etorreborre: This more or less corresponds to [`ExtLedgerState`](https://github.com/IntersectMBO/ouroboros-consensus/blob/01140b42005f03c7addae690b185ff9168f643f2/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/Extended.hs#L70) in the Haskell cardano node. For header validation we need a bit of chain data and a bit of ledger data. I'm personally fine with keeping the name `Ledger` here as an approximation of everything that gives us the possibility to validate block and headers but maybe you would have a better name in mind?

etorreborre: @KtorZ @rkuhn when can we discuss this?

crates/amaru-ledger/src/peers_data.rs
rkuhn: Ah, I followed the wrong train of thought, with the current design you’re right.

All this locking gives me headaches, it might be better to keep the State on a dedicated thread and communicate with that via channels — totally old school. But this is a discussion for another milestone ;-)

crates/amaru/src/stages/build_node.rs
KtorZ: I agree that this changes is a little weird? The nonces were out of the ledger for this reason. Also because they are actually independent of the ledger, but are very-much tied to a particular chain. The `ExtLedgerState` is not necessarily an example to follow ... 😬; this is the consensus patching the ledger state in a weird way; mostly because they only have one form of key value store.

crates/amaru-consensus/Cargo.toml
```diff
@@ -38,11 +36,11 @@ tracing.workspace = true
 tracing-opentelemetry.workspace = true
 
 amaru-kernel.workspace = true
+amaru-ledger.workspace = true
```
KtorZ: This is sort of a big thing 😅 ... we had gone quite some effort to precisely avoid doing this. The ouroboros-traits was the attempt at decoupling the two, so I suspect there's something wrong going on here.

crates/amaru-kernel/src/cardano/nonce.rs
```diff
@@ -12,9 +12,89 @@
...
+    pub fn new(hash: Hash<NONCE>) -> Self {
+        Self(hash)
+    }
+
+    /// Obtain the final nonce at an epoch boundary for the epoch from the stable candidate and the
+    /// last block (header) of the previous epoch.
+    ///
+    /// Return `None` if header has no parent (i.e. which never happens because all our blocks have
+    /// parents in Amaru).
+    pub fn make_epoch_nonce<H: IsHeader>(&self, header: &H) -> Option<Nonce> {
```
KtorZ, line 41: I'd say that this is `new` and what is current `new` should rather be a `From` instance.

crates/amaru-ledger/src/header_validator/assertions.rs
```diff
@@ -167,6 +109,53 @@ pub fn assert_all<'a>(
...
+            (Self::KnownLeaderVrf(l0), Self::KnownLeaderVrf(r0)) => l0 == r0,
+            (Self::VrfProof(l0), Self::VrfProof(r0)) => l0 == r0,
+            (Self::LeaderStake(l0), Self::LeaderStake(r0)) => l0 == r0,
+            (Self::KesSignature(l0), Self::KesSignature(r0)) => l0 == r0,
+            (Self::OperationalCertificate(l0), Self::OperationalCertificate(r0)) => l0 == r0,
+            (Self::PoolError(l0), Self::PoolError(r0)) => l0 == r0,
+            (Self::TryFromSliceError, Self::TryFromSliceError) => true,
+            (Self::UnknownPool { pool: l_pool }, Self::UnknownPool { pool: r_pool }) => l_pool == r_pool,
+
+            _ => false,
```
KtorZ, line 151: Let's not 😅 ... Please enumerate all the cases; this is otherwise a ticking bomb. It seems to be that the only case which needs a manual implementation is `TryFromSliceError`? Everything else can delegate back to `PartialEq`?

crates/amaru-ledger/src/header_validator/mod.rs
KtorZ, line 1: Not quite sure about this one. I don't think the responsibility belongs in the ledger. The ledger really only cares about `body`; anything that is `header` is a chain-concern. This one in particular is protocol-dependent (Praos). Header validation was different in T(ransitional) Praos and Ouroboros BFT. It's not excluded it also changes in Leios. So it's good to keep the header validation logic tied to a _protocol_; as it was here.

crates/amaru-ledger/src/header_validator/validator.rs
```diff
@@ -0,0 +1,365 @@
...
+}
+
+impl<S: Store + Send, HS: HistoricalStores + Send> HasPools for HeaderValidator<S, HS> {
+    #[expect(clippy::unwrap_used)]
+    fn get_pool_summary(
+        &self,
+        slot: amaru_kernel::Slot,
+        pool_id: &amaru_kernel::PoolId,
+    ) -> Result<Option<PoolSummary>, amaru_ouroboros_traits::pools::GetPoolError> {
+        self.state.lock().unwrap().get_pool_summary(slot, pool_id)
```
KtorZ: Not okay! Interactions between the ledger state and the outside MUST go through the existing observer mechanism. We do not want to introduce more points of contention on the ledger mutex; especially because it would here happen on every header. 

The current design ensured that the mutex on which the consensus/chain depends for validation is only rarely modified (once per epoch); this is why we have this whole indirection with the `StakeDistributionObserver` and no immediate dependency on the ledger state itself.

crates/amaru-kernel/src/cardano/nonce.rs
etorreborre: `new` taking `self` and returning an `Option` is weird, isn't it?

crates/amaru-consensus/Cargo.toml
etorreborre: Indeed, that's a mistake because of an import for an error type. I removed it.

crates/amaru-ledger/src/header_validator/assertions.rs
etorreborre: I just moved that code around. I try to not touch code too much when it's outside my refactoring. I propose that we address that in another PR.

KtorZ: Ho, okay. Then a `TODO` will likely do!

crates/amaru-kernel/src/cardano/nonce.rs
KtorZ: Well, it would only take the header; no? I see `new` as the "normal" way of constructing a type, the one we expect to see in production. And it can totally yield an error if the arguments are invalid.

crates/amaru-ledger/src/header_validator/validator.rs
etorreborre: This can be done ok. The issue is with the opcert sequence numbers. They need to be rolled-back when we switch forks. This is why I piggy-backed on the ledger, since it already has all the machinery. 

I agree that it would be better to keep consensus data separate. In the Haskell node, it looks like the "ledger" is responsible for rolling back the whole "extended ledger state" which is the ledger state proper and some chain state.

The mutex is indeed problematic, and not only here, since our current "switch to fork" is not atomic (#967).
This is why I started [this draft PR](https://github.com/pragma-org/amaru/pull/968) to investigate a possible way forward (so that readers can read an isolated snapshot lock-free most of the time, but I think there are improvements that can be made).

KtorZ: I guess now we remember why we had left the operational certificates out for so long 😅 ... 

If we flush out the situation a little bit:

- The chain/consensus stages which need a "read" access on the operational certificates
- The ledger which needs to persist the operational certificate as part of its state (with the same bells and whistles as other elements, to allow rollbacks of such elements)
- Yet, we do not want to expose a mutex of the whole ledger state to the consensus; but rather, some kind of "view" which can be updated independently. 

The big issue I see lies in the concurrent aspect of these. The consensus will need to know _the latest_ operational certificates likely before the ledger has "had time" to process the block which isn't acceptable. Also, the ledger has strictly no business whatsoever with the operational certificates; that they exist or not really doesn't change much. 

So in essence, it seems that the ledger really isn't the right "owner" of this data; even if it can be rolled back. It's also a MUCH easier piece to rollback because it's fundamentally just a map between pool id an some number; and there's no special calculation made on them. So it seems to me that like the nonces, it's probably better to keep this data in a store distinct of the ledger, and re-implement the bits of rollback logic necessary to its lifecycle.

etorreborre:
Closing this PR for now. I will open the new one with [the approach presented by @KtorZ](https://github.com/pragma-org/amaru/pull/965#discussion_r3460353116).

PR #966, opened 2026-06-16
feat: track operational cert sequence numbers

### LINKED ISSUE #882: Fully process an AssertOperationalCertificateError

etorreborre:
This PR fixes #882 by:

- Dealing with the case of a pool that is known but has never issued a block, hence, a pool that is providing an operational cert sequence number for the first time.
- Persisting  operational cert sequence numbers in the ledger database (volatile + on-disk).
- Augmenting the bootstrap with the retrieval of operational cert sequence numbers from the ledger snapshots.

Notes: 

 - The testing support for headers validation has also been extended.
 - Since the opcert sequence numbers are part of the volatile DB, there is currently a risk of invalidating a header during a fork since this happens as 2 separate operations, a rollback + some roll forwards. This is tracked with issue #967


etorreborre:
@coderabbit review

crates/amaru-ledger/src/state.rs
```diff
@@ -212,6 +212,22 @@ impl<S: Store, HS: HistoricalStores> State<S, HS> {
...
+        // Iterate from most recent to least recent
+        for fragment in self.volatile.iter().rev() {
+            if fragment.issuer() == *pool_id {
+                return Ok(Some(fragment.operational_cert_sequence_number()));
+            }
+        }
+        Ok(self
+            .stable
+            .lock()
+            .unwrap()
```
rkuhn, line 224: we should really switch to parking_lot throughout, if only to get rid of these mandatory code line references embedded in the executables at all optimisation levels

crates/amaru-stores/src/rocksdb/ledger/columns/opcerts.rs
```diff
@@ -0,0 +1,68 @@
...
+        unsafe_decode,
+    },
+};
+use amaru_observability::trace_span;
+use rocksdb::{DBPinnableSlice, Transaction};
+
+use crate::rocksdb::common::{PREFIX_LEN, as_key, as_value};
+
+/// Name prefixed used for storing last opcerts sequence numbers entries. UTF-8 encoding for "opce"
+pub const PREFIX: [u8; PREFIX_LEN] = [0x6F, 0x70, 0x63, 0x65];
```
rkuhn: please write this in readable fashion
```suggestion
pub const PREFIX: [u8; PREFIX_LEN] = b"opce";
```

crates/amaru-stores/src/rocksdb/ledger/columns/opcerts.rs
```diff
@@ -0,0 +1,68 @@
...
+        db_operation_name = "get".to_string(),
+        db_collection_name = "opce".to_string()
+    );
+    let _guard = _span.enter();
+
+    let key = as_key(&PREFIX, key);
+    Ok(db_get(&key).map_err(|err| StoreError::Internal(err.into()))?.map(|d| unsafe_decode::<Value>(&d)))
+}
+
+pub fn put<DB>(db: &Transaction<'_, DB>, key: &Key, value: Value) -> Result<(), StoreError> {
```
rkuhn, line 47: please document that and why this can only be done inside a transaction (which feels correct to me); same for `remove`

rkuhn (commented):
looking good, just nitpicks

crates/amaru-ledger/src/header_validator/assertions.rs
```diff
@@ -405,20 +404,12 @@ impl AssertOperationalCertificateError {
...
-            }
-            None => {
-                // FIXME: Double-check whether we mustn't fail in this case or if it is acceptable
-                // to have no opcert available?
-            }
+        if declared_sequence_number < latest_sequence_number {
+            return Err(Self::SequenceNumberTooSmall { declared_sequence_number, latest_sequence_number });
+        }
+
+        if declared_sequence_number - latest_sequence_number > 1 {
```
KtorZ, line 411: Underflow potential?

crates/amaru-ledger/src/state.rs
KtorZ: I don't disagree; but I would even challenge the necessity for the ledger to use any mutex here. 

The ledger is single-threaded and in principle only need a synchronisation point with other components around very specific data pieces which are held in memory. 

Having the stable (and also a few others currently) behind a mutex seems unnecessary.

KtorZ (approved):
Nice work! One concern over a subtraction which seems unsafe.

crates/amaru-stores/src/rocksdb/ledger/columns/opcerts.rs
etorreborre: Yes, I forgot about that. Much better!

crates/amaru-stores/src/rocksdb/ledger/columns/opcerts.rs
etorreborre: Done.

crates/amaru-ledger/src/header_validator/assertions.rs
etorreborre: Not with the previous check that asserts that `declared_sequence_number >= last_sequence_number`.

rkuhn (approved):

KtorZ:
...

- Started looking at merging this PR. 
- Realized there's no changelog.
- Went into drafting requirements for a changelog linter check in CI
- Went back looking at this PR wondering why it's not merged yet 
- Forgot about the changelog
- Merged it

PR #968, opened 2026-06-16
fix: switch forks on the ledger atomically

etorreborre:
WIP, I still need to work on error cases.

### DIFF
--- a/crates/amaru-consensus/src/stages/validate_block/tests.rs
+++ b/crates/amaru-consensus/src/stages/validate_block/tests.rs
@@ -22,10 +22,10 @@ use crate::stages::{
     select_chain::SelectChainMsg,
     test_utils::{te_input, te_state},
     validate_block::test_setup::{
-        assert_trace, setup, te_ledger_contains, te_rollback_ledger, te_send, te_terminate, te_terminated,
-        te_validate_block, test_prep, tm_record_metrics,
+        assert_trace, setup, te_send, te_switch_to_fork_ledger, te_terminate, te_terminated, te_validate_block,
+        test_prep, tm_record_metrics,
     },
-}; // if needed for Literal
+};
 
 #[test]
 fn test_genesis_block_skips_validation() {
@@ -111,7 +111,6 @@ fn test_parent_in_ledger_skips_roll_forward() {
     );
     logs.assert_and_remove(Level::DEBUG, &["validating block"])
         .assert_and_remove(Level::INFO, &["rolling back ledger to common ancestor point"])
-        .assert_and_remove(Level::INFO, &["rolling forward ledger to reach parent"])
         .assert_no_remaining_at([Level::INFO, Level::WARN, Level::ERROR]);
 }
 
@@ -138,16 +137,15 @@ fn test_mock_rollback_fails_terminates() {
         &[
             te_state("vb-1", &prep.state),
             te_input("vb-1", &msg),
-            te_ledger_contains("vb-1", &parent),
-            te_rollback_ledger("vb-1", &parent),
+            te_switch_to_fork_ledger("vb-1", &prep.headers.h2a.point(), &tip.point()),
             te_terminate("vb-1"),
             te_terminated("vb-1", TerminationReason::Voluntary),
         ],
     );
     logs.assert_and_remove(Level::DEBUG, &["validating block"])
         .assert_and_remove(Level::INFO, &["rolling back ledger to common ancestor point"])
-        .assert_and_remove(Level::ERROR, &["ledger volatile DB contains point by rollback failed"])
-        .assert_and_remove(Level::INFO, &["terminated stage"])
+        .assert_and_remove(Level::ERROR, &["failed to switch to fork"])
+        .assert_and_remove(Level::INFO, &["terminated"])
         .assert_no_remaining_at([Level::INFO, Level::WARN, Level::ERROR]);
 }
 
@@ -182,7 +180,6 @@ fn test_grand_parent_in_ledger() {
     );
     logs.assert_and_remove(Level::DEBUG, &["validating block"])
         .assert_and_remove(Level::INFO, &["rolling back ledger to common ancestor point"])
-        .assert_and_remove(Level::INFO, &["rolling forward ledger to reach parent"])
         .assert_no_remaining_at([Level::INFO, Level::WARN, Level::ERROR]);
 }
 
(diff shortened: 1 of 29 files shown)


etorreborre:
@coderabbit review

crates/amaru-consensus/src/stages/validate_block/tests.rs
```diff
@@ -22,10 +22,11 @@ use crate::stages::{
     select_chain::SelectChainMsg,
     test_utils::{te_input, te_state},
     validate_block::test_setup::{
-        assert_trace, setup, te_ledger_contains, te_rollback_ledger, te_send, te_terminate, te_terminated,
+        assert_trace, setup, te_ledger_contains, te_send, te_switch_to_fork_ledger, te_terminate, te_terminated,
         te_validate_block, test_prep, tm_record_metrics,
     },
-}; // if needed for Literal
+};
+// if needed for Literal
```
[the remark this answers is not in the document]
etorreborre: Done.

crates/amaru-ledger/src/block_validator.rs
```diff
@@ -84,26 +94,40 @@ where
...
+                .ok_or(BlockValidationError::new(anyhow!("Block not found in chain store: {point}")))?
+                .decode()
+                .map_err(|e| BlockValidationError::new(e.into()))?;
+            result = self.forward_block(&point, block)?;
+            done += 1;
+            if done % 100 == 0 {
+                tracing::info!(%done, %to_do, "rolling forward ledger to reach fork tip");
+            }
+        }
+        Ok(result)
```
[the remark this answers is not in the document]
etorreborre: fixed

etorreborre:
@CodeRabbit review

crates/amaru-ledger/src/block_validator.rs
```diff
@@ -124,3 +157,104 @@ where
...
+            if result.is_err() {
+                return Ok(result);
+            }
+            done += 1;
+            if done % 100 == 0 {
+                tracing::info!(%done, %to_do, "rolling forward ledger to reach fork tip");
+            }
+        }
+        Ok(result)
+    }
```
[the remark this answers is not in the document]
etorreborre: fixed

crates/amaru-ledger/src/block_validator.rs
```diff
@@ -85,25 +95,50 @@ where
...
+        old_tip: &Point,
+        new_tip: &Point,
+    ) -> Result<Result<LedgerMetrics, BlockValidationError>, BlockValidationError> {
         let mut state = self.state.lock().unwrap();
-        state.rollback_to(to).map_err(|e| BlockValidationError::new(anyhow!(e)))
+
+        let (rollback_point, forward_points) = self.find_rollback_point(&mut state, new_tip)?;
+        let result = self.switch_to(&mut state, &rollback_point, forward_points)?;
+
+        // If we cannot apply a block on the fork, we need to revert the ledger state to what it was.
```
rkuhn: would it not be better for `switch_to` to retain the previous state and restore it upon error?

rkuhn (commented):
Not sure about this direction: yes, this is arguably more correct, but at the same time it locks the ledger for extended periods of time during which nothing else can happen (nothing of relevant, at least). Is it really impossible to just keep a snapshot of the current ledger around while computing the next ledger state in the background?

crates/amaru-ledger/src/block_validator.rs
etorreborre: Definitely but that requires more machinery. Here I was aiming for "slow but correct". Let me investigate if I can go all the way and do enough copying to make this work.

etorreborre: @rkuhn see my latest commits. I don't think that's the end game but this is better than re-applying all the blocks in case of a failed switch.

rkuhn: Will take a look this evening. Unfortunately, this week is a bit difficult for me, next two weeks will be much better.

crates/amaru-ledger/src/state.rs
```diff
@@ -184,724 +103,154 @@ impl<S: Store, HS: HistoricalStores> State<S, HS> {
...
+    /// Modify a *candidate* clone of the current ledger state and atomically
+    /// publish the result if the modifications is successful. The provided closure returns `(R, bool)`:
+    /// - The `R` value is returned to the caller.
+    /// - If `bool` is true then the modified state snapshot is swapped with the old version.
+    ///   If it is `false` that modified state snapshot is just dropped.
+    fn modify<R>(&self, f: impl FnOnce(&mut StateSnapshot<S, HS>) -> (R, bool)) -> R {
+        // Clone the live view into a candidate. The read lock is held only long enough
+        // to bump the inner Arc's refcount; the clone of `StateSnapshot` itself happens
+        // outside the lock because we drop the guard immediately after.
+        let mut candidate = (**self.current.read()).clone();
```
rkuhn, line 150: This is not atomic: the read lock is dropped here already, meaning that the write below can overwrite a different state than we cloned here.

crates/amaru-ledger/src/state.rs
```diff
@@ -184,724 +103,154 @@ impl<S: Store, HS: HistoricalStores> State<S, HS> {
...
-            return Cow::Owned(st.point());
+    /// Sibling of [`Self::modify`] that holds the swap `RwLock` for the **entire**
+    /// duration of the mutation, blocking `load()` calls until the candidate is published
+    /// (or dropped).
+    ///
+    /// Used for mutations that touch both the in-memory overlay *and* the stable store
+    /// in a way that must be observed atomically by readers.
+    fn modify_exclusive<R>(&self, f: impl FnOnce(&mut StateSnapshot<S, HS>) -> (R, bool)) -> R {
+        let mut guard = self.current.write();
+        let mut candidate = (**guard).clone();
```
rkuhn, line 167: This is actually cloning the state bytes, right? Isn’t that quite expensive?

crates/amaru-ledger/src/state.rs
```diff
@@ -62,84 +54,29 @@ pub mod volatile;
...
-        &self.overlay
-    }
+///  - The writer uses the `transaction` method to mutate a candidate clone of the live `StateSnapshot` and atomically
+///    publish it if the modification is successful.
+///
+pub struct State<S, HS> {
+    /// Live view of the ledger. It is wrapped in an `RwLock<Arc<...>>` so that
+    /// readers can take a brief read lock and `Arc::clone` the inner pointer, while the writer
+    /// takes the write lock long enough to publish the next view.
+    current: Arc<parking_lot::RwLock<Arc<StateSnapshot<S, HS>>>>,
```
rkuhn, line 74: This structure basically is the Rust equivalent of software-transactional memory, with all the drawbacks of that design plus the high cost of cloning a StateSnapshot (which even contains the foot gun of internal mutability). I don’t think these technologies should be mixed, it reminds me of the cranberry juice in Die Hard – With a Vengeance.

I think this problem deserves a dedicated meeting with @KtorZ and @yHSJ next week.

yHSJ: > I think this problem deserves a dedicated meeting with @KtorZ and @yHSJ next week.

Agreed! Tomorrow's meeting may be a good time to schedule this chat (or just have it)

crates/amaru-ledger/src/state.rs
etorreborre: That shouldn't be the case since there's only one writer ever by design.

crates/amaru-ledger/src/state.rs
etorreborre: The cloning is cheap here because `Arc` references have been introduced here and there. 
There is some cloning involved though. The idea is to use some `copy-on-write` some data like rewards and pool updates, but that happens only at specific times during an epoch.

etorreborre:
Closing this PR for now. I'll open another PR based on the latest changes for the ledger in #975.

PR #977, opened 2026-06-22
Populate ValidationContext from test vectors

yHSJ:
Previously, we have been passing `Default::default` for any non-UTxO state in our test vectors. This PR moves the decoding logic from the bootstrap to the kernel so we can reuse them here, and then provides real values to the `ValidationContext` we create for each fixture.

In practice, this is effectively the same, since all the vectors are considered a "genesis state", and any cert/gov state must happen during the series of transactions in the fixture. However, this change allows us to easily update the initial state if needed, without running into the foot-gun of `Default::default`.

We haven't yet wired up the `PassEpoch` transitions from the test vectors, so we do not have any state that is calculated in the epoch transition in our final state. Which is OK, since we're not even considering the final state in these tests yet. 

## Summary by CodeRabbit

* **New Features**
  * Added constitutional committee authorization and state support for governance data handling
  * Introduced snapshot conversion utilities for synthesizing ledger state (account state, committee members, proposals)

* **Refactor**
  * Streamlined snapshot bootstrap interface and consolidated decoder implementations


yHSJ:
@coderabbit review

Issue #971, opened 2026-06-17
PlutusV2 missing IntegerToByteString/ByteStringToInteger costs


yHSJ:
### What revision are you using?

9f2a14aa2c9fc85690f30e4177cb636073ac8c24

### What operating system are you using?

MacOS

### What architecture are you using?

aarch64/armv8

### What problem are you having? What should be the expected behaviour?

`is_available_in` permits the `IntegerToByteString` and `ByteStringToInteger` builtins for PlutusV2 (as we should), but there is no arm in `get_cost` for those builtins, meaning it will lead to a runtime error when evaluating a script.

### Please confirm that you've already searched issues, CHANGELOG and available documentation.

- [x] I searched, but no luck.

yHSJ:
This was completed in #988

PR #982, opened 2026-06-25
chore: add missing costs

jonathanlim222:
Fixes issue [971](https://github.com/pragma-org/amaru/issues/971), adds the missing costs and tests to verify.

## Summary by CodeRabbit

* **New Features**
  * Extended built-in cost modeling with new costing entries for integer/byte-string conversions and additional hashing/signature-related built-ins (including blake2b_224, keccak_256, ECDSA and Schnorr verification).
* **Tests**
  * Updated default builtin cost fixtures and assertions for both supported protocol versions to reflect the new cost keys and vector layout.
* **Bug Fixes**
  * Improved cost lookup behavior by replacing wildcard matching with explicit handling, so missing cost entries are more clearly surfaced when new built-ins are added.
* **Chores**
  * Fixed formatting in the unreleased changelog entry.


crates/amaru-uplc/src/machine/cost_model/builtin_costs/builtin_costs_v2.rs
```diff
@@ -968,6 +1002,14 @@ impl BuiltinCostModel for BuiltinCostsV2 {
...
             )),
+            DefaultFunction::IntegerToByteString => Some(ExBudget::new(
+                self.integer_to_byte_string.mem.cost([args[0], args[1], args[2]]),
+                self.integer_to_byte_string.cpu.cost([args[0], args[1], args[2]]),
+            )),
+            DefaultFunction::ByteStringToInteger => Some(ExBudget::new(
+                self.byte_string_to_integer.mem.cost([args[0], args[1]]),
+                self.byte_string_to_integer.cpu.cost([args[0], args[1]]),
+            )),
             _ => None,
```
yHSJ: I don't love the wildcard here (not part of your changes, I know).  We should have an explicit either cost or `None` for every builtin. That would also remove the need for the 100+ line test that was introduced here; the compiler would enforce we at *least* have a branch for the builtin and, in theory, conformance tests would ensure the value is correct.

crates/amaru-uplc/src/machine/cost_model/builtin_costs/mod.rs
```diff
@@ -341,6 +346,152 @@ mod tests {
         assert_eq!(budget, ExBudget::new(4, 116711 + 1957 * 5));
     }
 
+    const ALL_BUILTINS: &[DefaultFunction] = &[
```
yHSJ: If we don't have a wildcard in the get_cost match, we won't need this test. If we want to keep a test like this, we need to find a nicer way to iterate over the builtins, since this requires constant manual maintenance, and is exactly the kind of thing that often gets overlooked.

yHSJ (changes requested):
Instead of a long and manually maintained test, let's take advantage of the compiler and avoid using the wild card in the match expression.

Also, please add a changelog entry

KtorZ:
I started reviewing this PR and ended up writing #988. Many issues with the uplc crate at the moment; so this is a first pass to fix most of them; but also cleanup a bit the internal. I'll do another pass later.

PR #985, opened 2026-06-26
feat: upgrade extractor to match cardano-node version

jeluard:
Upgrade the haskell based extractor so that it can be used with ledger state produced by the db-analyser version we use.

Also:
* allow to customize the generated data location
* check in CI it actually works
* `make test` to test the final binary; not hooked in CI yet due to missing ledger state generated with proper cardano_node version

## Summary by CodeRabbit

* **New Features**
  * Added CI conformance testing for tooling against a preprod snapshot.
  * Added `--output DIR` support to the snapshot extractor, organizing generated results by network.
* **Bug Fixes**
  * Updated fixture output layout (pots, pools, nonces, DReps, rewards provenance, delegates) and adjusted snapshot loading to read ledger state from the snapshot’s `state/` directory.
  * Added explicit handling for unsupported snapshot eras (including Dijkstra).
* **Tests**
  * Added a `make test` target with snapshot-path validation and expected output checks.
* **Documentation**
  * Added a concrete “Run” example command to the extractor README.


.github/workflows/continuous-integration.yml
```diff
@@ -153,6 +153,53 @@ jobs:
...
+          sudo make -C conformance-tests/haskell-node-extractor secp256k1 PREFIX=/usr/local
+          sudo make -C conformance-tests/haskell-node-extractor blst PREFIX=/usr/local
+      - name: Test haskell-node-extractor
+        shell: bash
+        env:
+          LD_LIBRARY_PATH: /usr/local/lib:${{ env.LD_LIBRARY_PATH }}
+        run: |
+          make -C conformance-tests/haskell-node-extractor build
+          # Tests are disabled for now as we don't have snapshots compatible with the current version of the extractor yet.
+          # make -C conformance-tests/haskell-node-extractor test SNAPSHOT="$SNAPSHOT" NETWORK=preprod
```
KtorZ: Not sure about putting this one in the continuous-integration workflow; this isn't a project that changes often enough (and it'll likely not change at all for a long period once we're done with the tweaks). 

Perhaps we should move it as a workflow dispatch, store the binary on Cloudflare or on the download page (make it part of our downloadable artifacts) and only rebuild on-demand. We need not to rebuild this continuously on every PR.

crates/amaru/tests/stake-distributions/haskell-node-extractor/src/Data/NetworkName.hs
```diff
@@ -1,10 +1,12 @@
 module Data.NetworkName
     ( NetworkName (..)
     , networkNameToNetwork
+    , networkNameToText
     ) where
 
 import Relude
-    ()
+    ( Text
+    )
```
KtorZ, line 10:
```suggestion
```

Relude is a Prelude-like module, it's meant to be important unqualified.

conformance-tests/haskell-node-extractor/src/Query/DelegateRepresentatives.hs
```diff
@@ -92,12 +94,16 @@ mergeStateAndStake dRepStates dRepStakes =
...
     -> Map.Map DRep DelegateRepresentative
 mergeDelegations dReps dRepDelegations =
     Merge.merge
         (Merge.mapMissing $ \_ -> identity)
-        (Merge.mapMissing $ \drep _ -> error ("DRep has delegation but not stake or state: " <> show drep))
+        (Merge.mapMaybeMissing $ \drep delegators ->
+            case drep of
+                DRepAlwaysAbstain        -> Just (setDelegators delegators (predefinedDRep drep mempty))
+                DRepAlwaysNoConfidence   -> Just (setDelegators delegators (predefinedDRep drep mempty))
+                _                        -> Nothing)
```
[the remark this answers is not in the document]
KtorZ: I agree with the rabbit, we should crash loudly if that case it hit, not silently ignore the value.

.github/workflows/continuous-integration.yml
jeluard: We could move this to a nightly flow then. As we probably don't need to distribute this binary (hopefully it will be integrated in the automated snapshots flow) I would remove the complexity of dealing with publishing releases for this.

KtorZ: Agreed. But I wouldn't put this as a nightly, it will change very rarely. A workflow dispatch is probably all that we need.

PR #988, opened 2026-06-27
Cleanup and fix amaru uplc

KtorZ:
Supersedes #982 .

I was looking into confirming that the cost models handling is okay for the next hard fork, and ended up rewriting a quite large part of the amaru-uplc crate... What motivated the rewrite (and what's in this "small" PR):

1. Wrong cost models

    Many (still) wrong cost models, due to the lack of proper handling of the machine semantics. Before, we only distinguised between two semantics: V1 and V2; but we have 5: A, B, C, D and E -- even though we shall never really encounter A anymore since we don't run v9 and before. 

2. Difficult updates / maintenance

    Updating the cost model was really unpractical; the stringly-typed map that we used is extremely error prone: no exhaustiveness check, easy to mistype a key (invisible at compile type), and also, it introduces an unnecessary overhead in parsing and in the establishment of the cost models.

3. Confusing versions

    We have a plutus version, protocol version, machine version and semantics version. And it was really mixed up in a few place (e.g. the machine version referring to "v3" instead of "v1.1.0"). We would sometime take in a PlutusVersion when what we wanted was a protocol version, and vice-versa. The machine version was also unnecessarily allocated on the arena; adding its own set of problems.

4. Version 9 usages

     In many places, we were still relying on protocol v9 in a few places, whereas we dropped support for it a while ago already. Many tests were built on top of v9 and haven't been updated since. Yet cleaning up v9 required some finesse here and there; in particular because of some golden tests that depended on v9 without actually requiring it. 

5. Wrong handling of builtins

    Partly because of (3) and (4), we allowed the VM to handle builtins that shouln't have existed for a given protocol version. There was no coherence between the cost models and the runtime execution. Note that here, I am still not entirely satisfied with the current state: the VM will now crash with an integer overflow if one is trying to run a program with a builtin that isn't covered by the protocol. This should not happen _in practice_, because the program will simply fail to parse to begin with (test coverage increased slightly on that front). Adding a `if_protocol_v11` guard on every branch of the VM seems also ill-advised (although that might be a 'premature optimisation').  Anyway, we need better ways of guaranteeing which builtins are available based on a given protocol version. 

## Summary by CodeRabbit

* **New Features**
  * Added support for protocol version 11, including updated default protocol version constants.
  * Updated cost-model and builtin-availability behavior to be driven by protocol/version selection and the machine version used for evaluation.
* **Bug Fixes**
  * Fixed script integrity hashing and phase-one/phase-two validation to use the correct cost-model inputs.
  * Improved decoding/evaluation consistency for Plutus scripts across protocol versions.
* **Tests & Benchmarks**
  * Updated conformance, unit tests, and benchmarks to match the new decoding/evaluation flow and defaults.


crates/amaru-kernel/src/cardano/plutus_script.rs
```diff
@@ -12,43 +12,13 @@
...
-    let plutus_version = reify_plutus_version::<V>()
-        .ok_or_else(|| FlatDecodeError::Message(format!("unable to reify type-level Plutus version '{V:#?}' ??!")))?;
-
-    let program = match plutus_version {
-        PlutusVersion::V3 => flat::decode_strict::<DeBruijn>(arena, bytes, plutus_version, pv),
-        PlutusVersion::V1 | PlutusVersion::V2 => flat::decode::<DeBruijn>(arena, bytes, plutus_version, pv),
-    }?;
-
-    Ok((program, plutus_version))
-}
```
KtorZ, line 54: Moved to amaru-uplc, under the flag::decode module due to a circular dependency otherwise. This is also somewhat more coherent as it puts the kernel back at the root.

crates/amaru-kernel/src/cardano/script_integrity_data.rs
```diff
@@ -119,62 +117,64 @@ impl fmt::Display for ScriptIntegrityData {
...
+    #[test_case(fixture_cost_models(), &[Language::PlutusV3], Some("472370a6360002ce23e55c8e593754bca8cfb21c57c26eb16ff33049d61230ff") ; "v3 only")]
+    #[test_case(fixture_cost_models(), &[Language::PlutusV1, Language::PlutusV2], Some("b099d797177cb401334fa99d6fae4bb50c3789515865797bdddc7c00e43e0429") ; "v1 + v2")]
+    #[test_case(fixture_cost_models(), &[Language::PlutusV2, Language::PlutusV1], Some("b099d797177cb401334fa99d6fae4bb50c3789515865797bdddc7c00e43e0429") ; "v2 + v1 sorts to v1 + v2")]
+    #[test_case(fixture_cost_models(), &[Language::PlutusV1, Language::PlutusV1], Some("450734f2bd32c2272c31e6a3c781104ee69040c89da0308e613674203a457e86") ; "v1 + v1 dedups to v1")]
+    #[test_case(no_v1_cost_models(), &[Language::PlutusV1], Some("ac871208ce24bcee5ca68ab58c321750fa80f73bfb4e5d9a6acc2b1e3817dcc4") ; "v1 only with no v1 cost model")]
+    #[test_case(no_v2_cost_models(), &[Language::PlutusV2], Some("b96a45ab8016ded02c3fbf1939c695e86155a961c83fc695aba5316086ec47e3") ; "v2 only with no v2 cost model")]
+    #[test_case(no_v3_cost_models(), &[Language::PlutusV3], Some("302569f5de0c5cd3d1816ff8b95a9117398227639a0425915ee4f68b590382a1") ; "v3 only with no v3 cost model")]
```
KtorZ, line 163: I adjusted the values here; because the tests would otherwise continuously fail each time we would update the default values for the cost models; and that's just irrelevant. So they are now based on fixture cost models and thus, required new "golden" values.

crates/amaru-uplc/src/builtin/default_function.rs
```diff
@@ -366,145 +366,31 @@ impl DefaultFunction {
     ///
     /// Follows the Haskell's `builtinsIntroducedIn` mapping from plutus-ledger-api:
     /// <https://github.com/IntersectMBO/plutus/blob/master/plutus-ledger-api/src/PlutusLedgerApi/Common/Versions.hs>
-    pub fn is_available_in(&self, plutus_version: PlutusVersion, pv: u32) -> bool {
+    pub fn is_available_in(&self, protocol_version: ProtocolVersion) -> bool {
```
KtorZ, line 369: Vastly simplified by simply considering V10 and V11; since we should never need to parse programs in an older protocol version. Still; I've left an explicit `unreachable!` assertion in case this happen so that we don't accidentally swallow a v9 or earlier program and coerce it into a v10/v11. 

In the long run, we might want to revisit; especially if people are interested in doing historical validations with bits of Amaru. But at this point, we don't have to support more than what we need.

crates/amaru-uplc/src/machine/cost_model/builtin_costs/builtin_costs_v1.rs
KtorZ, line 1: Unified under a single `BuiltinCosts` struct

crates/amaru-uplc/src/machine/cost_model/builtin_costs/builtin_costs_v2.rs
KtorZ, line 1: Unified under a single `BuiltinCosts` struct

crates/amaru-uplc/src/machine/cost_model/builtin_costs/builtin_costs_v3.rs
KtorZ, line 1: Unified under a single `BuiltinCosts` struct

crates/amaru-uplc/src/machine/cost_model/builtin_costs.rs
```diff
@@ -0,0 +1,1389 @@
...
+    fn default() -> Self {
+        CostModel::v3().builtin_costs
+    }
+}
+
+impl BuiltinCosts {
+    #[allow(clippy::disallowed_types)]
+    pub fn new(
+        cost_map: &HashMap<ParamName, i64>,
+        semantics: Semantics,
```
KtorZ: This is the main reason for all this... ☝️

crates/amaru-uplc/src/machine/cost_model/costing.rs
```diff
@@ -44,134 +44,46 @@ impl Cost<1> for OneArgument {
...
-        TwoArguments::MultipliedSizes(MultipliedSizes { intercept, slope })
-    }
-
-    pub fn subtracted_sizes(intercept: i64, slope: i64, minimum: i64) -> TwoArguments {
-        TwoArguments::SubtractedSizes(SubtractedSizes { intercept, slope, minimum })
-    }
-
-    pub fn linear_on_diagonal(constant: i64, intercept: i64, slope: i64) -> TwoArguments {
-        TwoArguments::LinearOnDiagonal(ConstantOrLinear { constant, intercept, slope })
-    }
```
KtorZ, line 117: I got rid of all those smart constructors for the various costing functions because they are _dangerous_. Given that everything is typed as `i64` here, we should prefer labels and record construction over positional arguments. These are a recipe for mistakes.

crates/amaru-uplc/src/machine/runtime.rs
```diff
@@ -142,7 +120,7 @@ where
     }
 }
 
-impl<'a, B: BuiltinCostModel> Machine<'a, B> {
+impl<'a> Machine<'a> {
     pub fn call<V>(&mut self, runtime: &'a Runtime<'a, V>) -> Result<&'a Value<'a, V>, MachineError<'a, V>>
```
KtorZ, line 124: Many unfortunate changes in this file as I re-ordered the branches to match the order in which the builtins are declared in the cost model (v3). This was originally meant to simplify a bit the gatekeeping of builtins not available in specific protocol versions; but I ended up keeping the gatekeeping only at the parsing level (debatable but, happy to debate). 

There are a couple of other necessary changes which I'll try to point out.

crates/amaru-uplc/src/machine/runtime.rs
```diff
@@ -2848,7 +2788,15 @@ impl<'a, B: BuiltinCostModel> Machine<'a, B> {
...
                     .builtin_costs
-                    .get_cost(DefaultFunction::InsertCoin, &[ledger_value::value_max_depth(v)])
+                    .get_cost(
+                        DefaultFunction::InsertCoin,
+                        &[
+                            value::byte_string_ex_mem(ccy),
+                            value::byte_string_ex_mem(tok),
+                            value::integer_ex_mem(qty),
+                            ledger_value::value_max_depth(v),
+                        ],
```
KtorZ, line 2798: This is one important fix; the `InsertCoin` builtins uses a 4-arg costing function; so it needs all of 4 args, not just the depth.

PR #989, opened 2026-06-29
feat: test external tools in CI

jeluard:
## Summary by CodeRabbit

* **Chores**
  * Added a new CI integration check to validate end-to-end tool workflows.
  * Runs only for non-draft pull requests, starts a local node using preproduction bootstrap assets, and verifies it becomes reachable.
  * Executes the Oura tooling against the local endpoint and enforces expected results, including at least one output event.
  * When failures occur, captures and uploads relevant logs/artifacts to speed up troubleshooting.

### DIFF
--- a/.github/dolos.toml
+++ b/.github/dolos.toml
@@ -1,0 +1,20 @@
+[upstream]
+peer_address = "127.0.0.1:3000"
+
+[storage]
+version = "v3"
+path = "dolos-data"
+
+[genesis]
+byron_path = "cardano-node-config/byron-genesis.json"
+shelley_path = "cardano-node-config/shelley-genesis.json"
+alonzo_path = "cardano-node-config/alonzo-genesis.json"
+conway_path = "cardano-node-config/conway-genesis.json"
+
+[chain]
+type = "cardano"
+magic = 1
+is_testnet = true
+
+[sync]
+sync_limit = { MaxBlocks = 10 }
--- a/.github/workflows/tool-integrations.yml
+++ b/.github/workflows/tool-integrations.yml
@@ -1,0 +1,224 @@
+name: Tool Integrations
+
+on:
+  schedule:
+    - cron: '0 3 * * *'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  actions: write
+
+env:
+  LEDGER_CACHE_VERSION: "v13"
+
+jobs:
+  prepare:
+    name: Build Amaru & restore bootstrap DBs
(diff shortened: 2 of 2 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -153,6 +153,183 @@ jobs:
...
+      - uses: actions/checkout@v6
+        with:
+          submodules: True
+
+      - uses: ./.github/actions/cache
+
+      - name: Build Amaru
+        shell: bash
+        run: |
+          cargo build --release --bin amaru --locked
```
KtorZ: That is going to kill the CI time I believe; wouldn't it make sense to add this job as a sanity check on top of the nightly release builds? 

It failing may be independent of our will but simply signaling some changes upstream; so we should also not condition CI passing to that.

.github/workflows/continuous-integration.yml
```diff
@@ -153,6 +153,90 @@ jobs:
...
+      LEDGER_CACHE_VERSION: "v13"
+      AMARU_NETWORK: preprod
+      AMARU_CHAIN_DIR: ./chain.preprod.db
+      AMARU_LEDGER_DIR: ./ledger.preprod.db
+      AMARU_LISTEN_ADDRESS: 127.0.0.1:3000
+      AMARU_PEER_ADDRESS: preprod-node.play.dev.cardano.org:3001
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          submodules: True
```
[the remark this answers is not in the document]
KtorZ: Not a bad idea given that we then invoke untrusted bash scripts afterwards.

.github/workflows/continuous-integration.yml
```diff
@@ -153,6 +153,183 @@ jobs:
...
+                echo "conway_path = \"$(realpath cardano-node-config/conway-genesis.json)\""
+                echo ''
+                echo '[chain]'
+                echo 'type = "cardano"'
+                echo 'magic = 1'
+                echo 'is_testnet = true'
+                echo ''
+                echo '[sync]'
+                echo 'sync_limit = { MaxBlocks = 10 }'
+              } > dolos.toml
```
KtorZ: Isn't the goal to use the side config file instead?

.github/dolos.toml
KtorZ, line 1: I don't think we should put anything under `workflow` that isn't a worfklow; historically also, we've been putting stuff like this at the root of the `.github`

KtorZ (approved):

PR #994, opened 2026-06-30
doc: update the tracing EDR and its relationship with the observability EDR

etorreborre:
This PR updates the EDRs for tracing spans, EDR-26, and observability, EDR-07.

There are some formatting changes due to my IDE (line breaks) but the most relevant changes are:

 - The recommended use of `debug_span!` and `info_span!` for creating spans with a macro.
 - The description of the relation between spans schema definition and the creation of the target + name of a span.
 - Some adjustments for the target names (in particular, there's only `amaru::consensus` and no mention of stage names for that target.
 - I changed some previous span names to follow a bit more a pattern like `state.object.action` (in any case, I think that EDR-26 just shows guidelines for targets and span names, it doesn't have to reflect the exact situation IMO).
 
 The [tracing spans PR](https://github.com/pragma-org/amaru/pull/996) is an application of those 2 modified EDRs.

skip-changelog

## Summary by CodeRabbit

* **Documentation**
  * Updated observability guidance with a standardized, hierarchical tracing schema, including revised span construction/instrumentation examples for both synchronous and asynchronous code.
  * Refined tracing span naming conventions and logging-level expectations, clarifying uniqueness across targets and improved schema-based classification via tags.
  * Expanded tracing guidance across network, consensus, ledger, mempool, and protocol activity, including header trace-ID/attribute handling and updated performance-span naming.

### DIFF
--- a/engineering-decision-records/007-observability.md
+++ b/engineering-decision-records/007-observability.md
@@ -7,19 +7,13 @@ status: accepted
 
 ## Context
 
-This decision record documents a general organization for how the Amaru node will collect and report critical
-operational metrics, logs, and tracing data.
+This decision record documents a general organization for how the Amaru node will collect and report critical operational metrics, logs, and tracing data.
 
-A key component of operating any computing system at scale is observability: tracking what happened, how often something
-happened, how long it took, how many resources something is consuming, and where potential issues or bottlenecks might
-be found.
+A key component of operating any computing system at scale is observability: tracking what happened, how often something happened, how long it took, how many resources something is consuming, and where potential issues or bottlenecks might be found.
 
-Thus, we would like Amaru to track and report these traces and metrics. We would also like to use industry standards so
-that users of the Amaru node can swap out their own visualization and aggregation infrastructure as needed.
+Thus, we would like Amaru to track and report these traces and metrics. We would also like to use industry standards so that users of the Amaru node can swap out their own visualization and aggregation infrastructure as needed.
 
-We would like the codebase to organize the metrics it tracks in a simple, consistent and modular way, such that each
-module or crate can own a subset of the metrics, while at the same time being consistent and discoverable through
-reading the code.
+We would like the codebase to organize the metrics it tracks in a simple, consistent and modular way, such that each module or crate can own a subset of the metrics, while at the same time being consistent and discoverable through reading the code.
 
 ## Decision
 
@@ -27,9 +21,7 @@ reading the code.
 
 - By default, the application provides human-readable information on stderr in a pretty format (possibly TUI).
 
-- Additionally, the application will
-  embrace [observability](https://peter.bourgon.org/blog/2017/02/21/metrics-tracing-and-logging.html) and provide logs,
-  traces, spans and metrics using Rust's [tracing](https://docs.rs/tracing/latest/tracing/index.html) ecosystem.
+- Additionally, the application will embrace [observability](https://peter.bourgon.org/blog/2017/02/21/metrics-tracing-and-logging.html) and provide logs, traces, spans and metrics using Rust's [tracing](https://docs.rs/tracing/latest/tracing/index.html) ecosystem.
 
 - All traces and spans can be made available on stdout in a structured format (JSON).
 
@@ -38,11 +30,9 @@ reading the code.
         - [Jaeger](https://www.jaegertracing.io/)
         - [Grafana](https://grafana.com/) and [Tempo](https://grafana.com/oss/tempo/)
 
-    - We will use the [opentelemetry-rust](https://github.com/open-telemetry/opentelemetry-rust) crate to collect and
-      report traces and metrics.
+    - We will use the [opentelemetry-rust](https://github.com/open-telemetry/opentelemetry-rust) crate to collect and report traces and metrics.
 
-    - We will make judicious use of [Spans](https://opentelemetry.io/docs/concepts/observability-primer/#spans) to
-      expose the structured nature of the workload the Amaru node performs.
+    - We will make judicious use of [Spans](https://opentelemetry.io/docs/concepts/observability-primer/#spans) to expose the structured nature of the workload the Amaru node performs.
 
 - We define the frontier between logs and traces by following a simple rule:
     - Any event at the DEBUG level or above is considered a log
@@ -50,13 +40,11 @@ reading the code.
 
 ### Tracing Schemas
 
-To ensure consistency and enable compile-time validation of tracing instrumentation, we use a schema-based approach
-implemented in the `amaru-observability` crate.
+To ensure consistency and enable compile-time validation of tracing instrumentation, we use a schema-based approach implemented in the `amaru-observability` crate.
 
 #### Schema Definition
 
-Schemas are defined using the `define_schemas!` macro in a central location (`amaru-observability/src/schemas.rs`).
-They are organized hierarchically with five levels of nesting:
+Schemas are defined using the `define_schemas!` macro in a central location (`amaru-observability/src/schemas.rs`). They are organized hierarchically with five levels of nesting:
 
 - The first two levels define the target of a span, for example `amaru::consensus` or `amaru::ledger`.
 - The next three levels define the name of the span, for example `header.evolve_nonce` or `block.apply` (
@@ -73,15 +61,15 @@ define_schemas! {
                 VALIDATE {
                     required issuer_key: String
                 }
-            }            
+            }
         }
-        ledger {           
+        ledger {
             block {
                 APPLY {
                     required point_slot: u64
                     optional error: String
-                }              
-            }               
+                }
+            }
         }
     }
 }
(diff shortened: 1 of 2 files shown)


engineering-decision-records/007-observability.md
jeluard: Same for `info_span!` and `trace_span!` ?

etorreborre: Yes, I updated the doc to clarify.

rkuhn:
Please don’t reformat .md files. In particular, using some automatic line wrapping basically makes all diffs useless, even if everyone were to use the same configuration. Markdown is meant as a vastly simplified way of writing HTML, which is meant to capture semantics and display accordingly. Therefore, line breaks in the .md source should be semantically meaningful, not an artefact of the editor settings. Please configure your editor to display long lines in the way you want, without changing newline characters in the file.

engineering-decision-records/026-tracing-span-design.md
rkuhn, line 24: The mention of the span message was intentional and important.

rkuhn: Even if line wrapping were a thing, ending a line with an opening parenthesis is just wrong because what follows (within prose) is never whitespace.

rkuhn: I left out INFO deliberately because spans tend to be created from things happening frequently, which should not be at INFO level. What we show to the node admin should be messages (a.k.a. events), not spans. Spans are used for adding contextual information to messages.

engineering-decision-records/026-tracing-span-design.md
```diff
@@ -18,116 +20,179 @@ This section is structured by the kind of traces under consideration:
...
+> operation; `TRACE` level is removed from release binaries by compile-time options.
 >
-> **For these reasons all nominally emitted trace spans intended for production use MUST be emitted at `DEBUG` level!**
+> **For these reasons all nominally emitted trace spans intended for production use MUST be emitted at `DEBUG` or
+the `INFO` levels!**
 >
-> Since OpenTelemetry treats logs (i.e. tracing events a.k.a. log messages) as individually important, all debug logging that may occur at high rate MUST be emitted at `TRACE` level.
+> Since OpenTelemetry treats logs (i.e. tracing events a.k.a. log messages) as individually important, all debug logging
+> that may occur at high rate MUST be emitted at `TRACE` level.
+
```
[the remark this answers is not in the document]
rkuhn: No, high rate will overload the telemetry system and must therefore be at the (discarded) TRACE level. Otherwise the production system is broken.

engineering-decision-records/026-tracing-span-design.md
rkuhn: This span name seems to be intended for nesting within a span for running a certain stage, so that that information is also available in the trace, albeit not in this individual span. That’s fine, but it raises the question of what the `state` is then doing in the name, since decoding headers is stateless. And if we’re going in the direction of using this string to be displayed on trace graphs, then the double-colon separator isn’t typographically useful because it has the same visual height as most of the letters. I’d rather use a glyph with very little height, like dot or dash or underscore:

- `amaru.consensus.header.decode`
- `amaru-consensus-header-decode`
- `amaru_consensus_header_decode`

rkuhn: We’ll need spans matching the message processing of the stage (as oppposed to the logical processing steps) in order to see whether the pure-stage graph has performance problems or optimisation potential. The logical processing steps should then be spans nested into the per-stage spans as appropriate. This way we get both views, including the link between them. What is your reasoning for removing the per-stage spans?

rkuhn, line 136: Yes, protocols are interesting, but why remove peer management? It obviously only exists at this point in a rudimentary way, but that is supposed to change over the coming 1–2 months.

rkuhn: Network connections are not treated as first-class objects in Ouroboros, that role is taken by peer management, dictating how a connection is used (including when to establish or tear down).

engineering-decision-records/007-observability.md
rkuhn: Not sure whether this URL is the canonical one — isn’t `docs.rs/tracing` what is guaranteed to show the docs for the latest version?

rkuhn: I’ve always been confused by this overlap in naming with the `tracing` crate. Since we’re changing things quite thoroughly, wouldn’t it be better to name the macros differently, like `amaru_info!` etc.?

rkuhn: Not sure about `trace_span!` given our reasoning for how to make production Amaru observable.

rkuhn: What is the Sync module?

rkuhn (dismissed):
Not sure whether I missed an actual change, and the review felt like I’m probably missing some discussions that happened while I was incapacitated. Would love to talk about it tomorrow, let’s see when it fits.

engineering-decision-records/026-tracing-span-design.md
etorreborre: The ledger has `info_span!` statements that are created at that level on purpose, I validated that with @KtorZ. Note that spans also appears as log events in the console.

etorreborre: The `::` separator gets translated to a `.` in the actual span name. It's only a `::` when denoting a schema entry. 

Re. `state` in `state.header.decode` I can choose another example for the EDR. Let's discuss the EDR application in the other PR.

engineering-decision-records/026-tracing-span-design.md
```diff
@@ -18,116 +20,179 @@ This section is structured by the kind of traces under consideration:
...
+```
+amaru::consensus::state::header::decode
+```
+
+In this example, the target is `amaru::consensus` and the name is `state.header.decode`.
+Similarly for the ledger, we can find a schema entry like:
+
+```
+amaru::ledger::state::validation_context::create
+```
```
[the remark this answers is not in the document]
etorreborre: Done.

engineering-decision-records/026-tracing-span-design.md
etorreborre: The motivation was to get some consistency across the whole naming scheme. Something like `target1.target2.category.entity.action`. If we want a "stage-oriented" view we can use the input message as the "entity", like `state.validate_block_msg.process`.

etorreborre: I left it out because we don't have such spans at the moment (and I've added the network section because we have them)

rkuhn: If this is about printing something on the console with default logging configuration then `tracing::info!()` might be more appropriate — unless the span timings are actually meaningful to the casual observer and the intended message to be displayed.

rkuhn: The `state` issue doesn’t go away by using different examples: the screenshot you showed nicely illustrates that `*.state.*` in all spans doesn’t carry any information, which is the reason why I am arguing for its removal. A span named `header.decode` obviously starts before decoding and ends after finishing decoding — I have trouble imagining any variation on this theme that could be marked by different prefixes.

The justification discussed in and after Porto was the ability to filter for certain types of work. Back then, I didn’t have a better proposal, now I’d say that `cpu` would be a much better marker than `state`. Not sure whether that needs to be in the name, though — could it be a property? My goal is to make good use of screen real estate when looking at traces, and it feels wrong to waste space for six glyphs (or even four).

@KtorZ Since we’re rewriting this EDR, can we talk about improving this part?

rkuhn: The category wouldn’t be `state` here, it wouldn’t even be `cpu` if we were to agree on my proposal above, it would be `stage` or similar, which covers both CPU usage and waiting for async effects. And I’d want the stage name there, and the message variant name as well, probably as a property (since it is dynamic).

Checking [the docs](https://docs.rs/tracing/latest/tracing/#configuring-attributes) it is pretty clear that the target should be something like a module (i.e. a coherent collection of event sources) while the name usually is the filename and line number — and I also saw that the name doesn’t apply to spans, only events. So the natural target is `amaru::consensus::track_peers` and the message variant name or trace ID should be properties. Enlarging this to `amaru::consensus::stage::track_peers::msg` doesn’t add any value as far as I can see; filtering for all these could be done using the `msg_name` (or whatever we choose) property not being empty.

rkuhn: The EDR should cover the intended kinds of spans, I think.

KtorZ: This is the latter @rkuhn; we use `info_span` in the ledger mostly around epoch transitions, as a way to measure the time it takes to perform certain operations (stake distribution calculation, overlay on-disk saving, rewards payouts, ratification, etc...).

engineering-decision-records/007-observability.md
KtorZ:
```suggestion
- The next two levels define the name of the span, for example `header.evolve_nonce` or `block.apply` (
```

KtorZ, line 89: Does the filtering work even if we have a `tags` list?

KtorZ, line 105:
````suggestion
Amaru now prefers explicit span creation over function-wide instrumentation wrappers. New schema-based tracing should use `debug_span!` or `info_span!` to create a span at the point where the work actually begins, then either enter that span or attach it to a future with `.instrument(...)`.

Note that even though the schema compilation generates full names like `amaru_observability::amaru::consensus::header::EVOLVE_NONCE`, the `info_span!/debug_span!/trace_span!` macros only requires the second target name and the span name, e.g. `consensus::header::EVOLVE_NONCE`.

```rust
fn evolve_nonce(&self, hash: String) -> Result<Nonce, ConsensusError> {
    let span = debug_span!(consensus::header::EVOLVE_NONCE, hash = &hash);
let _guard = span.enter();
```
````

Comment (KtorZ) [inline engineering-decision-records/007-observability.md]:
```suggestion
.instrument(debug_span!(ledger::block::APPLY, point_slot = point_slot))
```

KtorZ:
```suggestion
    let span = debug_span!(ledger::block::APPLY, point_slot = point_slot);
```

KtorZ:
```suggestion
        ledger::block::APPLY,
```

engineering-decision-records/026-tracing-span-design.md
KtorZ: In a previous conversation, we agreed with @etorreborre to settle on what we _have_ currently, and get the code to already reflect that. Then later add more targets if necessary.

I don't mind documenting `amaru::peer` as a target since we already know it *will* end up there. But I also don't mind us adding it later.

engineering-decision-records/007-observability.md
etorreborre: Actually, I removed `state` where it doesn't make sense but kept 3 levels for protocols. So the bullet points become:
```
- The first two levels define the target of a span, for example `amaru::consensus` or `amaru::ledger`.
- The other levels (two or three in practice) define the name of the span, for example `header.evolve_nonce` or `block.apply` (
  see [EDR-026](./026-tracing-span-design.md) for more details).
```

etorreborre: Yes, I added 2 examples to show how to `AND` and how to `OR` tags:
```
- To select the spans carrying _either_ tag (OR), use one directive per tag: `AMARU_LOG='[{amaru.tag.db=true}]=trace,[{amaru.tag.io=true}]=trace'` enables all the spans tagged with `db` plus all the spans tagged with `io`, because each directive is evaluated independently.
- To select only the spans carrying _all_ the tags (AND), list the tags inside a single directive: `AMARU_LOG='[{amaru.tag.cpu=true,amaru.tag.db=true}]=trace'` only enables the spans tagged with both `cpu` and `db`.
```

etorreborre: Done.

etorreborre: Done.

etorreborre: Done.

etorreborre: Done.

etorreborre: Not sure, but right now only `LedgerMetrics` exists, so I fixed the sentence.

etorreborre: The intention is to direct to the open telemetry concepts (if you're referring to [this link](https://opentelemetry.io/docs/concepts/observability-primer/#spans)) not the actual code.

KtorZ:
@rkuhn can we merge this one or do you still have remarks/concerns?

PR #996, opened 2026-06-30
feat: tracing spans

etorreborre:
This PR brings the code base up to date with EDR-026 on tracing (in its updated version [here](https://github.com/pragma-org/amaru/pull/994)). The main changes are (sorry, pretty large PR!):

### Observability

 - The span name declaration has been simplified: no need for `amaru_observability::amaru`.
 - span names follow a `xxx.yyy.zzz` format where `xxx` is generally one of `state`, `setup`, `perf`; `yyy` is an entity name like `header` or `block`, and `zzz` is an action (I changed a few names in the original EDR for consistency).
 - `chain_sync_client` was an outlier in the schema definition. That struct has been moved to the `amaru` crate, since it is only used for bootstrapping and was using the `amaru::network` target name.

### Effects

 - `Store` and `Ledger` can be initialized with a `TraceContext` passed to effects built with those instances.
    - Another solution would be to make `Effects` `TraceContext`-aware, but that would add an OpenTelemetry dependency to the `amaru-pure-stage` crate.
 - Only a subset of all effects have a function `with_trace_context` that initialize them with a trace context.
 - When an effect has a trace context, that one is used to attach the effect execution to the trace context so that all spans created during the execution have the right parent

### Stages

 - A `TraceContext` has been introduced with a dual purpose:
   - Pass just the context of a trace from stage to stage in order to relate spans created in those stages to the same trace id
   - Pass a specific span, opened in one stage, to another stage (see the `perf` spans below).
   - It can be either instantiated from a `&Span` and then just contains a span context. Or it can be instantiated from a `Span`, and then contains the full span.
   - That does the job but also does not look totally right to me.
   
 - The main consensus stages have been instrumented so that a nominal trace is available end to end.
 - `header_hash` has been added as a span attribute everywhere possible, even if it was a bit redundant with `tip` or `point` sometimes. The goal is to have a unique attribute to correlate spans for the same header.
 - The messages sent to the `peer_selection` stage contain a `TraceContext` tagged with the `header_hash` but that context is not used yet.
 
### Ledger / Stores / Protocols

 - span name changes to follow the convention.
 - Some uses of `instrument` for async blocks.

### Pure-stage

 - The `tokio` poll span is now using a regular `tracing::trace_span` to avoid the dependency with the observability crate for just one span (@rkuhn do you want it back?)

### Tests

I tested manually against the `relay-1` demo that I will submit separately:

image

## Summary by CodeRabbit

* **New Features**
  * Added end-to-end tracing context propagation across consensus, networking, protocols, and storage.
  * Introduced new tracing record helpers for consistent INFO/DEBUG event emission.
* **Bug Fixes**
  * Improved consistency of adversarial peer notifications and updated related expectations.
  * Enhanced mempool INFO logging for transaction acceptance and batch insertion.
* **Documentation**
  * Refreshed trace schema documentation and fixtures to match the updated tracing naming/structure.
  * Updated changelog entries and trace-schema references accordingly.


etorreborre:
@coderabbit review

etorreborre:
Back to draft to take the time to address Coderabbit's comments.

crates/amaru-consensus/src/effects/ledger_effects.rs
```diff
@@ -80,30 +77,20 @@ impl LedgerOps for Ledger {
...
 
-    fn rollback(
-        &self,
-        peer: &Peer,
-        point: &Point,
-        ctx: opentelemetry::Context,
-    ) -> BoxFuture<'static, anyhow::Result<(), BlockValidationError>> {
-        self.effects.external(RollbackBlockEffect::new(peer, point, ctx))
+    fn rollback(&self, peer: &Peer, point: &Point) -> BoxFuture<'static, anyhow::Result<(), BlockValidationError>> {
+        self.effects.external(RollbackBlockEffect::new(peer, point))
```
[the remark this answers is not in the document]
etorreborre: Done.

crates/amaru-consensus/src/stages/fetch_blocks/mod.rs
```diff
@@ -227,11 +257,17 @@ impl FetchBlocks {
         }
 
         tracing::info!(tip = %tip.point(), "no blocks to fetch");
-        eff.send(&self.upstream, SelectChainMsg::FetchNextFrom(tip.point())).await;
+        self.fetch_next_from(eff, tip.point()).await;
```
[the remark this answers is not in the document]
etorreborre: Done.

crates/amaru-consensus/src/stages/select_chain/mod.rs
```diff
@@ -177,24 +244,51 @@ impl SelectChain {
...
             } else {
                 tracing::info!(%parent, %tip, "upstream tip depends on invalid block");
+                self.end_perf_spans(&tip.hash(), BlockOutcome::Abandoned);
             }
         }
 
+        self.perf_header_block_fetch_wait_trace_contexts.insert(
+            tip.hash(),
+            debug_span!(root, consensus::perf::header::BLOCK_FETCH_WAIT, tip = tip, header_hash = tip.hash()).into(),
+        );
```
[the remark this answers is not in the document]
etorreborre: Fixed

crates/amaru-consensus/src/stages/select_chain/mod.rs
```diff
@@ -242,7 +337,16 @@ impl SelectChain {
...
                         self.may_fetch_blocks = false;
-                        eff.send(&self.downstream, (new_best_tip.tip(), parent)).await;
+                        eff.send(
+                            &self.downstream,
+                            NewBestTip {
+                                tip: new_best_tip.tip(),
+                                parent,
+                                trace_context,
+                                perf_header_block_fetch_wait_trace_contexts: Default::default(),
+                            },
```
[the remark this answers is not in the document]
etorreborre: Fixed

crates/amaru-consensus/src/stages/track_peers/mod.rs
```diff
@@ -325,49 +337,52 @@ impl TrackPeers {
...
+                eff.send(&self.downstream, NewTip { tip, parent, trace_context, perf_context }).await;
             }
             Ok(None) => {
                 tracing::debug!(%peer, tip = %current_point, "roll forward, header already stored");
             }
             Err(error) => {
                 tracing::error!(%error, %peer, "chain_sync.store_header.failed");
                 self.upstream.remove(&peer);
-                eff.send(&self.peer_selection, PeerSelectionMsg::Adversarial(peer)).await;
+                eff.send(&self.peer_selection, PeerSelectionMsg::adversarial(peer)).await;
```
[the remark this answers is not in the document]
etorreborre: Fixed

crates/amaru-consensus/src/stages/peer_selection/mod.rs
```diff
@@ -359,7 +366,7 @@ pub async fn stage(mut state: PeerSelection, msg: PeerSelectionMsg, eff: Effects
                 .await;
             eff.send(&ledger_check, ()).await;
         }
-        PeerSelectionMsg::Adversarial(peer) => {
+        PeerSelectionMsg::Adversarial(peer, _trace_context) => {
             tracing::debug!(%peer, "peer_selection.adversarial");
             state.ban_peer(peer, &eff).await;
```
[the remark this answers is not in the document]
etorreborre: Fixed

crates/amaru-consensus/src/stages/select_chain/mod.rs
```diff
@@ -290,11 +406,97 @@ impl SelectChain {
...
+    }
+
+    /// If a trace context representing the performance for getting a block from the reception
+    /// of header exists, close its span.
+    fn end_perf_e2e_forward_span(&mut self, hash: &HeaderHash, outcome: &BlockOutcome) {
+        if let Some(mut perf_context) = self.perf_e2e_forward_trace_contexts.remove(hash) {
+            let span: tracing::Span = debug_span!(root, consensus::perf::header::FORWARD, header_hash = hash);
+            span.record("blah", "blah");
+            perf_context.record("outcome", outcome.as_str());
+            perf_context.close();
```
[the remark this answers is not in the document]
etorreborre: Ouch. Fixed

crates/amaru-consensus/src/stages/track_peers/mod.rs
```diff
@@ -389,74 +404,110 @@ impl TrackPeers {
...
-                self.execute_roll_forward(peer, handler, variant, header, tip, mode, eff).await;
+                tracing::debug!(%peer, highest = %tip.point(), "roll forward");
+                let peer_clone = peer.clone();
+                let span = debug_span!(
+                    root,
+                    consensus::state::roll_forward::PROCESS,
+                    tip = tip,
+                    peer = peer_clone,
+                    header_hash = tip.hash(),
+                );
```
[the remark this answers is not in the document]
etorreborre: Fixed

crates/amaru-consensus/src/stages/track_peers/mod.rs
```diff
@@ -389,74 +404,110 @@ impl TrackPeers {
...
+
+                eff.send(&handler, InitiatorMessage::RequestNext).await;
 
-                let store = Store::new(eff.clone());
                 if let Err(error) = self.roll_backward(&peer, current, tip, &store).await {
                     tracing::error!(%error, %peer, "chain_sync.roll_backward.failed");
                     self.upstream.remove(&peer);
-                    eff.send(&self.peer_selection, PeerSelectionMsg::Adversarial(peer)).await;
+                    eff.send(&self.peer_selection, PeerSelectionMsg::Adversarial(peer, trace_context)).await;
                 }
```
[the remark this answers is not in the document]
etorreborre: Done.

crates/amaru-observability/macros/src/traces.rs
```diff
@@ -52,10 +52,11 @@ impl SchemaMeta {
...
     ///
     /// Local schemas are defined with `define_local_schemas!` and their macros
-    /// are NOT exported with `#[macro_export]`. They must be called without a path
-    /// prefix since they're local to the module where they're defined.
+    /// are NOT exported with `#[macro_export]`. They are identified by a
+    /// leading `self::` or `crate::` segment in the user-supplied schema path,
+    /// which `parse_macro_module` collapses to `"self"` or `"crate"`.
     fn is_local_schema(&self) -> bool {
-        self.macro_module != "amaru" && !self.macro_module.ends_with("::amaru")
+        matches!(self.macro_module.as_str(), "self" | "crate")
```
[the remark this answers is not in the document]
etorreborre: Done.

docs/traces-schema.json
```diff
@@ -4,7 +4,110 @@
...
+      "required": [],
+      "optional": [],
+      "additionalProperties": false,
+      "name": "validate",
+      "level": "TRACE",
+      "target": "amaru::ledger::state::block",
+      "description": "Validate block against rules",
+      "public": true
+    },
+    "amaru::ledger::state::context::ADD_FEES": {
```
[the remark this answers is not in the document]
etorreborre: Done.

crates/amaru-observability/macros/src/traces.rs
```diff
@@ -496,6 +531,7 @@ pub fn expand_trace_span(input: TokenStream) -> TokenStream {
         Ok(args) => args,
         Err(err) => return err.to_compile_error().into(),
     };
+    let fields: Vec<_> = args.fields.iter().filter(|field| field.name != "category").collect();
```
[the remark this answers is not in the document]
etorreborre: Fixed, that was a left-over.

crates/amaru-ledger/src/context/default/preparation.rs
```diff
@@ -174,7 +174,7 @@ fn resolve_inputs<'a>(
     policy: UnresolvedInputPolicy,
     mut keys: impl Iterator<Item = &'a TransactionInput>,
 ) -> Result<BTreeMap<TransactionInput, MemoizedTransactionOutput>, ContextHydratationError> {
-    trace_span!(amaru_observability::amaru::ledger::state::HYDRATE_INPUTS).in_scope(|| {
+    trace_span!(ledger::state::inputs::HYDRATE).in_scope(|| {
```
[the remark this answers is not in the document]
etorreborre: Done.

crates/amaru-observability/src/trace_context.rs
```diff
@@ -0,0 +1,90 @@
...
+use opentelemetry::{
+    Context, ContextGuard,
+    trace::{SpanContext, TraceContextExt},
+};
+use tracing_opentelemetry::OpenTelemetrySpanExt;
+
+/// Parent context for spans that must cross stage boundaries.
+/// It encapsulates an OpenTelemetry `SpanContext` to provide a few helper functions.
+#[derive(Clone, Debug)]
+pub struct TraceContext {
```
rkuhn, line 33: It is unclear to me how and when this is supposed to be used: it does not implement `Serialize` and `Deserialize`, so it cannot be sent between stages. The `span` can be closed but no new span can be opened. `.attach()` always creates a new context, so it seems that this `TraceContext` is somewhat hollow, but it isn’t obvious how to create a child span. Perhaps all this is caused by opentelemetry just being weird with all its thread-locals; in that case we either need to use the opentelemetry API or create a meaningful new API that just uses opentelemetry for its implementation.

crates/amaru-observability/src/trace_context.rs
```diff
@@ -0,0 +1,90 @@
...
+
+    /// Close the corresponding span
+    pub fn close(&mut self) {
+        if let Some(span) = self.span.take() {
+            drop(span)
+        }
+    }
+
+    /// Append a field to the corresponding span
+    pub fn record(&self, name: &str, value: impl Display) {
```
rkuhn: This should probably take a `&'static str` and a `&dyn Display` instead.

scripts/run-until
```diff
@@ -44,7 +44,7 @@ function watchTargetEpoch() {
       (try fromjson catch empty)
       | select(.level == "INFO")
       | select(.fields.message == "exit")
-      | select(.span.name == "epoch_transition")
+      | select(.span.name == "state.epoch_transition.epoch_transition")
```
rkuhn: I know this systematic came from @KtorZ but I don’t see any value in this addition. I was also under the impression that somehow the NAME isn’t that useful, hence we fold it into the TARGET. I’m probably behind on some of the discussions.

CHANGELOG.md
```diff
@@ -37,29 +37,45 @@ Other guiding principles:
...
+### Added
+
+- **amaru**: add / modify tracing spans to conform
+  to [EDR-26](https://github.com/pragma-org/amaru/blob/main/engineering-decision-records/026-tracing-span-design.md). ([#996][])
+
 ### Changed
 
-- **amaru**: make sure bootstrap can't be done with unsupported snapshots; similarly prevent startup if state becomes unsupported ([#1000][])
+- **amaru**: make sure bootstrap can't be done with unsupported snapshots; similarly prevent startup if state becomes
+  unsupported ([#1000][])
```
rkuhn: Please don’t reformat existing logs. I seem to remember me doing that on the logbook about a year ago, and I apologise for that.

crates/amaru-consensus/src/stages/select_chain/mod.rs
```diff
@@ -125,14 +142,30 @@ pub async fn stage(mut state: SelectChain, msg: SelectChainMsg, eff: Effects<Sel
...
-        SelectChainMsg::TipFromUpstream(tip, parent) => {
-            state.handle_tip_from_upstream(tip, parent, eff).await;
+        SelectChainMsg::TipFromUpstream { tip, parent, trace_context } => {
+            let span = debug_span!(
+                parent_context: &trace_context,
+                consensus::state::chain::SELECT_FROM_TIP,
+                tip = tip,
+                header_hash = tip.hash(),
+            );
+            let child_trace_context = (&span).into();
```
rkuhn, line 152: The mechanics of this aren’t clear to me: shouldn’t a TraceContext just hold the trace ID, which in this case is tied to the header hash? Span nesting seems like it should be done using just the mechanisms of `tracing`, since that seems to have worked before, but I’m very likely missing something.

rkuhn (commented):
Sorry for the confused comments, @etorreborre, I’m probably lacking background knowledge. This may well be indicative of missing code docs in this PR as well because other people may well have the same deficit as me.

crates/amaru-observability/src/trace_context.rs
etorreborre: What's wrong with `Display`? I'm passing some enum values, for example `PerfHeaderForwardOutcome::DuplicateHeader` (in the [perf span pr](#1005))

CHANGELOG.md
etorreborre: Sorry, that's my editor. I changed the settings to prevent that from happening again.

crates/amaru-consensus/src/stages/select_chain/mod.rs
etorreborre: The trace id is not semantically significant, I've added a `header_hash` attribute everywhere we need to correlate spans by header hash (or just want to get the information for diagnosis).
For the span parenting, we need to provide the `child_trace_context` its proper parent, that comes from a message. There might be other ways to do that parenting by attaching first the incoming parent context, then creating a span locally, but I'm not sure that's simpler, more obvious.

scripts/run-until
etorreborre: Let's discuss that during our next dev. meeting.

crates/amaru-observability/src/trace_context.rs
etorreborre: > It is unclear to me how and when this is supposed to be used: it does not implement Serialize and Deserialize, so it cannot be sent between stages.

The trace context is marked as `#[serde(skip)]` to avoid having to serialize it. As far as I understand (and can see) that only prevents it from being persisted in the trace buffer, but not to be propagated from stage to stage.

> The span can be closed but no new span can be opened. .attach() always creates a new context, so it seems that this TraceContext is somewhat hollow, but it isn’t obvious how to create a child span.

As I wrote in the PR description I'm not a big fan of the `TraceContext` either. But it solves 2 use cases:

1. Pass a trace id + parent span so that a span opened in another stage with the `debug_span!` macro can be properly parented. This is done with either `.attach()` so that the current thread gets the right context implicitly (in effects implementations) or with `parent_context` when creating a span with `debug_span!`.
Internally that's done by keeping an `OpenTelemetry` `SpanContext`.

2. Pass a full span (and implicitly its context) if we need to keep around that span then close it in another stage (typically for a performance span)

rkuhn: Thanks, that explains some things. Given that we use a common trace context, why do we need to transport the parent span across the message flow? It was my understanding that the parent–child relationship is intended to be used for nested spans on the same call stack.

One thing I’m wondering about: I seem to recall getting into trouble when keeping a span active while switching threads, the span cannot be entered while doing that, it must be idle. Is my understanding correct? Does that break the perf spans defined in the EDR?

crates/amaru-observability/src/trace_context.rs
rkuhn: The difference is that my proposal avoids creating lots of instances of this method, or copying a lot of code in case this gets inlined. Dynamic dispatch is not our enemy, we should use it in such places.

(Taking `impl AsRef<str>` can be an exception because there will likely be only two or three types with which the template is instantiated.)

Another reason for taking a `&dyn` is that that makes it impossible to pass ownership of a possibly expensive data type — which the artificial idiot may easily do by “using `.clone()` to fix the compile error.” This is the same pattern used by string formatting in the std lib, or capturing values in `tracing`.

crates/amaru-consensus/src/stages/select_chain/mod.rs
rkuhn: The [`TraceId`](https://docs.rs/opentelemetry/latest/opentelemetry/struct.TraceId.html) is a 16 byte value that seems to be one half of the `SpanContext` (the other half basically being the `SpanId`), which looks like it is the serializable part of an OTel span (of which I don’t yet understand the relationship to a `tracing::Span`). It does seem like we can just make up those 16 bytes however we want — wouldn’t it be possible to use half of the `header_hash` for that purpose? This way, we don’t need to keep that `SpanContext` around in situations where we get the info otherwise. I’m aware that this is not a complete proposal as it lacks any understanding of what an OTel `SpanId` shall be, just thinking out loud.

crates/amaru-observability/src/trace_context.rs
etorreborre: I see. Good idea. I also changed the implementation to be:
```rustlang
span.record(name, tracing::field::display(value));
```
This way we reuse the `Display` instance directly without going via a `String`.

monitoring/README.md
```diff
@@ -26,30 +26,49 @@ Any event (trace, span or metric) can be filtered by target and severity using t
...
+```
+
+Directives can be combined to match several tags:
+
+- `[{amaru.tag.cpu=true}]=trace,[{amaru.tag.io=true}]=trace` selects spans with the `cpu` **or** the `io` tag;
+- `[{amaru.tag.cpu=true,amaru.tag.io=true}]=trace` selects spans with **both** tags.
+
+Like span filters, tag filters are scoped: any `span` or `event` created inside a matching span is also considered.
+
+Note that the value match (`=true`) is required: a field-presence directive such as `[{amaru.tag.cpu}]` only restricts *events*, and matches all spans regardless of their tags.
```
KtorZ, line 70: 🫠

crates/amaru-observability/macros/src/define_schemas.rs
```diff
@@ -1603,6 +1732,80 @@ mod tests {
...
+                }
+            }
+        "#;
+        let (schemas, errors) = extract_schemas(input);
+        assert!(errors.is_empty(), "Unexpected errors: {:?}", errors);
+        assert_eq!(schemas.len(), 3);
+        assert_eq!(schemas[0].tags, vec!["cpu".to_string(), "io".to_string()]);
+        assert_eq!(schemas[1].tags, vec!["setup".to_string()]);
+        assert!(schemas[2].tags.is_empty());
+    }
```
KtorZ, line 1769: 👍

crates/amaru-stores/src/rocksdb/ledger/columns/cc_members.rs
```diff
@@ -32,8 +32,8 @@ pub fn get<'a>(
     db_get: impl Fn(&[u8]) -> Result<Option<DBPinnableSlice<'a>>, rocksdb::Error>,
     credential: &Key,
 ) -> Result<Option<Row>, StoreError> {
-    let _span = trace_span!(
-        amaru_observability::amaru::stores::ledger::columns::CC_MEMBERS_GET,
+    let _span = debug_span!(
```
KtorZ: Note: these need to stay as `trace`; it'll get very very noisy otherwise. I'll push a commit.

PR #997, opened 2026-06-30
901 predicate delegatee stake pool not registered deleg

### LINKED ISSUE #901: Ledger predicate: DelegateeStakePoolNotRegisteredDELEG
### Abstract

Implement predicate: `DelegateeStakePoolNotRegisteredDELEG`

### Why?

Conformance

### How?

[Reference Implementation](https://github.com/IntersectMBO/cardano-ledger/blob/0cfbf861cfb456660a7b73281c6fb714a53d40f9/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Deleg.hs#L215-L219)

### Discussion points

_No response_

_No response_

### Checklist

jonathanlim222:
## Summary by CodeRabbit

* **Bug Fixes**
  * Delegation now fails when the target pool does not exist, instead of being accepted.
  * Improved validation for stake credential delegation and vote delegation edge cases, including deregistration and re-registration flows.

* **Tests**
  * Added broader transaction conformance coverage for delegation scenarios.
  * Added new pass/fail fixtures for registered, unregistered, and mixed-order delegation cases.


crates/amaru-ledger/src/rules/transaction/phase_one/certificates.rs
```diff
@@ -512,3 +512,95 @@ where
...
+
+    fn pointer() -> TransactionPointer {
+        TransactionPointer { slot: Slot::from(0), transaction_index: 0 }
+    }
+
+    fn account() -> AccountState {
+        AccountState { deposit: 2_000_000, pool: None, drep: None, rewards: 0 }
+    }
+
+    fn ctx_with(
```
KtorZ: What about using a `Default` trait instance for `DefaultValidationContext`, and overriding in place the fields? I don't think the helper buys us much here.

crates/amaru-ledger/src/rules/transaction/phase_one/certificates.rs
```diff
@@ -512,3 +512,95 @@ where
...
+            &mut ctx,
+            vec![
+                Certificate::StakeDeregistration(cred(1)),
+                Certificate::Reg(cred(1), 2_000_000),
+                Certificate::StakeDelegation(cred(1), pool(2)),
+            ],
+        );
+        assert!(result.is_ok());
+    }
+}
```
KtorZ: Use `test_case` here; this is fundamentally three times the same test with slightly different inputs. That also removes the need for the `run_certs` function entirely; and we can have one easy to read test matrix.

yHSJ (approved):

PR #998, opened 2026-06-30
Test MaxTxSizeUTxO

### LINKED ISSUE #893: Ledger Predicate: MaxTxSizeUTxO
### Abstract

Implement predicate `MaxTxSizeUTxO`

### Why?

Conformance

### How?

[Reference Implementation](https://github.com/IntersectMBO/cardano-ledger/blob/cd8b7fab8365c5376c4ca2929cfb455d31f4fde6/eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Utxo.hs#L563)

### Discussion points

_No response_

_No response_

### Checklist

yHSJ:
While `MaxTxSizeUTxO` has already been implemented, this PR provides an extra fixture to enforce the boundary behavior (tx size can be equal to the max size). 

Closes: #893
Skip-changelog

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed transaction size handling so conformance checks now match ledger fee calculation when testing standalone transactions.
  * Added support for transactions that are exactly at the configured maximum size.

* **Tests**
  * Added a new pass fixture covering a transaction at the size limit.
  * Improved test size calculations for transaction fixtures to better reflect the encoded format.

### DIFF
--- a/crates/amaru-ledger/src/rules/transaction/phase_one/mod.rs
+++ b/crates/amaru-ledger/src/rules/transaction/phase_one/mod.rs
@@ -352,6 +352,7 @@ mod tests {
     #[test_case(fixture!("fail/BabbageNonDisjointRefInputs/0"); "input appears as both spent and reference")]
     #[test_case(fixture!("fail/WrongNetworkInTxBody/0"); "body network_id disagrees with fixture network")]
     #[test_case(fixture!("pass/tx-size-at-limit"); "transaction size exactly at maxTransactionSize")]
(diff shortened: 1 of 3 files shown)


KtorZ (approved):
Would be nice to see the almost-same-test with `"maxTransactionSize": 199` to prove that it's really at the minimum. Especially when the raw bytes are `201` bytes long (due to the `isValid` being ignored for sizing). It's very easy to be off-by-one here.

PR #1000, opened 2026-07-01
feat: properly handle older non-supported states

jeluard:
Amaru only supports starting some well know protocol version. Make sure it only ever start when this is enforced. Specifically:
* do not bootstrap too old snapshots
* prevent startup if state is not supported anymore (due to upgrade)

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented startup and initial snapshot import from continuing when the provided protocol version is below the supported minimum.
  * Added clearer error reporting when snapshots/state are incompatible due to protocol version.
* **Chores**
  * Updated the changelog to document the new compatibility behavior.
  * Improved changelog reference validation to enforce numerically sorted reference links.


crates/amaru-ledger/src/bootstrap.rs
```diff
@@ -54,6 +55,9 @@ enum InitialSnapshotFormatError {
 
     #[error("invalid initial snapshot payload: expected a previous-blocks map immediately after the epoch")]
     MissingPreviousBlocksMap,
+
+    #[error("snapshot protocol version {snapshot_version} is too old; minimum supported version is {minimum_version}")]
+    ProtocolVersionTooOld { snapshot_version: String, minimum_version: String },
```
KtorZ: Please let's use real types here 🙏, we have a `ProtocolVersion`

crates/amaru-ledger/src/protocol_version_validation.rs
```diff
@@ -0,0 +1,38 @@
...
+) -> Result<(), ProtocolVersionTooOld> {
+    if version.0 < minimum_version.0 || (version.0 == minimum_version.0 && version.1 < minimum_version.1) {
+        return Err(ProtocolVersionTooOld {
+            snapshot_version: protocol_version::fmt(&version),
+            minimum_version: protocol_version::fmt(&minimum_version),
+        });
+    }
+
+    Ok(())
+}
```
KtorZ: To be moved under `kernel/cardano/protocol_version`

crates/amaru-ledger/src/protocol_version_validation.rs
```diff
@@ -0,0 +1,38 @@
...
+pub struct ProtocolVersionTooOld {
+    pub snapshot_version: String,
+    pub minimum_version: String,
+}
+
+pub fn validate_protocol_version(
+    version: ProtocolVersion,
+    minimum_version: ProtocolVersion,
+) -> Result<(), ProtocolVersionTooOld> {
+    if version.0 < minimum_version.0 || (version.0 == minimum_version.0 && version.1 < minimum_version.1) {
```
KtorZ: I believe we have an `Ord` instance on the `ProtocolVersion`, it's just a tuple of uint

KtorZ (approved):

PR #1001, opened 2026-07-01
fix: make sure that tips are removed from the select_chain map in case of a fork

etorreborre:
We don't completely clean-up the `select_chain` stage in case of a fork.

Skip-Changelog


etorreborre:
@coderabbit review

etorreborre:
skip-changelog

rkuhn (commented):
The `tips` map was supposed to hold a view of all chain tips seen on the network, to remove DB access load. Why is this change beneficial?

etorreborre:
@rkuhn aren't those tips going to accumulate forever if they correspond to forks that never win?

rkuhn:
The tips should eventually be cleaned up, which can easily be done by a check that runs once per day (evicting everything older than the anchor) — the number of such cases shouldn’t be more than a handful per day and each occurrence just costs less than 1kB. This is also why I didn’t add that logic when drafting that stage, I should have left a TODO instead.

etorreborre:
Ok, closing then

PR #1005, opened 2026-07-02
feat: add performance tracing spans to the consensus stages

etorreborre:
This PR adds a few spans dedicated to the measure of the node performance, w.r.t the network (as specified in EDR-26):
    - `perf.header.forward` measures the time between the reception of a header and its adoption as the best tip.
    - `perf.block.fetch` measures the time it takes to fetch a block from peers.
    - `perf.fork.switch` measures the time it takes to switch to a new fork from the moment it is detected to the moment its new tip is adopted.
    - `perf.header.block_fetch_wait` measures the time between the moment we receive a header and the moment we start fetching its block.
    
Note: the creation and passing of these spans is a bit tricky. If the logic is wrong:
   - They might not be closed at the right time.
   - They could accumulate and create memory leaks.

## Summary by CodeRabbit

* **New Features**
  * Added performance tracing for header forwarding, block-fetch waits, block fetching, and fork switching.
  * Added outcomes and timing details to help diagnose consensus and node performance.
* **Documentation**
  * Documented the new performance trace targets, spans, fields, and schemas.
  * Added an unreleased changelog entry for the performance tracing improvements.


etorreborre:
@coderabbit review

crates/amaru-consensus/src/stages/fetch_blocks/mod.rs
```diff
@@ -116,6 +116,13 @@ pub struct FetchBlocks {
     /// this stage are children of that context
     #[serde(skip, default)]
     trace_context: Option<TraceContext>,
+    /// Trace contexts sent by the select_chain stage tracking the time between the reception of a header
+    /// and the emission of a request for its block.
+    #[serde(skip, default)]
```
rkuhn: This annotation is always a smell, and there is a risk of adding code that depends on the contents of these maps for details of the behaviour, which will then break determinacy of the stage. Can this not be serialized?

crates/amaru-consensus/src/stages/fetch_blocks/mod.rs
```diff
@@ -342,6 +376,13 @@ impl FetchBlocks {
             return;
         }
 
+        // close the performance fetch span
+        if let Some(span) = self.perf_block_fetch_spans.remove(&point.hash()) {
+            drop(span);
+        } else {
+            tracing::warn!("received a block that was not requested: {}", point.hash());
```
rkuhn: This is something that will happen incorrectly during replay.

crates/amaru-consensus/src/stages/fetch_blocks/mod.rs
```diff
@@ -387,6 +431,19 @@ impl FetchBlocks {
         let trace_context = self.trace_context.take().unwrap_or_default();
         eff.send(&self.upstream, SelectChainMsg::FetchNextFrom(from, trace_context)).await;
     }
+
+    fn close_perf_spans(&mut self) {
```
rkuhn: Not sure I understand where this is useful, but the implementation is needlessly complicated: why not just clear both maps? The result should be the same.

crates/amaru-consensus/src/stages/fetch_blocks/mod.rs
```diff
@@ -387,6 +431,19 @@ impl FetchBlocks {
...
+        // Close the perf spans for fetched blocks
+        for span in std::mem::take(&mut self.perf_block_fetch_spans).into_values() {
+            drop(span)
+        }
+        self.perf_block_fetch_spans.clear();
```
rkuhn, line 478: this is a noop

rkuhn (commented):
not a full review yet, will need to catch up with @etorreborre to get background

crates/amaru-consensus/src/stages/fetch_blocks/mod.rs
etorreborre: The idea was to avoid serializing data that was not useful for replay because tracing should not interfere (to spare runtime cycles + storage). But I see your point that we could get a different replay outcome in case we do, by accident, change the stage behavior when processing traces.

crates/amaru-consensus/src/stages/fetch_blocks/mod.rs
etorreborre: Absolutely, fixed.

crates/amaru-consensus/src/stages/fetch_blocks/mod.rs
etorreborre: Right, I'm going to investigate how make it serializable (cf. the comment above).

crates/amaru-consensus/src/stages/fetch_blocks/mod.rs
rkuhn: I was also under the impression that the important part (the trace ID and header_hash) should be readily serializable. I haven’t yet pulled together the full story around TraceContext, though.

crates/amaru-consensus/src/stages/fetch_blocks/mod.rs
etorreborre: Done. One caveat though: spans are not serializable. So if a span was included in the trace context, it is deserialized as an empty span when the trace context is deserialized.

rkuhn: This code is still depending on the `self.perf_block_fetch_spans` map, which will be empty after deserialization, thus not deterministically replayable.

crates/amaru-consensus/src/stages/select_chain/mod.rs
```diff
@@ -219,17 +255,45 @@ impl SelectChain {
                 self.tips.insert(tip.hash(), ancestors);
             } else {
                 tracing::info!(%parent, %tip, "upstream tip depends on invalid block");
+                self.end_perf_header_trace_context(&tip.hash(), PerfHeaderForwardOutcome::AbandonedBlock);
+                self.end_perf_fork_switch_span(&tip.hash(), PerfForkSwitchOutcome::AbandonedBlock);
```
rkuhn: Shouldn’t this instead just be `self.end_perf_header_trace_context(...)`?

And shouldn’t the reason be InvalidBlock? (I haven’t yet read everything below)

crates/amaru-consensus/src/stages/select_chain/mod.rs
```diff
@@ -219,17 +255,45 @@ impl SelectChain {
...
+                // Close the perf spans for unvalidated blocks that are not part of the new best branch
+                for hash in best_tip_unvalidated_blocks.iter().filter(|h| !new_best_unvalidated_blocks.contains(h)) {
+                    self.end_perf_header_trace_context(hash, PerfHeaderForwardOutcome::AbandonedBlock);
+                }
+            }
             if self.may_fetch_blocks {
                 self.may_fetch_blocks = false;
-                eff.send(&self.downstream, NewBestTip { tip, parent, trace_context: parent_context }).await;
+                let perf_header_block_fetch_wait_trace_contexts =
+                    std::mem::take(&mut self.perf_header_block_fetch_wait_trace_contexts);
```
rkuhn: This looks brittle: how do we ensure that this map contains exactly and only the correct entries?

crates/amaru-consensus/src/stages/select_chain/mod.rs
```diff
@@ -219,17 +255,45 @@ impl SelectChain {
...
+                        parent,
+                        trace_context: parent_context,
+                        perf_header_block_fetch_wait_trace_contexts,
+                    },
+                )
+                .await;
             }
             self.best_tip = Some(header);
+        } else {
+            self.end_perf_header_trace_context(&tip.hash(), PerfHeaderForwardOutcome::LosingHeader);
```
rkuhn: How does the tracing deal with actual forks? During such a situation, the branch that will eventually end up losing may be ahead for some time, during which the other branch’s headers will get to this point. How are their spans re(opened|created) once we switch to the now winning branch?

crates/amaru-consensus/src/stages/select_chain/mod.rs
```diff
@@ -260,16 +324,20 @@ impl SelectChain {
                     v.drain(0..=idx);
                 }
             });
+            self.end_perf_header_forward_trace_context(&h, PerfHeaderForwardOutcome::ValidBlock);
+            self.end_perf_fork_switch_span(&h, PerfForkSwitchOutcome::ValidBlock);
```
rkuhn, line 330: At this point we’re getting the first block on the chain that we may be switching to, meaning that the switch is not necessarily complete. Why close the span already?

rkuhn (commented):
There are some pieces that may be wrong (or I misunderstood), and in general I get the feeling that especially in select_chain the tracking of span&context state changes is more complex than the weaving into the rest of the code (which is already non-trivial) permits. Perhaps the whole perf tracking should be factored out into a separate data structure that just gets informed of relevant events? This could then also have a specialised serde impl that only writes which hashes are “known” and supplies fresh spans upon deserialization. I’m not certain about any of this, just voicing my gut feeling — let me know if you want to join a quick brainstorming.

crates/amaru-consensus/src/stages/select_chain/mod.rs
etorreborre: Yes, this one is useless, I removed it.
It's `AbandonedBlock` because this block didn't directly failed validation.

crates/amaru-consensus/src/stages/select_chain/mod.rs
etorreborre: This only closes the span if the header hash `h` corresponds to the fork tip, which we store alongside the open span.

crates/amaru/build.rs
```diff
@@ -256,6 +262,52 @@ fn collect_workspace_type_aliases(
...
+    }
+
+    for relative_path in split_nul_terminated_paths(&output.stdout) {
+        let path = workspace_dir.join(relative_path);
+        let crate_name = crate_name_from_workspace_relative_path(Path::new(relative_path));
+
+        println!("cargo:rerun-if-changed={}", path.display());
+        let source = fs::read_to_string(&path)?;
+        collect_type_aliases_from_source(&source, aliases, crate_name.as_deref());
+    }
```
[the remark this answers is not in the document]
etorreborre: Fixed.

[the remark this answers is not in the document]
etorreborre: Removed from this PR

etorreborre:
@rkuhn I changed the processing of perf spans:

 - The spans associated to potential forks are preserved instead of being dropped (I did this as a first approximation to collect data in the steady state of growing one main branch)
 - The spans that could not be closed by the `fetch_block` stage and passed back to the `select_chain` stage.
 
 There might be a better solution, for example by sending events to a separate stage that could emit synthetic spans. I think that this is possible with OpenTelemetry `SpanBuilder` API. We can discuss this.

rkuhn (commented):
I still think that the logic inlined and mixed with the already non-trivial chain selection is too complex to be reasonable. Could you move all TraceContext and Span handling into a separate struct with optimised serde representation and method names matching the messaging semantics of select_chain? That way the impact on select_chain code is only one line per relevant activity and the trace handling can be reviewed without getting caught up in chain selection logic.

etorreborre:
Closing this one for now. I will propose something simpler.

PR #1007, opened 2026-07-03
fix(ledger): recompute missing rewards summary at epoch boundary

paolino:
### Problem

Running amaru as n2n relays under Antithesis fault injection (kill/restart faults), consensus terminates while validating the first block after an epoch boundary:

```
ERROR amaru_consensus::stages::validate_block: failed to validate block
      error=BlockValidationError: rewards summary not ready
ERROR amaru::cmd::node::run: Consensus died, this should not happen!
      Please report this incl. preceding logs to the Amaru team.
```

`restart: always` brings the node back, it re-validates the same block and dies again — a crash loop. Reproduced on `5923f085` and on `main` (`eb21a990`). Full evidence (runs, logs, timeline): https://gist.github.com/paolino/ff8d2c19dfc986c42aaf1e022e6ec18f

### Cause

The rewards summary is computed when a block at relative slot >= the stability window is applied (`try_compute_rewards`) and held only in the in-memory volatile overlay. A node restarting with its stable tip already past the stability window replays no block within the window, so the summary is missing when the epoch boundary is reached, and `epoch_transition` fails with `RewardsSummaryNotReady` — fatal for consensus.

### Fix

In `epoch_transition`: when no transition progress has been persisted and the summary is missing, recompute it. All inputs (the `e-1` snapshot and the `e-3` stake distribution) are persisted and frozen for the epoch (`MIN_LEDGER_SNAPSHOTS = 3`), so the recomputation is deterministic — same result as computing at the stability window. Every previously-working path is unchanged; a `warn!` traces the recovery.

Also adds a regression unit test reconstructing the post-restart state (empty volatile, stable tip past the stability window); it fails with `RewardsSummaryNotReady` without the fix. The test uses self-contained store mocks because `amaru-stores` cannot be used from `amaru-ledger` unit tests (cyclic dev-dependency links a second instance of the crate, breaking trait identity).

### Testing

- Regression test: fails on `main`, passes with the fix; full `amaru-ledger` suite green; clippy/fmt clean.
- 1h Antithesis fault-injection run on the patched build (same harness that reproduced the crash twice): `Consensus died` 0 occurrences (previously 50+); the recovery `warn` fired 50+ times, i.e. the previously-fatal state was entered repeatedly and survived.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved epoch transition handling during restarts when prior progress or cached rewards data is unavailable.
  * The ledger now recovers by recomputing the rewards summary instead of failing the transition.

* **Tests**
  * Added a regression test covering successful epoch transition after a restart with missing in-memory state.

### DIFF
--- a/crates/amaru-ledger/src/state.rs
+++ b/crates/amaru-ledger/src/state.rs
@@ -371,12 +371,34 @@ impl<S: Store, HS: HistoricalStores> State<S, HS> {
             // FIXME: This should eventually be a '.await', as we always expect to *eventually*
             // have some rewards summary being available. There's no way to continue progressing
             // the ledger if we don't.
-            let computed_rewards = self.volatile.take_computed_rewards();
+            let mut computed_rewards = self.volatile.take_computed_rewards();
+
+            let progress = {
+                #[allow(clippy::unwrap_used)]
+                let db = self.stable.lock().unwrap();
+                db.epoch_transition_progress().map_err(StateError::Storage)?
+            };
+
+            // NOTE: Recomputing missing rewards at the boundary
+            //
+            // The rewards summary is normally computed when the chain crosses the stability
+            // window, and only held in the in-memory volatile overlay. A node restarting with
+            // its stable tip already past the stability window replays no block within the
+            // window, so the summary is missing when the epoch boundary is reached — which used
+            // to kill consensus with 'RewardsSummaryNotReady'. All inputs to the computation
+            // (epoch snapshots and stake distributions) are persisted and frozen for the epoch,
+            // so recomputing here yields the same summary as computing it at the stability
+            // window.
+            if progress.is_none() && computed_rewards.is_none() {
+                warn!(
+                    epoch = u64::from(next_epoch - 1),
+                    "rewards summary missing at epoch boundary (restart past the stability window?); recomputing",
+                );
+                computed_rewards = Some(self.compute_rewards(next_epoch - 1)?.into());
+            }
 
             #[allow(clippy::unwrap_used)]
             let db = self.stable.lock().unwrap();
-
-            let progress = db.epoch_transition_progress().map_err(StateError::Storage)?;
 
             let protocol_parameters = self.protocol_parameters();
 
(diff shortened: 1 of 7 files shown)


crates/amaru-ledger/src/state.rs
```diff
@@ -371,13 +371,35 @@ impl<S: Store, HS: HistoricalStores> State<S, HS> {
...
+            // (epoch snapshots and stake distributions) are persisted and frozen for the epoch,
+            // so recomputing here yields the same summary as computing it at the stability
+            // window.
+            if progress.is_none() && computed_rewards.is_none() {
+                warn!(
+                    epoch = u64::from(next_epoch - 1),
+                    "rewards summary missing at epoch boundary (restart past the stability window?); recomputing",
+                );
+                computed_rewards = Some(self.compute_rewards(next_epoch - 1)?.into());
+            }
```
KtorZ: I don't think this is _the right fix_; as in, this shouldn't happen to begin with. The logic should have already computed the rewards when arriving here; and if they aren't there, that means the previous logic is wrong. From a quick look, we have:

```rs
            // 1. Rewards calculation
            BlockValidation::from(self.try_compute_rewards(*point))?;

            // 2. Epoch transition
            BlockValidation::from(self.try_epoch_transition(*point))?;
```

So there's indeed an attempt at computing rewards prior to the transition. In `try_compute_rewards` we do:

```rs
        let is_stake_distribution_stable = next_relative_slot >= self.global_parameters.stability_window();
        if self.volatile.rewards_not_ready() && is_stake_distribution_stable {
          ...
        }
```

which ensures rewards are only computed once they are "stable enough". That means, when we are `3*k/f` deep in the epoch. But then, in `try_epoch_transition` we gate the epoch transition with:

```
        if next_epoch > current_epoch {
          ...
        }
```

Which triggers an epoch transition when the next block would belong to the next epoch. So the core issue likely comes from how we look at the stability window for the next slot, rather than the current tip. If we stop Amaru right at the last slot of an epoch and restart, we've lost the previously calculated rewards and yet `is_stake_distribution_stable` is going to be false because the next slot is not beyond the stability window (it's the first slot of the new epoch). 

But it's also not as simple as a one-off mistake here. Suppose that we stop Amaru within the `k` first blocks of an epoch. We need to recompute the entire overlay / epoch transition (and thus we need the rewards). 

So yes, it is the `try_compute_rewards` logic which is needs to be adjusted here.

crates/amaru-ledger/src/state.rs
```diff
@@ -1062,3 +1084,365 @@ impl From<governance::Error> for StateError {
...
+    // transactional context so the whole test is self-contained (using
+    // `amaru-stores` from unit tests would link a second instance of this
+    // crate and break trait identities).
+    #[derive(Clone)]
+    struct EmptyStore {
+        tip: Point,
+        epoch: Epoch,
+    }
+
+    impl ReadStore for EmptyStore {
```
KtorZ: We already have: https://github.com/pragma-org/amaru/blob/main/crates/amaru-ledger/tests/state_rollback.rs#L149 which I suppose would be nice to turn into a proper helper / utility for tests. Exposed by the store module. Instead of re-defining mocks everywhere.

crates/amaru-ledger/src/state.rs
```diff
@@ -436,16 +436,28 @@ impl<S: Store, HS: HistoricalStores> State<S, HS> {
...
 
+        // The next block may also close the current epoch while the rewards are still
+        // missing: on a node restarting with its stable tip already past the stability
+        // window, the previously computed summary (held only in the in-memory volatile
+        // overlay) is lost, and no block within the window remains to be replayed. The
+        // relative slot of the next block is then *below* the stability window (it belongs
+        // to the next epoch), so the check above never triggers, and the epoch transition
+        // used to die with 'RewardsSummaryNotReady'. The epoch being closed is complete, so
+        // its inputs are final and the rewards can (and must) be computed now.
+        let is_epoch_ending = next_epoch > current_epoch;
```
KtorZ: This _works_ but makes the rewards calculation happens at the worse possible moment (just before an epoch boundary). In practice, we simply want to check: 

`current_relative_slot > self.global_parameters.stability_window()`; 

That covers the new situation and the previous one.

paolino:
Verified under Antithesis fault injection — 1h of kill/restart faults on the `cardano_amaru` testnet, the conditions that reproduced the crash. No `Consensus died` or `rewards summary not ready` over the run.

Report: https://amaru-cardano.antithesis.com/report/6GSpxlKM-dHijif5prjqxAd3/xPC8BlNcRyxYlB25tbWdm_bY8sV97EgAHjwnPUI1MTg.html?auth=v2.public.eyJuYmYiOiIyMDI2LTA3LTA0VDA5OjM3OjI1Ljk1NTkwMTI5M1oiLCJzY29wZSI6eyJSZXBvcnRTY29wZVYxIjp7ImFzc2V0IjoieFBDOEJsTmNSeXhZbEIyNXRiV2RtX2JZOHNWOTdFZ0FIanduUFVJMU1UZy5odG1sIiwicmVwb3J0X2lkIjoiNkdTcHhsS00tZEhpamlmNXByanF4QWQzIn19fQAn6ZKgxsj9BEkiCzb76oT4DbK9YN6tw4zewWZwLGhKsdDCNyX93B4IOKx2PaMRCWxOg86-ibKZfy1-R0K2EAY

KtorZ:
Note: I had to rework the logic around rewards calculation greatly in #1010 for other but connected reasons. So the core of this PR is likely now superseded. I am interested though in preserving the tests from this PR possibly.

PR #1009, opened 2026-07-03
feat: atomic switch to fork

### LINKED ISSUE #967: Switch to a fork atomically in the ledger database
### Abstract

We currently switch to a fork with 2 separate effects: a rollback and n roll forwards.
We need this to be atomic.

### Why?

Incoming headers are validated against the volatile DB state. If we switch to a fork non-atomically we risk the possibility of trying to validate a header right after a rollback, and before the subsequent roll forwards. 
In that case the header might fail to be validated, for example because it doesn't have the right opcert sequence number.

### How?

By using an atomic effect and rely on the current mutex on the ledger state.

This kind of issue should be found via simulation testing with a minimum of ledger modeling.

### Discussion points

_No response_

_No response_

### Checklist

etorreborre:
This PR makes sure that, in the case of a switch to a new fork, we fork the ledger atomically.
This way, processes which require the ledger state, like the mempool, cannot access some transient ledger state.

In order to do this:

- A new `SwitchToFork` effect has been introduced.
- A recovery procedure has been introduced in order to put back the ledger state to what it was in case a block from the new fork fails to apply.

Note: during the switch, the ledger is locked. We could relax this constraint and provide some read-only state to validate incoming transactions but this will be done in a subsequent PR.

## Summary by CodeRabbit

* **New Features**
  * Added fork-switching support during ledger validation for smoother chain recovery.
  * Improved node startup and sync flows with clearer, more structured ledger configuration handling.

* **Bug Fixes**
  * Validation now recovers more reliably when a proposed fork is not accepted.
  * Updated rollback behavior to better preserve and restore ledger state during recovery.
  * Improved test coverage and mock behavior to match the new validation flow.


etorreborre:
@coderabbit review

etorreborre:
@coderabbit review

crates/amaru-ledger/src/state/volatile/db.rs
```diff
@@ -503,46 +516,263 @@ mod tests {
...
+        // Rollback to slot 25 (between 20 and 30)
+        let rollback_point = Point::Specific(Slot::from(25), Hash::new([0u8; 32]));
 
         let result = db.rollback_to(&rollback_point);
 
-        // This should succeed
-        assert!(result.is_ok(), "Rolling back to a point after the sequence should succeed");
+        assert_eq!(result.unwrap_err(), &rollback_point);
         assert_eq!(db.len(), 3, "All elements should be retained");
     }
```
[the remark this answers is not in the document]
etorreborre: Fixed

crates/amaru-ledger/src/state/volatile/db.rs
```diff
@@ -401,6 +339,81 @@ impl VolatileDB {
 
         Ok(())
     }
+
+    /// Empty the volatile window and return its previous contents, leaving behind an empty window.
+    /// The current metadata (epoch, protocol parameters, governance activity) is kept rather than
+    /// reset to defaults. The returned `VolatileDB` can be used to fully restore the volatile state
+    /// (with [`Self::recover`]) in the case where we fail to fully switch to a new fork.
+    pub fn take(&mut self) -> VolatileDB {
```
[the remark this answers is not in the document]
etorreborre: Done.

crates/amaru-ledger/src/epoch_transition/rewards_state.rs
```diff
@@ -52,33 +52,38 @@ use amaru_kernel::{Lovelace, StakeCredential};
...
 
     /// Rewards have been computed but we haven't crossed the epoch boundary _yet_, so they are
     /// pending until ready to be applied.
-    Computed(Rewards<Computed>),
+    ///
+    /// Held behind an `Arc` so that snapshotting the volatile overlay (e.g. when capturing a
+    /// rollback recovery) doesn't require a deep copy of the per-account rewards map.
+    /// The rewards are only ever read or replaced wholesale while in the overlay, so
+    /// no copy-on-write mutation is needed.
+    Computed(Arc<Rewards<Computed>>),
```
yHSJ, line 72: I assume this is inherited from pure-stage, but it would be nice to not have `Arc`s in the ledger; the ledger *is* synchronous.

crates/amaru-consensus/src/stages/validate_block/mod.rs
```diff
@@ -94,152 +90,101 @@ pub async fn stage(mut state: ValidateBlock, msg: ValidateBlockMsg, eff: Effects
...
-        current_hash = current_header
-            .parent()
-            .or_terminate(eff, async move |_| {
-                // NOTE: parent links are validated by track_peers already, and we are younger than ledger_tip
-                tracing::error!(%current_hash, "reached genesis block while searching for rollback point");
-            })
-            .await;
-    }
-}
+// async fn roll_back_to_ancestor(
```
yHSJ: We can just delete any commented out code. It exists in the git history if it's needed later.

yHSJ (commented):
My first read through, this makes logical sense and I like the change generally. I'll go through it again now that I think I generally understand it for more granular feedback

yHSJ:
One thing I've been considering over lunch is whether or not it is possible to have a `switch_to_fork` call get reversed *after* blocks have been written to the stable db. I'm not even sure it's possible and, if it were possible, it would require a very specific case: two competing chains that are both maintained for *nearly* 2160 blocks, such that by the time our node is told to switch to the longer chain, that chain is >2160 blocks. It also must contain an invalid block at any point after 2160 blocks, so that the block(s) in the stable DB are not common to both forks.


If such a scenario is even possible, and I tend to assume it's not, it would cause issues in our state, since we never undo writes to the stable DB.

crates/amaru-ledger/src/epoch_transition/rewards_state.rs
etorreborre: That comes from the requirement that the ledger state needs to be `Send + Sync` for now to be shared across stages (via the `Arc<Mutex<State>>` in the `BlockValidator`).

crates/amaru-consensus/src/stages/validate_block/mod.rs
etorreborre: Of course, that was a left-over.

etorreborre:
@yHSJ no, it's not possible. When there's a fork, the new chain is at most one block longer than the previous one. If that last new block fails to validate, nothing gets written in the stable db.

crates/amaru-consensus/src/block_validator.rs
```diff
@@ -0,0 +1,144 @@
...
+}
+
+#[async_trait::async_trait]
+impl<S: Store + Send + Sync, HS: HistoricalStores + Send + Sync> CanValidateBlocks for BlockValidator<S, HS> {
+    #[expect(clippy::unwrap_used)]
+    async fn roll_forward_block(
+        &self,
+        point: &Point,
+        block: Block,
+    ) -> Result<Result<LedgerMetrics, BlockValidationError>, BlockValidationError> {
```
KtorZ, line 76: Kinda weird; why not preserve the `BlockValidation` type that's meant for that?

etorreborre: I guess that the original intention was to have only consensus error types at that level. I agree that it is weird but since it was already there in previous functions, I didn't touch it within this PR.

crates/amaru-ledger/Cargo.toml
```diff
@@ -44,7 +43,7 @@ amaru-uplc.workspace = true
 
 [dev-dependencies]
 # External dependencies ───────────────────────────────────────────────────────┐
-once_cell.workspace = true
+pallas-primitives.workspace = true
```
KtorZ: I don't think so no 😂 , this is now utterly forbidden. Please use `amaru-kernel` and expose from the kernel instead. I want to limit the amount of chasing we have to do for these.

crates/amaru/src/stages/config.rs
```diff
@@ -117,6 +111,46 @@ impl Default for Config {
...
+    pub ledger_vm_alloc_arena_count: usize,
+
+    // Initial size (in bytes) of each allocation arena to use for script evaluation in the ledger
+    // virtual machine. Higher sizes means less re-allocations but more resident memory footprint
+    // since the arena is leaking memory on purpose.
+    pub ledger_vm_alloc_arena_size: usize,
+}
+
+impl LedgerConfig {
+    pub fn consensus_parameters(&self) -> ConsensusParameters {
```
KtorZ:
```suggestion
    pub fn to_consensus_parameters(&self) -> ConsensusParameters {
```

Given that this is cloning underneath, it's good to make this appear in the name using the usual conventions.

crates/amaru-ledger/src/state.rs
```diff
@@ -787,6 +800,18 @@ impl<S: Store, HS: HistoricalStores + Send> State<S, HS> {
         })
     }
 
+    /// Restore the volatile db to its state prior to a rollback.
+    pub fn recover(&mut self, state_recovery: StateRecovery) {
```
KtorZ: I believe this should be a faillible function. Either explicitly by returning a result, or with a big runtime assertion/panic.

If the immutable tip has changed, we cannot guarantee the recovery. In the case of the full recovery, it is certain that we are now on a corrupted path; in the case of the partial recovery, I need to check the implementation of `.recover` but it's not obvious that it is a safe operation on _any_ volatile.

crates/amaru-ledger/src/state/volatile/overlay.rs
```diff
@@ -44,7 +44,7 @@ use crate::{
 /// This lives inside the [`crate::state::volatile::VolatileDB`]: it is part of the volatile state
 /// (it can be rolled back), and co-locating it with the two volatile series keeps reads and
 /// rollback cohesive.
-#[derive(Default)]
+#[derive(Default, Debug, Clone)]
```
KtorZ: I disagree with the `Clone`. If only needed for test, introduce this behind a cfg pragma. We do not want *ever* to clone the overlay in production.

crates/amaru-ledger/src/state/volatile/fragment.rs
```diff
@@ -257,7 +257,7 @@ impl VolatileFragment {
 // --------------------------------------------------------------------------- AnchoredVolatileFragment
 
 /// A [`VolatileFragment`] anchored to a specific point and block issuer.
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
KtorZ: Same here; if needed for test, gate it behind a cfg PRAGMA. This is too risky to have in the normal path otherwise and soon, as we say in French, it will be "la foire a la saucisse".

crates/amaru-ledger/src/state/volatile/fragment.rs
```diff
@@ -88,7 +88,7 @@ impl<L, R, V> Existence<Bind<L, R, V>> {
 }
 
 /// Resulting state change coming from processing a block.
-#[derive(Debug, Default)]
+#[derive(Debug, Default, Clone)]
```
KtorZ: ☝️

crates/amaru-ledger/src/epoch_transition/rewards_state.rs
```diff
@@ -101,7 +116,7 @@ impl RewardsState {
 /// computed and effective rewards that occur at the epoch boundary. It ensures that we don't
 /// misuse computed rewards too early, and it reduces the amount of boilerplate in having to create
 /// multiple types.
-#[derive(Debug)]
+#[derive(Debug, PartialEq, Eq, Clone)]
```
KtorZ: ☝️

crates/amaru-ledger/src/epoch_transition/rewards_state.rs
```diff
@@ -52,33 +52,38 @@ use amaru_kernel::{Lovelace, StakeCredential};
...
 
-#[derive(Debug)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub struct Computed;
 impl KnownRewardState for Computed {
     type UnclaimedRewards = ();
 }
 
-#[derive(Debug)]
+#[derive(Debug, Clone, PartialEq, Eq)]
```
KtorZ: ☝️

crates/amaru-ledger/src/epoch_transition/rewards_state.rs
KtorZ, line 72: Hmmm. Not sure about this; but I assume this was done because of the `Clone`? If it's only for tests, then I rather not have the `Arc` at all here and do a full clone in tests.

crates/amaru-consensus/src/block_validator.rs
```diff
@@ -0,0 +1,149 @@
...
+            }
+            FindAncestorOnBestChainResult::Found { fork_point, forward_points } => {
+                let mut state = self.state.lock().unwrap();
+                let mut ledger_metrics = LedgerMetrics::default();
+                let state_recovery =
+                    state.rollback_to(&fork_point).map_err(|error| BlockValidationError::from(anyhow!(error)))?;
+                for point in forward_points.iter() {
+                    let block = self
+                        .chain_store
+                        .load_block(&point.hash())
```
KtorZ, line 104: Would be nice to get an iterator here rather than fetching point after point, no? That would allow to concurrently fetch and roll forward.

crates/amaru-ledger/src/state/volatile/db.rs
```diff
@@ -437,6 +375,81 @@ impl VolatileDB {
...
+    /// The current metadata (epoch, protocol parameters, governance activity) is kept rather than
+    /// reset to defaults. The returned `VolatileDB` can be used to fully restore the volatile state
+    /// through the whole-volatile recovery path if switching to the new fork fails.
+    pub fn take(&mut self) -> VolatileDB {
+        let current = mem::take(&mut self.current);
+        let draining = mem::take(&mut self.draining);
+        let overlay = self.overlay.clone();
+        if !draining.is_empty() {
+            self.overlay.rollback();
+        }
```
KtorZ: Not quite sure of what's going on here? If we are copying/taking the volatile, then why not just take the overlay with it? The draining is also being `mem::take` too, so if there's an overlay, it should stay; no?

crates/amaru-ledger/src/state/volatile/db.rs
```diff
@@ -437,6 +375,81 @@ impl VolatileDB {
...
+        if self.draining.has_point(point) {
+            // If we are rolling back to a point in the draining sequence, we need to
+            // promote it as current while discarding the entire current series.
+            let old_current = mem::take(&mut self.current);
+            self.current = mem::take(&mut self.draining);
+            let drained = self.current.rollback_to(point)?;
+
+            // We must also rollback the overlay since we are crossing the epoch boundary again.
+            let overlay = self.overlay.clone();
+            self.overlay.rollback();
```
KtorZ: We need the same approach for the overlay as we have for the volatile; `.rollback()` needs to emit an object that allows for recovery so that can avoid cloning something big here.

crates/amaru-ledger/src/state/volatile/db.rs
```diff
@@ -437,6 +375,81 @@ impl VolatileDB {
...
+            Ok(VolatileDBRecovery::RecoverAcrossEpoch { fork_point: *point, old_current, drained, overlay })
+        } else {
+            let discarded = self.current.rollback_to(point)?;
+            let overlay = self.overlay.clone();
+            Ok(VolatileDBRecovery::RecoverInEpoch { fork_point: *point, discarded, overlay })
```
KtorZ: We shouldn't need the overlay if we have rolled-back within the epoch. It should suffice to restore the discarded blocks.

crates/amaru-ledger/src/state.rs
```diff
@@ -824,7 +849,7 @@ impl<S: Store, HS: HistoricalStores + Send> State<S, HS> {
         }
     }
 
-    pub fn rollback_to(&mut self, to: &Point) -> Result<(), BackwardError> {
+    pub fn rollback_to(&mut self, to: &Point) -> Result<StateRecovery, BackwardError> {
```
KtorZ: I have to say that I find it a bit scary to have a 2-step API here. I wonder if we couldn't try a continuation passing style with a closure to apply block or _rollback/undo_ automatically on failure, or provide a blocks iterator directly in there. 

At least to reduce the chances of discrepancy between a rollback_to followed by a recovery. There is currently nothing that guarantees those things happen in sequence and in an atomic fashion; so it's a nice footgun.

crates/amaru-ledger/src/state/volatile/db.rs
```diff
@@ -437,6 +375,81 @@ impl VolatileDB {
...
+        }
+    }
+
+    /// Restore the volatile DB to its pre-rollback state, undoing both the rollback and any
+    /// roll-forwards replayed since (a fork switch replays blocks before it may recover).
+    pub fn recover(&mut self, recovery: VolatileDBRecovery) {
+        match recovery {
+            VolatileDBRecovery::RecoverInEpoch { fork_point, discarded, overlay } => {
+                self.start_recovery_from(&fork_point);
```
KtorZ: I don't think this function (`start_recovery_from`) should be reused between the two branch? If we `RecoverInEpoch` then, necessarily it shouldn't be possible to ever have: `self.draining.has_point(point)`; so I find it confusing (and a potential room for logic bugs) to go through that path here.

crates/amaru-ledger/src/state.rs
etorreborre: Done.

crates/amaru-consensus/src/block_validator.rs
etorreborre: Done.

crates/amaru-ledger/src/epoch_transition/rewards_state.rs
etorreborre: Removed the `Clone`

crates/amaru-ledger/src/epoch_transition/rewards_state.rs
etorreborre: Removed the `Clone`

crates/amaru-ledger/src/state/volatile/fragment.rs
etorreborre: Removed the `Clone`.

crates/amaru-ledger/src/state/volatile/fragment.rs
etorreborre: Remove the `Clone`.

> it will be "la foire a la saucisse".

IYKYK :-D

crates/amaru-ledger/src/state/volatile/overlay.rs
etorreborre: Removed the `Clone`

crates/amaru/src/stages/config.rs
etorreborre: Done.

crates/amaru-ledger/Cargo.toml
etorreborre: Of course, done.

crates/amaru-ledger/src/state/volatile/db.rs
etorreborre: We are going to apply roll-forwards on top of that volatile part, so if there was a `draining` part, we need to rollback the overlay. This is similar to what the current `VolatileDB::rollback_to` does.

crates/amaru-ledger/src/state/volatile/db.rs
etorreborre: See the recent changes and the comments in the `StateOverlay::snapshot` method.

crates/amaru-ledger/src/state/volatile/db.rs
etorreborre: See the recent changes and the comments in the `StateOverlay::snapshot` method.

crates/amaru-ledger/src/state/volatile/db.rs
etorreborre: It's actually possible if we end up crossing an epoch when rolling blocks forward. I added a full test case showing that situation.

crates/amaru-ledger/src/epoch_transition/rewards_state.rs
etorreborre: This is done to cheaply snapshot the `StateOverlay` and share it with the ledger as it is rolling forward blocks but also keep it for recovery.

crates/amaru-ledger/src/state.rs
etorreborre: I've added an assert to check that when we recover the immutable tip hasn't changed.

etorreborre:
@KtorZ I made adjustments to remove cloning. The strategy is now to:

- Take a "snapshot" of the `StateOverlay` when rolling back. That snapshot is cheap because it's mostly cloning `Arc` values.
- During roll forwards, we have 3 cases:
   1. Pool updates might be replaced wholesale -> in the case the recovery instance keeps the original pool updates value.
   2. Governance updates follow the same pattern.
   3. Rewards have been restructured to contain: the list of accounts + a set of unclaimed accounts. Then, they are passed around by reference and the only copy that can occur is the list of unclaimed accounts if it needs to be modified (but it's arguably small).
   
So, overall, the approach to recovery for the `StateOverlay` is:

- Keep a snapshot.
- Copy on write just the minimal amount of data.
- Use the snapshot if recovery is required.

I also:

 - Encapsulated the whole switch to fork as one function on the ledger API.
 - Added an `assert` testing that we won't modify the stable state during a switch to fork by having too many blocks to roll forward.
 - Added another `assert` to check that the immutable tip hasn't moved before doing the recovery.

crates/amaru-ledger/src/epoch_transition/rewards_state.rs
```diff
@@ -60,36 +64,65 @@ pub enum RewardsState {
...
 impl RewardsState {
+    /// Snapshot the rewards state for rollback recovery. The rewards are held behind an `Arc`, so
+    /// this only bumps the reference count rather than deep-copying the per-account map.
+    pub fn snapshot(&self) -> RewardsState {
+        match self {
+            RewardsState::NotReady => RewardsState::NotReady,
+            RewardsState::Computed(rewards) => RewardsState::Computed(Arc::clone(rewards)),
+            RewardsState::Effective(rewards) => RewardsState::Effective(Arc::clone(rewards)),
+        }
+    }
```
KtorZ, line 108: Isn't that simply `.clone` ? I am picking on it because `snapshot` is already an overloaded term. I suppose this exists in this form to avoid deriving `Clone` (cf my previous remarks 😅).

etorreborre: Yes, that's to avoid a blanket `Clone` instance. Would you prefer something else like `checkpoint`?

crates/amaru-ledger/src/epoch_transition/rewards_state.rs
```diff
@@ -60,36 +64,65 @@ pub enum RewardsState {
...
+
     /// Consume computed rewards from the state, if available.
     pub fn take_computed_rewards(&mut self) -> Option<Rewards<Computed>> {
         match std::mem::replace(self, Self::NotReady) {
             Self::NotReady | Self::Effective(_) => None,
-            Self::Computed(computed) => Some(computed),
+            Self::Computed(computed) => Some(Arc::try_unwrap(computed).unwrap_or_else(|rewards| rewards.snapshot())),
```
KtorZ, line 121:
```suggestion
            // NOTE: non-expensive clone of rewards
            // 
            // The `Rewards<T>` object wraps maps inside `Arc` internally, making clone relatively 
            // cheap here. Most of the time, there should be only a single reference to that `Arc`, but in 
            // case where we are attempting to switch to a new fork, there will be two: the one now being 
            // taken due to an epoch transition, and the one we stashed away to restore the state in case 
            // the new candidate chain is invalid and we have to switch back.
            Self::Computed(computed) => Arc::unwrap_or_clone(computed),
```

crates/amaru-ledger/src/epoch_transition/rewards_state.rs
```diff
@@ -60,36 +64,65 @@ pub enum RewardsState {
...
+        }
+    }
+
+    /// Rollback the rewards state when rolling back across an epoch
+    pub fn rollback(self) -> Self {
+        match self {
+            st @ (RewardsState::NotReady | RewardsState::Computed(..)) => st,
+            RewardsState::Effective(effective) => {
+                let effective = Arc::try_unwrap(effective).unwrap_or_else(|rewards| rewards.snapshot());
```
KtorZ, line 130:
```suggestion
                let effective = Arc::into_inner().unwrap_or_else(|| unreachable!("RewardState has more than one owner at the moment of a rollback?"))
```

crates/amaru-ledger/src/epoch_transition/rewards_state.rs
```diff
@@ -112,104 +145,152 @@ pub struct Rewards<STEP: KnownRewardState> {
...
+    /// boundary. For `Effective` this is the set of such account keys (their amounts stay in
+    /// `accounts` and are summed back into the treasury); for `Computed` it is `()`. Rolling an
+    /// `Effective` back to a `Computed` is therefore just dropping this set, and the reverse is
+    /// re-attaching it. Neither touches `accounts`.
     unclaimed: STEP::UnclaimedRewards,
 }
 
+impl<STEP: KnownRewardState> Rewards<STEP>
+where
+    STEP::UnclaimedRewards: Clone,
```
KtorZ, line 172: I don't think we do need Clone for the `Effective` step (and thus, for the unclaimed rewards)?

crates/amaru-ledger/src/epoch_transition/rewards_state.rs
```diff
@@ -112,104 +145,152 @@ pub struct Rewards<STEP: KnownRewardState> {
...
-    fn extract_rewards(&mut self, account: &StakeCredential) -> Option<Lovelace> {
-        self.accounts.remove(account)
-    }
-
-    /// Return leftovers rewards that couldn't be allocated to account because they no longer
-    /// exist. This method consumes (i.e. takes ownership) of the item because it is meant to be
-    /// called last.
-    fn unclaimed_rewards(&mut self) -> BTreeMap<StakeCredential, Lovelace> {
-        std::mem::take(&mut self.accounts)
-    }
```
KtorZ, line 143: Where have these gone? The `extract_rewards` is quite instrumental to the logic; this is how we determine unclaimed rewards.

crates/amaru-ledger/src/epoch_transition/rewards_state.rs
```diff
@@ -112,104 +145,152 @@ pub struct Rewards<STEP: KnownRewardState> {
...
+    pub fn new(computed_rewards: Rewards<Computed>, accounts: impl Iterator<Item = StakeCredential>) -> Self {
+        // The set of accounts still registered at the boundary, used to tell claimed from unclaimed.
+        let registered: BTreeSet<StakeCredential> = accounts.collect();
+
+        let unclaimed: BTreeSet<StakeCredential> = computed_rewards
+            .accounts
+            .iter()
+            .filter(|(account, reward)| **reward > 0 && !registered.contains(*account))
+            .map(|(account, _)| account.clone())
+            .collect();
```
KtorZ, line 230: I see. Not a huge fan of the double-iteration over `accounts` here; it's big. Nor the internal cloning of the stake credential, although this one is more minor. But it was nicely avoided before.

crates/amaru-ledger/src/epoch_transition/rewards_state.rs
etorreborre: There's currently no `Clone` on `Rewards`, hence the `Arc::try_unwrap(computed).unwrap_or_else...`. 
Do you propose to add `Clone` on `Rewards`?

crates/amaru-ledger/src/epoch_transition/rewards_state.rs
KtorZ: Also, this will likely mess with our snapshots and stake distribution down the line because this now stores under `registered` accounts with no rewards; they were omitted on purpose before.

crates/amaru-ledger/src/epoch_transition/rewards_state.rs
etorreborre: It has more than one owner since we snapshot the overlay before rollbacking it.

crates/amaru-ledger/src/epoch_transition/rewards_state.rs
etorreborre: We need it for `snapshot` that's used in `take_computed_rewards` above.

crates/amaru-ledger/src/epoch_transition/rewards_state.rs
KtorZ: Hmmm. Then we can check precisely that! There should be a single strong ref and we know by construction the clone is necessary; no need to `try_unwrap`

crates/amaru-ledger/src/epoch_transition/rewards_state.rs
KtorZ: Well, in this situation, yes. My preferred approach would still be to avoid the Arc entirely, because they make some bits of logic harder to resolve now (cf the unclaimed rewards). 

And we need to take some more time thinking about the consequences, but I believe that not using Arc will only be an issue if we have a situation where we have to cross two epoch boundaries during a switch; which would have _anyway_ many other problems...

crates/amaru-ledger/src/epoch_transition/rewards_state.rs
etorreborre: This has gone to the initialization of the `unclaimed` accounts in the `Rewards<Effective>::new` method. In that approach we don't consume anything in order to be able to cheaply clone the recovery data since `accounts` become `Arc<BTreeMap<StakeCredential, Lovelace>>`.

crates/amaru-ledger/src/epoch_transition/rewards_state.rs
etorreborre: > the double-iteration over accounts here

That shouldn't be double. It's a lazy iterator and `filter` + `map` are fused.

etorreborre: > Also, this will likely mess with our snapshots and stake distribution down the line because this now stores under registered accounts with no rewards; they were omitted on purpose before.

How so? This data doesn't seem to be serialized anywhere. 
However I realized that the instantiation via the `RewardSummary` could introduce accounts with zero rewards via leader rewards. I added a commit to fix that.

KtorZ: > That shouldn't be double. It's a lazy iterator and filter + map are fused.

I wasn't referring to the .filter and .map; but to (1) the .collect on the stable accounts iterator, and then, the second iteration on all the accounts in the computed rewards. Both have nearly 1.5M elements on Mainnet.

KtorZ: > How so? This data doesn't seem to be serialized anywhere.

Not directly, but this feeds into the next rewards summary which gets serialized. It's not a _major_ problem, I think, because it would simply add accounts with no rewards to the map, whereas they were previously left out. But it's annoying for the conformance, because we would now diverge from what the Haskell node does and we would need a way to distinguish accounts that earned 0 rewards from accounts that didn't earn any rewards.

KtorZ: The reason I don't like to collect and filter, is because we are effectively duplicating the memory. Before, we ensured that the computed rewards were consumed and turned into the effective rewards; and we would possible add a tiny overhead of the unclaimed accounts. 

Now, we will momentarily cause a spike in memory to store the set of all accounts in order to check for `.contains`. 

On the other hand, iterating over all accounts is already a problem to begin with (as stated in one of the TODO/FIXME note) and we should instead record de-registration that occurs during an epoch and use that to filter out the unclaimed rewards. 

So maybe I am just making a story out of nothing (or, something we ought to fix differently _anyway_).

PR #1011, opened 2026-07-07
Withdrawal Checks

### LINKED ISSUE #890: Ledger Predicate: ConwayWdrlNotDelegatedToDRep
### Abstract

Implement the predicate `ConwayWdrlNotDelegatedToDRep`

### Why?

Conformance

### How?

[Reference Implementation](https://github.com/IntersectMBO/cardano-ledger/blob/0cfbf861cfb456660a7b73281c6fb714a53d40f9/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ledger.hs#L492)

### Discussion points

_No response_

_No response_

### Checklist

### LINKED ISSUE #895: Ledger predicate: ConwayIncompleteWithdrawals
### Abstract

Implement predicate `ConwayIncompleteWithdrawals`

### Why?

Conformance

### How?

[Reference Implementation](https://github.com/IntersectMBO/cardano-ledger/blob/3448adc634eac8f97ec6616dc86a6c96dedab504/eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Ledger.hs#L341-L360)

### Discussion points

_No response_

_No response_

### Checklist

yHSJ:
Closes #890, #895

The sync fails due to an unrelated issue exposed by the logic introduced here. That is fixed by the changes in #1010 (confirmed after testing locally), and out of scope of these changes. This PR is still targeting main to avoid stacking PRs very deep, making merging and review easier.

This PR introduces the two following predicates:
- [x] `ConwayWdrlNotDelegatedToDRep`
- [x] `WithdrawalsNotInRewardsCERTS` (which, in pv11, is split into `ConwayWithdrawalsMissingAccounts` and `ConwayIncompleteWithdrawals`)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Conway-era withdrawal validation.
  - Withdrawals now require a registered reward account, full reward-balance withdrawal, and required DRep delegation.
  - Added clearer handling for invalid withdrawal and deregistration scenarios.

- **Tests**
  - Expanded conformance coverage for valid withdrawals, unregistered accounts, incomplete withdrawals, and missing DRep delegation.
  - Added coverage for withdrawal-and-deregister transactions.

- **Documentation**
  - Updated the unreleased changelog with the new withdrawal validation behavior.


crates/amaru-ledger/src/rules/transaction/phase_one/withdrawals.rs
```diff
@@ -56,15 +64,29 @@ where
...
-                    })
-                } else {
-                    Ok((credential, st))
+                    });
+                };
+
+                let account =
+                    context.lookup(&credential).ok_or(InvalidWithdrawals::AccountNotRegistered(credential.clone()))?;
+
+                if matches!(credential, StakeCredential::AddrKeyhash(_)) && account.drep.is_none() {
```
KtorZ, line 78: We need a test case for withdrawals from script credential without drep. Those are authorized in v10+ (as this line specifies)

crates/amaru-ledger/src/rules/transaction/phase_one/withdrawals.rs
```diff
@@ -56,15 +64,29 @@ where
...
+                    return Err(InvalidWithdrawals::MissingAccountDRepDelegation(credential.clone()));
                 }
+
+                if amount != account.rewards {
+                    return Err(InvalidWithdrawals::IncompleteWithdrawal {
+                        balance: account.rewards,
+                        withdrawal: amount,
+                    });
+                }
```
KtorZ, line 87: I believe those checks need to only be performed when the transaction `is_valid`. Only the network is being validated on the failed path (part of the UTXO rule). Other checks are part of `withdrawalsThatDoNotDrainAccounts` which is under the CERTS rule, gated by `isValid True`.

The withdrawn amount is still collected for the value preservation check, but whether the account is drained (or even exists) is left out of the checks. 

We need proper test coverage for those as well.

crates/amaru-ledger/src/state/volatile/series.rs
```diff
@@ -190,14 +209,64 @@ impl VolatileSeries {
...
+    fn decrement(counts: &mut BTreeMap<StakeCredential, u32>, credential: &StakeCredential) -> bool {
+        if let Some(count) = counts.get_mut(credential) {
+            *count -= 1;
+            if *count == 0 {
+                counts.remove(credential);
+                return true;
+            }
+        }
+        false
+    }
```
KtorZ, line 166: Not sure I fond of this whole idea here. For one main reason: this only solves the problem for accounts, but we likely have it for dreps as well, and maybe pools. 

So I rather settle on an approach that's a bit more general and can solve this nicely for all the models. We briefly discussed this, and I suppose that it could be solved nicely by keeping a list of (anchored) updates in the aggregate rather than a single aggregated `DiffBind`. 

If we need to resolve the state, we can quickly aggregate the vector of updates, and when a fragment is dropped, we can simply drop it from the updates (if any).

yHSJ: Agreed. But is that in scope for these changes? That feels like it should be it's own PR

KtorZ: We can no longer introduce code that we dislike and that we wanna "fix later". We are beyond that period; so yes. If we need to tackle a new problem, we must tackle it properly. We have accumulated enough technical debt.

yHSJ:
@KtorZ thank you for picking this up and making these improvements while I was out. These changes look good to me, but I can't approve my own PR, so that is what this comment is.

KtorZ:
Couple of TODOs still IMO:

I haven't tested this outside of the unit tests and quick e2e sanity check. 

But I just uploaded hundreds of epoch snapshots for preprod (and we had many for previews already). 

So it'd be nice to check against epoch snapshots and maybe add a few comments here and there about what's going on. I've had several iterations on this and some parts are scarves with comments. 

The newly introduced individual methods on DiffSet and DiffBind could also deserve some unit tests I think. 

And finally, if we manage to cleanup the pool registrations properly too; then we could get rid of the artificial aggregate recompute every 1080 blocks

crates/amaru-ledger/src/state/volatile/aggregate.rs
```diff
@@ -0,0 +1,290 @@
...
+    /// This exactness is what lets a series stabilize purely incrementally, with no periodic
+    /// recompute to fall back on. It is not automatic: a naive collapse loses information every
+    /// time two fragments merge (e.g. an account that registers, deregisters, then re-registers
+    /// within the window collapses to a single verdict, and retracting the front then reads the
+    /// wrong existence). Each field is therefore shaped so that retracting the front is exact,
+    /// each for its own reason:
+    ///
+    /// - `utxo`: utxos, by definition, are unique, so cleaning up specific UTxOs can never overwrite older state.
+    /// - `pools`: existence is monotonic-additive, a registration counts up and retirement is
+    ///   deferred to the epoch boundary, so retracting just decrements the count this fragment added.
```
KtorZ, line 169: Not sure about that one.

crates/amaru-ledger/src/state/volatile/aggregate.rs
```diff
@@ -0,0 +1,290 @@
...
+/// the oldest fragment (stabilization) and the newest (rollback) are therefore the same decrement.
+#[derive(Debug, Default, Clone)]
+struct PoolRegistrations {
+    counts: BTreeMap<PoolId, usize>,
+}
+
+impl PoolRegistrations {
+    /// Record a fragment's registrations, one increment per pool it registered.
+    fn extend<V>(&mut self, diff: &DiffEpochReg<PoolId, V>) {
+        for pool_id in diff.registered.keys() {
```
KtorZ: For what's worth; if seeing a pool de-registration certificate still means that the pool exists for this epoch :)

PR #1012, opened 2026-07-07
chore: make bootstrap leaner and more complete

jeluard:
Overall simplification of the bootstrap process. Noticeably, do not use an embedded set of `snapshot.json` but rely on R2 hosted index file (anonymous connections can't list R2 buckets).

* move publication to rust code
* snapshots are now compressed with zstd
* do not use a manifest but rely on R2 listing via index.json
* use same progress bar everywhere
* align make targets, env variables, etc..
* introduce a `reindex` command allowing to force rebuild the R2 hosted index file so that it matches actual files
* get rid of `pallas-hardano`

Usage:

```shell
AMARU_NETWORK=mainnet AMARU_EPOCH=600 ./target/release/amaru snapshot create
AMARU_NETWORK=mainnet ./target/release/amaru snapshot publish
```

## Summary by CodeRabbit

* **New Features**
  * Added bootstrap snapshot publishing via the CLI (including manifest-based 3-epoch uploads).
  * Snapshot archives are now **Zstandard-compressed** using **`.tar.zst`**.
  * Bootstrap snapshots support optional **SHA-256 checksums** and **published URL** tracking.
* **Bug Fixes**
  * Bootstrap downloads now **fail fast** on checksum mismatches.
  * Bootstrap bootstrapping improved for **locally generated** snapshots and packaged block import.
* **Improvements**
  * More detailed **progress reporting** for snapshot DB analysis and Mithril downloads.  
  * Updated Make targets/env guidance for bootstrap snapshot creation/publishing.


Makefile
```diff
@@ -78,23 +76,12 @@ help:
...
-		exit 1; \
-	fi; \
-	AMARU_NETWORK="$(AMARU_NETWORK)" \
-	AMARU_DIST_DIR="$(AMARU_DIST_DIR)" \
-	AWS_ACCESS_KEY_ID="$(AWS_ACCESS_KEY_ID)" \
-	AWS_SECRET_ACCESS_KEY="$(AWS_SECRET_ACCESS_KEY)" \
-	AWS_DEFAULT_REGION="$(AWS_DEFAULT_REGION)" \
-	BUCKET_NAME="$(BUCKET_NAME)" \
-	ENDPOINT="$(ENDPOINT)" \
-	bash ./scripts/publish-bootstrap-snapshots "$(BOOTSTRAP_SNAPSHOT_EPOCH)"
```
KtorZ: I don't think the Makefile commands are any useful anymore.

crates/amaru/src/bin/amaru/cmd/snapshot/publish/mod.rs
```diff
@@ -0,0 +1,173 @@
...
+        .stdout(Stdio::inherit())
+        .stderr(Stdio::inherit())
+        .status()?;
+
+    if !status.success() {
+        return Err(format!("aws s3 cp failed with status {status}").into());
+    }
+
+    Ok(())
+}
```
KtorZ: I am not sure this is a big win. Really not fond of wrapping other executables in our executable. It makes the command-line brittle and annoying to work with. I rather produce artifacts needed by other command-line to operate. In particular, the `aws s3` cli is quite okay on its own and I don't think it's ever been part of the annoyance with the snapshots.

crates/amaru/src/bin/amaru/cmd/snapshot/create/db_analyser.rs
```diff
@@ -81,25 +84,69 @@ fn run_logged_command(
...
+                if let Ok(s) = s.lock()
+                    && let Some(pb) = s.bar.as_ref()
+                {
+                    pb.tick(0);
+                }
+                thread::sleep(std::time::Duration::from_millis(100));
+            }
+        });
+        (handle, stop_tx)
+    });
```
KtorZ: Not sure what's going on here but this code screams it should be encapsulated in its own abstraction?

crates/amaru/src/bin/amaru/cmd/snapshot/publish/mod.rs
```diff
@@ -0,0 +1,173 @@
...
+    /// S3-compatible bucket name.
+    #[arg(long, env = "BUCKET_NAME")]
+    bucket: String,
+
+    /// S3-compatible endpoint URL (e.g. https://<id>.r2.cloudflarestorage.com).
+    #[arg(long, env = "ENDPOINT")]
+    endpoint: String,
+
+    /// Public base URL at which uploaded objects are reachable (e.g. https://pub-xxx.r2.dev).
+    /// If omitted, inferred from the first existing URL in the manifest.
```
KtorZ: This is _too magic_; can we simply use a hard coded default that is also displayed automatically through clap?

crates/amaru/src/bin/amaru/cmd/snapshot/publish/mod.rs
KtorZ: If we want to get away from having people use the aws s3 cli directly, then we can at least integrate the sdk: https://docs.rs/aws-sdk-s3/latest/aws_sdk_s3/

crates/amaru/src/bin/amaru/cmd/snapshot/publish/mod.rs
```diff
@@ -0,0 +1,173 @@
...
+        long,
+        value_name = amaru::value_names::NETWORK,
+        env = amaru::env_vars::NETWORK,
+    )]
+    network: NetworkName,
+
+    /// Starting epoch of the three-epoch window to publish. Defaults to the latest unpublished
+    /// epoch in the manifest.
+    #[arg(long, value_name = amaru::value_names::UINT)]
+    epoch: Option<u64>,
```
KtorZ: Everywhere else, we call "epoch" the arrival epoch; so we should find a better name for this one; maybe `start_epoch` or `first_epoch`

crates/amaru/src/bin/amaru/cmd/snapshot/publish/mod.rs
```diff
@@ -0,0 +1,173 @@
...
+        }
+
+        if is_reachable(&client, &object_url).await {
+            info!(epoch = target_epoch, url = %object_url, "snapshot already published, skipping upload");
+        } else {
+            info!(epoch = target_epoch, archive = %archive_path.display(), "uploading snapshot");
+            upload_to_s3(&archive_path, &bucket, &archive_name, &endpoint)?;
+        }
+
+        if !is_reachable(&client, &object_url).await {
```
KtorZ: This should be nested in the previous branch.

crates/amaru/src/bin/amaru/cmd/snapshot/publish/mod.rs
```diff
@@ -0,0 +1,173 @@
...
+        info!(epoch = target_epoch, url = %object_url, "snapshot published");
+
+        if let Some(entry) = entries.iter_mut().find(|e| u64::from(e.epoch) == target_epoch) {
+            entry.url = Some(object_url);
+        }
+    }
+
+    let tmp_path = manifest_path.with_extension("json.tmp");
+    fs::write(&tmp_path, serde_json::to_vec_pretty(&entries)?)?;
+    fs::rename(&tmp_path, &manifest_path)?;
```
KtorZ: Why the tmp file?

crates/amaru/src/bin/amaru/cmd/snapshot/publish.rs
```diff
@@ -0,0 +1,173 @@
...
+            )
+            .into());
+        }
+
+        info!(epoch = target_epoch, url = %object_url, "snapshot published");
+
+        if let Some(entry) = entries.iter_mut().find(|e| u64::from(e.epoch) == target_epoch) {
+            entry.url = Some(object_url);
+        }
+    }
```
KtorZ, line 149: The bootstrap code is full of thing like this; It would benefit a lot from having a proper `Manifest` data-structure with associated methods of this kind. This would (a) avoid repeating these methods and a variation thereof in multiple places and (b) it would help with both testing and readability of the code.

crates/amaru/src/bin/amaru/cmd/snapshot/create/mod.rs
```diff
@@ -446,32 +496,20 @@ fn infer_start_epoch(current_epoch: Epoch) -> Result<Epoch, Box<dyn std::error::
...
-    if fs::symlink_metadata(path)?.is_dir() {
-        fs::remove_dir_all(path)?;
-    } else {
-        fs::remove_file(path)?;
-    }
-
-    Ok(())
+fn wipe_immutable_dir_and_wrap(err: Box<dyn std::error::Error>, cardano_node_db: &Path) -> Box<dyn std::error::Error> {
+    let immutable_dir = cardano_node_db.join("immutable");
+    match fs::remove_dir_all(&immutable_dir) {
```
KtorZ: Wow wow wow! No. If this happens to me I'll be very upset. I can't stand tools that delete my own files without asking. Especially when it takes multiples hours / days to obtain those files. This is also against our [command-line guidelines/policy](https://github.com/pragma-org/amaru/blob/main/engineering-decision-records/019-guidelines-for-writing-cli.md): no destructive operations.

crates/amaru/src/bin/amaru/cmd/snapshot/create/mod.rs
```diff
@@ -346,38 +323,89 @@ struct SnapshotBuildContext<'a> {
...
-    target.archive_path = Some(prepared_archive_path.to_string_lossy().into_owned());
-    target.snapshot_path = Some(prepared_snapshot_path.to_string_lossy().into_owned());
-    write_epoch_metadata(context.metadata_dir, &target)?;
-
     info!(epoch = %target.epoch, slot = %target.slot, snapshot = %prepared_snapshot_path.display(), archive = %prepared_archive_path.display(), "finished epoch snapshot");
 
     Ok(target.slot)
 }
 
+fn write_manifest(
```
KtorZ: Please move as an associated method of a `Manifest` object.

crates/amaru-mithril/src/lib.rs
```diff
@@ -398,6 +365,126 @@ pub fn from_chunk_for_resume_point(
     Ok(chunk_for_slot(network, resume_point.slot_or_default().into())?.saturating_sub(1))
 }
 
+/// A block read from the cardano-node immutable store.
+#[derive(Debug)]
+pub struct ImmutableBlock {
+    pub hash: HeaderHash,
+    pub header_cbor: Vec<u8>,
+    pub raw_block: Vec<u8>,
+}
```
KtorZ: Let's try to stick with the pattern with been adopting across the codebase: one type per module (as much as reasonably possible). Avoid putting everything under `lib.rs` because it gets out of hand rapidly.

KtorZ (changes requested):

KtorZ:
Ola! Few remarks from playing with this:

- `node bootstrap --help` shows all the newly introduce `--s3` options under `Network Global Parameters Overrides`

- All the meta var for these options are unhelpful: `S3_BUCKET`, `S3_ENDPOINT`, `S3_REGION`, `S3_PUBLIC_URL`; those should help understand the expected format (e.g. show enum variants when relevant), instead of examples being given in description. 

- We've lost the snapshot download progress bar.

- Similarly, there's no progress report from uploading snapshots during `publish`. It just hangs there. 

- I don't think showing the `point` and the `key` in the trace events is useful; those are mostly the same information and it creates very long and redundant log lines:

  ```
  2026-07-26T08:40:04.290125Z  INFO amaru::bootstrap: snapshot.download epoch=999 point=86399953.5d37babd87c35b9d5464b8e6c81f655a2fae44a422a1104ca67a7f72476286af key=preview/86399953.5d37babd87c35b9d5464b8e6c81f655a2fae44a422a1104ca67a7f72476286af.tar.zst
  ```

- The log events on `snapshot create` are also quite noisy:

  ```
  2026-07-26T08:51:34.220181Z  INFO amaru::cli: snapshot.create snapshot_output_dir=/path/to/amaru/snapshots/mainnet config_dir=/path/to/amaru/data/mainnet/epoch-snapshots/work/cardano-node-config/mainnet cardano_node_db=/path/to/cardano-node/mainnet.db network=mainnet dist_dir=/path/to/amaru/data/mainnet/epoch-snapshots epoch="640" snapshots=
  2026-07-26T08:53:29.585999Z  INFO amaru::cmd::snapshot::create: finished epoch snapshot epoch=637 slot=190252774 snapshot=/path/to/amaru/snapshots/mainnet/190252774.2a019e6d6f524233fe0fa754736c3a8e027e7cafcfa0747a2887bda87fdc1f04 archive=/path/to/amaru/snapshots/mainnet/190252774.2a019e6d6f524233fe0fa754736c3a8e027e7cafcfa0747a2887bda87fdc1f04.tar.zst
  ```

  We should at least show relative paths using: [utils::path::relative_path](https://github.com/pragma-org/amaru/blob/main/crates/amaru-kernel/src/utils/path.rs#L18-L20); and showing both the `archive` and the `snapshot` is a little redundant. 

  Same remark for one of the `publish` log event, which shows both the snapshot and the key, that's redundant:

  ```
  2026-07-26T08:57:43.711791Z  INFO amaru::cmd::snapshot::publish: uploading archive=/path/to/amaru/snapshots/mainnet/190252774.2a019e6d6f524233fe0fa754736c3a8e027e7cafcfa0747a2887bda87fdc1f04.tar.zst key=mainnet/190252774.2a019e6d6f524233fe0fa754736c3a8e027e7cafcfa0747a2887bda87fdc1f04.tar.zst
  ```

- Some event targets are wrong, e.g. `amaru::cmd::snapshot::create` and `amaru::cmd::snapshot::publish` should be `amaru::cli`.

- Option names between `node bootstap` and `snapshot publish` are inconsistent:

  - `--s3-endpoint` vs `--endpoint` 
  - `--s3-bucket` vs `--bucket`
  - `--s3-region` vs `--region`
  - `--s3-public-url` vs `--public-url`

  I have a preference for the `--s3-*` versions, as the prefix helps (1) grouping those parameters nicely and (2) avoid future clashes and ambiguity by scoping them.

- I don't think `--aws-access-key-id` and `--aws-secret-access-key` should exist; we don't want to provide secret materials like this via command-line (which gets written in shell history and the like). Those should be requested from env, and a clear error message shown when not present.

- Having both an archive and a flat snapshot also seems redundant (and takes needless space locally). What about getting read of the decompressed form entirely and allowing bootstrap to happen from a compressed archive (while still allowing to ingest uncompressed ones)?

PR #1021, opened 2026-07-09
feat: track opcert sequence numbers

### LINKED ISSUE #950: Track the pools opcert sequence numbers
### Abstract

We don't currently track the pool opcerts as found in block data but hard code them in `ConsensusParameters`.
This means that we can't enforce the rule checking that the new opcert sequence number must be equal to last sequence number, or be last sequence number + 1.

### Why?

To apply the exact same rules as the ones followed by the Haskell node (and abide by the Praos specification).

### How?

- Move the opcert counters out of `ConsensusParameters` into their own store.
- Use a function reading data from that store during header validation.
- Update the store during block validation.
- Update the store during rollback.
- Integrate the store with snapshots (when persisting or bootstrapping).

There should be stricter cases in the `header-test-cases` + the end-to-end snapshot tests should pass without the workaround added in #946.

### Discussion points

_No response_

_No response_

### Checklist

etorreborre:
This PR:

 1. Collects opcert sequence numbers from snapshots when bootstrapping.
 2. Stores opcert sequence numbers in the chain store when a header declares a new sequence number for a given pool id.
 3. Checks the validity of a new sequence number for a given header (the missing check in #882).
 4. Migrates the database from V3 to V4 on startup to recover the opcert sequence numbers for already stored headers.

It has been tested by:

 - Adding more test cases to the headers validation test cases to cover errors like `SequenceNumberTooFarAhead` (resulting from the check of the header's opcert sequence number)
 - Passing the e2e tests in both `preprod` and `preview` where all the existing headers should still be correctly validated.

## Summary by CodeRabbit

* **New Features**
  * Added end-to-end support for tracking per-pool operational certificate sequence numbers, including chain-store persistence and lookup behavior.
  * Chain dump can optionally include op-cert sequence numbers via a new CLI flag.
  * Snapshot imports now persist richer bootstrap chain state (nonce state + sequence counters).
* **Bug Fixes**
  * Improved header validation by deriving issuer/pool details from the header and validating against stored per-pool sequence information, with clearer unknown-pool reporting.
* **Refactor**
  * Reworked Praos nonce/sequence-counter handling into dedicated components and updated chain-store consensus parameter plumbing.


etorreborre:
@coderabbit review

CHANGELOG.md
```diff
@@ -35,6 +35,15 @@ Other guiding principles:
   ```
 -->
 
+## v10.10.20260716 _[unreleased; planned for 2026-07-16]_
+
+ ### Added
```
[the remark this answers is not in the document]
etorreborre: Done.

crates/amaru-ouroboros-traits/src/praos/opcert_sequence_numbers.rs
```diff
@@ -0,0 +1,46 @@
...
+#[cbor(transparent)]
+pub struct OpcertSequenceNumbers {
+    #[n(0)]
+    sequence_numbers: BTreeMap<PoolId, u64>,
+}
+
+impl OpcertSequenceNumbers {
+    pub fn get(&self, pool_id: &PoolId) -> Option<u64> {
+        self.sequence_numbers.get(pool_id).cloned()
+    }
```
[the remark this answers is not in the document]
etorreborre: Done

crates/amaru-ouroboros-traits/src/stores/chain_store/overriding_chain_store.rs
```diff
@@ -48,13 +50,18 @@ struct Overrides {
...
+    >,
     has_header: Option<Box<dyn FnMut(&dyn BaseReadChainStore, &HeaderHash) -> bool + Send>>,
     store_header: Option<Box<dyn FnMut(&dyn ChainStore, &BlockHeader) -> Result<(), StoreError> + Send>>,
     set_anchor_hash: Option<Box<dyn FnMut(&dyn ChainStore, &HeaderHash) -> Result<(), StoreError> + Send>>,
     set_best_chain_hash: Option<Box<dyn FnMut(&dyn ChainStore, &HeaderHash) -> Result<(), StoreError> + Send>>,
     store_block: Option<Box<dyn FnMut(&dyn ChainStore, &HeaderHash, &RawBlock) -> Result<(), StoreError> + Send>>,
     set_block_valid: Option<Box<dyn FnMut(&dyn ChainStore, &HeaderHash, bool) -> Result<(), StoreError> + Send>>,
     put_nonces: Option<Box<dyn FnMut(&dyn ChainStore, &HeaderHash, &Nonces) -> Result<(), StoreError> + Send>>,
+    put_opcert_seed:
+        Option<Box<dyn FnMut(&dyn ChainStore, &OpcertSequenceNumbers, &Point) -> Result<(), StoreError> + Send>>,
```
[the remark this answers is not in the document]
etorreborre: Done.

crates/amaru-ouroboros-traits/src/praos/nonces.rs
```diff
@@ -0,0 +1,54 @@
+// Copyright 2025 PRAGMA
```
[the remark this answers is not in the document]
etorreborre: Done.

crates/amaru-ouroboros/tests/validation.rs
```diff
@@ -54,36 +227,36 @@ struct GeneratorContext {
...
 
-#[derive(Debug)]
 pub struct KesKeyWrapper {
     bytes: Vec<u8>,
 }
 
+impl std::fmt::Debug for KesKeyWrapper {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("KesKeyWrapper").field("bytes", &hex::encode(&self.bytes)).finish()
+    }
```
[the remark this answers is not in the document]
etorreborre: It's only for tests, it's ok.

crates/amaru/src/bin/ledger/cmd/sync.rs
```diff
@@ -155,19 +157,32 @@ async fn process_block(
...
         consensus_parameters,
         block_header.header(),
+        issuer,
+        last_opcert_sequence_number,
+        pool_summary,
         to_cbor(&block_header.header_body()).as_slice(),
-        Arc::new(block_validator.state.lock().unwrap().view_stake_distribution()),
         &epoch_nonce.active,
     )
     .and_then(|assertions| assertions.into_par_iter().try_for_each(|assert| assert()))?;
```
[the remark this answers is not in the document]
etorreborre: The blocks should be valid

crates/amaru-stores/src/rocksdb/consensus/tests.rs
```diff
@@ -1097,6 +1102,109 @@ fn read_snapshot_supports_child_tips_skip_invalid() {
     );
 }
 
+#[test]
+fn opcert_sequence_number_is_none_for_an_unknown_pool() {
+    with_db(|db| {
+        let pool_id: PoolId = run_strategy(any_pool_id());
+        let header = run_strategy(any_header());
+        db.store_header(&header).unwrap();
+        assert_eq!(db.get_latest_opcert_sequence_number(&pool_id, &header).unwrap(), None);
```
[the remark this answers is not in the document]
etorreborre: Done.

crates/amaru-ouroboros-traits/src/praos/opcert_counters.rs
```diff
@@ -0,0 +1,46 @@
...
+use std::collections::BTreeMap;
+
+use amaru_kernel::{PoolId, cbor, cbor as minicbor};
+
+#[derive(
+    Debug, PartialEq, Eq, PartialOrd, Ord, Clone, serde::Serialize, serde::Deserialize, cbor::Encode, cbor::Decode,
+)]
+#[cbor(transparent)]
+pub struct OpcertCounters {
+    #[n(0)]
```
rkuhn: What does this do in conjunction with `#[cbor(transparent)]`?

crates/amaru-ouroboros-traits/src/stores/chain_store/in_memory_chain_store.rs
```diff
@@ -298,6 +341,78 @@ impl BaseReadChainStore for InMemConsensusSnapshot {
...
+        let volatile = Slot::from(u64::from(anchor_slot) + 1)..=as_of_slot;
+        let floor = entries.range(volatile.clone()).next().map(|(slot, _)| *slot).unwrap_or(as_of_slot);
+        let candidates: BTreeMap<HeaderHash, u64> =
+            entries.range(volatile).flat_map(|(_, by_hash)| by_hash.iter().map(|(hash, seq)| (*hash, *seq))).collect();
+        (floor, candidates)
+    } else {
+        (as_of_slot, BTreeMap::new())
+    };
+
+    // 2. resolve candidates against the actual lineage of the parent
```
rkuhn, line 396: It seems that I’m missing something fundamental here: we get headers from peers along properly linked chains, meaning a later header is only validated after all its ancestors are already stored and processed. Given this, how could it ever be valid to receive a header whose parent does not have an entry in the opcert storage?

crates/amaru-ouroboros-traits/src/stores/chain_store/in_memory_chain_store.rs
```diff
@@ -298,6 +341,78 @@ impl BaseReadChainStore for InMemConsensusSnapshot {
...
+            pool_id,
+            header,
+        )
+    }
+}
+
+fn latest_opcert_sequence_number(
+    headers: &BTreeMap<HeaderHash, BlockHeader>,
+    anchor: &HeaderHash,
+    opcert_sequence_numbers: &BTreeMap<PoolId, BTreeMap<Slot, BTreeMap<HeaderHash, u64>>>,
```
rkuhn, line 371: How could there be multiple header hashes within the same slot by the same pool?

crates/amaru-ouroboros-traits/src/stores/chain_store/in_memory_chain_store.rs
```diff
@@ -323,6 +438,16 @@ impl DiagnosticChainStore for InMemoryChainStore {
         Box::new(inner.nonces.clone().into_iter())
     }
 
+    #[expect(clippy::unwrap_used)]
+    fn load_opcert_sequence_numbers(&self) -> Box<dyn Iterator<Item = (PoolId, Slot, HeaderHash, u64)> + '_> {
+        let inner = self.inner.lock().unwrap();
```
rkuhn, line 450: This smells dangerous: allowing arbitrary other code to run while holding a lock sooner or later will lead to deadlocks.

crates/amaru-ouroboros/src/praos/header.rs
```diff
@@ -58,10 +59,6 @@ pub enum AssertHeaderError {
     OperationalCertificate(#[from] AssertOperationalCertificateError),
     #[error("could not convert slice to array")]
     TryFromSliceError,
-    #[error("Unknown pool: {}", hex::encode(&pool[0..7]))]
-    UnknownPool { pool: PoolId },
-    #[error("{0}")]
-    PoolError(#[from] GetPoolError),
```
rkuhn, line 64: This will lead to interesting merge conflicts.

crates/amaru-stores/src/rocksdb/consensus/base_read_chain_store.rs
```diff
@@ -149,12 +154,106 @@ where
...
+                    break;
+                }
+                floor = floor.min(slot);
+                if let Some(sequence_number) = from_cbor(&value) {
+                    candidates.insert(hash, sequence_number);
+                }
+            }
+        }
+
+        // 2. Resolve candidates against the actual lineage of `as_of`
```
rkuhn, line 195: why is this logic duplicated?

crates/amaru-stores/src/rocksdb/consensus/migration.rs
```diff
@@ -117,6 +118,23 @@ fn migrate_to_v3(store: &RocksDBStore<DB>) -> Result<(), StoreError> {
     set_version(store, 3)
 }
 
+fn migrate_to_v4(store: &RocksDBStore<DB>) -> Result<(), StoreError> {
+    tracing::warn!(
+        "migrating chain DB to version 4: opcert sequence numbers are reconstructed from stored \
+           headers only; counters from before this database was bootstrapped are unknown, which can \
+           lead to incorrectly rejected headers from pools that have not produced a block since. \
+           Re-bootstrapping from a snapshot is the reliable option."
```
rkuhn, line 126: Reading this makes me think: how are new pool registrations handled during normal operation? Shouldn’t a header be accepted if the pool wasn’t yet known? It will be rejected if the pool doesn’t have stake anyway.

rkuhn (commented):
sorry, for the delay and for the incurred merge conflicts …

crates/amaru-ouroboros-traits/src/praos/opcert_counters.rs
etorreborre: The `derive` macro requires all fields to be annotated.

crates/amaru-stores/src/rocksdb/consensus/base_read_chain_store.rs
etorreborre: It's the same structure as the in-memory one but using different primitives to access data. One is test code, the other is production code, and they serve as a model of each other in the tests.

crates/amaru-ouroboros-traits/src/stores/chain_store/in_memory_chain_store.rs
etorreborre: That doesn't seem like a big deal here, cloning data that we return? That's quite consistent with other implementations like in `impl BaseReadChainStore for InMemoryChainStore`. You'd prefer the lock to be just scoped to the `clone`?

crates/amaru-ouroboros-traits/src/stores/chain_store/in_memory_chain_store.rs
etorreborre: That's because we don't store all the opcerts sequence numbers for all the pools at each header. For a given pool presenting an opcert sequence number we need to go back in time to find the previous header that modified the opcert sequence number for that same pool.

crates/amaru-stores/src/rocksdb/consensus/migration.rs
etorreborre: On a registration the "last_opcert_sequence_number" of a pool is retrieved as `0`. So the first header where that pool declares an initial `opcert_sequence_number` is accepted.

crates/amaru-ouroboros-traits/src/stores/chain_store/in_memory_chain_store.rs
etorreborre: I'm following the Haskell implementation here. This could happen with a fork and we retain the header data by slot.

crates/amaru-ouroboros-traits/src/praos/opcert_counters.rs
rkuhn: That makes me suspicious of that macro’s code quality — please ensure that there are serialization format tests.

crates/amaru-ouroboros-traits/src/stores/chain_store/in_memory_chain_store.rs
rkuhn: So what we need is a map `PoolId -> Slot -> u64` with a simple lookup for the pool followed by a range+last for the Slot. The algorithm in the code is a lot more complex and iterates, and I don’t yet understand why.

crates/amaru-ouroboros-traits/src/stores/chain_store/in_memory_chain_store.rs
rkuhn: How could the same pool create multiple headers in the same slot? If we see this, then we need to consider this header to be invalid, don’t we?

crates/amaru-ouroboros-traits/src/stores/chain_store/in_memory_chain_store.rs
rkuhn: No, the issue has nothing to do with cloning, it only has to do with the lock guard moved into the Iterator, holding the lock. The code using the Iterator can then easily call some code path that attempts to obtain the same lock, which deadlocks the stage. Therefore, all code that runs while holding a lock MUST be carefully controlled and audited, which is why returning a lock guard is a very strong anti-pattern. Sorry if I missed this in earlier PRs.

crates/amaru-stores/src/rocksdb/consensus/base_read_chain_store.rs
rkuhn: It would only be a model if there were two very different ways to express the same time, especially when one of them is obviously correct. This duplication only means that changing one file requires the same change to the other file, which I don’t consider to be a win.

crates/amaru-stores/src/rocksdb/consensus/migration.rs
rkuhn: Where in the code does this happen? I might have missed it, I didn’t see opcert tracking within the ledger rules for pool registration.

crates/amaru-ouroboros-traits/src/stores/chain_store/in_memory_chain_store.rs
etorreborre: As far as I can see this is not considered as invalid in the Haskell node. Also why would the first header received on that slot be preferred to the second one? This is why the full mapping is `PoolId -> Slot -> Header -> u64`

crates/amaru-ouroboros-traits/src/stores/chain_store/in_memory_chain_store.rs
etorreborre: See my comment below.

crates/amaru-stores/src/rocksdb/consensus/base_read_chain_store.rs
etorreborre: > when one of them is obviously correct

Which one? We have many tests for the RocksDB storage testing that a separate in-memory implementation agrees with the RocksDB one. If there's a failure, the error could be in either one. They could also be wrong in the same way but it's less likely to happen if they don't share code.

crates/amaru-stores/src/rocksdb/consensus/migration.rs
etorreborre: This happens [here](https://github.com/pragma-org/amaru/pull/1021/changes#diff-66d2fcf371e73c8383309b32315fb258482a344b01d03e52e04e49bb3e0f9d94R409):
```rustlang
// A registered pool with no recorded opcert behaves as if its counter were 0, so its
// first certificate may carry sequence number 0 or 1.
let latest_sequence_number = latest_sequence_number.unwrap_or(0);
```

crates/amaru-ouroboros-traits/src/stores/chain_store/in_memory_chain_store.rs
etorreborre: Isn't that code equivalent to
```rustlang
let inner = self.inner.lock().unwrap();
let sequence_numbers = inner.opcert_sequence_numbers.clone();
Box::new(sequence_numbers.into_iter().flat_map(|(pool_id, slots)| {
    slots.into_iter().flat_map(move |(slot, hs)| {
        hs.into_iter().map(move |(header_hash, counter)| (pool_id, slot, header_hash, counter))
    })
}))
// guard is dropped here
```

crates/amaru-ouroboros-traits/src/praos/opcert_counters.rs
etorreborre: I think that we will fix this for good when we move off the Pallas types and be able to upgrade our minicbor version. That issue has been fixed upstream some time ago now: https://github.com/twittner/minicbor/issues/14.

crates/amaru-ouroboros-traits/src/stores/chain_store/in_memory_chain_store.rs
KtorZ: > How could the same pool create multiple headers in the same slot? If we see this, then we need to consider this header to be invalid, don’t we?

People often have backup block producers running as "ghost" ready to be connected if needs be; and it happened in the past that an SPO misconfigured their firewalls and have their backup node diffuse (different) blocks than their main BP to the network; mostly because the backup will typically see different transactions.

crates/amaru-ouroboros-traits/src/praos/praos_trait.rs
```diff
@@ -0,0 +1,62 @@
...
+Provides an interface around Praos nonces & operational certificates. Praos maintains an ever
+evolving nonce to which every stake pool contributes when producing blocks using their VRF.
+
+Hence, the VRF output of each block is combined with a rolling nonce. Once a certain point
+within the epoch is reached, the nonce is anchored for the epoch. Nevertheless, the evolving
+nonce keeps evolving until the epoch ends for the next epoch.
+
+Summarizing:
+
+```
```
yHSJ, line 25: This is never closed, if we're trying to get it to render nicely in other views

etorreborre: Thanks.

crates/amaru-stores/src/rocksdb/consensus/base_read_chain_store.rs
```diff
@@ -149,12 +154,106 @@ where
...
+                    None => break,
+                }
+            }
+        }
+
+        // 3. Immutable fallback: get the newest entry at slot <= min(anchor, as_of)
+        // that sits on the best chain.
+        // Note: the case where as_of < anchor would only exist if we try to revalidate an old header
+        // when ingesting old blocks for example.
+        let bound = anchor_slot.min(as_of_slot);
```
rkuhn, line 217: It worries me slightly that we are tracking the end of the immutable part of the chain in two different databases — how certain are we that they will always hold the same information? This is an issue outside the scope of this PR, just noting it here so if I’m not wrong then we can open an issue to fix that.

rkuhn (approved):
With the recent clarification and communication as well as independent verification of the research on how the Haskell node does this, I agree with this PR and think that the storage scheme looks quite good. Sorry for the delay, @etorreborre !

I still think that the Haskell implementation is flawed, but that is an entirely different topic.

crates/amaru-stores/src/rocksdb/consensus/base_read_chain_store.rs
etorreborre: I guess that's why in Haskell it's called the `ExtLedgerState` for `Extended Ledger State` I guess.

PR #1025, opened 2026-07-09
Treasury Value Mismatch

### LINKED ISSUE #888: Ledger predicate: ConwayTreasuryValueMismatch
### Abstract

Implement the `ConwayTreasuryValueMismatch` predicate

### Why?

Conformance

### How?

[Reference Implementation](https://github.com/IntersectMBO/cardano-ledger/blob/0cfbf861cfb456660a7b73281c6fb714a53d40f9/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Ledger.hs#L461)

### Discussion points

_No response_

_No response_

### Checklist

yHSJ:
This PR introduces the `TreasuryValueMismatch` predicate, which needs the current epoch's treasury value at validation time. The treasury only changes at the epoch boundary but during the window in which an epoch transition is not yet stable, the stable store still holds the closing epoch's value. So the net boundary treasury delta is computed at the epoch transition and stashed on the overlay, letting reads resolve the new epoch's treasury before the flush.

## Summary by CodeRabbit

- **New Features**
  - Added validation ensuring a transaction’s declared treasury value matches the ledger’s current treasury.
  - Added unified support for treasury, reserves, fees, and donations.
  - Improved treasury tracking across epoch transitions, including rewards, donations, refunds, withdrawals, and rollbacks.

- **Bug Fixes**
  - Treasury values are now preserved correctly across pending transitions, flushing, and rollback.

- **Tests**
  - Added scenarios covering matching and mismatched treasury values.

- **Documentation**
  - Updated the unreleased changelog with Conway treasury consistency validation.


KtorZ:
> @yHSJ  It required some changes to the overlay; notably the unclaimed rewards (and gov proposals, etc.) were only calculated at apply

This is not right? The overlay is calculated at the epoch transition, not during `apply` 🤔

yHSJ:
> This is not right? The overlay is calculated at the epoch transition, not during apply 🤔

Poorly phrased, updated the wording in the PR description.

The overlay only contains the `rewards.delta_treasury`, which doesn't include donations, and pool/withdrawal missing account refunds requires a DB read. Computing and storing the value here *is* duplicated work, but is lower risk (without a significant added cost) than reworking the epoch transition logic in my opinion.

crates/amaru-ledger/src/state.rs
```diff
@@ -475,9 +476,37 @@ impl<S: Store, HS: HistoricalStores + Send> State<S, HS> {
...
+            //
+            // The treasury only changes at the boundary, and the whole change is knowable here. It
+            // mirrors exactly what `StateOverlay::apply` writes when the boundary is later flushed.
+            //
+            // We stash it on the overlay so the validation context can resolve the new epoch's
+            // treasury during the straddle window, without recomputing per-account work on the hot
+            // path. Account existence is resolved on the volatile-over-stable view (the same view
+            // the boundary already works against), not a raw stable read.
+            let closing_epoch_donations = db.pots()?.donations
+                + self.volatile.iter().map(|anchored| anchored.fragment.donations).sum::<Lovelace>();
```
KtorZ: This should be handled when building the VolatileView already,.

crates/amaru-ledger/src/state.rs
```diff
@@ -475,9 +476,37 @@ impl<S: Store, HS: HistoricalStores + Send> State<S, HS> {
...
+
+            let mut refund_leftovers: Lovelace = 0;
+            for (credential, amount) in pools_updates.refunds().chain(governance_updates.deposit_refunds.iter()) {
```
KtorZ: This is a bit more complexe though that's about the spirit. We have already computed the unclaimed rewards in the overlay; but what we are only missing here are the refunds from the beginning of epoch.

crates/amaru-ledger/src/state.rs
yHSJ: It can be nicer; the `VolatileView` doesn't currently have it, but that's a nicer place to put it, I think

crates/amaru-ledger/src/state.rs
yHSJ: I am not entirely sure what you mean here

KtorZ (approved):

PR #1041, opened 2026-07-16
add more powerful defer logic to track_peers

### LINKED ISSUE #947: Explain and tweak max header fetch headstart
### Abstract

Currently arbitrary limit of 2160 blocks that fetching headers can be ahead of block validation, should be shrunk to 500 and explained.

### Why?

so that future us knows what the number means

### How?

by making it a constant with a doc comment

code review

### Discussion points

_No response_

_No response_

### Checklist

### LINKED ISSUE #1019: Defer header validation until required stake distribution is available
Currently, we assume that the stake distribution for epoch “now − 2” is available if we bound the lead of header fetching relative to block adoption to some fixed number. This is not reliable, as exercised by the `preview` network last November.

## The Fix

Upon encountering the error that the stake distribution is not yet available (`GetPoolError::StakeDistributionNotAvailable` wrapped in other error types), that header is placed in a deferral queue and that peer is not asked for the next header (due to pipelining this will happen multiple times for a peer in a row). The deferral queue is rechecked every time the set of stake distributions known by the ledger changes.

To make the above asynchronous, the ledger is outfitted with a message sender towards the `track_peers` stage for notifying it of changes. The actual list of stake distributions is accessible via an effect and only contains the `pools` map, which is rather small — the `accounts` map is forbiddingly large and shall not be cloned. The effect only provides the `PoolSummary` for the given pool at the given slot.

This allows the lead to be bounded by a reasonable number (like 300 blocks) under nominal conditions, but falling back to a more conservative approach under exceptional conditions.

rkuhn:
fixes #1019 

This also adds proper handling of wall-clock time in order to check incoming headers for clock skew (and reject or defer validation as appropriate).

Please review especially the ledger part and take a look at the core track_peers logic — this needs to be understandable to all of us, and for some reason I had great difficulty to find a formulation that I can reason about.

I’ll fix the merge while you review, will push that only once you are done.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved header validation in scenarios with unavailable stake data, clock skew, or peers ahead of the ledger, with corrected deferral/retry behavior and proper handling when switching based on block height.
  * Improved stake-distribution update handling and deferred peer-request sequencing.
* **New Features**
  * Enhanced simulation/protocol time handling with global epoch offsets for more consistent timestamps and scheduling.
* **Documentation**
  * Added step-by-step guidance for fixing failing pure-stage trace-matching assertions.
  * Updated the changelog and marked a related architecture decision record as obsolete.


.grok/skills/stage-trace-test/SKILL.md
```diff
@@ -0,0 +1,96 @@
+---
```
etorreborre, line 1: Is there a way to add this skill to other agents as well?

crates/amaru-consensus/src/validate_header.rs
```diff
@@ -12,80 +12,58 @@
...
+/// (ValidateHeaderEffect) so that up-to-date resources (in particular PoolSummaries) can be
+/// obtained on each use.
+pub fn validate_header(
+    header: &BlockHeader,
+    consensus_parameters: Arc<ConsensusParameters>,
+    store: Arc<dyn ChainStore>,
+    pool_summaries: Arc<PoolSummaries>,
+    era_history: Arc<EraHistory>,
+) -> Result<(), ValidateHeaderError> {
+    let _span = debug_span!(consensus::header::VALIDATE, header_hash = &header.hash());
```
etorreborre: `in_scope` would be nicer here.

crates/amaru-ledger/src/block_validator.rs
```diff
@@ -87,6 +87,14 @@ impl<S: Store + Send, HS: HistoricalStores + Send> BlockValidator<S, HS> {
         let state = self.state.lock().unwrap();
         state.tip().into_owned()
     }
+
+    /// Set callback invoked when a new stake distribution is computed/available.
+    /// The provided PoolSummaries should be used to update resources for header validation.
+    #[expect(clippy::unwrap_used)]
+    pub fn set_on_stake_dist_updated(&self, cb: Arc<dyn Fn(PoolSummaries) + Send + Sync>) {
+        let mut state = self.state.lock().unwrap();
+        state.set_on_stake_dist_updated(cb);
```
etorreborre, line 96: I wouldn't mind the long name `set_on_stake_distribution_updated` personally.

crates/amaru-ledger/src/state.rs
```diff
@@ -114,6 +114,10 @@ where
     /// Which network are we connected to. This is mostly helpful for distinguishing between
     /// behavious that are network specifics (e.g. address discriminant).
     network: NetworkName,
+
+    /// Optional callback invoked whenever a new stake distribution snapshot is added.
+    /// Used to update resources and notify stages (e.g. track_peers) about fresh PoolSummaries.
+    on_stake_dist_updated: Option<Arc<dyn Fn(PoolSummaries) + Send + Sync>>,
```
etorreborre, line 119: This callback feels like a wart to an overall nice design. I think we still want to have a proper ledger stage, right? In that case we could have a comment here explaining that this is temporary (how hard is it to do the right thing, you think?)

crates/amaru/src/bin/ledger/cmd/sync.rs
```diff
@@ -162,12 +163,15 @@ async fn process_block(
     chain_store.store_block(&point.hash(), &network_block.raw_block())?;
     let epoch_nonce = praos_chain_store.evolve_nonce(&block_header)?;
 
-    // Verify block headers
+    // Verify block headers using the extracted pool summaries and the high-level validate path is
+    // exercised via the free function in consensus, but here we call assert_all directly (as before).
```
etorreborre: "the free function" is not super clear. That's `validate_header::validate_header` I suppose?

"(as before)"  is also not clear, before what?

crates/amaru/src/stages/build_node.rs
```diff
@@ -124,24 +126,62 @@ pub fn build_node(
...
+        #[expect(clippy::expect_used)]
+        if let Ok(rt) = tokio::runtime::Handle::try_current() {
+            rt.spawn(send);
+        } else {
+            let rt =
+                tokio::runtime::Builder::new_current_thread().build().expect("cannot build current thread runtime");
+            rt.block_on(send);
+        }
+    }));
+    // TODO: The runtime spawn/block_on hack above is required by the current Tokio integration
```
etorreborre, line 165: I think that answer my question about having a ledger stage :-).

engineering-decision-records/012-interaction-between-consensus-and-ledger.md
```diff
@@ -1,13 +1,18 @@
 ---
 type: architecture
-status: accepted
+status: obsolete
```
etorreborre, line 3: 🥇

crates/amaru/src/tests/setup.rs
```diff
@@ -196,8 +203,17 @@ fn set_resources(node_config: &NodeTestConfig, stage_graph: &mut impl StageGraph
     stage_graph.resources().put::<Arc<dyn DiagnosticChainStore>>(node_config.chain_store.clone());
     stage_graph.resources().put::<ResourceBlockValidation>(block_validation.clone());
     stage_graph.resources().put::<ResourceHasStakePools>(block_validation);
-    stage_graph.resources().put::<ResourceHeaderValidation>(Arc::new(MockCanValidateHeaders));
     stage_graph.resources().put::<ResourceTxValidation>(Arc::new(MockCanValidateTxs));
+
+    // Resources for ValidateHeaderEffect (which now calls the free validate fn directly).
```
etorreborre: > (which now calls the free validate fn directly)

I think this should be removed because that refers to a version of the code that does not exist anymore.
This can be added in one of the agent files (depending on the agent). I have
```
- Never reference the implementation plan, step numbers, ticket IDs, or in-progress refactor names in commit messages or code comments.
- Describe what the code does and why it exists, not where it sits in the current work.
```

crates/amaru-ledger/src/state.rs
etorreborre, line 119: Also same thing about the name, I'd rather have `on_stake_distribution_updated`.

crates/amaru-consensus/src/stages/track_peers/mod.rs
```diff
@@ -182,23 +217,23 @@ pub struct NewTip {
 
 impl NewTip {
     pub fn new(tip: Tip, parent: Point) -> Self {
+        // FIXME: shouldn’t the trace context come from upstream?
         NewTip { tip, parent, trace_context: Default::default() }
```
etorreborre: That function is unused, it can be removed.

crates/amaru-consensus/src/stages/track_peers/mod.rs
```diff
@@ -400,41 +511,114 @@ impl TrackPeers {
...
+
+                    let header_height = header.block_height();
+                    let limit = header_height - self.max_peer_lead;
+                    // maybe update ledger applied block height (rate-limited to 500ms or initial)
+                    if limit > self.ledger_applied_block_height
+                        && (now.saturating_since(self.ledger_last_checked_at) > Duration::from_millis(500)
+                            || self.ledger_applied_block_height == BlockHeight::from(0))
+                    {
+                        self.ledger_last_checked_at = now;
+                        self.ledger_applied_block_height = eff.external(VolatileTipEffect).await.block_height();
```
etorreborre, line 612: This departs from using the `Ledger` like we do for example on line 453, and provides a `volatile_tip` function
```
let ledger = Ledger::new(eff.clone()).with_trace_context(trace_context);
```
Is it on purpose?

crates/amaru-consensus/src/stages/track_peers/mod.rs
```diff
@@ -330,57 +369,129 @@ impl TrackPeers {
         Ok(())
     }
 
-    #[expect(clippy::too_many_arguments)]
-    async fn execute_roll_forward(
+    /// Try to defer this header validation due to missing stake distribution.
+    /// Returns true if deferred (and not adversarial).
+    /// Rejects (returns false to let caller do adversarial) if the missing dist is >1 epoch ahead.
+    fn try_defer_for_stake(&mut self, args: &RollForwardArgs, error: &ConsensusError) -> Option<DeferredHeader> {
+        let ConsensusError::InvalidHeader(_, vhe) = error else {
```
etorreborre: Can we encapsulate this as a method on `ConsensusError` directly?

crates/amaru-consensus/src/stages/track_peers/mod.rs
```diff
@@ -330,57 +369,129 @@ impl TrackPeers {
...
 
-        let ledger = Ledger::new(eff.clone()).with_trace_context(&trace_context);
-        let store = Store::new(eff.clone()).with_trace_context(&trace_context);
+    fn is_deferred(&self, peer: &Peer) -> bool {
+        self.deferred.iter().any(|d| d.peer == *peer)
+    }
 
-        let result = self.validate_header(&peer, variant, &header, tip, &ledger).await;
+    /// Try to execute a roll forward from a peer. Preconditions like maximum distance from applied
+    /// ledger height have already been checked. Stake distribution availability leads to deferral,
```
etorreborre: "Stake distribution unavailability" I suppose?

crates/amaru-consensus/src/stages/track_peers/mod.rs
```diff
@@ -330,57 +369,129 @@ impl TrackPeers {
...
+        let RollForwardArgs { peer, variant, header, tip, trace_context, .. } = &args;
+
+        let ledger = Ledger::new(eff.clone()).with_trace_context(trace_context);
+        let store = Store::new(eff.clone()).with_trace_context(trace_context);
+
+        let result = self.validate_header(peer, *variant, header, *tip, &ledger, now).await;
         let parent = match result {
             Ok(parent) => parent,
             Err(error) => {
+                if let Some(dh) = self.try_defer_for_stake(&args, &error) {
```
etorreborre, line 503: That's a matter of taste but I would separate the extraction of the error, from the making of the deferred header:
```
match validation_error_kind(error) {
    UnavailableStake(target) => return Err(make_deferred_header_for_unavailable_stake(target)),
    ClockSkew => return Err(make_deferred_header_for_clock_skew()),
    Other => ...
}
```

crates/amaru-consensus/src/stages/track_peers/mod.rs
```diff
@@ -168,9 +131,81 @@ struct PerPeer {
     highest: Tip,
 }
 
+#[derive(Default, Debug, Clone, PartialEq, serde::Serialize, serde::Deserialize)]
+enum DeferReason {
+    /// Wait until the ledger has reached at least this applied block height before asking the peer for more.
+    LedgerHeight { min_height: BlockHeight, header: BlockHeader, tip: Tip, variant: EraName },
+    /// The header's validation requires a stake distribution that is not yet available; hold the
+    /// data needed to re-validate and store once it arrives (via StakeDistUpdated).
+    StakeDistribution { epoch: Epoch, header: BlockHeader, tip: Tip, variant: EraName, rn_sent: bool },
```
etorreborre, line 151: `rn_sent` could be more explicit.

crates/amaru/src/stages/build_stage_graph.rs
```diff
@@ -97,10 +98,11 @@ pub fn build_stage_graph(
...
     let k = global_parameters.consensus_security_param;
-    // `track_peers` can safely look ahead by at most 4*k/f slots: one stability window for the
-    // previous epoch stake distribution to become due, plus one more window to observe a block
-    // that advances the ledger clock far enough to compute it.
-    let max_forecast = Slot::new(global_parameters.randomness_stabilization_window());
+
+    // The number of headers fetched via chainsync ahead of the applied ledger tip.
+    // This value should be large enough to avoid stalling block fetches, but not much
+    // larger because those headers also consume resources.
+    let max_peer_lead = 1000;
```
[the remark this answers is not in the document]
rkuhn: It doesn’t make a difference whether this value comes from a name with small or capital letters, the doc comment is the important part.

crates/amaru-consensus/src/stages/track_peers/mod.rs
```diff
@@ -168,9 +131,81 @@ struct PerPeer {
...
+/// A header (or request) that was deferred. The reason indicates what is blocking and what data
+/// (if any) must be retained to resume.
+#[derive(Debug, Clone, PartialEq, serde::Serialize, serde::Deserialize)]
+struct DeferredHeader {
+    peer: Peer,
+    handler: StageRef<chainsync::InitiatorMessage>,
+    reason: DeferReason,
+    trace_context: TraceContext,
+}
+impl Default for DeferredHeader {
```
etorreborre, line 170: Do we really need a `Default` instance? It looks like it's only needed to `take` elements. 
We could take the whole `Vec` instead and rebuild it:
```
  let mut blocked = BTreeSet::new();
let mut remaining = Vec::with_capacity(self.deferred.len());
for d in take(&mut self.deferred) {
    let defer = blocked.contains(&d.peer)
        || match &d.reason {
            DeferReason::LedgerHeight { min_height, .. } => curr_height < *min_height,
            DeferReason::StakeDistribution { epoch, .. } => self.max_epoch < *epoch,
            DeferReason::ClockSkew { min_time, .. } => current_time < *min_time,
            DeferReason::FollowUp { .. } => false,
            DeferReason::Placeholder => false,
        };
    if defer {
        blocked.insert(d.peer.clone());
        remaining.push(d);
        continue;
    }
    if let Err(dh) = self.try_roll_forward(d.into(), eff, current_time).await {
        remaining.push(dh);
    }
}
self.deferred = remaining;
```

crates/amaru-consensus/src/stages/track_peers/mod.rs
```diff
@@ -330,57 +369,129 @@ impl TrackPeers {
...
+    /// Try to execute a roll forward from a peer. Preconditions like maximum distance from applied
+    /// ledger height have already been checked. Stake distribution availability leads to deferral,
+    /// so that this method can be called again later with the same inputs. `Ok(())` is also
+    /// returned in case the peer was removed due to an unrecoverable error.
+    async fn try_roll_forward(
+        &mut self,
+        args: RollForwardArgs,
+        eff: &Effects<TrackPeersMsg>,
+        now: Instant,
+    ) -> Result<(), DeferredHeader> {
```
etorreborre, line 493: Why not return an `Option<DeferredHeader>`?

crates/amaru-consensus/src/stages/track_peers/mod.rs
```diff
@@ -330,57 +369,129 @@ impl TrackPeers {
...
+    ) -> Option<DeferredHeader> {
+        if !matches!(error, ConsensusError::HeaderSlotInNearFuture(_)) {
+            return None;
         }
+        // compute accurate wait using current clock and header onset from era; last clock check was before validation calculations
+        let now = eff.clock().await;
+        let elapsed = now.duration_since_global_epoch();
+        let onset = self.era_history.slot_to_relative_time_unchecked_horizon(args.header.slot()).unwrap_or_default();
+        let wait = onset.saturating_sub(elapsed);
+        eff.schedule_after(TrackPeersMsg::RecheckLedgerHeight, wait).await;
```
etorreborre: Why do we only schedule a `RecheckLedgerHeight` after a clock skew? Why not get a message from `adopt_chain` telling us that a new block has been applied?

etorreborre (approved):
👍 + possible nitpicks to consider (the ledger height recheck is the biggest concern)

.grok/skills/stage-trace-test/SKILL.md
rkuhn: I suppose the description should work, not sure where in the repo they expect to find this file. You’re using Claude, correct? Could you try it and if necessary add a symlink in the right place?

crates/amaru-consensus/src/validate_header.rs
rkuhn: you’re right, this doesn’t look nice — I prefer EnteredSpan for the outer one and in_scope for the inner; do you recall why we bother with these fine-grained spans? the durations are tiny

crates/amaru-ledger/src/block_validator.rs
rkuhn: Hmm, to my mind the current name is too long already — absent strong arguments I feel like this must be an acceptable compromise because we’re both unhappy ;-)

crates/amaru-ledger/src/state.rs
rkuhn: This callback has precisely the shape that is needed by header validation, so I don’t consider it temporary or interim, handing out full snapshots binds a lot more memory. We may later implement the invocation of this callback based on ledger state snapshots, though.

crates/amaru/src/bin/ledger/cmd/sync.rs
rkuhn: sorry, left-over comment from collaborating with the agent

crates/amaru/src/stages/build_node.rs
rkuhn: yes, partly — this PR is large enough already

crates/amaru-ledger/src/state.rs
rkuhn: same reply as above

crates/amaru-consensus/src/stages/track_peers/mod.rs
rkuhn: Yes, I see no value in wrapping Effects and cloning it, the pure-stage API is nice and clear already. I know that you introduced these wrappers and their traits so that the functions called by a stage can be independently unit tested, but none of that exists in track_peers, so I removed it.

crates/amaru-consensus/src/stages/track_peers/mod.rs
rkuhn: good question — the issue is that pattern matching doesn’t work on Box, which we are using to please The Clippy (whom I hate for this); will try it

crates/amaru-consensus/src/stages/track_peers/mod.rs
rkuhn: yeah, that’s the other way; no strong enough preference for me to change anything

crates/amaru-consensus/src/stages/track_peers/mod.rs
rkuhn: Yes, and it was, but then I got annoyed by those very lengthy lines. Perhaps that future in which humans basically stop interacting with the actual code will have its upsides.

crates/amaru-consensus/src/stages/track_peers/mod.rs
rkuhn: No, in Rust there are no optimisations done magically like by GHC, so I won’t build a new vector when there is a perfectly usable one already available.

crates/amaru-consensus/src/stages/track_peers/mod.rs
rkuhn: Because it has the wrong semantics: the error case must be treated, a full option can be silently dropped.

crates/amaru-consensus/src/stages/track_peers/mod.rs
rkuhn: Right, this was an open point that I forgot to fix before opening the PR: track_peers should ensure checking even while back-pressuring during high upstream message influx. If we hook this up with adopt_chain, then back-pressure will extend to that part of the system, which can lead to deadlocks. So I’ll need to make the mechanism more robust.

crates/amaru-consensus/src/stages/track_peers/mod.rs
etorreborre: Ok, I was not interpreting that as an error but a normal outcome.

> a full option can be silently dropped.

Wouldn't that trigger a warning?

crates/amaru-consensus/src/validate_header.rs
etorreborre: > the durations are tiny

Maybe just to double-check that they really are, that we don't accidentally pull more than we need for the validation.

.grok/skills/stage-trace-test/SKILL.md
etorreborre: Yes, you're right, symlinking should work. I'll add a link for Claude when this PR lands.

crates/amaru-consensus/src/stages/track_peers/mod.rs
rkuhn: see #1066

KtorZ (commented):
Nice.

PR #1051, opened 2026-07-17
improve logging when fetch-blocks has no peers

### LINKED ISSUE #1050: fix block-fetch ERROR immediately after startup
### Abstract

```
2026-07-16T21:09:24.178335Z ERROR amaru_consensus::stages::fetch_blocks: timeout fetching blocks req_id=1
```

### Why?

Errors should only be printed in anomalous situations, not as part of normal startup.

### How?

Defer initial block fetching (which happens when interrupting while syncing) until some peer is connected.

It would be great to create a simulation test for this … which will be quite some effort because many stages are involved and we need to keep it deterministic.

### Discussion points

_No response_

_No response_

### Checklist

rkuhn:
fixes #1050

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved block fetching: when no upstream peers are available, fetching is paused without error logs and automatically retried once peers connect.
  * Timeout logging is downgraded while paused, and returns to error-level after peers are contacted.
  * Stale “no peers available” replies are safely ignored.
* **Documentation**
  * Updated block-fetch documentation and the changelog to reflect the new pause/retry behavior.
* **Tests**
  * Added coverage for peer-unavailable pause, retry behavior, and stale request-id handling.
* **Chores**
  * Updated merge behavior for `CHANGELOG.md`.


etorreborre (approved):
🏎️

PR #1054, opened 2026-07-17
break amaru-node out of amaru crate

rkuhn:
We’ll want people to use Amaru as a library, which means offering a crate that cleanly provides
everything needed for running a node, without dependencies like `clap`. Unfortunately, this is only
possible with `cargo` by creating a separate crate, which this PR does.

I’m doing this now so that I can then add another front-end for running the e2e snapshot tests,
which takes the begin/end epochs as cmdline parameters and sets up logging appropriately.

This PR only moves things around, it does not contain any API changes (that will be necessary) or
behaviour changes, so you can focus the review on the build process only.

## Summary by CodeRabbit

* **New Features**
  * Amaru node functionality is now available as an embeddable library for integration into other applications.
  * Added/standardized HTTP transaction submission endpoint (`/api/submit/tx`) with CBOR support and clear status handling for accepted, duplicate, invalid, and unavailable cases.
  * Expanded in-memory node simulation and testing support, including multi-node state verification utilities.
* **Bug Fixes**
  * Improved shutdown signaling integration to reliably trigger graceful cancellation on exit signals.
* **Documentation**
  * Updated the unreleased changelog to reflect the library split and new capabilities.


jeluard:
Using amaru as a lib w/o clap has been doable almost from the get go using `default-features = false` as demonstrated [here](https://github.com/jeluard/amaru-wear/blob/main/rust/Cargo.toml#L11)
This approach has the upside of allowing end user to just use `amaru` as crate.

rkuhn:
@jeluard If we want to offer an artifact that is meant to be consumed as a library, then it would be highly irregular (and unexpected) that the user is required to put `default-features = false` into their dependency list to avoid spurious dependencies. We’ll probably trim things down more over time and move the particulars into `amaru`, at least I hope so.

KtorZ (approved):

PR #1059, opened 2026-07-19
track chainsync proto lifecycle in track_peers to release resources

### LINKED ISSUE #906: Record connections in `track_peers`
### Abstract

There is a `FIXME` in the `track_peers` `handle_from_upstream` function:
```
record this connection and create a mechanism for removing upon disconnect
```

### Why?

The stage creates some state entries  for a new peer, it should remove them when the peer disconnects.

### How?

A new notification from the connection Manager needs to be added to inform `track_peers` of disconnects — in particular this means that whenever the chainsync mini-protocol terminates (error or `Done`), a notification needs to be sent. This matches the `Initialize` message that will be sent upon next start of the mini-protocol. The notifications both need to contain the ConnectionId for proper correlation.

pure-stage tests verifying that all per-peer state is purged upon receiving the notification.

### Discussion points

_No response_

_No response_

### Checklist

rkuhn:
fixes #906

This also corrected the mistake of `track_peers` only using `Peer` to identify a connection. With
other node implementations using `bind()` before `connect()` we might actually have two connections
with the same `Peer`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cleanup when peer synchronization stops or a connection terminates.
  * Prevented stale synchronization state from affecting replacement or re-established connections.
  * Preserved correct ordering for deferred synchronization work on each connection.
  * Improved handling of terminated chain-sync sessions so they stop cleanly without producing extra actions.

* **Reliability**
  * Added supervised lifecycle handling across connection and network protocol sessions, improving detection and cleanup of stopped components.

### DIFF
--- a/crates/amaru-consensus/src/stages/track_peers/mod.rs
+++ b/crates/amaru-consensus/src/stages/track_peers/mod.rs
@@ -23,7 +23,7 @@ use amaru_kernel::{
     from_cbor_no_leftovers, num::CheckedSub,
 };
 use amaru_observability::{TraceContext, debug_record, debug_span};
-use amaru_ouroboros::praos::header::AssertHeaderError;
+use amaru_ouroboros::{ConnectionId, praos::header::AssertHeaderError};
 use amaru_ouroboros_traits::has_stake_distribution::GetPoolError;
 use amaru_protocols::{
     chainsync::{self, ChainSyncInitiatorMsg, HeaderContent},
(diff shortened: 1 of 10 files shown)


rkuhn:
@coderabbitai review

crates/amaru-consensus/src/stages/track_peers/mod.rs
```diff
@@ -115,7 +119,7 @@ use crate::{
 #[derive(Debug, Clone, PartialEq, serde::Serialize, serde::Deserialize)]
 pub struct TrackPeers {
     era_history: EraHistory,
-    upstream: BTreeMap<Peer, PerPeer>,
+    upstream: BTreeMap<ConnectionId, PerPeer>,
```
etorreborre, line 131: I was about to comment about that in the other PR :-).

crates/amaru-consensus/src/stages/track_peers/mod.rs
```diff
@@ -354,21 +435,29 @@ impl TrackPeers {
...
-        per_peer.highest = tip;
+        *current_ref = current_tip;
+        *highest_ref = tip;
         Ok(())
     }
 
+    /// Remove all per-connection state for a chainsync session. Idempotent.
+    fn purge_connection(&mut self, conn_id: ConnectionId) {
+        self.upstream.remove(&conn_id);
+        self.deferred.retain(|d| d.conn_id != conn_id);
```
etorreborre, line 464: Maybe a `BTreeMap` for `self.deferred` would be appropriate now since a few behaviors depend on `conn_id`?

etorreborre (approved):

crates/amaru-consensus/src/stages/track_peers/mod.rs
rkuhn: It could be a multimap, yes, but in practice the number of entries will be quite low almost all the time, so I prefer to keep it simple until proven inadequate.

PR #1071, opened 2026-07-21
test: phase-1 validation rules conformance tests

### LINKED ISSUE #1039: Run conformance tests against the Haskell node
### Abstract

We have a set of conformance tests, specified in JSON files, that we use to validate the ledger rules.
We need to run those tests against the Haskell to make sure that:

1. The tests are correct: if `amaru` succeeds and `cardano-node` fails, the test might be incorrect and not usable in a conformance suite.
2. `amaru` is correct: maybe `amaru` passes but it should actually fail (or the opposite). 

### Why?

Two objectives:
- Improve the confidence in our conformance test suite.
- Validate that this test suite can be used against any Cardano node.

### How?

There are several tasks:
 - [ ] Extract a separate binary from the `haskell-node-extractor` that is used on the `haskell-conformance-phase-one` branch prototyping the implementation of this issue.
 - [ ] Make the output JSON so that it can be parsed in CI.
 - [ ] Make the `phase_one` tests discoverable in `amaru` to simplify their maintenance
 - [ ] Fix the gaps
         - An `amaru` failure must be a subset of the `Haskell` failures
         - Recreate the tests that fail the deserialization (or fix the decoder)
         - Change the decoding strategy in `amaru`. We currently decode lazily parts of the transaction and end up with a different error than the Haskell node. Possibly for those tests, provide another fixture that will fail but not because of a deserialization failure.
         - Validation mismatch: a fixture uses a transaction marked as invalid, but the transaction itself is actually valid.

Running the executable must not raise any error.

### Discussion points

_No response_

_No response_

### Checklist

etorreborre:
This PR provides a way to test phase one validation rules against a Haskell node:

image

Errors are reported like this:

image

The PR itself gathers:

1. Test framework improvements

 - Moved with other tests in the `amaru` crate. 
 - Extracted a separate executable to run the tests (accepts both a single fixture, or a directory of fixtures).
 - Made a nicer output for human execution + JSON output to be used in CI.

2. Test framework fixes

Changes to the test framework that made some tests pass:

 - Decoding failures can be expected and some test cases assert that data cannot be decoded.
 - The Plutus cost models are loaded.
 - Phase two errors from the Haskell node are ignored.
 - Past-horizon failure `(UtxosFailure … TimeTranslationPastHorizon)` becomes `OutsideForecast`.
 - `IncorrectDepositDELEG` and `StakeKeyHasNonZeroAccountBalanceDELEG` are supported.
 - `InputSetEmptyUTxO` is properly handled and not as `BadInputsUTxO`.

  2. Test fixture fixes

  - `InvalidWitnessesUTXOW/{1,2,4}.json` have been reclassified decoding failures (and the corresponding code fixed, see below). As a result three orphaned CBOR fixtures (invalid-signature-length, invalid-key-length, invalid-signature-size) were dropped.
  - `OutsideForecast/0.json` has been regenerated to incorporate a real, signed Plutus V3 spend whose ttl (`500000`) exceeds the forecast horizon (`432000`).
  - `withdrawal.json` now registers the withdrawing account (with rewards = 0 and no DRep delegation).
  - `script-integrity-hash/0.json` registers the two script credentials (with rewards = 0 and 1234087, and no DRep delegation).  -

  4. Code fixes

  - `transaction.rs` `Transaction::decode` now checks the witness sizes, matching the Haskell ledger (this will be done differently once we migrate from Pallas).


crates/amaru/tests/conformance/stake-distributions/haskell-node-extractor/stake-distributions-preview.sh
KtorZ, line 1: I should not have committed this 🙈

KtorZ (approved):

PR #1073, opened 2026-07-21
feat: experiment with agentic workflow

jeluard:
Use agentic workflows to investigate failures of nightly jobs

## Summary by CodeRabbit

* **New Features**
  * Added an automated CI Doctor to investigate failed or cancelled nightly and continuous integration checks.
  * Generates structured failure reports with suspected causes, affected steps, and recommended actions.
  * Helps avoid duplicate reports and can provide follow-up comments or issues when appropriate.
  * Includes safeguards to prevent sensitive information and excessively large log excerpts from appearing in reports.

* **Bug Fixes**
  * Improved changelog automation so it responds to the correct CI workflow results.


KtorZ:
I am not sure about giving our GITHUB_TOKEN to AI agent 😶?

jeluard:
The `GITHUB_TOKEN` is manipulated in the yml file created by `gh aw` which is the official gh tool. So I would expect that it's safe. It apparently is only used to communicate back to gh mcp gateway and create issues.

If we are not confortable with this it's fine, but we can't use gh aw then AFAIK.

jeluard:
Parking for now

Issue #754, opened 2026-04-15
Add extra metrics: support gLiveView.sh


jeluard:
### Abstract

[gLiveView](https://cardano-community.github.io/guild-operators/Scripts/gliveview/) is a script used by SPOs to get a CLI based dashboard for `cardano-node`. Make sure it can also support amaru.

gLiveView relies on prometheus [metrics](https://github.com/cardano-community/guild-operators/blob/master/scripts/cnode-helper-scripts/env#L596) to get details about the running node. We want to ensure all exposed by amaru.

## Already implemented

*  cardano_node_metrics_blockNum_int
*  cardano_node_metrics_txsInMempool_int
*  cardano_node_metrics_mempoolBytes_int
*  cardano_node_metrics_txsProcessedNum_(int|counter)
*  cardano_node_metrics_slotNum_int
*  cardano_node_metrics_epoch_int
*  cardano_node_metrics_slotInEpoch_int
*  cardano_node_metrics_density_real

## To be implemented

*  [x] cardano_node_metrics_cardano_build_info
*  [x] cardano_node_metrics_served_block_count(er|_int)
*  [x] cardano_node_metrics_currentKESPeriod_int
*  [x] cardano_node_metrics_remainingKESPeriods_int
*  [ ] cardano_node_metrics_blockfetchclient_blockdelay_(s|real)
*  [ ] cardano_node_metrics_blockfetchclient_lateblocks(_counter|)
*  [ ] cardano_node_metrics_blockfetchclient_blockdelay_cdfOne(_real|)
*  [ ] cardano_node_metrics_blockfetchclient_blockdelay_cdfThree(_real|)
*  [ ] cardano_node_metrics_blockfetchclient_blockdelay_cdfFive(_real|)
*  [x] cardano_node_metrics_connectionManager_(incoming|inbound)Conns(_int|)
*  [x] cardano_node_metrics_connectionManager_(outgoing|outbound)Conns(_int|)
*  [x] cardano_node_metrics_connectionManager_unidirectionalConns(_int|)
*  [ ] cardano_node_metrics_slotsMissed(Num|)_int

## Cannot yet be implemented

Those metrics depend on amaru being a block producer.

* cardano_node_metrics_Forge_node_is_leader_(int|counter)
* cardano_node_metrics_Forge_adopted_(int|counter)
* cardano_node_metrics_Forge_forged_(int|counter)
* cardano_node_metrics_Forge(_forge|)_about_to_lead_(int|counter)
[...]

PR #1075, opened 2026-07-22
feat: automate prom metrics checks

jeluard:
Augment ledger epoch snapshots tests so that prometheus metrics exposed by amaru (via oltp-collector) are compared with the one exposed by `cardano-node`.

`scripts/compare-metrics.json` allows to define metrics/metrics prefixes to ignore.

You can see an example summary generated by the CI here: https://github.com/pragma-org/amaru/actions/runs/30457091133?pr=1075

## Summary by CodeRabbit

* **New Features**
  * Enhanced the ledger epoch snapshot workflow with OpenTelemetry export, including Prometheus exposure and an OTLP collector step.
  * Added automated metrics conformance checking between the Cardano node and Amaru, producing per-network JSON reports and optional Markdown coverage output.
  * Introduced configurable exclusions for metric comparisons (prefix-based and specific metrics), plus a check mode that flags missing metrics.
  * Added readiness checks to wait for both metrics endpoints before evaluating results.

* **Bug Fixes**
  * Improved nightly workflow failure artifacts to include both run logs and the tool’s output.


.github/workflows/nightly-tool-integrations.yml
```diff
@@ -215,3 +215,107 @@ jobs:
...
+
+      - name: Wait for cardano-node metrics endpoint
+        shell: bash
+        run: |
+          timeout 120 bash -c \
+            'until curl -sf http://localhost:12798/metrics >/dev/null; do sleep 2; done' || {
+            echo "cardano-node metrics endpoint did not become ready within 120s."
+            docker logs cardano-node 2>&1 | tail -30 || true
+            exit 1
+          }
```
KtorZ: Couldn't we simply do that during the epoch-snapshots tests? Since we have a running node, we might as well query the metrics and do the comparison as a follow-up?

jeluard: As in capture the expected metrics at this time, then use the nightly to check amaru conformance? That would require allowing the gh workflow to update the repo, which I'm always a bit worried about. 

An alternative might be to just manually refresh expected metrics every once in a while, as new releases are made. We could even keep expected metrics per cardano-node versions.

KtorZ: > then use the nightly to check amaru conformance?

No, then check the conformance as part of that test too, continuously. While the cardano-node metrics may not change often, we may break it accidentally.

.github/workflows/template-ledger-epoch-snapshots.yml
```diff
@@ -139,9 +156,46 @@ jobs:
...
+          timeout 120 bash -c \
+            'until curl -sf http://localhost:12798/metrics | grep -q "^cardano_node_metrics_"; do sleep 2; done' || {
+            echo "cardano-node metrics did not become available within 120s."
+            docker logs cardano-node 2>&1 | tail -30 || true
+            exit 1
+          }
+
+      - name: Check metrics conformance
+        if: ${{ inputs.network == '' || inputs.network == matrix.network.name }}
+        continue-on-error: true
```
KtorZ: This sends wrong signal with a green CI actually failing; if the check fails then the job should fail

KtorZ:
Seems like we are missing most of the metrics actually:

```json
{
  "cardano_total": 57,
  "cardano_filtered": 20,
  "amaru_total": 35,
  "missing_count": 17,
  "missing": [
    "cardano_node_metrics_Mem_resident_int",
    "cardano_node_metrics_Stat_cputicks_int",
    "cardano_node_metrics_Stat_threads_int",
    "cardano_node_metrics_connectionManager_duplexConns",
    "cardano_node_metrics_connectionManager_fullDuplexConns",
    "cardano_node_metrics_connectionManager_incomingConns",
    "cardano_node_metrics_connectionManager_outgoingConns",
    "cardano_node_metrics_connectionManager_unidirectionalConns",
    "cardano_node_metrics_forging_enabled",
    "cardano_node_metrics_inboundGovernor_cold",
    "cardano_node_metrics_inboundGovernor_hot",
    "cardano_node_metrics_inboundGovernor_idle",
    "cardano_node_metrics_inboundGovernor_warm",
    "cardano_node_metrics_nodeStartTime_int",
    "cardano_node_metrics_served_block_count_int",
    "cardano_node_metrics_served_block_latest_count_int",
    "cardano_node_metrics_served_header_counter_int"
  ]
}
```

Some of them seems like easy reach.

.github/workflows/template-ledger-epoch-snapshots.yml
jeluard: For now it will always fail due to some missing metrics. Or we can ignore them and create issues to track their implementation.

jeluard:
@KtorZ I suspect not all cardano-node metrics are exposed (due to conf? or not yet published?) as I know we have more in common.

jeluard:
Also see https://github.com/pragma-org/amaru/issues/754#issuecomment-5118852482

scripts/compare-metrics.json
```diff
@@ -0,0 +1,155 @@
...
+    "cardano_node_metrics_mempoolBytes_int",
+    "cardano_node_metrics_remainingKESPeriods_int",
+    "cardano_node_metrics_slotInEpoch_int",
+    "cardano_node_metrics_slotNum_int",
+    "cardano_node_metrics_tipBlock",
+    "cardano_node_metrics_txsInMempool_int",
+    "cardano_node_metrics_txsProcessedNum_int",
+    "cardano_node_metrics_txsSyncDuration_int"
+  ]
+}
```
KtorZ, line 153: 🙌

PR #1076, opened 2026-07-23
fix: incorrect default logs

jeluard:
Skip-Changelog

## Summary by CodeRabbit

* **Improvements**
  * Reduced default logging verbosity when `AMARU_LOG` / `AMARU_TRACE` are not set, focusing on errors and `amaru` at info level or higher.
  * Simplified default trace filtering to reduce low-level detail (while still capturing more critical events).
  * Updates tracing subscriber initialization behavior, including JSON trace output.

### DIFF
--- a/crates/amaru/src/observability.rs
+++ b/crates/amaru/src/observability.rs
@@ -43,11 +43,11 @@ use tracing_subscriber::{
 
 const AMARU_LOG_VAR: &str = "AMARU_LOG";
 
-const DEFAULT_AMARU_LOG_FILTER: &str = "info,amaru::consensus=debug,amaru::ledger=debug,amaru_pure_stage=warn";
+const DEFAULT_AMARU_LOG_FILTER: &str = "error,amaru=info";
 
 const AMARU_TRACE_VAR: &str = "AMARU_TRACE";
 
-const DEFAULT_AMARU_TRACE_FILTER: &str = "amaru=trace,amaru_pure_stage=trace,amaru_protocols=warn,amaru_consensus=info";
+const DEFAULT_AMARU_TRACE_FILTER: &str = "error,amaru=debug";
 
 const OTEL_ERROR_THROTTLE_MS: u64 = 5_000;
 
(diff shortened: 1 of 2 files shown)


crates/amaru/src/observability.rs
```diff
@@ -43,11 +43,11 @@ use tracing_subscriber::{
 
 const AMARU_LOG_VAR: &str = "AMARU_LOG";
 
-const DEFAULT_AMARU_LOG_FILTER: &str = "info,amaru::consensus=debug,amaru::ledger=debug,amaru_pure_stage=warn";
+const DEFAULT_AMARU_LOG_FILTER: &str = "info,amaru_pure_stage=warn";
 
 const AMARU_TRACE_VAR: &str = "AMARU_TRACE";
 
-const DEFAULT_AMARU_TRACE_FILTER: &str = "amaru=trace,amaru_pure_stage=trace,amaru_protocols=warn,amaru_consensus=info";
+const DEFAULT_AMARU_TRACE_FILTER: &str = "amaru=info,amaru_pure_stage=info,amaru_protocols=warn,amaru_consensus=info";
```
KtorZ:
```suggestion
const DEFAULT_AMARU_TRACE_FILTER: &str = "error,amaru=debug";
```

crates/amaru/src/observability.rs
```diff
@@ -43,11 +43,11 @@ use tracing_subscriber::{
 
 const AMARU_LOG_VAR: &str = "AMARU_LOG";
 
-const DEFAULT_AMARU_LOG_FILTER: &str = "info,amaru::consensus=debug,amaru::ledger=debug,amaru_pure_stage=warn";
+const DEFAULT_AMARU_LOG_FILTER: &str = "info,amaru_pure_stage=warn";
```
KtorZ:
```suggestion
const DEFAULT_AMARU_LOG_FILTER: &str = "error,amaru=info";
```

crates/amaru/src/observability.rs
KtorZ: It should boils down to that in principle. If not, then we have to fix the trace calls.

PR #1081, opened 2026-07-23
Track account unregistrations to avoid O(n) scan on all accounts during epoch transition calculations.

### LINKED ISSUE #1035: Track unregistered accounts to determine rewards
### Abstract

Track accounts registrations / unregistrations across epochs to compute unclaimed rewards

### Why?

Otherwise we need to do a full pass on all the accounts in `Rewards<Effective>::new` to determine which accounts have unclaimed rewards. 

### How?

By keeping an additional data structure tracking the deregistrations.
The difficulty is that the window of deregistrations used for rewards computations is large, so this data must be preserved across volatile -> stable flushes + rollbacks / switch to fork.

One thing we can do it do both methods for getting unclaimed accounts: the current one + the new one then assert that they provide the same results with the end to end tests.

### Discussion points

_No response_

_No response_

### Checklist

KtorZ:
@etorreborre my take on #1038; I figured that instead of keeping something in-memory (which may grow unbounded...) we may as well just put it in the store. It also simplifies the management of it, in particular around epoch transitions, quite a lot IMO. 

## Summary by CodeRabbit

* **Performance**
  * Improved epoch transition reward calculations by focusing on recently unregistered accounts instead of scanning all accounts.
* **Bug Fixes**
  * Fixed effective reward accounting for accounts unregistered during the current epoch.
  * Recently unregistered account tracking is now persisted, pruned at epoch boundaries, and kept in sync with account add/remove.
* **Observability**
  * Added trace events for recently unregistered account record insert, remove, and prune.
  * Added a bootstrap trace event when local snapshot reads fail.
* **Chores**
  * Updated the default database bootstrap cache version when not explicitly set.

### DIFF
--- a/crates/amaru-ledger/src/state/volatile/overlay.rs
+++ b/crates/amaru-ledger/src/state/volatile/overlay.rs
@@ -201,8 +201,8 @@ impl StateOverlay {
                 Span::current().record("should_begin_epoch", should_begin_epoch);
 
                 let updated = if should_begin_epoch {
+                    batch.clear_recently_unregistered_accounts()?;
                     reset_blocks_count(batch)?;
-
                     reset_fees_and_donations(batch)?;
 
                     if let Some(pools_updates) = mem::take(&mut self.pools_updates) {
@@ -343,7 +343,7 @@ impl StateOverlay {
 
 #[cfg(test)]
 mod test {
-    use std::collections::BTreeMap;
+    use std::collections::{BTreeMap, BTreeSet};
 
     use amaru_kernel::{Hash, PREPROD_DEFAULT_PROTOCOL_PARAMETERS, ProposalsRoots};
 
@@ -392,7 +392,7 @@ mod test {
     /// the epoch, so its rewards are unclaimed and returned to the treasury.
     fn effective_rewards() -> Rewards<Effective> {
         let computed = Rewards::<Computed>::new(1_000, 7, BTreeMap::from([(credential(1), 100), (credential(2), 42)]));
-        Rewards::<Effective>::new(computed, std::iter::once(credential(1)))
+        Rewards::<Effective>::new(computed, &BTreeSet::from([credential(2)]))
     }
 
     fn governance_updates() -> GovernanceUpdates {
--- a/crates/amaru-ledger/src/state/volatile/overlay.rs
+++ b/crates/amaru-ledger/src/state/volatile/overlay.rs
@@ -201,7 +201,7 @@ impl StateOverlay {
                 Span::current().record("should_begin_epoch", should_begin_epoch);
 
                 let updated = if should_begin_epoch {
-                    batch.clear_recently_unregistered_accounts()?;
+                    batch.prune_recently_unregistered_accounts(self.epoch)?;
                     reset_blocks_count(batch)?;
                     reset_fees_and_donations(batch)?;
 
@@ -391,8 +391,9 @@ mod test {
     /// Effective rewards where `credential(1)` is still registered while `credential(2)` unregistered during
     /// the epoch, so its rewards are unclaimed and returned to the treasury.
     fn effective_rewards() -> Rewards<Effective> {
-        let computed = Rewards::<Computed>::new(1_000, 7, BTreeMap::from([(credential(1), 100), (credential(2), 42)]));
-        Rewards::<Effective>::new(computed, &BTreeSet::from([credential(2)]))
+        let computed =
+            Rewards::<Computed>::new(1_000, 7, 142, BTreeMap::from([(credential(1), 100), (credential(2), 42)]));
+        Rewards::<Effective>::new(computed, BTreeSet::from([credential(2)]))
     }
 
     fn governance_updates() -> GovernanceUpdates {
--- a/crates/amaru-ledger/src/state/volatile/view/iter_unregistered_accounts.rs
+++ b/crates/amaru-ledger/src/state/volatile/view/iter_unregistered_accounts.rs
@@ -16,18 +16,16 @@ use std::{collections::BTreeSet, mem};
 
 use amaru_kernel::StakeCredential;
 
-use crate::store::columns::accounts::Row as Account;
-
-/// Similar to [`crate::state::volatile::IterPools`], but for accounts; It provides an
-/// unordered iterator over accounts that patches a read-only stable store with pending updates
-/// such as registrations or de-registrations.
-pub(crate) struct IterAccounts<'volatile, DBIter: Iterator<Item = (StakeCredential, Account)>> {
+/// Similar to [`crate::state::volatile::IterPools`], but for accounts; It provides an unordered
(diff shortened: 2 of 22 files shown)


crates/amaru-stores/src/rocksdb/ledger/columns/accounts.rs
```diff
@@ -42,6 +45,9 @@ pub fn add<DB>(db: &Transaction<'_, DB>, rows: impl Iterator<Item = (Key, Value)
                     let mut row = Row { deposit, pool: None, drep: None, rewards };
                     pool.set_or_reset(&mut row.pool);
                     drep.set_or_reset(&mut row.drep);
+
+                    recently_unregistered_accounts::remove(db, &credential)?;
```
KtorZ, line 49: Removed from the set when re-registering the account.

crates/amaru-stores/src/rocksdb/ledger/columns/accounts.rs
```diff
@@ -129,6 +135,7 @@ pub fn set_rewards<DB>(
 pub fn remove<DB>(db: &Transaction<'_, DB>, rows: impl Iterator<Item = Key>) -> Result<(), StoreError> {
     trace_span!(stores::ledger::accounts::REMOVE).in_scope(|| {
         for credential in rows {
+            recently_unregistered_accounts::insert(db, &credential)?;
```
KtorZ: Added to the set when removing the account.

crates/amaru-ledger/src/state/volatile/view/iter_unregistered_accounts.rs
```diff
@@ -37,19 +35,19 @@ impl<'volatile, DBIter: Iterator<Item = (StakeCredential, Account)>> IterAccount
...
                 continue;
             }
 
             return Some(credential);
         }
 
-        if let Some(credential) = self.registrations.pop_first() {
+        if let Some(credential) = self.deregistrations.pop_first() {
             return Some(credential.clone());
         }
```
KtorZ, line 53: Here, we still resolve the bits that are unstable and make sure that an account deregistration or re-registration within the last k blocks is handled properly.

crates/amaru-ledger/src/state/volatile/overlay.rs
```diff
@@ -199,8 +199,8 @@ impl StateOverlay {
                 Span::current().record("should_begin_epoch", should_begin_epoch);
 
                 let updated = if should_begin_epoch {
+                    batch.clear_recently_unregistered_accounts()?;
```
KtorZ: And importantly, we clear the set at the beginning of every epochs. Like we do for fees and blocks counts.

etorreborre:
@KtorZ that is indeed a simplification but I think there are some possible issues:

1. Recent re-registration

Say we are calculating rewards at the 100 -> 101 boundary from the stake distribution at epoch 97. If we clear the set at each epoch boundary, we will clear at the beginning of 100 (the 99 to 100 transition). But if an account deregistered inside epoch 99, we would lose this information and still include it in the rewards. We can possibly fix this by storing an associated epoch to the last deregistration and only prune deregistrations that are too old.

2. A pool account that's not registered gets automatically rewarded. 

My understanding is that pools accounts are not necessarily registered and they need to be in order to get rewards. Since we only track de-registrations we would always consider those accounts as claimable for rewards. We can just query them directly since it's a small number.

3. Similar situation at bootstrap

We could just do a full scan during the first 3 epochs (like my PR was doing) in order to properly populate the unregistered column.

I'm wondering if an option would be to stream them in order from the database in lockstep with the "accounts to reward" (already pruned for volatile deregistrations) for a given epoch, and skip the ones that aren't in the database. 
That would still be a full scan but just accumulating the set of unclaimed accounts in memory. 

I get the impression that this scan anyway is a fraction of the overall computation time for a given epoch. So maybe this last option would be worth it and would remove the complexity of maintaining an additional column with the issues mentioned above.

KtorZ:
> @etorreborre: Say we are calculating rewards at the 100 -> 101 boundary from the stake distribution at epoch 97. If we clear the set at each epoch boundary, we will clear at the beginning of 100 (the 99 to 100 transition). But if an account deregistered inside epoch 99, we would lose this information and still include it in the rewards. We can possibly fix this by storing an associated epoch to the last deregistration and only prune deregistrations that are too old.

This does not impact rewards calculation at all; only rewards _distribution_ which happens on top of an already-calculated set of rewards. The rewards are distributed at the end of an epoch, and the account unregistration cleared at the beginning of the following epoch; so there's no problem here; the information is used right at the moment it needs to be used. 

> @etorreborre: My understanding is that pools accounts are not necessarily registered and they need to be in order to get rewards. Since we only track de-registrations we would always consider those accounts as claimable for rewards. We can just query them directly since it's a small number.

That is correct, and that is already taken into account during rewards calculation; or more exactly, during the stake distribution calculation. Only registered accounts can have stake. So a pool with unregistered reward account will simply have no rewards at all. 

This is independent of this PR. 

> @etorreborre:  Similar situation at bootstrap

Nope; bootstrapping snapshots are always taken *after* rewards have been distributed (after the "snapshot" was taken). So, the first thing we'll do on restart is to clear the recently unregistered account set. In practice, that means there's nothing extra to do for bootstrap.

KtorZ:
> @etorreborre: Say we are calculating rewards at the 100 -> 101 boundary from the stake distribution at epoch 97. If we clear the set at each epoch boundary, we will clear at the beginning of 100 (the 99 to 100 transition). But if an account deregistered inside epoch 99, we would lose this information and still include it in the rewards. We can possibly fix this by storing an associated epoch to the last deregistration and only prune deregistrations that are too old.

Coming back to this; I have indeed overlooked something: the effect of an account un-registering has effect on more than one epoch! 

- in epoch `e`, we compute rewards using the stake distribution at the end of `e-3`. 
- in the transition from `e` to `e+1`, rewards are distributed
- at the beginning of `e+1`, the counters are reset. 

So, an account which de-registers in `e-2`, `e-1` or `e` should be part of the "recently unregistered" set of `e`

KtorZ:
Fucking hell. The bootstrapping was indeed a lot more complexe than I thought due to this recent de-registration impacting multiple epochs. Since the information doesn't exists out of the box in the Haskell snapshots, we have to recover it _somehow_ and it's not particularly easy (at least I got it wrong 4 times 😶 ...). We need information from the previously restored snapshot in order to resolve the next one. I was able to land back on my feet by also leveraging the `mark` stake snapshot present in the Haskell's snapshots (which materialises the accounts that existed in the previous epoch, which we can compare to the current state to figure out which one have unregistered). 

Preview and Preprod snapshots are now agreeing and I've synced over a hundred epochs on Preview without issues. 

I've also strengthen the rewards assertion; instead of looking at a number of accounts, we know look at the actual rewards amount (expected at epoch boundary vs actually paid out k blocks later). Had to do this because (1) it was actually easier to do that than to count accounts and (2) counting led to the guard passing when the state was actually wrong but just happened to align on the number of accounts. Happened once.

PR #1087, opened 2026-07-24
perf: skip the validation of already-stored headers

etorreborre:
This PR skips the verification of a header when it has already been stored in the `ChainStore`.
This situation can happen when there are multiple upstream peers sending us the same headers.

## Summary by CodeRabbit

* **New Features**
  * Enhanced nonce-aware header validation so duplicate headers are handled more efficiently when evolved nonces are already stored.
  * Improved tip propagation by storing a newly validated header before forwarding downstream.
* **Bug Fixes**
  * Fixed scenarios where headers missing evolved nonces were not correctly revalidated and propagated after import/bootstrap.
  * Prevented duplicate-header reprocessing from triggering unnecessary storage or forwarding.
* **Tests**
  * Added nonce evolution and regression coverage, plus updated validation/tracing expectations for deferred and peer-related flows.

### DIFF
--- a/crates/amaru-consensus/src/stages/track_peers/mod.rs
+++ b/crates/amaru-consensus/src/stages/track_peers/mod.rs
@@ -25,6 +25,7 @@ use amaru_kernel::{
 use amaru_metrics::consensus::ConsensusMetrics;
 use amaru_observability::{TraceContext, debug, debug_record, debug_span, error};
 use amaru_ouroboros::ConnectionId;
+use amaru_ouroboros_traits::Nonces;
 use amaru_protocols::{
     chainsync::{self, ChainSyncInitiatorMsg, HeaderContent},
     metrics_effects::{Metrics, MetricsOps},
@@ -384,12 +385,13 @@ impl TrackPeers {
     ///
     /// The received `tip` is the highest advertised tip for the peer as part of the RollForward message.
     ///
-    /// If the store already holds evolved nonces for this header (this is the durable proof that it
-    /// went through full validation once before) the header is skipped and `None` is returned.
-    /// Otherwise the header is validated and the point of its parent is returned.
+    /// If the store already holds evolved nonces for this header, it went through full validation
+    /// before (nonces are only stored together with a validated header), so the header is skipped
+    /// and `None` is returned. Otherwise the header is validated and the point of its parent is
+    /// returned, together with the nonces to store alongside it.
     ///
-    /// Note: a header could already sits in the chain store, as is the case for headers imported
-    /// during bootstrap, but not carry a nonce yet.
+    /// Note: a header can already sit in the chain store without carrying any nonces, as is the
+    /// case for headers imported during bootstrap. Those still need to be validated.
     #[expect(clippy::too_many_arguments)]
     async fn validate_header(
         &mut self,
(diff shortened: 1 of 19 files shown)


crates/amaru-consensus/src/stages/track_peers/mod.rs
```diff
@@ -382,9 +382,12 @@ impl TrackPeers {
         });
     }
 
-    /// Validate an incoming header for protocol conformance and store it in the chain store.
+    /// Validate an incoming header for protocol conformance.
     ///
     /// The received `tip` is the highest advertised tip for the peer as part of the RollForward message.
+    ///
+    /// Returns the parent point of a newly validated header, or `None` when the header is already
+    /// present in the chain store.
```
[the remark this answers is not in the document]
etorreborre: Fixed

rustc-ice-2026-07-24T23_13_19-6152.txt
KtorZ, line 1: Probably not meant to be committed 😅 ?

KtorZ:
Will that not conflict with the op certs? A header could be valid for a given chain but not for another just because the opcerts don't align?

rustc-ice-2026-07-24T23_13_19-6152.txt
etorreborre: Thanks! Removed.

etorreborre:
> Will that not conflict with the op certs? A header could be valid for a given chain but not for another just because the opcerts don't align?

@KtorZ that shouldn't be a problem if you have a look at #1021. We validate headers based on their ancestors, whether they are on the best chain or not.

crates/amaru-consensus/src/stages/track_peers/mod.rs
```diff
@@ -448,8 +456,14 @@ impl TrackPeers {
             return Err(ConsensusError::HeaderSlotInNearFuture(header.slot()));
         }
 
+        // The evolved nonces are the durable proof that a header was fully validated.
```
rkuhn: No, this is not the case: `validate_header` first evolves the nonces, then evaluates the Praos rules. Currently, the header is only stored if it was valid, we never store invalid ones, so presence of the header is the correct proof to use.

crates/amaru-consensus/src/stages/track_peers/tests.rs
```diff
@@ -367,6 +366,53 @@ fn test_roll_forward_known_peer_header_already_stored() {
     ]);
 }
 
+/// A header imported during bootstrap is present in the chain store but has no evolved nonces.
+/// When re-received from a peer, its nonces must still be computed (via `validate_header`) so
+/// that descendant headers can be validated. Nonce absence means the header was never fully
+/// validated, so it is treated like a new header: stored (idempotent) and propagated downstream.
```
rkuhn: Okay, this sounds reasonable as well, so even when skipping header validation (because the header is already stored) we need to check for the nonces. Perhaps the evolve_nonces effect should just be idempotent in the most efficient way.

crates/amaru-consensus/src/stages/track_peers/mod.rs
etorreborre: @rkuhn Indeed I missed that. In order to fix the issue, and improve the consistency, I added a new commit where:

1. Evolving the nonces is separated from storing them.
2. We store a validated header and its nonces atomically. So the presence of nonces really means that the header was validated.

crates/amaru-consensus/src/stages/track_peers/tests.rs
etorreborre: Please see my new commit.

rkuhn (approved):
yes, looking good!

PR #1088, opened 2026-07-24
Volatile benchmarks + related fixes

KtorZ:
```
configuration
├─ volatile_size=2160 (AMARU_BENCH_VOLATILE_SIZE=)
╰─ block_size=90112 (AMARU_BENCH_BLOCK_SIZE=)

volatile_db                               median  │ mean    
├─ bench_hydrate_context                          │         
│  ├─ accounts [total_items=2994491]     163 ms   │ 165.2 ms
│  ├─ committee [total_items=4325376]    113 ms   │ 114.3 ms
│  ├─ dreps [total_items=1946419]        76.62 ms │ 76.5 ms 
│  ├─ mixed [total_items=629909]         32.32 ms │ 32.76 ms
│  ├─ pools [total_items=1946419]        96.91 ms │ 96.98 ms
│  ├─ proposals [total_items=97320]      4.252 ms │ 4.411 ms
│  ├─ utxo [total_items=3538944]         171.1 ms │ 172.8 ms
│  ├─ votes [total_items=2780598]        188.3 ms │ 189.3 ms
│  ╰─ withdrawals [total_items=4866048]  113.5 ms │ 113.6 ms
├─ bench_roll_forward                             │         
│  ├─ accounts [total_items=2994491]     128.9 ms │ 130.4 ms
│  ├─ committee [total_items=4325376]    125.1 ms │ 125.6 ms
│  ├─ dreps [total_items=1946419]        61.98 ms │ 62.46 ms
│  ├─ mixed [total_items=629909]         31.34 ms │ 31.86 ms
│  ├─ pools [total_items=1946419]        94.55 ms │ 95.74 ms
│  ├─ proposals [total_items=97320]      5.173 ms │ 5.345 ms
│  ├─ utxo [total_items=3538944]         171.9 ms │ 181.7 ms
│  ├─ votes [total_items=2780598]        187.3 ms │ 187 ms  
│  ╰─ withdrawals [total_items=4866048]  106.6 ms │ 106.5 ms
╰─ bench_switch_to_fork                           │         
   ├─ accounts [total_items=2994491]     2.603 s  │ 2.636 s 
   ├─ committee [total_items=4325376]    2.447 s  │ 2.455 s 
   ├─ dreps [total_items=1946419]        1.39 s   │ 1.421 s 
   ├─ mixed [total_items=629909]         244.9 ms │ 248 ms  
   ├─ pools [total_items=1946419]        1.11 s   │ 1.11 s  
   ├─ proposals [total_items=97320]      22.6 ms  │ 22.9 ms 
   ├─ utxo [total_items=3538944]         2.117 s  │ 2.132 s 
   ├─ votes [total_items=2780598]        1.646 s  │ 1.67 s  
   ╰─ withdrawals [total_items=4866048]  1.911 s  │ 1.99 s  

memory usage (volatile footprint)
├─ accounts=56MB
├─ committee=561MB
├─ dreps=104MB
├─ mixed=121MB
├─ pools=392MB
├─ proposals=31MB
├─ utxo=1009MB
├─ votes=762MB
╰─ withdrawals=429MB
```

## Summary by CodeRabbit

* **New Features**
  * Added volatile database benchmarks (roll-forward, switch-to-fork, context hydration) with scenario configuration and per-scenario memory reporting, plus CI execution coverage.
  * Added a state rollback integration test.
  * Added updated proposal-roots state handling.

* **Improvements**
  * Streamlined volatile governance/DRep/account hydration and increased the forced volatile recomputation threshold.
  * Improved rewards handling to favor cheaper cloning.

* **Testing**
  * Expanded reusable test generators and simplified the rollback mock store.

* **Breaking Changes**
  * Removed rewards snapshot helper APIs.
  * Removed/changed account-preparation and committee term-related interfaces/exports.


crates/amaru-ledger/tests/state_rollback.rs
```diff
@@ -227,158 +227,10 @@ fn point(slot: u64, tag: u8) -> Point {
 
 struct MockStore(RocksDB);
 
-#[expect(unused_variables)]
 impl ReadStore for MockStore {
     fn tip(&self) -> amaru_ledger::store::Result<Point> {
         Ok(Point::Origin)
     }
-
-    fn epoch_transition_progress(&self) -> amaru_ledger::store::Result<Option<EpochTransitionProgress>> {
```
KtorZ, line 236: This is what the changes in the ledger/store are for. We now have a default dummy implementation of the store in tests; so we can define mocks much more easily.

crates/amaru-ledger/src/state/volatile/fragment.rs
```diff
@@ -156,7 +156,7 @@ impl VolatileFragment {
                 value: None,
             })
         } else if self.committee.consumed.contains(credential) {
-            Existence::Gone
```
KtorZ, line 159: This was wrong; we've been mixing up CC members being removed from them simply dropping their hot credential delegation.

crates/amaru-ledger/src/context/default/preparation.rs
```diff
@@ -44,8 +46,7 @@ use crate::{
 pub struct DefaultPreparationContext<'a> {
     pub utxo: BTreeSet<&'a TransactionInput>,
     pub pools: BTreeSet<&'a PoolId>,
-    pub accounts: BTreeSet<&'a StakeCredential>,
-    pub withdrawals: BTreeSet<&'a RewardAccount>,
+    pub accounts: BTreeSet<Cow<'a, StakeCredential>>,
```
KtorZ, line 49: Withdrawals and accounts have the same effects on the preparation context here; We have to resort to a `Cow` here however because withdrawals are still present as `RewardAccounts` in transactions, so we have to parse them and therefore own them. 

This should be simplified once we represent them as proper stake credentials.

crates/amaru-ledger/benches/state/volatile_db/benches/hydrate_context.rs
```diff
@@ -0,0 +1,64 @@
...
+                let ix = rng.random_range(0..self.scale.volatile_size);
+                let mut ctx = DefaultPreparationContext::new();
+                let fragment = &db.iter().nth(ix).expect("db must have fragment").fragment;
+                self.scenario.prepare_fragment(&mut ctx, fragment);
+                (db.clone(), ctx, roots.clone())
+            })
+            .bench_local_values(|(db, ctx, roots)| {
+                assert!(
+                    ctx.into_validation_context(UnresolvedInputPolicy::Defer, roots, &db, &self.scenario.mock_store())
+                        .is_ok()
```
KtorZ, line 61: In the "HydrateContext" benchmark, we create a validation context from a "fat" block (materialized by a fragment here) where all the data to resolve are in the volatile. This allows to exercise the resolution logic without having to re-implement much of it. We can simply provide a mock store that can respond in case the preparation context goes down to the stable store (happens in some cases still).

crates/amaru-ledger/benches/state/volatile_db/benches/roll_forward.rs
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    pub fn run(self, bencher: Bencher<'_, '_>) -> u64 {
+        let mut rng = StdRng::seed_from_u64(self.scenario.seed());
+        let (db, rss) = memory::rss_delta(|| self.scenario.new_volatile_db(&mut rng, &self.scale));
+        let fragment = self.scenario.new_fragment(&mut rng, &self.scale);
+
+        bencher.with_inputs(|| (db.clone(), fragment.clone())).bench_local_values(|(mut db, next)| {
+            let _previous = db.pop_front();
+            db.push_back(next);
```
KtorZ, line 51: The RollForward bench exercises an actual roll forward using the volatile public API. Note that the volatile is first pre-filled with A LOT of data, so the `pop_front` induces an incremental cleanup, and the `push_back` will also be added to the aggregate.

crates/amaru-ledger/benches/state/volatile_db/benches/switch_to_fork.rs
```diff
@@ -0,0 +1,57 @@
...
+        let db = self.scenario.new_volatile_db(&mut rng, &self.scale);
+        let point = db.iter().nth(self.scale.volatile_size / 2).expect("volatile MUST have fragment").point();
+        bencher
+            .with_inputs(|| {
+                let fragments = db.iter().skip(self.scale.volatile_size / 2).cloned().collect::<Vec<_>>();
+                (db.clone(), point.clone(), fragments)
+            })
+            .bench_local_values(|(mut db, point, fragments)| {
+                assert!(db.rollback_to(&point).is_ok());
+                fragments.into_iter().for_each(|fragment| db.push_back(fragment));
```
KtorZ, line 62: Finally, the `switch_to_fork` is doing a long fork of `k / 2` blocks and then immediately re-apply `k / 2` fragments. Why `k / 2`? So that we still have an aggregate to compute from the existing volatile that is non-trivial. We could probably tweak this to see what's actually worse (small rollback with big aggregate or large rollback with small aggregate). 

Maybe the `/ 2` could be a bench parameter eventually, because that doesn't fundamentally changes the bench setup and execution.

crates/amaru-ledger/benches/state/volatile_db/common/scenario.rs
```diff
@@ -0,0 +1,405 @@
...
+            // the overhead.
+            Self::Votes => 70,
+
+            Self::Mixed => {
+                let all = Self::round_robin();
+                let len = all.len();
+                all.iter().map(|entity| entity.per_item_size()).sum::<usize>() / len
+            }
+        }
+    }
```
KtorZ, line 195: ☝️ All those numbers are debatable; the point is simply to trying to measure a somewhat "realistic" worse load. For example, there's no point to benchmarks for millions of proposals because this cannot physically happen anyway. The numbers are still chosen to be quite pessimistic (e.g. we end up with 3M+ accounts which is more than mainnet has ever seen in its lifetime...).

crates/amaru-ledger/src/rules.rs
```diff
@@ -92,15 +92,17 @@ fn prepare_votes<'a>(context: &mut impl PreparationContext<'a>, transaction: &'a
 
 fn prepare_governance_action<'a>(context: &mut impl PreparationContext<'a>, proposal: &'a Proposal) {
     if let GovernanceAction::TreasuryWithdrawals(withdrawals, _) = &proposal.gov_action {
-        withdrawals.iter().for_each(|withdrawal| context.require_withdrawal(&withdrawal.0));
+        withdrawals
+            .iter()
+            .for_each(|withdrawal| context.require_account(Cow::Owned(expect_stake_credential(&withdrawal.0))));
```
yHSJ: Is there any gate on the format of these bytes before this point? I don't believe so, and that opens a simple attack vector to crash Amaru nodes on the network

crates/amaru-ledger/src/context/default/preparation.rs
```diff
@@ -400,50 +386,39 @@ fn resolve_committee<'a>(
...
-                    cc_members.insert(credential.clone(), member);
-
-                    Ok(cc_members)
+                    None
                 }
 
-                (Existence::Exists(Bind { value: None, left, .. }), valid_until) => {
-                    if let Some(row) = db.cc_member(credential).map_err(ContextHydratationError::ResolveCommittee)? {
+                (Existence::Exists(Bind { left, .. }), valid_until) => {
+                    db.cc_member(credential).map_err(ContextHydratationError::ResolveCommittee)?.map(|mut row| {
```
yHSJ: What about when a value is in the state overlay and not yet in the DB? This would either error or disappear from reads, right?

crates/amaru-ledger/benches/state/volatile_db/benches/switch_to_fork.rs
```diff
@@ -0,0 +1,57 @@
...
+            })
+    }
```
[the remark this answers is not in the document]
yHSJ: @KtorZ  I think coderabbit is right here. Not that it's that urgent, we're off by less than 1% in regards to timing. The size reports will also not be accurate to a real world scenario (due to the duplicate fragment). Not tragic by any means, but we might as well get it right in a bench.

crates/amaru-ledger/benches/state/volatile_db/benches/hydrate_context.rs
```diff
@@ -0,0 +1,65 @@
...
+        let mut rng = StdRng::seed_from_u64(self.scenario.seed());
+        let db = self.scenario.new_volatile_db(&mut rng, &self.scale);
+        let roots = fixture::proposals_roots(&mut rng);
+        bencher
+            .with_inputs(|| {
+                let ix = rng.random_range(0..self.scale.volatile_size);
+                let mut ctx = DefaultPreparationContext::new();
+                let fragment = &db.iter().nth(ix).expect("db must have fragment").fragment;
+                self.scenario.prepare_fragment(&mut ctx, fragment);
+                (db.clone(), ctx, roots.clone())
```
yHSJ, line 56: why do we have to clone the DB here? Does `bench_local_values` require it to be owned? It's in a benchmark (so not critical), but we can probably just pass a reference to it (since `into_validation_context` doesn't need it to be owned).

crates/amaru-ledger/benches/state/volatile_db/common/scenario.rs
yHSJ: It's also worth noting that our numbers are *slightly* above reality due to the practically of creating such transactions. For example, one cannot fill a block with just proposals, there must also be at least one input, a witness, and some bytes that are CBOR overhead. These benchmarks are good enough, especially since we will be overestimating worst case scenarios, but I don't want anyone to cite these numbers as exact measurements of reality.

crates/amaru-ledger/src/state/volatile/series.rs
```diff
@@ -52,9 +52,10 @@ use crate::state::{
 /// it is highly likely that *at least one* would lead to an observable rollback. Finally, when
 /// syncing, the impact should be negligible as only one block every 1080 would cause an aggregate
 /// recompute.
```
yHSJ, line 54: This comment is now out of sync with the value below. Not that urgent, since we are removing this in the withdrawal checks PR anyway, but worth checking on other doc comments that assert things about the code that have changed (especially around the hydration functions that resolve state since there was a change to CC members).

crates/amaru-ledger/benches/state/volatile_db/benches/roll_forward.rs
```diff
@@ -0,0 +1,56 @@
...
+        let scale = BenchScale::from_env();
+        Self { scenario, scale }
+    }
+
+    pub fn run(self, bencher: Bencher<'_, '_>) -> u64 {
+        let mut rng = StdRng::seed_from_u64(self.scenario.seed());
+        let (db, rss) = memory::rss_delta(|| self.scenario.new_volatile_db(&mut rng, &self.scale));
+        let fragment = self.scenario.new_fragment(&mut rng, &self.scale);
+
+        bencher.with_inputs(|| (db.clone(), fragment.clone())).bench_local_values(|(mut db, next)| {
```
yHSJ, line 49: Same thing with this db.clone, do we need it?

yHSJ (commented):
Left a few comments, most are very nit-picky, but the one I'm concerned about is the unreachable that I think is actually very reachable.

crates/amaru-ledger/benches/state/volatile_db/common/scenario.rs
yHSJ: I was looking at our generators, and they don't match up with these numbers. For example, the utxo says we have an address with no delegation and min ada, but our generator produces an address with delegation *and* the value is a random u64 instead of the min value. This is super nit-picky, but it could lead us to potentially incorrect conclusions from benchmarks.

I wonder if it's better to construct an in-memory object for each case that we feel meets the "average" on-chain, then serialize it with `to_cbor` and measure that size. It doesn't address my above nit pick, but it gives us a accurate value that is easier for readers to understand. It's also a bit more resilient to encoding changes, since if the type changes we'd get compilation errors.

crates/amaru-ledger/src/rules.rs
KtorZ: This is correct and I should've left a note. These should be removed _soon_ as proper stake credential in our base model. So that's one more on the list of "Pallas rewrite". 

I hope to get a good stab at this one of this weekend.

crates/amaru-ledger/src/context/default/preparation.rs
KtorZ: The overlay is already taken into account when we resolve from the volatile, so all good here.

crates/amaru-ledger/benches/state/volatile_db/benches/switch_to_fork.rs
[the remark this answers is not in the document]
KtorZ: Ho it really doesn't matter, it's one droplet amongst the size of the benchmark. I will fix it just for the sake of keeping the logic clear but it should have no visible effect on the results.

crates/amaru-ledger/benches/state/volatile_db/benches/hydrate_context.rs
KtorZ: It's faster. We are putting millions of elements in the volatile which are randomly generated. If we don't clone, we have to regenerate those same elements (using a deterministic seed anyway!) on each bench iteration. This makes the benchmarks considerably slower. 

But I also did not want to just allow cloning the volatile, which is why I've gated it behind a test-utils config attr.

crates/amaru-ledger/benches/state/volatile_db/common/scenario.rs
KtorZ: So actually, that's on purpose. 

The estimated sizes are only there to constraint how many elements we want to generate. And so they take a lower bound of size which disregards cbor overhead and extra glue in transactions. This slightly inflates the number of elements compared to what's actually feasible in practice. 

On the other hand, we generate values that aren't particularly constrained and will tend to be actually much larger than the estimate. Yet we will still fit the same number. So the consequence of that is our measurement of the memory is also inflated compared to reality (because we measure _actual resident memory_, not an estimate). 

These are just "safety margins". We are making mistakes, but mistakes that go in the safe direction. 

For example, the worse case for the memory seems to be the UTxO, which can add up to 1GB; just for the volatile. That means, whatever number we communicate as Amaru's memory requirement, we have to consider at least 1GB above that to be safe. And we know we need likely not much more because we already made approximations in the right direction in those benchmarks. Of course, we could theoretically come up with a precise number but rooms for mistake are important.. so I prefer measuring with some safety margins and draw conclusions from the most pessimistic case.

crates/amaru-ledger/benches/state/volatile_db/benches/switch_to_fork.rs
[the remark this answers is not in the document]
KtorZ: Addressed; and made rollback point configurable.

crates/amaru-ledger/src/rules.rs
KtorZ: I've put back the `parse_reward_account` at this stage for now; ignoring parse errors as before. So an invalid withdrawal will show up later as missing from the context. It's a bit confusing, but at least it'll not crash needlessly.

crates/amaru-ledger/src/context/default/preparation.rs
KtorZ: I see what you meant.

PR #1101, opened 2026-07-28
fix: leader unclaimed rewards

etorreborre:
@KtorZ #1081 introduced an issue with leader accounts and unclaimed rewards.
A reward account might not be in the list of owner accounts and end-up getting some rewards even if it is not registered.

This PR fetches leader accounts from the database to make sure we scan all of them to compute unclaimed rewards (but this is not a large set).

I'm still wondering if a merge-join check for registered accounts would not have been the easiest implementation and just a few milliseconds compared to the full epoch transition computation.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed leader rewards handling when the pool reward credential has no registered/unreachable account, ensuring correct unclaimed reward resolution at epoch boundaries.
  * Improved epoch-transition reachability so only eligible pool reward accounts are used to compute unclaimed rewards, even when volatile overrides apply.
* **Tests**
  * Added coverage to ensure repeated or rewardless credentials are not retained for unclaimed calculations.
  * Expanded tests for credited leader rewards when the reward credential has no registered account and for volatile vs stable registration/unregistration interactions.


crates/amaru-ledger/src/epoch_transition/rewards_state.rs
```diff
@@ -153,6 +153,15 @@ pub struct Rewards<STEP: KnownRewardState + Clone> {
...
     accounts: Arc<BTreeMap<StakeCredential, Lovelace>>,
 
+    /// The subset of `accounts` credited with a leader reward, i.e. the reward accounts of pools
+    /// that earned one.
+    ///
+    /// This used to compute the unclaimed rewards. Leader accounts might not be registered and
+    /// the registration/deregistration tracking we do for regular accounts does not inform us of their registration
+    /// status. This means that, for these accounts specifically we need to look up in the stable db if they are currently registered
+    /// and can claim a reward.
+    leader_reward_accounts: Arc<BTreeSet<StakeCredential>>,
```
KtorZ: Still not fond of this fellow here 😅 ...

Can we perhaps get rid of it entirely and instead, pass it from the PoolsEpochUpdates ? We have to iterate through all the pools at the epoch boundary _anyway_, so we might as well retain their reward account at that time, and feed it when constructing the rewards summary? This way, we don't need to store it anywhere and to modify the summary / rewards model. Things remain "local" to the transition.

etorreborre: @KtorZ I investigated that but I think that using PoolUpdates would be incorrect since they iterate on pools for the current epoch transition while we want to get the pool data from e - 3 (the one used to compute rewards). In the meantime a pool might have changed its reward account. Do you agree?

etorreborre: Putting this PR back to draft because I'm attempting an improvement.

KtorZ: >  In the meantime a pool might have changed its reward account.

That is precisely why we want the current pools I believe. It does not matter that the account existed when the rewards were calculated; it matters it exists when they are distributed.

crates/amaru-ledger/src/state/volatile/view.rs
```diff
@@ -121,23 +120,59 @@ impl<'volatile, 'db, DB: ReadStore> VolatileView<'volatile, 'db, DB> {
...
+            }
+            Some(accounts) => accounts,
+        };
+
+        // Recently unregistered, unless they have since re-registered within the volatile window.
+        let mut unclaimed = self
+            .db
+            .iter_recently_unregistered_accounts()?
+            .filter(|credential| !registered.contains(credential))
+            .collect::<BTreeSet<StakeCredential>>();
```
KtorZ: We should avoid `.collect` here; these can get arbitrarily large. This is why we currently use an iterator!

etorreborre: Good point. Fixed.

CHANGELOG.md
```diff
@@ -60,6 +60,7 @@ Other guiding principles:
 - **amaru-node**: support Cardano ledger peer snapshots via `--peer-snapshot` / `AMARU_PEER_SNAPSHOT` for cold-start big-ledger peers in peer selection (complements `--peer-address`) ([#1047](https://github.com/pragma-org/amaru/pull/1047))
 - **amaru-node**: embed best-effort peer snapshots for known networks (for example mainnet, preprod, preview) at build time from cardano-foundation/cardano-configurations; used by default when `--peer-snapshot` is omitted
 - **amaru**: fix the start/restart of a node ([#1095][], [#1098][])
+- **amaru-ledger**: fix the handling of leader accounts for unclaimed rewards ([#1095][])
```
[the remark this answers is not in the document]
etorreborre: Done.

etorreborre:
@KtorZ I updated the PR and it successfully passes on mainnet at epoch `603` where it was previously failing.

etorreborre:
@KtorZ looks good!

PR #1109, opened 2026-07-30
fix: fix pool updates

etorreborre:
This PR fixes an issue with pool updates found on `mainnet` at epoch `615`.
The [logic used to update the pool params](https://github.com/pragma-org/amaru/compare/main...etorreborre/fix/fix-pool-updates#diff-edf82145bb92346aff32133985c7661766e0aec6e181a439859ae6409624b2ffL154) was not the same than the [logic used to fold them](https://github.com/pragma-org/amaru/compare/main...etorreborre/fix/fix-pool-updates#diff-edf82145bb92346aff32133985c7661766e0aec6e181a439859ae6409624b2ffL79):

 - The first commit provides a fix.
 - The second commit refactors the code into `PoolCertificates` and `PendingPoolCertificates` in order to provide a more meaningful API than a direct `Vec<(Option<PoolParams>, Epoch)>`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed stake pool update processing across epoch transitions.
  * Correctly applies pool registrations, metadata changes, and retirements.
  * Re-registration now cancels a later scheduled retirement when applicable.
  * Improved persistence and recovery of pending pool updates, including legacy stored data.
  * Fixed initial stake pool imports to avoid incorrect epoch-based certificate information.

* **Documentation**
  * Added an unreleased changelog entry for this fix.


etorreborre:
@coderabbit review

crates/amaru-ledger/src/epoch_transition/pools_updates/pool_certificates.rs
KtorZ, line 1: I don't think this belongs in the kernel as it's likely ledger-specific; but whatever!

crates/amaru-ledger/src/epoch_transition/pools_updates/pool_certificates.rs
```diff
@@ -0,0 +1,256 @@
...
+    pub fn is_empty(&self) -> bool {
+        self.0.is_empty()
+    }
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub enum PoolCertificate {
+    Registration(PoolParams, Epoch),
+    Retirement(Epoch),
+}
```
KtorZ, line 134: Yes, this type ought to have existed a long time ago.. not sure what was the obsession with tuples of options back then.

KtorZ: We should probably remove the `Epoch` parameter from the `Registration` as it creates room for mistake. A re-registration certificate is necessarily applied at an epoch boundary. The epoch is fully irrelevant.

KtorZ (approved):

PR #1110, opened 2026-07-30
feat: don't display amaru tags in the compact console output

jeluard:
Hide tags (e.g. perf.cpu) from the console output. Tags are only used at the open telemetry level to filter spans.

Code courtesy of @etorreborre 

no-changelog

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved console log readability by hiding internal `amaru.tag.*` fields.
  * Preserved other useful span details, such as transaction identifiers, in console output.


etorreborre (approved):
🏎️

PR #1115, opened 2026-07-31
improve peer snapshot cache invalidation

### LINKED ISSUE #1114: small fix: peer snapshots should be retrieved when base commit timestamp changes
With #1108 the github API request is rate limited to less than 1 per 12h, which was triggered by the observation that each LSP in a worktree will run this logic upon every check. That quick fix makes it easy to forget cleaning the cache when switching to a different branch or older commit, leading to peer snapshots that don’t match that commit’s time.

The commit metadata cache should track the `until` timestamp that is used in the github API request so that the cache can be invalidated when that timestamp changes too much (exact equality is not required because the data change infrequently and the main worry is when checking out a much older or newer commit).

rkuhn:
cache is not used when base commit timestamp changes by more than 1h

fixes #1114

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved peer snapshot caching when switching between commits.
  - Cached snapshots are now reused only when both cache age and commit timestamp compatibility checks pass.
  - Updated cached metadata correctly during conditional refreshes and successful cache responses.
  - Prevented stale snapshots from being reused when commit timestamps are incompatible.

- **Documentation**
  - Clarified peer snapshot cache reuse and refresh conditions.
  - Added a changelog entry describing cache invalidation improvements.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,10 @@ Other guiding principles:
 
 - **amaru-kernel**: dependency on `pallas-*`
 
+### Fixed
+
+- **amaru-node**: invalidate peer snapshot commit metadata cache when switching to older or newer commits. ([#1114](https://github.com/pragma-org/amaru/issues/1114))
+
 ## [v10.11.20260730](https://github.com/pragma-org/amaru/releases/tag/v10.11.20260730)
 
 ### Added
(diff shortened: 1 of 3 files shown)


rkuhn:
@KtorZ This fixes an issue with the 12h no-refresh rule added yesterday to avoid github API throttling that I had decided to ignore, which was pointed out by the rabbit as well, and which after a good night’s sleep looks fix-worthy. Merging now, please let me know if you see any more trouble with peer snapshot downloading.

PR #1117, opened 2026-07-31
ensure bounded schedule usage in peer_selection

### LINKED ISSUE #1112: Too many header validation failures trigger a pure-stage panic
### What revision are you using?

2f8cc386a2b12c03a047a84f1371238a894a6335

### What operating system are you using?

MacOS

### What architecture are you using?

aarch64/armv8

### What problem are you having? What should be the expected behaviour?

Here is an example of running `amaru` on `mainnet`:
```
2026-07-30T17:50:23.130941Z ERROR amaru::consensus: perf.header.lifecycle peer=46.4.53.238:6000 header_hash=6d18159dd21f1783ea259aec266b9536cc854af5fdd0b016375c8a00f68dd727 error=Failed to validate header at 175787097.6d18159dd21f1783ea259aec266b9536cc854af5fdd0b016375c8a00f68dd727: header validation failed: Operational certificate sequence number (23) is too far ahead of the latest known sequence number (21). outcome=invalid header
2026-07-30T17:50:23.130973Z  WARN amaru_consensus::stages::peer_selection: removing peer (outbound) peer=46.4.53.238:6000 peer_state=Connected(Connection { id: ConnectionId(13), full_duplex_capable: false, full_duplex: false }) is_static=false
2026-07-30T17:50:23.131047Z  INFO amaru_consensus::stages::peer_selection: peer_selection.add_peer peer=relay1.thevikingpool.com:6000 was_banned=false
2026-07-30T17:50:23.131069Z  INFO amaru_protocols::manager: adding peer peer=relay1.thevikingpool.com:6000
2026-07-30T17:50:23.131064Z ERROR amaru::consensus: perf.header.lifecycle peer=46.4.53.238:6000 header_hash=d6684be5f80ca78af00f91fd41d455348cabe8cf98d4550ef559c91643dea03e error=Unknown peer 46.4.53.238:6000, bailing out outcome=invalid header
2026-07-30T17:50:23.131087Z  INFO amaru_protocols::manager: disconnecting outbound connection peer=46.4.53.238:6000 conn_id=13
2026-07-30T17:50:23.131107Z  WARN amaru_pure_stage::tokio: stage `13-46.4.53.238:6000-259` terminated
2026-07-30T17:50:23.131111Z  INFO amaru_pure_stage::tokio: terminated stage_name=13-46.4.53.238:6000-259
2026-07-30T17:50:23.131124Z  INFO amaru_consensus::stages::track_peers: chainsync terminated, purging connection state peer=46.4.53.238:6000 conn_id=13

   amaru::fatal::error

   This is almost certainly a bug, and we'd appreciate a report so we can improve Amaru.

   Please report this error at https://github.com/pragma-org/amaru/issues/new.

   In your bug report please provide the information below and if possible the code
   that produced it.

   Operating System: macos
   Architecture:     aarch64
   Version:          v10.11.0+92fe9a4


thread 'amaru-node' (627623125) panicked at /Users/etorreborre/projects/amaru/amaru-1/crates/amaru-pure-stage/src/tokio.rs:549:25:
stage `peer_selection-2` exceeded priority mailbox size (10): too many outstanding scheduled messages
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

This is preceded by lots of header validation failures and the relevant error message at the end is:
```
stage `peer_selection-2` exceeded priority mailbox size (10): too many outstanding scheduled messages
```
Which means, I guess, that the `peer_selection` should do some clean-up when peers are removed. 

### If possible, provide reproductions steps.

`AMARU_NETWORK=mainnet amaru node bootstrap` and `AMARU_NETWORK=mainnet amaru node run` should be enough to reproduce the issue (until the opcert tracking problem is fixed)

### If any, provide logs / error outputs from the system when the problem occurs.

_No response_

### Please confirm that you've already searched issues, CHANGELOG and available documentation.

- [x] I searched, but no luck.

rkuhn:
fixes #1112

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed consensus peer selection to use a single shared cooldown timer, improving reliability and resource usage.
  * Prevented priority mailbox growth from becoming unbounded.
  * Improved handling of multiple temporarily unavailable peers, including timer reuse, rescheduling, and cancellation.
  * Ensured peers remain excluded during active cooldowns and become eligible again after cooldown expiration.
  * Improved handling of repeated bans and overlapping cooldown periods.

### DIFF
--- a/crates/amaru-consensus/src/stages/peer_selection/mod.rs
+++ b/crates/amaru-consensus/src/stages/peer_selection/mod.rs
@@ -13,7 +13,8 @@
 // limitations under the License.
 
 use std::{
-    collections::{BTreeMap, BTreeSet, btree_map::Entry},
+    cmp::Reverse,
+    collections::{BTreeMap, BTreeSet, BinaryHeap, btree_map::Entry},
     time::Duration,
 };
 
@@ -21,7 +22,7 @@ use amaru_kernel::{BlockHeight, Peer};
 use amaru_observability::{TraceContext, debug_span};
 use amaru_ouroboros::{ConnectionDirection, ConnectionId};
 use amaru_protocols::manager::ManagerMessage;
-use amaru_pure_stage::{Effects, ScheduleId, StageRef};
+use amaru_pure_stage::{Effects, Instant, ScheduleId, StageRef};
 use rand::{SeedableRng, rngs::StdRng, seq::IteratorRandom};
 use tracing::Instrument;
 
@@ -43,7 +44,7 @@ const STATIC_PEER_BAN_PERIOD: Duration = Duration::from_secs(10);
 ///
 /// Outbound regulation uses a random refill (via `GenerateRandomSeed` external effect)
 /// preferring static peers, then snapshot candidates, then ledger candidates, while
-/// skipping any peer currently tracked in `outbound_peers` or `cooldown_timers`.
+/// skipping any peer currently tracked in `outbound_peers` or `cooldown_until`.
 /// Inbound connections are accepted up to `target_downstream_peers` (excess are
 /// immediately rejected with a `Disconnect`).
 ///
@@ -58,12 +59,15 @@ const STATIC_PEER_BAN_PERIOD: Duration = Duration::from_secs(10);
 /// - `manager`: `StageRef<ManagerMessage>` for all outbound commands.
 /// - `static_peers`, `snapshot_candidates`, `ledger_candidates`: candidate pools (`BTreeSet<Peer>`).
 /// - `peer_removal_cooldown`: duration for non-static bans.
-/// - `cooldown_timers`: `BTreeMap<Peer, ScheduleId>` for active bans (self-scheduled
-///   `CooldownEnded` messages).
+/// - `cooldown_until`: `BTreeMap<Peer, Instant>` of active bans (end time per peer).
+/// - `cooldown_heap`: min-heap of pending `(Instant, Peer)` cool-down entries
+///   (may contain stale entries after re-ban or early `AddPeer` lift).
+/// - `cooldown_timer`: at most one outstanding `schedule_at(CheckCooldowns)` for the
+///   earliest heap entry (`None` when no cool-downs are pending).
 /// - `inbound_peers`: `BTreeMap<Peer, Connection>` (downstream tracking).
 /// - `outbound_peers`: `BTreeMap<Peer, PeerState>` (`Connecting` or `Connected(Connection)`).
 ///
-/// ## Message Handling (core of `stage()` at mod.rs:188)
+/// ## Message Handling
 ///
 /// All behaviour is implemented in the single `match msg` inside `pub async fn stage`.
 /// The stage is purely message-driven; there are no background loops outside scheduled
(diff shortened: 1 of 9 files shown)


crates/amaru-consensus/src/stages/peer_selection/test_setup.rs
```diff
@@ -114,39 +140,63 @@ pub fn setup(prep: &TestPrep, msg: PeerSelectionMsg) -> (SimulationRunning, Dese
...
-            // child created on Initialize does not require a real Ledger resource.
             running
                 .override_external_effect::<VolatileTipEffect>(usize::MAX, |_| OverrideResult::handled(Tip::origin()));
             running.override_external_effect::<TipEffect>(usize::MAX, |_| OverrideResult::handled(Tip::origin()));
             running.override_external_effect::<RegisteredRelaySocketAddrsEffect>(usize::MAX, |_| {
                 OverrideResult::handled(Ok(BTreeSet::new()))
             });
 
-            // Make peer selection's random choices fully deterministic in tests.
+            // NOTE: Make peer selection's random choices fully deterministic in tests.
```
etorreborre: Why not a `TODO`?

crates/amaru-consensus/src/stages/peer_selection/mod.rs
```diff
@@ -304,10 +341,56 @@ impl PeerSelection {
...
+
+        match self.cooldown_timer {
+            None if was_empty => {
+                let id = eff.schedule_at(PeerSelectionMsg::CheckCooldowns, when).await;
+                self.cooldown_timer = Some(id);
+            }
+            None => {
+                self.arm_next_cooldown(eff).await;
+            }
+            Some(id) if id.time() > when => {
```
etorreborre, line 332: I was surprised at first with the condition `id.time() > when`, then I realized that the ban_period can vary depending on the peer. Maybe a comment and / or a small example could help present the situation here.

crates/amaru-consensus/src/stages/peer_selection/mod.rs
```diff
@@ -272,7 +307,9 @@ impl PeerSelection {
             static_peers,
             snapshot_candidates,
             peer_removal_cooldown: Duration::from_secs(peer_removal_cooldown_secs),
-            cooldown_timers: BTreeMap::new(),
+            cooldown_until: BTreeMap::new(),
```
etorreborre: Since those 2 fields are the same data accessed differently, I guess it could be a good idea to give it its own struct like `Cooldowns` and a uniform API.

etorreborre (approved):
Nice idea! We indeed just need one `ScheduleId` at the time.

crates/amaru-consensus/src/stages/peer_selection/test_setup.rs
rkuhn: ah, so I failed: this is the explanation of what the following line does, not a TODO

crates/amaru-consensus/src/stages/peer_selection/mod.rs
rkuhn: While the data are not always the same, I’ll try to see whether a new data structure simplifies the code — in this case it doesn’t look like there is much to gain, though.

crates/amaru-consensus/src/stages/peer_selection/mod.rs
rkuhn: This code by itself shouldn’t be surprising: schedule needs to be updated if the new wakeup is earlier than the previously next wakeup.

crates/amaru-consensus/src/stages/peer_selection/mod.rs
rkuhn: I take that back: it is more readable even though the overall structure is not simpler.

PR #1119, opened 2026-07-31
test: fix a flaky test checking spans

etorreborre:
This PR fixes a flaky test execution (observed [here](https://github.com/pragma-org/amaru/actions/runs/30620201652/job/91122712769?pr=1117)).

The test using `assert_send` is just there to typecheck a value but it actually executes a function invoking the `trace_span!` macro. This starts executing the internal tracing machinery without a default subscribe (what is done in the other tests like  `test_trace_span_instrument_runs_inside_span`). As a result the second test doesn't get a proper context for its span.

## Summary by CodeRabbit

* **Tests**
  * Updated compile-time validation tests to correctly verify that traced asynchronous functions meet sendability requirements.
  * Improved test coverage without invoking asynchronous functions during validation.

### DIFF
--- a/crates/amaru-observability/macros/tests/validation.rs
+++ b/crates/amaru-observability/macros/tests/validation.rs
@@ -381,7 +381,8 @@ mod async_functions {
         }
     }
 
-    fn assert_send<T: Send>(_: T) {}
+    /// Just check that T has type `Send` but don't evaluate the argument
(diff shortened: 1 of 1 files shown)


rkuhn (approved):
Thanks for the explanation, would have been difficult just with the github diff :-)

PR #1124, opened 2026-07-31
fix: only send new tip from track peers for truly new headers from peers

etorreborre:
This PR fixes a consequence of #1087 (validation of already stored headers). It is possible on bootstrap to have headers with no associated nonces. In that case we should not consider those headers as "new" in `track_peers` and not send them to `select_chain` because that stage will see their block as already validated and stop.

We prevent this by avoiding to send downstream headers that are already stored, whether they have, or not, associated nonces in the database.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of previously stored and bootstrap-imported headers during peer tracking.
  * Prevented duplicate headers from being forwarded unnecessarily.
  * Ensured stored headers are revalidated correctly while preserving duplicate-header metrics and logging.
  * Improved processing of new and deferred headers by checking whether they are already present before storage.

* **Tests**
  * Expanded coverage for duplicate headers, header presence checks, revalidation, and deferred header processing.

### DIFF
--- a/crates/amaru-consensus/src/stages/track_peers/mod.rs
+++ b/crates/amaru-consensus/src/stages/track_peers/mod.rs
@@ -79,9 +79,9 @@ pub const HEIGHT_RECHECK_INTERVAL: Duration = Duration::from_millis(200);
 ///
 /// `TrackPeers::try_roll_forward` runs protocol checks (parent, consecutive height, monotonic
 /// slot, clock skew) and Praos validation via ledger `validate_header`. On success it advances
-/// the peer tip, stores the header if new, and notifies `downstream` with [`NewTip`]. If
-/// `RequestNext` was not already sent (e.g. after a height defer is released), it is sent after
-/// success.
+/// the peer tip, stores the header with its nonces, and notifies `downstream` with [`NewTip`]
+/// when the header was not in the store yet. If `RequestNext` was not already sent (e.g. after a
+/// height defer is released), it is sent after success.
 ///
 /// # Deferral
 ///
@@ -382,7 +382,8 @@ impl TrackPeers {
     /// returned, together with the nonces to store alongside it.
     ///
     /// Note: a header can already sit in the chain store without carrying any nonces, as is the
-    /// case for headers imported during bootstrap. Those still need to be validated.
+    /// case for headers imported during bootstrap. Those still need to be validated, but they are
+    /// not forwarded downstream again (see `TrackPeers::try_roll_forward`).
     #[expect(clippy::too_many_arguments)]
     async fn validate_header(
         &mut self,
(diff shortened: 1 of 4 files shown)


crates/amaru-consensus/src/stages/track_peers/mod.rs
```diff
@@ -652,18 +653,15 @@ impl TrackPeers {
...
-                debug!(
-                    consensus::perf::header::LIFECYCLE,
-                    peer = peer.clone(),
-                    header_hash = current.hash(),
-                    outcome = PerfHeaderForwardOutcome::DuplicateHeader.as_str()
-                );
-                record_header_rejected(eff, PerfHeaderForwardOutcome::DuplicateHeader).await;
-            }
+        let parent_of_new_header = match validated {
+            None => None,
```
jeluard: if let instead of match looks more readable here?

etorreborre: Yes, I changed it.

crates/amaru-consensus/src/stages/track_peers/mod.rs
```diff
@@ -652,7 +653,40 @@ impl TrackPeers {
         let RollForwardArgs { peer, header, tip, sent_request_next, handler, trace_context, received_at, .. } = args;
         let header_tip = header.tip();
         let current = header_tip.point();
-        match validated {
+        let parent_of_new_header = if let Some((parent, nonces)) = validated {
+            // A header can already be in the store without carrying nonces, as is the case for
+            // headers imported during bootstrap. Such a header may have had its block validated
+            // by the startup recovery already, and chain selection only accepts tips whose block
+            // has no validity verdict yet, so only headers new to the store are forwarded.
+            let is_new = !store.has_header(&current.hash()).await;
```
rkuhn, line 668: While the comment is true, how does it match this line of code? The case under discussion does have the header but no nonces yet, right?

rkuhn:
From the PR description it sounds like an issue that would still be there if #1127 is merged. Not sure I agree with the solution, though: would it not be more robust to just discard uninteresting messages in `select_chain`?

crates/amaru-consensus/src/stages/track_peers/mod.rs
etorreborre: The comment explains that the header might have been stored without nonces at bootstrap yes. 
In that case we want to make sure that we don't propagate such header. This line of code sets the flag for this situation and allows L.695 to trigger then.

etorreborre:
@rkuhn 
> From the PR description it sounds like an issue that would still be there if https://github.com/pragma-org/amaru/pull/1127 is merged

Yes the problem remains.

> would it not be more robust to just discard uninteresting messages in select_chain?

Why send messages that we know we are going to discard?

rkuhn:
> Why send messages that we know we are going to discard?

Sure, no point in sending them. I was more worried (perhaps I misunderstood) that these messages would cause damage in `select_chain` — that recipient must be robust against useless messages.

etorreborre:
@rkuhn those messages do not respect the contract set by `select_chain`, so it's not damage, it's a breach of contract:
```
tracing::error!(tip = %tip.point(), %valid, "got tip from upstream that was already validated");
return eff.terminate().await;
```

rkuhn:
Thanks for your perseverance and sorry for taking so long! After some deeper review of the existing state I see two issues that need to be fixed:

1. The “nonces present ⇔ header validated” guarantee should be 100% enforced, with no possibility to violate it — leaky guarantees always produce bugs.
2. `select_chain` handling of NewTip must be idempotent (i.e. graceful).

The first one allows solid code with solid reasoning, but it requires upholding the invariant also during bootstrap and migrations. The second one is required because during startup fetch_blocks recovery may contribute a validation status after track_peers has checked it and before select_chain processes the resulting NewTip message — this is fundamentally unfixable.

This PR fixes only part of the problem and if we fix the above then this approach is no longer needed.

etorreborre:
@rkuhn I'm not sure where exactly you want to go. Basically lift the error case in `select_chain`:
```
tracing::error!(tip = %tip.point(), %valid, "got tip from upstream that was already validated");
return eff.terminate().await;
```
and make sure that we adopt the new tip?

We should also rescue the atomic save of header and nonce from this PR I suppose (`store_validated_header`).

rkuhn:
I have a sketch on a branch, currently in the meeting and then traveling back home, should be ready later today — that’s easier to discuss I think.

etorreborre:
Closed in favor of #1198

PR #1127, opened 2026-07-31
add performance resource

rkuhn:
add Performance resource managed by a dedicated thread

Skip-changelog

---

## Summary by CodeRabbit

* **New Features**
  * Added consensus performance monitoring for peer behavior, header and block lifecycles, fetching, forks, latency, bandwidth, and failures.
  * Added peer availability tracking, coverage queries, peer selection, scoring, snapshots, pruning, rollback, and churn ranking.
  * Added performance events, queries, and queue-depth diagnostics for consensus operations.
* **Documentation**
  * Updated comment guidance and added warning-free documentation build requirements.
* **Tests**
  * Added comprehensive coverage for performance tracking, peer behavior, lifecycle transitions, fork handling, and shutdown.

### DIFF
--- a/crates/amaru-consensus/src/performance/effects.rs
+++ b/crates/amaru-consensus/src/performance/effects.rs
@@ -15,17 +15,18 @@
 //! External-effect constructors and implementations for [`Performance`] events.
 //!
 //! Stages use factory methods on [`Performance`] and pass the result to `eff.external(...)`.
+//! Each effect is enqueued as a [`PerformanceOp`] on the worker thread.
 
 use std::{sync::Arc, time::Duration};
 
 use amaru_kernel::{BlockHeight, HeaderHash, Peer, Tip};
-use amaru_metrics::Meter;
 use amaru_protocols::metrics_effects::ResourceMeter;
 use amaru_pure_stage::{BoxFuture, ExternalEffect, ExternalEffectAPI, Instant, Resources, SendData};
+use tokio::sync::oneshot;
 
 use super::{
-    ClaimKind, FetchPeerSet, HeaderLifecycleOutcome, PeerScores, PeerSnapshot, Performance, ResourcePerformance,
-    SelectPeersParams,
+    ClaimKind, FetchPeerSet, HeaderLifecycleOutcome, PeerScores, PeerSnapshot, Performance, PerformanceOp,
+    ResourcePerformance, SelectPeersParams,
 };
 
 fn require_perf(resources: &Resources) -> ResourcePerformance {
@@ -33,8 +34,26 @@ fn require_perf(resources: &Resources) -> ResourcePerformance {
     resources.get::<ResourcePerformance>().expect("Performance effect requires ResourcePerformance").clone()
 }
 
-fn optional_meter(resources: &Resources) -> Option<Arc<Meter>> {
+fn optional_meter(resources: &Resources) -> Option<Arc<amaru_metrics::Meter>> {
     resources.get::<ResourceMeter>().ok().map(|m| m.clone())
+}
+
+/// Fire-and-forget: enqueue and return immediately (does not wait for the worker).
+fn enqueue(perf: &Performance, op: PerformanceOp) {
+    perf.submit(op);
+}
+
+/// Enqueue a query and await the oneshot reply without blocking a multi-thread Tokio worker.
+async fn enqueue_query<T: Send + 'static>(
+    perf: &Performance,
+    make: impl FnOnce(oneshot::Sender<T>) -> PerformanceOp,
+) -> T {
+    let (reply_tx, reply_rx) = oneshot::channel();
+    perf.submit(make(reply_tx));
+    #[expect(clippy::expect_used)]
+    {
+        reply_rx.await.expect("performance worker dropped reply")
+    }
 }
 
 // ---------------------------------------------------------------------------
@@ -155,16 +174,16 @@ impl Performance {
 
 #[derive(Debug, Clone, PartialEq, serde::Serialize, serde::Deserialize)]
 pub struct RecordIntersectionEffect {
-    peer: Peer,
-    current: Tip,
-    parent: Option<HeaderHash>,
-    at: Instant,
+    pub(crate) peer: Peer,
+    pub(crate) current: Tip,
+    pub(crate) parent: Option<HeaderHash>,
+    pub(crate) at: Instant,
 }
 
 impl ExternalEffect for RecordIntersectionEffect {
     fn run(self: Box<Self>, resources: Resources) -> BoxFuture<'static, Box<dyn SendData>> {
         let perf = require_perf(&resources);
-        perf.apply_intersection(self.peer, self.current, self.parent, self.at);
+        enqueue(&perf, PerformanceOp::RecordIntersection { effect: *self, done: None });
         Self::wrap_sync(())
     }
 }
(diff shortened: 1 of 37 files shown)


crates/amaru-consensus/src/performance/mod.rs
```diff
@@ -0,0 +1,380 @@
...
+    /// Enqueue an operation. Updates the pending counter and logs WARN/ERROR thresholds.
+    pub(crate) fn submit(&self, op: PerformanceOp) {
+        let depth = self.pending.fetch_add(1, Ordering::Relaxed) + 1;
+        if depth == QUEUE_WARN_THRESHOLD + 1 {
+            warn!(
+                target: "amaru_consensus::performance",
+                queue_depth = depth,
+                "performance op queue exceeded {QUEUE_WARN_THRESHOLD}"
+            );
+        }
```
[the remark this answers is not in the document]
rkuhn: No, AtomicUsize guarantees that someone must have seen 101 if someone sees 102.

crates/amaru-consensus/src/performance/mod.rs
```diff
@@ -0,0 +1,380 @@
...
+        #[expect(clippy::panic)]
+        if depth == QUEUE_ERROR_THRESHOLD + 1 {
+            error!(
+                target: "amaru_consensus::performance",
+                queue_depth = depth,
+                "performance op queue exceeded {QUEUE_ERROR_THRESHOLD}"
+            );
+            panic!();
+        }
```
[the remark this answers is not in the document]
rkuhn: No, queue > 1000 is a design violation that needs to fail loudly.

crates/amaru-consensus/src/performance/peer.rs
```diff
@@ -0,0 +1,569 @@
...
+
+        ranked.sort_by(|a, b| b.0.partial_cmp(&a.0).unwrap_or(std::cmp::Ordering::Equal).then_with(|| a.1.cmp(&b.1)));
+
+        if ranked.is_empty() {
+            return FetchPeerSet { peers: Vec::new(), weak: true };
+        }
+
+        let peers: Vec<Peer> = ranked.into_iter().take(max_peers).map(|(_, p)| p).collect();
+        FetchPeerSet { peers, weak: false }
+    }
```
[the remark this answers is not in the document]
rkuhn: min_peers has been removed

crates/amaru-consensus/src/performance/effects.rs
```diff
@@ -0,0 +1,581 @@
...
+    pub(crate) current: Tip,
+    pub(crate) parent: Option<HeaderHash>,
+    pub(crate) at: Instant,
+}
+
+impl ExternalEffect for RecordIntersectionEffect {
+    fn run(self: Box<Self>, resources: Resources) -> BoxFuture<'static, Box<dyn SendData>> {
+        let perf = require_perf(&resources);
+        enqueue(&perf, PerformanceOp::RecordIntersection { effect: *self });
+        Self::wrap_sync(())
```
etorreborre: That's equivalent but why not directly this?
```
Self::wrap_sync(enqueue(&perf, PerformanceOp::RecordIntersection { effect: *self }))
```

crates/amaru-consensus/src/performance/mod.rs
```diff
@@ -0,0 +1,337 @@
...
+            );
+        }
+        #[expect(clippy::panic)]
+        if depth == QUEUE_ERROR_THRESHOLD + 1 {
+            error!(
+                target: "amaru_consensus::performance",
+                queue_depth = depth,
+                "performance op queue exceeded {QUEUE_ERROR_THRESHOLD}"
+            );
+            // NOTE: Amaru fails loudly and early when design assumptions are dynamically
```
etorreborre, line 210: That makes me wonder if we should have 2 queues: one for headers performance, one for peers performance. Dropping telemetry events is already something that we already do when OpenTelemetry fails to send events to a collector. That's graceful degrading. On the other hand, dropping peer performance data puts the node at risk of taking the wrong decisions.

Also, now that I'm thinking about it, I wonder if the peer performance should not be a stage on its own given that it hold important (and non-trivial) logic driving peer selection (for block fetching) that retroacts to other stages. If we put all that in a resource we won't be able to simulate it.

crates/amaru-consensus/src/performance/mod.rs
```diff
@@ -0,0 +1,337 @@
...
+    Snapshot { effect: SnapshotEffect, reply: oneshot::Sender<Option<PeerSnapshot>> },
+    RecordRollback { effect: RecordRollbackEffect },
+    RecordHeaderRejected { effect: RecordHeaderRejectedEffect, meter: Option<Arc<Meter>> },
+    RecordHeaderAbandoned { effect: RecordHeaderAbandonedEffect, meter: Option<Arc<Meter>> },
+    RecordForkStarted { effect: RecordForkStartedEffect, meter: Option<Arc<Meter>> },
+    RecordBlockValid { effect: RecordBlockValidEffect, meter: Option<Arc<Meter>> },
+    RecordBlockPruned { effect: RecordBlockPrunedEffect, meter: Option<Arc<Meter>> },
+}
+
+impl Performance {
```
etorreborre, line 155: I would extract this in its own file.

crates/amaru-consensus/src/performance/peer.rs
```diff
@@ -0,0 +1,566 @@
...
+#[derive(Clone, Debug, Default, PartialEq, serde::Serialize, serde::Deserialize)]
+struct PeerState {
+    tips: BTreeMap<HeaderHash, ClaimMeta>,
+    scores: PeerScores,
+}
+
+/// Peer performance map (availability + scores). Owned by the performance worker thread.
+#[derive(Debug, Default)]
+pub struct PeerPerformance {
+    parents: BTreeMap<HeaderHash, ParentInfo>,
```
etorreborre, line 117: Some comments would be nice here (and where missing on some structs in the file)

crates/amaru-consensus/src/performance/effects.rs
rkuhn: Indeed, that’s an oversight.

crates/amaru-consensus/src/performance/mod.rs
rkuhn: If the node cannot digest the performance ops (header & peer) then the design is broken — this must be included in the capacity requirements. What we should avoid is that the external telemetry emission blocks the performance thread, perhaps that should be moved to a dedicated Tokio runtime?

crates/amaru-consensus/src/performance/mod.rs
rkuhn: It is rather small and the purpose matches the name of the module, so why not keep it here? Performance and PerformanceOps is the point of this module.

Or is it our longstanding disagreement about (not) placing logic in non-leaf modules? I’d file this particular module as exhibit A for my cause :-)

crates/amaru-consensus/src/performance/mod.rs
rkuhn: Another point we discussed which I’d like to persist here is that pure-stage offers a solution to the problem of resource-bounded information flow, enforced via a rigorous back-pressure mechanism. Designing that part of the problem such that deadlocks are avoided (i.e. carefully ensuring that no unbounded cycles exist in the graph) takes some effort and adds complexity — this complexity is inherent in the desire for bounded resource usage.

With this preamble it should be obvious that those parts which are not the target of resource bounds should be kept separate so that the complexity of the flow graph remains as low as possible. The performance metrics are bounded by virtue of the information flow being bounded, but adding them into the messages flowing between stages would greatly complicate the design due to new cycles. This was the main reason for choosing the `pure-stage` resource approach. We could also have used a secondary actor-like system without back-pressure, which we’re basically open-coding here with a dedicated thread, and the `pure-stage` mechanism for crossing this boundary would always be an `ExternalEffect`.

Now another reason for why header and peer performance are tracked inside the same infrastructure is that some input events are intrinsically relevant to both purposes and it makes sense to keep the probe points as semantic events like “header announced” instead of separating the peer tracking from the header lifecycle.

PR #1130, opened 2026-07-31
use Performance resource to select peers for fetch_blocks

### LINKED ISSUE #1093: Limit block fetching to bounded subset of peers
Fetching blocks needs to use enough peers (at least five if available) but should not use too many either due to bandwidth constraints. The response timeout needs to be short enough so that a retry with a different set of peers is possible, which at longer range queries may require adding a timeout for first response in addition to the overall response timeout.

Assumptions:

- 10Mbps was the classical reservation, we may consider increasing that to 100Mbps
- at 90kB blocks and 25 blocks per range query and a 2sec overall timeout this requires 9.2Mbps for that single peer
- syncing is limited by throughput, live mode should have stricter timeouts but also no large range queries — timeout based on range size?

rkuhn:
Use the Performance resource to select upstream peers for fetching blocks.

fixes #1093

---

## Summary by CodeRabbit

* **New Features**
  * Block fetching can now select up to three high-performing peers for each request.
  * Requests fall back to all eligible connections when peer selection is unavailable or insufficient.

* **Bug Fixes**
  * Improved timeout handling by avoiding failure attribution to peers that successfully contributed blocks.
  * Enhanced tracking of contacted and contributing peers during block retrieval.


rkuhn:
@coderabbitai review

PR #1133, opened 2026-08-01
Rework stake distribution and rewards calculations (asynchronous calculation + slim summary in memory)

### LINKED ISSUE #876: Compute rewards asynchronously
### Abstract

There is a `FIXME` in `state.rs` to compute rewards:
```
compute rewards in a thread, or in a non-blocking manner to carry on with other
tasks while rewards are being computed; they only need to be available at the epoch
boundary.
```
And later
```
This should eventually be a '.await', as we always expect to *eventually*
have some rewards summary being available. There's no way to continue progressing
the ledger if we don't.
```            

### Why?

Because it takes a long time and is a blocking operation

### How?

By spawning the computation in a separate thread.

The end to end snapshot tests should be enough.

### Discussion points

_No response_

_No response_

### Checklist

KtorZ:
See EDR-029 in the PR.

## Summary by CodeRabbit

* **New Features**
  * Added memory-efficient stake summaries and on-demand rewards computation.
  * Rewards and stake-distribution processing now runs asynchronously during epoch transitions.
  * Pool voting correctly applies fallback choices, including No Confidence and Abstain.
  * Added runtime memory and allocation measurements for diagnostics.

* **Bug Fixes**
  * Improved historical data handling during validation and rewards processing.
  * Added clearer reporting when background computation fails.

* **Documentation**
  * Documented the runtime stake-summary and rewards architecture.
  * Updated rewards computation trace fields.


rkuhn (commented):
only reviewed the EDR: looks good to me!

PR #1134, opened 2026-08-01
prune performance tracking when no longer relevant

rkuhn:
- peers are removed when banned
- headers for which blocks are not fetched are pruned when older than
  anchor

Skip-changelog

---


rkuhn:
@coderabbitai review

PR #1138, opened 2026-08-03
Introduce VotersDoNotExist logic

### LINKED ISSUE #923: Ledger predicate: VotersDoNotExist
### Abstract

Implement predicate`VotersDoNotExist`

### Why?

Conformance

### How?

[Reference Implementation](https://github.com/IntersectMBO/cardano-ledger/blob/0cfbf861cfb456660a7b73281c6fb714a53d40f9/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs#L581-L613)

### Discussion points

_No response_

_No response_

### Checklist

yHSJ:
Closes #923

This PR introduces logic to validate that any voters present in the transaction are actually registered entities, of whatever they claim to be. (`VotersDoNotExist` predicate). We had to update the test schema to support testing this new logic.

I also discovered a minor bug that could've led to unexpected behavior; a hot key can be registered by many cold credentials, so we have to allow for more than one `CCMember` to be resolved by the reverse lookup (hot key -> `CCMember`)

## Summary by CodeRabbit

* **New Features**
  * Added validation to reject votes from voters not present in ledger state.
  * Added support for constitutional committee hot-key authorization, rotation, resignation, and same-transaction voting.
  * Added serialization support for proposal kinds and proposal-root data.
* **Bug Fixes**
  * Improved committee membership resolution across pending and epoch-boundary changes.
  * Corrected voting behavior for registered DReps and stake pools.
* **Documentation**
  * Expanded changelog and phase-one scenario documentation with voting coverage.


crates/amaru-ledger/tests/data/phase-one/fail/StakePoolNotRegisteredOnKeyPOOL/0.json
```diff
@@ -18,6 +18,8 @@
     "pools": [],
     "accounts": [],
     "dreps": [],
+    "committee": [],
+    "proposals": [],
```
KtorZ: We gotta stop with this :( ... I have added default decoders specifically for this. It's annoying to constantly see 150+ files modified when it's just default values being added. It's not necessary, we don't have to touch the files.

KtorZ:
(rebasing and solving conflicts)

crates/amaru-observability/src/schemas.rs
```diff
@@ -463,6 +463,8 @@ define_schemas! {
                     public HYDRATE {
                         optional from_volatile: u64
                         optional from_db: u64
+                        /// Members scanned to match a vote's hot credential, which is not a store key.
+                        optional scanned: u64
```
KtorZ: That is weird and the comment feels off; "scanned" is roughly the same as `from_db` in this context; and the key is definitely in the store!

crates/amaru-ledger/src/context/default/validation.rs
```diff
@@ -340,6 +341,27 @@ impl CommitteeSlice for DefaultValidationContext {
...
+    /// See also [`authorizedHotCommitteeCredentials`][haskell] from the Haskell.
+    ///
+    /// The length of this set is *NOT* an authoritative count as CCMembers with the same hot key
+    /// and expiration collapse into one.
+    ///
+    /// [haskell]: https://github.com/IntersectMBO/cardano-ledger/blob/0cfbf861cfb456660a7b73281c6fb714a53d40f9/libs/cardano-ledger-core/src/Cardano/Ledger/State/CertState.hs#L335-L337
+    fn lookup_by_hot_credential(&self, hot_credential: &StakeCredential) -> BTreeSet<CCMember> {
+        self.committee
+            .keys()
+            .chain(self.state.committee.produced.keys())
```
KtorZ: This does not seem correct as it bundles old/overridden hot credentials with the others? Any key from `self.committee` may have been removed by a resignation or by a new delegation in produced.

Before fixing, make sure there's a test added for this case.

crates/amaru-ledger/src/context/default/preparation.rs
```diff
@@ -376,61 +382,97 @@ fn resolve_dreps<'volatile>(
 // member (Haskell's `cgceCommitteeProposals`), which lets a not-yet-elected member pre-declare
 // its hot key. That source needs the proposals read-path, so it is deferred until proposals are
 // exposed.
-fn resolve_committee<'block, 'volatile>(
+fn resolve_committee<'volatile>(
     volatile: &'volatile impl VolatileState<CCMember<'volatile> = Existence<CommitteeMemberBind<'volatile>>>,
     db: &impl ReadStore,
-    mut keys: impl Iterator<Item = &'block StakeCredential>,
+    cold_credentials: impl Iterator<Item = StakeCredential>,
+    hot_credentials: &BTreeSet<StakeCredential>,
```
KtorZ:
```suggestion
    voters: &BTreeSet<StakeCredential>,
```

crates/amaru-ledger/src/context/default/preparation.rs
```diff
@@ -376,61 +382,97 @@ fn resolve_dreps<'volatile>(
...
-                            from_db += 1;
-                            hot_credential.owned().set_or_reset(&mut row.hot_credential);
-                            CCMember { hot_credential: row.hot_credential, valid_until: row.valid_until }
-                        })
-                    }
+        // A vote names its member by hot credential, which no layer is indexed by. Scan the members
+        // reachable at block start -- the stable rows, plus any the volatile window touched -- and keep
+        // the ones a vote refers to. A member whose hot key is authorized by the *ongoing* block is
+        // reached through its cold credential instead, which that certificate already required.
+        if !hot_credentials.is_empty() {
```
KtorZ: To keep things optimal, we should really defer resolving cc_members from cold credentials from after this. Since we're going to do a full scan of all cc members, we may as well take the opportunity to resolve the cold credentials state during that iteration. That means, if any hot credential is required, then we can resolve the cold credential in the same iteration. 

This would also avoid the slightly awkward `touched_cc_members` method... 

So really, this entire resolver should be transformed so that:

- If no hot credentials are required, we hand-pick and resolve CC members individually. 
- If any hot credential is required, we then resolve required CC members while scanning them all, retaining only those explicitly required or with hot credentials we care about.

crates/amaru-ledger/tests/data/phase-one/fail/StakePoolNotRegisteredOnKeyPOOL/0.json
yHSJ: I was actually being very intentional about requiring every field, as I feel like it makes it much simpler to consume as an external user of this suite. It also means people are less likely to make mistakes when creating new fixtures, since they are required to provide every field. I felt that value outweighed the cost of updating these fixtures, especially since once they are settled this will happen much less frequently.

That being said, as long as we carefully maintain the `schema.json`, there should be enough to reduce ambiguity for consumers.

crates/amaru-observability/src/schemas.rs
yHSJ: The *hot key* is not a key in the store. But, if it's not a worthwhile addition, removing it is totally reasonable too.

KtorZ: You mean key as in, key/value key?

crates/amaru-ledger/src/context/default/validation.rs
yHSJ: This function just returns cold credentials.

crates/amaru-ledger/src/context/default/preparation.rs
```diff
@@ -364,64 +372,52 @@ fn resolve_dreps<'volatile>(
...
+
+        let mut voter_candidates = BTreeSet::new();
+        if !voters.is_empty() {
+            // A member changed inside the volatile window may have no store row yet, so the store's
+            // keys alone do not name the whole committee.
+            voter_candidates.extend(volatile.cc_members().copied());
+            voter_candidates.extend(
+                db.iter_cc_members()
+                    .map_err(ContextHydratationError::ResolveCommittee)?
+                    .map(|(credential, _)| credential),
```
KtorZ: This does not look right and deviates from the design we've adopted so far. The volatile owns the _current state_, so it's the volatile's responsibility to tell whether an entitiy is present, gone or unknown. Only when it's unknown should we reach for the stable store.

crates/amaru-ledger/src/state/volatile.rs
```diff
@@ -99,6 +100,10 @@ pub trait VolatileState {
         Self: 'a;
     fn resolve_cc_member<'a>(&'a self, credential: &StakeCredential) -> Self::CCMember<'a>;
 
+    /// Every cold credential these layers can resolve to a member for. A hot key authorized or a seat
+    /// granted inside the window has no stable row yet, so iterating the store cannot enumerate these.
+    fn cc_members(&self) -> impl Iterator<Item = &StakeCredential>;
```
KtorZ: I dont think this should exists. Or we should clarify the semantic. Does this also return retired members?

crates/amaru-ledger/src/context/default/validation.rs
```diff
@@ -398,6 +396,24 @@ impl ProposalsSlice for DefaultValidationContext {
...
+            for (cold_credential, _) in added.iter() {
+                if !matches!(self.state.committee.get(cold_credential), Existence::Exists(..)) {
+                    self.state
+                        .committee
+                        .bind_left(*cold_credential, None)
+                        .unwrap_or_else(|_| unreachable!("committee members are never 'unregistered'"));
+                }
+            }
+        }
+
```
KtorZ, line 416: I need to double-check here and have a test either way; but this may lead to discrepancy for the validations of the _current transaction_. Depending on when the certificate validation is done, the outcome will change: 

- If certificates are validated _before_, then a CC member registered in the same transaction cannot immediately delegate to a hot credential; 
- if certificates are validated _after_, then that becomes possible because of this now. 

If I recall correctly, the Haskell node validates proposals _after_ certificates, so that is fine. But a test should capture that.

crates/amaru-ledger/src/state/volatile/aggregate.rs
```diff
@@ -18,22 +18,21 @@ use std::{
...
 mod indexed_bind;
 pub use indexed_bind::IndexedBind;
 
 mod indexed_epoch_reg;
 pub use indexed_epoch_reg::IndexedEpochReg;
 
-mod indexed_set;
-pub use indexed_set::IndexedSet;
+// mod indexed_set;
+// pub use indexed_set::IndexedSet;
```
KtorZ, line 35: To be cleaned up entirely.

KtorZ:
Merged manually.

PR #1141, opened 2026-08-04
feat: added some more metrics

jeluard:
Fill the gaps of missing metrics to reach cardano_node parity. Tweak CI so that metrics missing amaru side are clearly stated.

## Summary by CodeRabbit

* **New Features**
  * Added metrics for node, network, filesystem, memory, forging, startup, and basic node information.
  * Added served-block and served-header metrics, including latest block position and chain-sync header counts.
  * Added transaction-processing metrics and streamlined metrics initialization.

* **Bug Fixes**
  * Renamed mempool duration and served-block metrics for consistency.
  * Updated the relay dashboard to use the renamed mempool metric.

* **Documentation**
  * Documented the new metrics in the unreleased changelog.


KtorZ:
Arg; this is going to clash with the TUI as I moved the process metrics into the amaru-metrics crate now, where they belong with the other metrics. Not a big deal.

jeluard:
Who doesn't enjoy a nice merge!

crates/amaru-metrics/src/mempool.rs
```diff
@@ -22,6 +22,25 @@ use opentelemetry::KeyValue;
 use crate::{Counter, Gauge};
 use crate::{Meter, MetricRecorder, MetricsEvent};
 
+#[cfg(not(target_arch = "wasm32"))]
+static TXS_PROCESSED: OnceLock<Counter<u64>> = OnceLock::new();
+
```
KtorZ, line 27: Why not a LazyLock directly?

crates/amaru-metrics/src/protocol.rs
```diff
@@ -15,13 +15,75 @@
...
+            .build()
+    })
+}
+
+#[cfg(not(target_arch = "wasm32"))]
+fn served_block_latest(meter: &Meter) -> &Gauge<u64> {
+    SERVED_BLOCK_LATEST.get_or_init(|| {
+        meter
+            .u64_gauge("cardano_node_metrics_served_block_latest_int")
+            .with_description("number of blocks served at the highest slot observed so far")
```
KtorZ, line 51: I don't understand this? And I don't think the implementation recording a slot is correct either.

crates/amaru/src/metrics.rs
```diff
@@ -130,3 +142,107 @@ fn record_build_info(provider: &SdkMeterProvider) {
...
+    number_of_cpus: u64,
+    meter: Option<&Meter>,
+    cardano_metrics: Option<&CardanoMetrics>,
+) {
+    let disk_usage = process.disk_usage();
+    let metrics = SystemMetrics {
+        runtime_seconds: process.run_time(),
+        cpu_percent: process.cpu_usage() as f64 / number_of_cpus as f64,
+        process_memory_bytes: process.memory(),
+        rss_bytes: process.memory(),
```
KtorZ: Just noticed this duplication here.

crates/amaru/src/metrics.rs
```diff
@@ -130,3 +142,107 @@ fn record_build_info(provider: &SdkMeterProvider) {
...
+    notify_subscribers(&metrics.into());
+}
+
+struct CardanoMetrics {
+    cpu_ticks: Gauge<u64>,
+    network_read_bytes: Gauge<u64>,
+    network_written_bytes: Gauge<u64>,
+    filesystem_read_bytes: Gauge<u64>,
+    filesystem_written_bytes: Gauge<u64>,
+    memory_resident_bytes: Gauge<u64>,
```
KtorZ: Apart from the network metrics, all these are already in the System metrics.

jeluard: Hmm I think I will switch the metrics lists (in the json) to a different format. Current one makes it to miss duplication when metrics are renamed.
Recent `cardano_node` renamed a number of metrics, and some of what we used to have are not the right ones anymore.

crates/amaru-metrics/src/protocol.rs
jeluard: It's the count of how many times the current `tip` has been served. If the tip moves the count resets.

Unclear if this is what is expected.

crates/amaru/src/metrics.rs
KtorZ: It makes me wonder how much of those we then actually want to support. If there's some degree of freedom in the metrics API, I rather not use names like `cardano_node_metrics_ChainSync_HeadersServed_counter` but something a bit less Haskell-tainted and more idiomatic with [Prometheus's conventions](https://prometheus.io/docs/practices/naming/) like `amaru_headers_served_total` 😬

jeluard:
@rkuhn is the idea behind the newly introduced `Telemetry::install` to encapsulate everything telemetry? Should it be used by amaru CLI?

crates/amaru-metrics/src/consensus.rs
```diff
@@ -15,18 +15,82 @@
...
 #[cfg(not(target_arch = "wasm32"))]
-use crate::{Counter, Histogram};
+use crate::{Counter, Gauge, Histogram};
 use crate::{Meter, MetricRecorder, MetricsEvent};
 
-/// Performance measurements emitted by the select_chain stage, computed from `perf.*` events
-/// but aggregatable as OpenTelemetry metrics (a duration histogram plus an outcome-labelled
-/// counter per measurement kind).
+#[derive(Debug, Clone, Copy, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
+pub enum GsmState {
```
KtorZ, line 29: I have no idea what `Gsm` means in this context, and I am afraid to ask 😶

KtorZ: Genesis State Machine

KtorZ: So that is a bit ... odd.

crates/amaru-metrics/src/mempool.rs
```diff
@@ -149,10 +165,9 @@ impl MetricRecorder for MempoolMetrics {
                 .build()
         });
 
-        // cumulative mempool sync duration.
-        let txs_sync_duration_total = TXS_SYNC_DURATION_TOTAL.get_or_init(|| {
+        let txs_sync_duration_total = TXS_SYNC_DURATION_TOTAL_COUNTER.get_or_init(|| {
             meter
-                .u64_counter("cardano_node_metrics_txsSyncDurationTotal_int")
+                .u64_counter("cardano_node_metrics_txsSyncDurationTotal_counter")
```
KtorZ, line 170: *sigh*

crates/amaru-metrics/src/protocol.rs
```diff
@@ -87,6 +155,16 @@ pub struct ServedBlockCountMetrics {
     pub count: u64,
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
+pub struct ServedBlockLatestMetrics {
```
KtorZ, line 159: How is that different from `TipBlock`?

crates/amaru-metrics/src/protocol.rs
```diff
@@ -159,24 +239,71 @@ impl MetricRecorder for ServedBlockCountMetrics {
...
+    fn record_to_meter(&self, _meter: &Meter) {}
+}
+
 #[cfg(not(target_arch = "wasm32"))]
-impl MetricRecorder for ServedBlockCountMetrics {
+#[derive(Default)]
+struct ServedBlockLatestState {
+    max_slot: Slot,
+    count: u64,
+}
```
KtorZ, line 252: Seems like this is redundant with ServedBlockCount and TipBlock already?

crates/amaru-metrics/src/protocol.rs
```diff
@@ -159,24 +239,71 @@ impl MetricRecorder for ServedBlockCountMetrics {
...
+        }
+        self.count
+    }
+}
+
+#[cfg(not(target_arch = "wasm32"))]
+impl MetricRecorder for ServedBlockLatestMetrics {
     fn record_to_meter(&self, meter: &Meter) {
-        static SERVED_BLOCK_COUNT: OnceLock<Counter<u64>> = OnceLock::new();
+        static STATE: OnceLock<Mutex<ServedBlockLatestState>> = OnceLock::new();
```
KtorZ, line 270: A OnceLock of a Mutex? That does not seem right.

KtorZ (commented):
Overall I don't know if I am super fond of the Haskell's node metrics... a lot of them can be derived from the traces already and I see little value in collecting that internally and re-emit them... 

Perhaps we should accept the non-parity over the metrics there and focus our metrics on elements that are hard or unpractical to derive from the traces?

KtorZ:
@rkuhn happy to have a second opinion here. We'll likely need to take the time one day to properly go through the metrics and telemetry we emit in general to see if this is reasonable.

jeluard:
Main plan was to support all cardano_node metrics not directly tied with haskell to ensure maximum compatibility with existing tooling.
Some could be dropped like the gsm one.

KtorZ:
> Main plan was to support all cardano_node metrics not directly tied with haskell to ensure maximum compatibility with existing tooling.

Indeed, but the more I see of those metrics, the less I think this was a good plan 🙈

KtorZ:
While we have to re-implement protocol bugs and implementation quirks, perhaps we can avoid re-implementing inadequate choices when we can; unless enough people complain about it...

jeluard:
Which one specifically do you think are irrelevant?

* cardano_node_metrics_GSM_state_int
* cardano_node_metrics_blockReplayProgress_real

More? `served*` ones?

KtorZ:
Most of them 🙈 ...

- "cardano_node_metrics_ChainSync_HeadersServed_counter" => redundant with `amaru_consensus_header_total`? 
- "cardano_node_metrics_GSM_state_int" => we don't do genesis and we don't really have any distinct syncing mode, so that is weird. 
- "cardano_node_metrics_Mem_resident_int" => readily available from `/proc` or any process monitoring tool
- "cardano_node_metrics_Stat_blkIOticks_int" => readily available from `/proc`
- "cardano_node_metrics_Stat_cputicks_int" => also process-specific, so available by watching the process
- "cardano_node_metrics_Stat_fsRd_int" => same
- "cardano_node_metrics_Stat_fsWr_int" => same
- "cardano_node_metrics_Stat_netRd_int" => same 
- "cardano_node_metrics_Stat_netWr_int" => same
- "cardano_node_metrics_basicInfo" => no idea what this refers to
- "cardano_node_metrics_blockReplayProgress_real" => we don't do block replay
- "cardano_node_metrics_blockNum_int" => can be derived from traces, but why not.. 
- "cardano_node_metrics_cardano_build_info" => emitted on startup
- "cardano_node_metrics_cardano_version_major_int" => emitted on startup
- "cardano_node_metrics_forging_enabled_int" => weird, could be emitted on startup
- "cardano_node_metrics_node_start_time_int" => weird but likely harmless
- "cardano_node_metrics_served_block_counter" => can be derived from traces
- "cardano_node_metrics_served_block_latest_int" => can be derived from traces
- "cardano_node_metrics_served_header_counter" => redundant with `cardano_node_metrics_ChainSync_HeadersServed_counter` ?
- "cardano_node_metrics_slotNum_int" => can be derived from traces, but why not..
- "cardano_node_metrics_tipBlock" => can be derived from traces but why not...
- "cardano_node_metrics_txsProcessedNum_int" => can be derived from traces
- "cardano_node_metrics_txsSyncDurationTotal_counter" => irrelevant
- "cardano_node_metrics_txsSyncDuration_int" => irrelevant

Of course, some of those "derived from traces" only holds if traces are at the DEBUG level. So maybe it is important to have this redundancy in the metrics; but it's ... awkward.

crates/amaru-metrics/src/protocol.rs
rkuhn: Well, it kind of makes sense if that is really what we need … the OnceLock protects initialization while the Mutex guards mutation.

rkuhn:
My opinion: the aggregations make sense to keep because otherwise the operator would need to keep all the logs to get the same info (blockNum*, served_*, txsProcessed*). The others are indeed questionable — we should not try to improve on the surrounding O/S, for example.

rkuhn:
> @rkuhn is the idea behind the newly introduced `Telemetry::install` to encapsulate everything telemetry? Should it be used by amaru CLI?

@jeluard It currently only simplifies the life of external users of `amaru-node`, unification with `amaru` binary requires a better plan on how to handle the TUI, for example.

jeluard:
During last call we agreed to only keep counter based metrics, so:

* cardano_node_metrics_ChainSync_HeadersServed_counter
* cardano_node_metrics_served_block_counter
* cardano_node_metrics_served_header_counter
* cardano_node_metrics_txsProcessedNum_int
* cardano_node_metrics_txsSyncDurationTotal_counter
* cardano_node_metrics_txsSyncDuration_int

It it unclear to me that any of those are valuable. If we decide that not, I'll just drop this PR.

cc @KtorZ @rkuhn

rkuhn:
I still think they are valuable and computing them outside Amaru is not worth it, @KtorZ thinks that they are computable outside Amaru and therefore should not be computed inside Amaru. I’m not aware of other opinions, so please speak up.

KtorZ:
I think what I mostly dislike is the amount of code we have to introduce a bit everywhere to count those things, on top of the traces we already emit. So perhaps, an idea: can't we derive those metrics through a dedicated subscriber on the traces? Since we cannot easily hook a connector onto the OTLP collector, we may as well just hook directly _in process_. 

Similar to what the TUI does (and ideally, the TUI should become a subscriber of this other component) but we could simply get the traces we already emit, and keep all the metric aggregation local to one component; then if we want a metric to be reported, we can simply emit the traces that will allow its computation. 

Note that peers metrics are completely out of the picture here because they are part of the internal logic and must be owned by the peer management stages.

crates/amaru-consensus/src/stages/adopt_chain/mod.rs
```diff
@@ -238,6 +240,13 @@ pub async fn stage(mut state: AdoptChain, msg: AdoptChainMsg, eff: Effects<Adopt
             );
             state.suppressed += 1;
         }
+
+        // Amaru does not implement cardano-node's Honest Availability Assumption. Approximate its
+        // GSM using the applied height and the best height observed by the block-fetch pipeline:
+        // https://github.com/IntersectMBO/ouroboros-consensus/blob/main/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Node/GSM.hs
+        let gsm_state = GsmState::from_chain_progress(msg.block_height(), state.max_block_height, None);
+        Metrics::new(&eff).record(ConsensusMetrics::GsmState(gsm_state).into()).await;
```
KtorZ, line 248: I don't think we should be emitting GsmState at all; it does not apply to Amaru.

crates/amaru-kernel/src/utils/process.rs
```diff
@@ -12,10 +12,45 @@
...
+/// Linux exposes this as field 42 (`delayacct_blkio_ticks`) in
+/// `/proc/self/stat`. Other Unix platforms do not provide the equivalent
+/// Linux process statistic and report zero, matching `cardano-node`.
+///
+/// `cardano-node` delegates resource collection to this Linux sampler:
+/// <https://github.com/IntersectMBO/hermod-tracing/blob/d322eb06fa2b351c0943ff2b80a94bf5ed954cd0/hermod-trace-resources/src/Hermod/Tracing/Resources/Linux.hs#L103-L120>
+pub fn sample_process_block_io_ticks() -> u64 {
+    #[cfg(target_os = "linux")]
+    {
+        std::fs::read_to_string("/proc/self/stat")
```
KtorZ, line 29: I don't see the benefits here; operators can sample that themselves, no?

crates/amaru-metrics/src/protocol.rs
KtorZ: Well okay, but the `ServedBlockLatestState` is just two integers... we can initialize it in place.

jeluard:
Parking for now

PR #1154, opened 2026-08-05
Fixture cleanup

yHSJ:
This PR cleans up fixtures which had multiple failure cases. We avoid those, so that implementations that may run validations in a different order can still use our fixtures.

It also committed the keys we use to sign fixtures, so that it's easier to create new ones, modify existing ones, and just generally stay consistent. As part of that, I edited and resigned existing fixtures that used other keys previously.

no-changelog

## Summary by CodeRabbit

- **Tests**
  - Expanded phase-one transaction validation coverage across successful, validation-failure, and decoding-failure scenarios.
  - Refined test fixtures for delegation, withdrawals, fees, transaction limits, credentials, and UTxO behavior.
  - Added stronger checks for expected failure outcomes and clearer handling of invalid test data.
  - Automated fixture discovery and duplicate-title detection.

- **Documentation**
  - Added guidance for shared test credentials, fixture naming, initial-state fields, and credential usage.


KtorZ:
Since we are touching all the fixtures, is this a good time to suggest:

- flattening the fixtures (why do we sometimes have them nested in a folder, and sometimes not?)
- use a more conventional naming scheme such as `{fixture-id}-(pass|fail)-{human-readable-name}.json`; for example: `00001-fail-stake-delegation-to-unregistered-pool.json` ? Just so we can refer to them more easily and they also get easier to consume and organise. 

🙈 ?

yHSJ:
> flattening the fixtures (why do we sometimes have them nested in a folder, and sometimes not?)
Two different conventions that both had reasonable reasons to exist, but neither one was authoritative. I think flat is fine, probably better.

> use a more conventional naming scheme such as {fixture-id}-(pass|fail)-{human-readable-name}.json; for example: 00001-fail-stake-delegation-to-unregistered-pool.json ? Just so we can refer to them more easily and they also get easier to consume and organise.

Yeah, that's a good idea. I think we should drop the `(pass|fail)`, since the folder it's in already says if it's pass or fail.

KtorZ (approved):

PR #1155, opened 2026-08-05
add `amaru node rollback` command

### LINKED ISSUE #1072: Have one dev command to rollback a recently ingested invalid block
### Abstract

Currently, when we ingest a (usually wrongly) invalid block, we mark peers as adversarial and get stuck. Usually, when blocks are invalid, it's because we made a mistake. 

Having a single (dev) command to recover from that situation would be handy. 

### Why?

Currently, rolling back manually is a bit cumbersome as it demands multiple internal commands. So while it is nice to have the flexibility offered by the other commands (although they would also deserve a little tutorial!), the case of rolling back just before the last "invalid" block ingested is common enough to deserve its own helper. 

### How?

There are multiple ways to approach this.

- One way could be to simply have a command to drop from the consensus anything that was considered "volatile" by the ledger. Or said differently, have a command that rollback to the immutable tip. 

- A more general approach could be to allow rolling back to _any_ tip; although that is not current possible due to the ledger architecture. But a weaker (and still useful) version of that is possible: rolling back to any _epoch start_. 

So in a sense, a command that can do both would be nice:

```console
amaru node rollback 
  (--immutable-tip | --epoch EPOCH_NUMBER) 
  [--ledger-db  DIR]
  [--chain-db DIR]
  --network NETWORK
```

Left as an exercise for the reader. 

### Discussion points

- Such a command should replace entirely the current `dev ledger reset-to-epoch`

_No response_

### Checklist

rkuhn:
fixes #1072

## Summary by CodeRabbit

* **New Features**
  * Added `amaru node rollback` for immutable-tip and epoch-based rollback.
  * Added chain-state realignment after ledger rollback.
  * Added configurable network and database locations.
  * Added rollback observability and trace details.

* **Changed**
  * Updated legacy reset commands to recommend the new rollback workflow.
  * Preserved compatibility for epoch-based reset usage.
  * Added command-line help coverage for rollback options.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@ Other guiding principles:
 
 ### Added
 
+- **amaru**: add `amaru node rollback` to recover after a wrongly invalidated block or to rewind to an epoch start. Supports `--immutable-tip` (chain store only) and `--epoch` (ledger snapshot reset + chain realign). Clears all descendant validation flags, sets the anchor/best tip, and culls the best-chain fragment. ([#1072](https://github.com/pragma-org/amaru/issues/1072))
 - **amaru**: add `amaru mithril sync` to download verified Mithril immutable files and replay their blocks directly into the chain and ledger stores.
 - **amaru**: add `amaru node rm --wipe-all-dbs` to remove the ledger and chain databases resolved from the selected network.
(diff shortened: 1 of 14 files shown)


rkuhn:
merging now  —  @KtorZ the comments on `reset_ledger_to_epoch` seem out of scope of this PR

PR #1157, opened 2026-08-05
Introduce UnelectedCommitteeVoters logic

### LINKED ISSUE #922: Ledger predicate: UnelectedCommitteeVoters
### Abstract

Implement predicate `UnelectedCommitteeVoters` 

### Why?

Conformance

### How?

[Reference Implementation](https://github.com/IntersectMBO/cardano-ledger/blob/0cfbf861cfb456660a7b73281c6fb714a53d40f9/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs#L486-L489)

### Discussion points

_No response_

_No response_

### Checklist

yHSJ:
Closes #922

This PR doesn't allow votes from committee members who have not yet been elected (pending election). It is possible to authorize a hot key from a pending cold key. Importantly, this logic does NOT validate the valid_until of the committee member. That is validated separately, and will be implemented as part of that work.

---

### DIFF
--- a/crates/amaru-ledger/src/rules/transaction/phase_one/voting_procedures.rs
+++ b/crates/amaru-ledger/src/rules/transaction/phase_one/voting_procedures.rs
@@ -15,8 +15,8 @@
 use std::collections::{BTreeMap, BTreeSet};
 
 use amaru_kernel::{
-    EraHistory, HasOwnership, MemoizedDatum, NonEmptyKeyValuePairs, ProposalId, RedeemerTag, RequiredScript,
-    StakeCredential, TransactionPointer, Voter, VotingProcedure,
+    EraHistory, HasOwnership, MemoizedDatum, NonEmptyKeyValuePairs, ProposalId, ProtocolVersion, RedeemerTag,
+    RequiredScript, StakeCredential, TransactionPointer, Voter, VotingProcedure,
 };
 use thiserror::Error;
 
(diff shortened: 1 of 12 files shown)


crates/amaru-ledger/src/rules/transaction/phase_one/voting_procedures.rs
```diff
@@ -57,6 +61,10 @@ where
         let mut unknown_proposals = BTreeSet::new();
 
         for (voter, votes) in voting_procedures.iter() {
+            if protocol_version.0 > 10 && is_unelected_committee_voter(context, voter) {
```
KtorZ:
```suggestion
            if protocol_version.major() > PROTCOL_VERSION_10 && is_unelected_committee_voter(context, voter) {
```

Otherwise it makes finding those branches later needlessly hard.

crates/amaru-ledger/tests/data/phase-one/pass/vote-committee-member-with-expired-term.json
```diff
@@ -0,0 +1,54 @@
+{
+  "title": "vote cast by an elected committee member whose term has already ended",
+  "description": "A member the elected committee still names, voting in epoch 1 with a term that ran out at the end of epoch 0.",
```
KtorZ, line 3: Is that a thing?

KtorZ: So yes, I've double-checked and... it is apparently a thing. Expired members can still vote and their vote is only ignored later during ratification. That's interesting.

KtorZ (approved):
Yes, modulo the PROTOCOL_VERSION_10 hard coded magic constant.

crates/amaru-ledger/tests/data/phase-one/pass/vote-committee-member-with-expired-term.json
yHSJ: Yes, that was my reaction too 😅 

This definitely feels more like a consequence of the two different maps Haskell uses and they simply didn't consider this case, than intentional.

rkuhn (commented):
also looks reasonable to me

PR #1159, opened 2026-08-06
fix: successfully validate lists of BLS values if empty

etorreborre:
On `preprod` a transaction contains the following script:
```
 (con (list bls12_381_G1_element) []))
```

The Haskell `cardano-node` allows this list element type as long as there are no values of that type in the list. On the other hand `amaru` rejects any list with that element type.

## Summary by CodeRabbit

* **Bug Fixes**
  * Empty Plutus lists/containers containing BLS12-381 elements are now validated and decoded correctly.
  * The flat decoder now recognizes BLS12-381 G1/G2 and Miller loop result type tags, including in nested type structures.
  * Unsupported decoding of actual BLS12-381 values now returns clearer, more accurate errors.
* **Tests**
  * Added decoding tests for empty/nested containers and expected failures when BLS values are encountered.
* **Documentation**
  * Updated the unreleased changelog entry for the validation fix.


KtorZ:
What? We G1 and G2 points can now be encoded to flat; did we miss that in Van Rossem 🤨 ?

crates/amaru-uplc/src/flat/decode/error.rs
```diff
@@ -44,8 +44,8 @@ pub enum FlatDecodeError {
     UnknownTypeTags(Vec<u8>),
     #[error("Missing type tag")]
     MissingTypeTag,
-    #[error("BLS type not supported")]
-    BlsTypeNotSupported,
+    #[error("Flat decoding is not supported for BLS12-381 element values: use uncompress on a bytestring instead")]
+    BlsValueNotSupported,
```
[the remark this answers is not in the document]
etorreborre: Changed the message.

crates/amaru-uplc/tests/preprod_script_decode.rs
```diff
@@ -0,0 +1,71 @@
...
+/// `(program 1.1.0 (con bls12_381_G1_element …))`, flat-encoded by hand: version words 01 01 00,
+/// then term tag `con` (0100), the type tag list [9] (1 1001 0), padded with zeroes. Decoding
+/// must fail on the *value*, mirroring the Haskell `Flat` instances for BLS elements (which
+/// always fail), before any value bytes are even read.
+#[test]
+fn bls_g1_element_value_fails_to_decode() {
+    let flat = [0x01, 0x01, 0x00, 0x4C, 0x80];
+    let arena = Arena::new();
+    let error = decode::<DeBruijn>(&arena, &flat, (11, 0)).expect_err("BLS element values must not decode");
+    assert!(error.to_string().contains("BLS12-381"), "unexpected error: {error}");
```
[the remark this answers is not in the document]
etorreborre: Done.

yHSJ:
Interesting, find and mildly annoying.

> What? We G1 and G2 points can now be encoded to flat; did we miss that in Van Rossem 🤨 ?

If we can now in Van Rossem, then yes we missed that. The last time I looked at this was before pv11.

CHANGELOG.md
```diff
@@ -71,6 +68,7 @@ Other guiding principles:
 - **amaru-consensus**: fix peer_selection to only schedule a single cool-down timer and thus properly bound priority mailbox usage. ([#1112](https://github.com/pragma-org/amaru/issues/1112))
 - **amaru-ledger**: validate provided treasury value matches actual treasury value (ConwayTreasuryValueMismatch) ([#1025][], [#888][])
 - **amaru-ledger**: reject governance proposals whose deposit return account is not registered. ([#928][])
+- **amaru-ledger**: fix the validation of Plutus scripts containing lists of BLS elements when they are empty, since the Haskell node accepts them ([#1159](https://github.com/pragma-org/amaru/issues/1159))
```
yHSJ: This should probably say amaru-uplc, for correctness.

crates/amaru-uplc/tests/decode_bls_containers.rs
yHSJ, line 1: I don't *love* adding test files like this, ideally we can represent them with fixtures that have the same shape as the Plutus conformance tests. That allows us to eventually upstream those tests too.

etorreborre: Good point. Let me do it.

etorreborre: Actually I kept the detailed unit tests because they are exactly at the level of what should be checked but I added 2 fixtures (valid / invalid) to test this kind of situation at the level of a transaction.

crates/amaru-uplc/tests/decode_bls_containers.rs
```diff
@@ -0,0 +1,189 @@
...
+        }
+    }
+
+    fn filler(&mut self) {
+        while self.used != 7 {
+            self.push(0, 1);
+        }
+        self.push(1, 1);
+    }
+}
```
KtorZ, line 189: This feels like a very brittle of writing UPLC programs; Hopefully the flat format doesn't change much that this doesn't require a lot of maintenance...

KtorZ (approved):

PR #1163, opened 2026-08-06
log build version on startup

### LINKED ISSUE #1161: running code version should be logged at startup
Getting a log file from a user is extremely helpful. It would be even better if that file included the code version that produced it so that we can be certain without having to ask.

rkuhn:
fixes #1161

## Summary by CodeRabbit

* **New Features**
  * Added build metadata logging after tracing starts, including version, Git commit, repository state, operating system, and architecture.
  * Added structured observability details to help identify the running binary and its build environment.

* **Documentation**
  * Documented the new build-version trace event and its schema.

* **Tests**
  * Added coverage verifying emitted build, platform, and Git metadata.

### DIFF
--- a/crates/amaru-observability/src/schemas.rs
+++ b/crates/amaru-observability/src/schemas.rs
@@ -1575,6 +1575,16 @@ define_schemas! {
                     required with_colors: bool
                 }
             }
+            build {
+                /// Running binary build/version identity (package version, git commit, target).
+                public VERSION {
+                    required version: String
+                    required git_commit: String
+                    required git_dirty: bool
+                    required os: String
+                    required arch: String
+                }
+            }
             trace {
                 /// Resolution of a trace filter from the environment
                 public FILTER {
--- a/crates/amaru/src/version.rs
+++ b/crates/amaru/src/version.rs
@@ -14,6 +14,8 @@
 
 use std::sync::LazyLock;
 
+use amaru_observability::info;
+
 mod built_info {
     include!(concat!(env!("OUT_DIR"), "/built.rs"));
 }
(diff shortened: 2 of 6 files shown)


crates/amaru/src/version.rs
```diff
@@ -58,3 +64,106 @@ pub fn target_os() -> &'static str {
...
+/// record which build produced them.
+pub fn log_build_version() {
+    info!(
+        setup::build::VERSION,
+        version = package_version(),
+        git_commit = git_commit_hash().unwrap_or("unknown"),
+        git_dirty = git_dirty().unwrap_or(false),
+        os = target_os(),
+        arch = target_arch(),
+        git_commit_short = git_commit_hash_short().unwrap_or("unknown"),
```
KtorZ:
```suggestion
```

pointless

crates/amaru-observability/src/schemas.rs
```diff
@@ -1575,6 +1575,17 @@ define_schemas! {
...
             }
+            build {
+                /// Running binary build/version identity (package version, git commit, target).
+                public VERSION {
+                    required version: String
+                    required git_commit: String
+                    required git_dirty: bool
+                    required os: String
+                    required arch: String
+                    optional git_commit_short: String
```
KtorZ:
```suggestion
```

crates/amaru/src/version.rs
rkuhn: Yeah, I kinda agreed, but then remembered that we’re emitting data more structured than strictly needed for easy consumption of other tools (that apparently can’t do substr() ;-) ) and I saw that the `amaru --version` uses the short hash. Will remove.

KtorZ:
Merged manually after resolving conflict.

PR #1166, opened 2026-08-07
Implement InvalidGuardrailsScriptHash

### LINKED ISSUE #931: Ledger predicate: InvalidGuardrailsScriptHash
### Abstract

Implement predicate `InvalidGuardrailsScriptHash`

### Why?

Conformance

### How?

[Reference Implementation](https://github.com/IntersectMBO/cardano-ledger/blob/0cfbf861cfb456660a7b73281c6fb714a53d40f9/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Gov.hs#L428-L434)

### Discussion points

_No response_

_No response_

### Checklist

yHSJ:
Closes #931

This PR introduce logic to validate that the guardrails script hash referenced in the `TreasuryWithdrawal` and `ParameterChange` gov actions matches the expected one defined in the on-chain constitution. It also requires some updates to the state logic to expose to the guardrails script hash, as well as changes to the fixture schema to allow test coverage for this logic

## Summary by CodeRabbit

* **New Features**
  * Governance proposals now require a policy matching the enacted constitution’s guardrails script.
  * Proposals are rejected when the script is missing, unexpected, or mismatched.
  * Added support for tracking guardrails scripts across constitution updates.

* **Bug Fixes**
  * Treasury withdrawal proposals now preserve and validate their policy correctly.

* **Documentation**
  * Documented the guardrails script in transaction schemas, test guidance, and the changelog.

### DIFF
--- a/crates/amaru-ledger/src/state/volatile/overlay.rs
+++ b/crates/amaru-ledger/src/state/volatile/overlay.rs
@@ -15,8 +15,8 @@
 use std::{cell::RefCell, collections::BTreeMap, mem, sync::Arc};
 
 use amaru_kernel::{
-    ConstitutionalCommitteeUpdate, Epoch, Lovelace, PoolId, ProposalId, ProposalsRoots, ProtocolParameters,
-    RatificationStatus, StakeCredential, TreasuryDelta,
+    Constitution, ConstitutionalCommitteeUpdate, Epoch, Hash, Lovelace, PoolId, ProposalId, ProposalsRoots,
+    ProtocolParameters, RatificationStatus, StakeCredential, TreasuryDelta, size::SCRIPT,
 };
 use amaru_observability::{debug, info_span};
 use tracing::Span;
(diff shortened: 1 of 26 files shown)


crates/amaru-ledger/src/state/volatile/overlay.rs
```diff
@@ -345,6 +349,11 @@ impl StateOverlay {
         self.governance_updates.as_ref().map(|update| &update.protocol_parameters)
     }
 
+    /// The constitution enacted by an in-flight epoch transition, if that transition enacts one.
```
rkuhn, line 351: Do I assume correctly that this returns `Some(_)` from the first block of the new epoch to k blocks later? (might be worth clarifying)

rkuhn (commented):
looks reasonable to me

crates/amaru-ledger/src/state/volatile/overlay.rs
yHSJ: Yes, that is an effect of the design of our volatile db. The `overlay` exists to provide a view of the state *after* an epoch transition, but before it is persisted to disk.

So, if there is a new constitution enacted at an epoch transition, this will return `Some(_)`. I wouldn't say that it ever *stops* returning that value, it's more that the overlay is removed once the state is applied to disk.

PR #1172, opened 2026-08-07
test: fix the serialization conformance tests gaps

### LINKED ISSUE #970: Fix the serialization test cases that are currently non-conformant
### Abstract

The PR https://github.com/pragma-org/amaru/pull/969 introduced test cases showing that `amaru` does not have full correct encoders/decoders. This issue is about fixing all those cases.

### Why?

To make sure that we inter-operate correctly with other nodes in the network.

### How?

By fixing the cases one by one. Some analysis will need to be done in order to understand if this issue needs to be further divided by sub-issues for similar cases.

All the tests should pass. Every fixed fixture should be modified to show that it's not divergent anymore.

### Discussion points

_No response_

_No response_

### Checklist

etorreborre:
This PR fixes the conformance gaps tracked by the `test_cbor_serialization` test.

The previous count, executed with `cargo test -p amaru-kernel --test test_cbor_serialization -- --no-capture`, was reporting:
```
cbor fixtures: 137 positive, 142 negative, 279 total (132 non-conformant)
```

The major changes are:

* The decoding of `metada` maps as association lists, not maps.
* The decoding of the protocol version (a new type is introduced to support this).
* The decoding of arrays to support definite and indefinite lengths.
* The decoding of strings with a maximum size.
* The decoding of `ScriptNOfK` native scripts.
* The decoding of lists of bytes to account for definite or indefinite lengths.
* The decoding of specific types like `AssetName`, `VKey`, `Ed25519Signature` that are lists of bytes with size constraints (the `MaxBytes` and `FixedBytes` types have been introduced for this).
* The encoding of auxiliary data, to properly account for Plutus versions.

The test framework itself has been fixed or improved:

* We can execute a test by fixture name.
* A flag specifies if the fixture supports re-encoding (decoding -> encoding = original bytes). This will evolve once we rely on a canonical CBOR check.
* The Conway `cddl` is supplemented with an `override` file to specify more precisely valid addresses and pool metadata. This might change in the future by using the same tool chain as the Haskell node with the `huddle` DSL on top of the cddl generation.
* After mutation to generate invalid cases, a cddl validator is used to check if we don't end-up by accident on a valid fixture (with this pass an invalid case will really be invalid, a valid one might not be since we're only using cddl constraints). The number of relabeled cases tell us if we generated enough invalid cases.

NOTE: this only fixes the gaps determined by the current set of generated fixtures. There are other gaps, some of them identified by `FIXME` comments that are still not covered by this PR.
The next step is specified by https://github.com/pragma-org/amaru/issues/1174.


etorreborre:
@coderabbit review

etorreborre:
@CodeRabbit review

yHSJ (approved):
I've reviewed the rust code specifically, and went through the decoders with the CDDL  pulled up to validate that they are implemented correctly.

This seems like a positive change and we are moving in the right direction. I am a bit curious about the behavior of `MaxBytes` given it pads. That does spook me *a bit*, given it could lead to some very frustrating sneaky edge cases, but not enough to push back entirely on it's existence.

crates/amaru-kernel/src/cardano/certificate.rs
```diff
@@ -189,136 +189,143 @@ impl<C> cbor::encode::Encode<C> for Certificate {
...
-                let b = d.decode_with(ctx)?;
-                Ok(Self::UpdateDRepCert(a, b))
-            }
-
-            _ => Err(cbor::decode::Error::message("unknown variant id for certificate")),
-        }
+        // NOTE: the array length is not asserted here.
+        //
+        // Each variant carries a different number of fields, so validating the length means
+        // asserting it per variant. That is worth doing, but it is a separate change; what matters
```
rkuhn: It was my understanding that we have no choice regarding length validation: either Haskell does it, then we must do it, or Haskell doesn’t, then we must not do it. Shouldn’t the fuzzing tests tell us which one it is?

crates/amaru-kernel/src/cardano/max_bytes.rs
```diff
@@ -0,0 +1,166 @@
...
+
+use std::{fmt, ops::Deref, str::FromStr};
+
+use crate::{Bytes, cbor};
+
+/// A list of bytes whose length is bounded by the given maximum.
+///
+/// `Eq`, `Ord` and `Hash` are written by hand rather than derived, because the array is padded past
+/// `len` and a derived impl would compare the padding: `b"a"` and `b"a\0"` share the same backing
+/// array and would wrongly compare equal.
```
rkuhn: Sure, the arrays are equal, but the `len` field would differ, so it should give the correct result, should it not? Unless, of course, the padding can be something other than zero.

crates/amaru-kernel/src/cardano/max_bytes.rs
```diff
@@ -0,0 +1,166 @@
...
+#[derive(Debug, Clone, Copy, serde::Serialize, serde::Deserialize)]
+#[serde(into = "String")]
+#[serde(try_from = "String")]
```
rkuhn: Since we’re using serde also for binary encoding, I’m not fond of this scheme. I’d rather encode `.as_slice()` and let the encoder worry about the representation.

crates/amaru-kernel/src/cardano/max_bytes.rs
```diff
@@ -0,0 +1,166 @@
...
+    fn try_from(value: String) -> Result<Self, Self::Error> {
+        Self::checked(&hex::decode(value)?)
+    }
+}
+
+impl<const N: usize> FromStr for MaxBytes<N> {
+    type Err = MaxBytesError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        Self::try_from(s.to_string())
```
rkuhn: This is the wrong way around: `hex::decode` here and use that in `try_from`.

crates/amaru-kernel/src/cardano/max_bytes.rs
```diff
@@ -0,0 +1,166 @@
...
+
+impl<const N: usize> From<MaxBytes<N>> for Bytes {
+    fn from(bytes: MaxBytes<N>) -> Self {
+        bytes.to_vec().into()
+    }
+}
+
+impl<const N: usize> From<MaxBytes<N>> for String {
+    fn from(bytes: MaxBytes<N>) -> Self {
+        hex::encode(bytes.to_vec())
```
rkuhn: this should use `.as_slice()`

crates/amaru-kernel/src/cardano/max_bytes.rs
```diff
@@ -0,0 +1,166 @@
...
+    type Err = MaxBytesError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        Self::try_from(s.to_string())
+    }
+}
+
+impl<const N: usize> fmt::Display for MaxBytes<N> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str(&hex::encode(self.to_vec()))
```
rkuhn: this should use `.as_slice()`

crates/amaru-kernel/src/cardano/max_string.rs
```diff
@@ -0,0 +1,86 @@
...
+            Ok(Self(value))
+        }
+    }
+}
+
+impl<const MAX: usize> FromStr for MaxString<MAX> {
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        Self::try_from(s.to_string())
```
rkuhn: similar here: express `try_from` in terms of `from_str` to avoid needless allocation

crates/amaru-kernel/tests/test_cbor_serialization.rs
```diff
@@ -32,27 +47,32 @@ fn test_cbor_serialization() {
...
+    let mut antigen_accepted: Vec<(&Fixture, String)> = Vec::new();
+    let mut other_divergence: Vec<(&Fixture, String)> = Vec::new();
+
     for fixture in &fixtures {
         if fixture.expectations.well_formed {
-            positive += 1;
+            positive += 1
         } else {
-            negative += 1;
+            negative += 1
```
rkuhn, line 59: ?

rkuhn (commented):
This is a lot, and I skimmed some parts. One thing I wonder: do we really need to have all those tiny files in our git history? Could we somehow zip them in useful bundles?

crates/amaru-kernel/src/cardano/vkey.rs
```diff
@@ -0,0 +1,17 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::cardano::{fixed_bytes::FixedBytes, hash::size::VRF_KEY};
+
+pub type VKey = FixedBytes<VRF_KEY>;
```
KtorZ:
```suggestion
pub type VerificationKey = FixedBytes<VRF_KEY>;
```

I am in a personal vendetta against short(er than necessary) names. We have auto-completion, we have AI, there's no need to be cheap with words!

crates/amaru-kernel/src/cardano/memoized/datum.rs
```diff
@@ -96,7 +97,7 @@ impl<'b, C> cbor::Decode<'b, C> for MemoizedDatum {
                     if d.tag()? != IanaTag::Cbor.tag() {
                         return Err(cbor::decode::Error::message("unknown tag for datum tag"));
                     }
-                    let plutus_data: MemoizedPlutusData = cbor::decode_with(d.bytes()?, ctx)?;
```
KtorZ, line 99: We should probably consider adding a clippy rules for these; since it's easy to miss.

crates/amaru-kernel/tests/data/cbor.decode/block/0df40008e40348c40cdc3b92a1e31d0e55675ddf2bb05ff7683b38a837048bca/meta.json
KtorZ, line 1: We should likely mark all those files as generated in the .gitattributes, to get them out of the diff because no one is reviewing this 😅 ...

KtorZ (approved):
I've looked at the decoders one by one and that looks _plausible_. I still see a couple of FIXME and stuff that likely needs revision afterwards but I believe it's left out on purpose.

crates/amaru-kernel/src/cardano/vkey.rs
etorreborre: I actually agree, just didn't want to go around possible previous conventions.

crates/amaru-kernel/tests/test_cbor_serialization.rs
etorreborre: Why not remove it?

crates/amaru-kernel/src/cardano/max_bytes.rs
etorreborre: Done.

crates/amaru-kernel/src/cardano/max_bytes.rs
etorreborre: Done

crates/amaru-kernel/src/cardano/max_string.rs
etorreborre: Done.

crates/amaru-kernel/src/cardano/max_bytes.rs
etorreborre: Done.

crates/amaru-kernel/src/cardano/max_bytes.rs
etorreborre: Fixed by using derived implementations and commenting that the array is always 0-padded.

crates/amaru-kernel/src/cardano/max_bytes.rs
etorreborre: Done.

crates/amaru-kernel/src/cardano/memoized/datum.rs
etorreborre: Adding this tentatively on my todo list.

crates/amaru-kernel/src/cardano/certificate.rs
etorreborre: Haskell checks it and I fixed it. Unfortunately this doesn't show up yet in the generated tests. We need to work on the generation to improve that aspect.

crates/amaru-kernel/tests/data/cbor.decode/block/0df40008e40348c40cdc3b92a1e31d0e55675ddf2bb05ff7683b38a837048bca/meta.json
etorreborre: Good idea.

etorreborre:
@KtorZ yes the FIXMEs will be addressed by https://github.com/pragma-org/amaru/issues/1174.

crates/amaru-kernel/tests/test_cbor_serialization.rs
rkuhn: It is very unidiomatic, I’m surprised rust-fmt and clippy don’t scream. There are lints about returning `()` in certain circumstances.

PR #1178, opened 2026-08-07
Run phase two validations in ledger conformance tests

### LINKED ISSUE #894: Ledger predicate: ValidationTagMismatch
### Abstract

Implement predicate: `ValidationTagMismatch`

### Why?

Conformance

### How?

[Reference Implementation](https://github.com/IntersectMBO/cardano-ledger/blob/0cfbf861cfb456660a7b73281c6fb714a53d40f9/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs#L224) (`isValid = true`)

[Reference Implementation](https://github.com/IntersectMBO/cardano-ledger/blob/0cfbf861cfb456660a7b73281c6fb714a53d40f9/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/Rules/Utxos.hs#L260) (`isValid = false`)

### Discussion points

_No response_

_No response_

### Checklist

yHSJ:
Closes #894

In order to cover the `ValidationTagMismatch` logic in our fixtures, we had to support running phase two validations in the test suite. This is a logical thing to do, regardless of that specific predicate.

This PR is mostly mechanical; renaming from the suite of fixtures `phase-one` to `transaction` (and updating related text that depends on the file structure). Mixed into those mechanical changes are some new fixtures to provide coverage of the required logic. There is also a bug fix that I snuck in there, which handles an edge case when there are no provided scripts but the transaction is marked as invalid.


crates/amaru-ledger/src/rules/transaction/phase_one/mod.rs
```diff
@@ -358,86 +355,3 @@ fn fail_on_network_mismatch(provided: Option<Network>, expected: Network) -> Res
...
-            (Expected::Pass, Ok(_)) => (),
-            (Expected::Pass, Err(actual)) => panic!("expected pass, got error: {actual:?}"),
-            (Expected::Fail(expected), Err(actual)) => {
-                assert_eq!(actual, expected, "expected {expected:?}, got {actual:?}");
-            }
-            (Expected::Fail(expected), Ok(_)) => panic!("expected fail ({expected:?}), got pass"),
-            (Expected::DecodingFailure, _) => unreachable!("handled before decoding the transaction"),
-        }
-    }
-}
```
rkuhn, line 447: would be nice if gh could show the diff with the other file you moved this code to

rkuhn (commented):
Looks reasonable to me, but I can’t check the actual details.

PR #1182, opened 2026-08-09
taking a stab at structured logging

rkuhn:
There is already quite some structured logging that we can do, with the notable exception of emitting structured attributes (`AnyValue`) in OTLP spans--here the `opentelemetry` crate is not yet far enough.

This PR is a proposal, I tried to push Grok as hard as I could to get it working, and I fully admit that I don’t understand all of this PR because I didn’t take the time to dive into tracing/otel/layer infrastructure. The state of affairs as far as I checked is:

- normal console logging prints CBOR diagnostic for structured fields (mostly looks like JSON)
- JSON traces have proper JSON objects, arrays, etc. when applicable (we don’t use that yet, but I tried it)
- OTel outputs emit arrays of primitives natively, all other structured fields are byte strings (not currently used, though)

The TUI also needed small tweaks to receive values correctly, and I rate-limited the initial stake distribution when using JSON traces (was emitted thousands per second).

---

## Summary by CodeRabbit

* **New Features**
  * Structured logging now supports complex values across JSON, console, and OpenTelemetry outputs.
  * Nested objects, arrays, numeric values, and other serialized fields are rendered with improved type fidelity.
  * Invalid or unsupported values receive readable diagnostic fallbacks.
  * OpenTelemetry logs and trace attributes now preserve structured field information where supported.

* **Bug Fixes**
  * Stake-distribution progress messages are throttled to reduce excessive console output.
  * TUI telemetry displays decoded numeric values using their correct types.


crates/amaru-observability/src/field.rs
```diff
@@ -0,0 +1,455 @@
...
+pub const DIAG_TRUNCATION_LIMIT: usize = 512;
+
+/// Serialize `value` as CBOR for transport through `tracing` as `record_bytes`.
+///
+/// Returns an owned `Box<[u8]>` so the result implements [`tracing::Value`] via
+/// `Box<[u8]>` → `[u8]` → `record_bytes`.
+pub fn encode_cbor<T: Serialize>(value: &T) -> Box<[u8]> {
+    let mut buf = Vec::new();
+    // Encoding failures are treated as empty CBOR null so instrumentation never panics.
+    if cbor4ii::serde::to_writer(&mut buf, value).is_err() {
```
KtorZ, line 45: Not fond of this one; can we stick to `minicbor` and the hand-written cbor definitions instead? This otherwise create a weird spot where we trace serde instances (which are often auto-derived) as CBOR instead of using the actual CBOR shape we may be used to see.

crates/amaru-observability/src/field.rs
```diff
@@ -0,0 +1,455 @@
...
+}
+
+/// Decode CBOR into a **trace** attribute [`TraceValue`] when the payload fits the
+/// classic span attribute model: a scalar or a **homogeneous** array of bool / i64 / f64 / string.
+///
+/// Returns [`None`] for maps, mixed arrays, byte strings, or invalid CBOR — callers should
+/// leave those to the stock `tracing-opentelemetry` string/hex fallback (or log-side `AnyValue`).
+///
+/// This is what the current Rust trace API can express without waiting for complex attributes.
+pub fn cbor_to_trace_value(bytes: &[u8]) -> Option<TraceValue> {
```
KtorZ: That `None` part will be triggered at runtime, not compile time right? So we may accidentally shoot oneself in the foot with blind traces if we happen to give a value that isn't convertible (e.g. a tuple as an array, a record, ... ). 

Shouldn't this function or the parent `cbor_to_json` handles the fallback to cbor diagnostic directly to ensure there's at least something logged?

crates/amaru-observability/src/field.rs
rkuhn: Logging should work with all the types we use, which is basically the same set as types sent between pure-stages, which is all based on serde, which of course is the ecosystem standard for serialization. `minicbor` only exists for network data types in Ouroboros, which is a different (and smaller) set of types. We could choose a different serde-cbor library (there are several, cbor4ii was most suitable at the time I designed pure-stage), but we should definitely use serde for derivation.

KtorZ: > which is basically the same set as types sent between pure-stages

Not at all; there's a larger superset from everything logged by the ledger. Currently, those are broken down to int/string/bool but we introduce the ability to log complete values, we might revise that. 

> We could choose a different serde-cbor library (there are several, cbor4ii was most suitable at the time I designed pure-stage), but we should definitely use serde for derivation.

I just find the discrepancy potentially annoying. Maybe it's only me. Most people won't be familiar with the underlying on-chain CBOR representation anyway, so a diagnostic or another would mean the same.

crates/amaru-observability/src/field.rs
rkuhn: As I said, I’m not 100% certain that I understand all the layers: this is used in an extra layer to log the `TraceValue` in addition to whatever the generic layers have already done. So yes, if you log a structured value that does not fit the bill then it will be logged only as CBOR hex string — we’ll notice that as soon as we start exercising these new facilities.

KtorZ (approved):
After discussing it on a call; we agreed that using serde for cbor-serialization in the observability layer may sometimes yield different cbor representations than those on the wire but, the complexity and annoyance of dealing with minicbor at this level is not worth the effort. 

Incidentally, direct cbor-encoded values in the observability layer should be the exception rather than the norm, so that is likely not a major concern.

crates/amaru-observability/src/otel_trace_arrays.rs
```diff
@@ -0,0 +1,193 @@
...
+    fn on_enter(&self, id: &tracing::span::Id, ctx: tracing_subscriber::layer::Context<'_, S>) {
+        let Some(span_ref) = ctx.span(id) else {
+            return;
+        };
+        let pending = span_ref.extensions_mut().remove::<PendingTraceAttributes>();
+        if let Some(pending) = pending {
+            // Entering makes this the current span.
+            apply_pending(&tracing::Span::current(), pending);
+        }
+    }
```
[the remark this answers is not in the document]
rkuhn: This is unfortunately not possibly with the existing API because the Span is not available.

crates/amaru-observability/src/json_format.rs
```diff
@@ -0,0 +1,497 @@
...
+};
+
+use crate::field::cbor_to_json;
+
+/// Root object keys reserved by the envelope; span fields must not overwrite them.
+const RESERVED_ROOT_KEYS: &[&str] = &["timestamp", "level", "fields", "target", "id", "parent_id", "span", "spans"];
+
+fn is_reserved_root_key(key: &str) -> bool {
+    RESERVED_ROOT_KEYS.contains(&key)
+}
```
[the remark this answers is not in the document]
rkuhn: I don’t get it: this reserved list is only forbidden for user fields, precisely because the layer does emit these properties.

crates/amaru-observability/src/otel_trace_arrays.rs
```diff
@@ -0,0 +1,209 @@
...
+        }
+        if let Some(span_ref) = ctx.span(id) {
+            let mut ext = span_ref.extensions_mut();
+            if let Some(pending) = ext.get_mut::<PendingTraceAttributes>() {
+                pending.attrs.extend(visitor.attrs);
+            } else {
+                ext.insert(PendingTraceAttributes { attrs: visitor.attrs });
+            }
+        }
+    }
```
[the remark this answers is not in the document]
rkuhn: Calling `.record` while the `Span` is not current is not only weird, it should be unsupported — what does that even mean?

PR #1189, opened 2026-08-09
push forwarded embedded Amaru

rkuhn:
This is an experiment in advancing the embedding of Amaru as a library. Certainly not yet perfect,
but the potential is demonstrated.

@KtorZ please review the EDR contained in the PR in light of the TUI as well as Kupo. The facilities
offered should be somewhat powerful while hopefully not burdening the node too much; simple
processing can happen within the ledger to profit from cache locality, more costly processing can be
moved to another thread.

## Summary by CodeRabbit

* **New Features**
  * Added cold-start bootstrapping from Cardano snapshot archives, including validation, downloading, and state restoration.
  * Added an embeddable node builder with configurable storage, peers, listeners, metrics, and lifecycle control.
  * Added ledger observers for adopted blocks, undo events, and ledger snapshots.
  * Added structured telemetry capture and configurable logging/export.
  * Added address-watching and run-until node examples.

* **Improvements**
  * Updated the run-until script to use the node runner directly.
  * Improved telemetry integration across the node and TUI.
  * Added support for importing snapshots from local directories.


crates/amaru-ledger/src/state.rs
```diff
@@ -1125,7 +1150,15 @@ impl<HS: HistoricalStores> BackgroundTasks<HS> {
             .unwrap_or(true);
 
         if should_push_summary {
-            let distr = compute_stake_summary(&snapshot, self.network, &self.era_history, |_| {})?.stake_distribution;
+            let summary = compute_stake_summary(&snapshot, self.network, &self.era_history, |_| {})?;
+
+            // Opt-in: hand the full summary (incl. accounts) to the embedder before we only
+            // retain the slim in-memory distribution.
+            if let Some(notify) = &self.on_ledger_snapshot {
+                notify(summary.clone());
```
KtorZ: 😬

crates/amaru-ledger/src/summary/stake_distribution.rs
```diff
@@ -37,8 +37,7 @@ use crate::{
 ///
 /// Note that the `accounts` field only contains _active_ accounts; that is, accounts
 /// delegated to a registered stake pool.
-#[derive(Debug)]
-#[cfg_attr(test, derive(Clone))]
+#[derive(Debug, Clone)]
```
KtorZ: Nope; I don't like that.

crates/amaru-ledger/src/observers.rs
```diff
@@ -0,0 +1,103 @@
...
+/// in-memory [`crate::summary::stake_distribution::StakeDistribution`].
+///
+/// Opt-in only: the outer app owns the copy and may index or drop it.
+pub type LedgerStateSnapshot = StakeSummary;
+
+/// Callbacks installed on ledger [`crate::state::State`].
+#[derive(Clone, Default)]
+pub struct LedgerObservers {
+    /// Invoked after each successful block application at the tip (roll-forward path).
+    pub on_adopted_block: Option<Arc<dyn Fn(AdoptedBlock) + Send + Sync>>,
```
KtorZ: I am really not sure about this. That plus the cloning of EVERY block and stake distribution. This isn't the primary use case and we don't have enough benchmarks in place to properly assess the impact here. 

Maybe behind a feature flag? But as such, it am afraid it becomes quite some unnecessary overhead.

KtorZ (dismissed):

crates/amaru-ledger/src/observers.rs
rkuhn: There is no overhead unless someone actually installs an observer — what am I missing?

crates/amaru-ledger/src/summary/stake_distribution.rs
rkuhn: This is required, it’s basic physics, so you’ll need to get over that emotion.

rkuhn:
So you don’t like the cloning. There are exactly two choices when it comes to handing data to an observer:

- the observer can see it while the ledger is showing it (which keeps the ledger from doing other things)
- the observer owns the data and can process it independently of what the ledger is doing

The currently proposed API forces option two. As a compromise we could also hand references to the observer, who can then decide whether to block the ledger for some time or _clone the data_ and send it to another thread for further processing. I hope this explains why the `impl Clone` must be there in any case.

crates/amaru-ledger/src/summary/stake_distribution.rs
KtorZ: I don't think it is required no; so far we have managed to do everything _without this_

crates/amaru-ledger/src/observers.rs
KtorZ: Hmmm. Maybe I misunderstood (review from mobile phone are not great), but I assumed that the main execution loop also switched to using this pattern. Now if this is done _optionally_ on top of the ledger main thread, I guess it is acceptable.

KtorZ:
> So you don’t like the cloning. There are exactly two choices when it comes to handing data to an observer:
>
> 1. the observer can see it while the ledger is showing it (which keeps the ledger from doing other things)
> 2. the observer owns the data and can process it independently of what the ledger is doing

Interestingly enough, I much prefer option (2), and not because of cloning; but because from experience, it is incredibly useful to be able to control the execution of the node when building indexer and services on top. Being able to pause and resume can save a lot of headache. I wish this was possible with the Cardano node. 

For the cloning, I would still not allow it on the stake distribution. Every field is clonable and/or iterable easily; so a consumer can choose to clone them if needs be. Having a global clone on this object just makes it _too easy_ to clone it accidentally due to Rust magic and automatic cloning from time to time. Cloning this should require some effort and explicit desire; not be an unfortunate accident.

crates/amaru-ledger/src/observers.rs
```diff
@@ -0,0 +1,125 @@
...
+//! Optional hooks for embedders observing ledger progress.
+//!
+//! Observers are **outside** the consensus stage graph: the node never stops
+//! itself based on these callbacks. The outer application decides lifecycle.
+//!
+//! # Semantics
+//!
+//! - Callbacks receive **short-lived references** and run on the ledger path.
+//!   Long work or panics delay or abort ledger progress; clone needed pieces and
+//!   hand them to another task if processing is non-trivial.
```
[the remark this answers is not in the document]
rkuhn: panic will kill the process, we never defend against it

PR #1190, opened 2026-08-10
fix: lift restrictive assertion on the fork length when switching to a fork

### LINKED ISSUE #1162: Bug: validation failure followed by illegal fork switch
from Discord:
```
2026-08-05T22:21:19.909270Z  INFO epoch_transition.compute: amaru::ledger: close time.busy=45.9ms time.idle=2.57µs [3mfrom[0m[2m=[0m302 [3minto[0m[2m=[0m303
2026-08-05T22:21:20.417887Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129256534 tip.hash=641d4e5935c49e2c5dff6858373d1adb14376f9f20323676fa8b5ca3a2c7b0e5 tip.block_height=4975391 max_block_height=4975602 suppressed=229
2026-08-05T22:21:21.431391Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129261359 tip.hash=1d92fd19adc07189f3bd5bacfdb939c3f6ed76d6a6cbd5abfe35d403c01edab5 tip.block_height=4975598 max_block_height=4975837 suppressed=206
2026-08-05T22:21:22.439506Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129265882 tip.hash=507675a6b64dba74e6aa925bc852671a6d173fe013b86a5b571f964edd7b7a0d tip.block_height=4975782 max_block_height=4976112 suppressed=183
2026-08-05T22:21:23.442182Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129269833 tip.hash=3d9010d41513bc7a25413c34aa72d50ea4745b8c4ec587a0766c06ba4e099b73 tip.block_height=4975960 max_block_height=4976353 suppressed=177
2026-08-05T22:21:24.442502Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129273761 tip.hash=a8b0db35beae71cda7977600263409829ab19e67aec6e3b7656b897436523c5a tip.block_height=4976140 max_block_height=4976597 suppressed=179
2026-08-05T22:21:25.447422Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129277933 tip.hash=54760e1208994f30f929b837064f3335c10226cb4bd5856f630832e831476b9b tip.block_height=4976311 max_block_height=4976855 suppressed=170
2026-08-05T22:21:26.452876Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129281544 tip.hash=d6e3735a39aad45f5be671a892dde1257006d992c7d8a380769dd950709ba71b tip.block_height=4976480 max_block_height=4977071 suppressed=168
2026-08-05T22:21:27.462802Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129284967 tip.hash=55f28875fde22f1468bfa9b8936b8f70d5047eddd92717d08d732c34a7083e07 tip.block_height=4976657 max_block_height=4977363 suppressed=176
2026-08-05T22:21:28.463515Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129288677 tip.hash=b26004bdcf6a28efc5b20c0d00e37f37b90830b4e591d5fb4811ad3cc3e4ae50 tip.block_height=4976820 max_block_height=4977600 suppressed=162
2026-08-05T22:21:29.473666Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129293527 tip.hash=369feddd49b9e36c8e0f0fe427ce86408cb4b9475d9c23679817344823e6d852 tip.block_height=4977021 max_block_height=4977851 suppressed=200
2026-08-05T22:21:30.533566Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129298921 tip.hash=3a647b19ab4735a91dee84db67da4d39f6e27b01e6b1da055c401ac666843717 tip.block_height=4977232 max_block_height=4978115 suppressed=210
2026-08-05T22:21:31.542490Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129303513 tip.hash=10ab6581483104c9622c10e2cb05612a4470c99d7d6cb5be7d80ed001545349e tip.block_height=4977435 max_block_height=4978330 suppressed=202
2026-08-05T22:21:31.784357Z  INFO epoch_transition.apply:ledger.epoch.create_snapshot: amaru::stores: close time.busy=142ms time.idle=7.75µs [3mepoch[0m[2m=[0m303 [3mshould_end_epoch[0m[2m=[0mtrue [3mshould_snapshot[0m[2m=[0mtrue [3mepoch[0m[2m=[0m302
2026-08-05T22:21:31.860595Z  INFO epoch_transition.apply: amaru::ledger: close time.busy=270ms time.idle=2.55µs [3mepoch[0m[2m=[0m303 [3mshould_end_epoch[0m[2m=[0mtrue [3mshould_snapshot[0m[2m=[0mtrue [3mshould_begin_epoch[0m[2m=[0mtrue
2026-08-05T22:21:31.861280Z  INFO ledger.epoch.prune_old_snapshots: amaru::stores: close time.busy=650µs time.idle=3.87µs [3mfunctional_minimum[0m[2m=[0m300 [3mdesired_minimum[0m[2m=[0m300
2026-08-05T22:21:32.552957Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129307000 tip.hash=3eb31d0200aca7c3c0317b695671fe45fdd636e06567f706f3225f4a9fbaa2fc tip.block_height=4977601 max_block_height=4978507 suppressed=165
2026-08-05T22:21:33.554741Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129311533 tip.hash=80d9f0a5a66a9fa6d2991f4d4775a12fd126d6544fdb887bacd9f8eb435aae82 tip.block_height=4977807 max_block_height=4978724 suppressed=205
2026-08-05T22:21:34.614841Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129316655 tip.hash=b0e1b6bd6387d8e428c2a9a301205151c939784c244df4da747cae53886c0bef tip.block_height=4978032 max_block_height=4978910 suppressed=224
2026-08-05T22:21:35.686690Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129322344 tip.hash=463c8905ab87fef75efab206b82b41bd0f197b2fa3effacce15d975fcc9d6cc5 tip.block_height=4978257 max_block_height=4979134 suppressed=224
2026-08-05T22:21:36.760237Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129327575 tip.hash=4bd16e62fd0753c019d4a6908590d94f309587ba2d333575189e32d5b476180a tip.block_height=4978482 max_block_height=4979359 suppressed=224
2026-08-05T22:21:37.766100Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129331917 tip.hash=45212ceb4363f1518a2e8ef4b8a6599cced5c581fde40b78776dfbc143f04bc1 tip.block_height=4978665 max_block_height=4979554 suppressed=182
2026-08-05T22:21:38.784313Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129335537 tip.hash=46a777b6bae991932ca45cfcc188eeac134032b7ed78313a231dae10933ab5e5 tip.block_height=4978832 max_block_height=4979767 suppressed=166
2026-08-05T22:21:39.788300Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129339794 tip.hash=377d3097e89a16155547097453adca1d4cf25406f2778ae55d5695c19c94c098 tip.block_height=4979025 max_block_height=4979959 suppressed=192
2026-08-05T22:21:40.793283Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129344080 tip.hash=132ca62c14d4fcd25252a1ad7f6f818096502555bb372afa72c4d6f09a2691a4 tip.block_height=4979230 max_block_height=4980099 suppressed=204
2026-08-05T22:21:41.793930Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129349085 tip.hash=0f891269253501f1b0d0f02ce33bf11a405c8210bb0dc4b891f25a24205ad4d0 tip.block_height=4979454 max_block_height=4980307 suppressed=223
2026-08-05T22:21:42.799344Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129354951 tip.hash=3ab5f4b1bfeb1d29972d3acf8e7e2e85afd4cc53c6e4cbbb8e86069cc538416b tip.block_height=4979701 max_block_height=4980512 suppressed=246
2026-08-05T22:21:43.800624Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129360489 tip.hash=3ce725090e5cefb7d0c5dcf4c0e6eff102054166009ba02d619d9a19e851a078 tip.block_height=4979966 max_block_height=4980777 suppressed=264
2026-08-05T22:21:44.802751Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129366222 tip.hash=0bbc75941181a62d0a662fdb574fcd3b4265527f1086b836313a326922b8a40a tip.block_height=4980221 max_block_height=4981032 suppressed=254
2026-08-05T22:21:45.812923Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129370316 tip.hash=34456c85d3a786804e217f7e4fa83d16cbdbf6ef5854b1f34553ff165da4eafc tip.block_height=4980425 max_block_height=4981272 suppressed=203
2026-08-05T22:21:46.814019Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129374463 tip.hash=b49689526c916b28a0722167eb2d3ce0e3509ea8f6be03ae641a1cc8c3921470 tip.block_height=4980637 max_block_height=4981505 suppressed=211
2026-08-05T22:21:47.817407Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129378186 tip.hash=128f6cc366a2993f0d7b1d59288d501840e26cf849e621a928f8f37eac912f00 tip.block_height=4980809 max_block_height=4981712 suppressed=171
2026-08-05T22:21:48.825999Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129382644 tip.hash=90594436cf633aa0d79f4b1b6b44585adb9e40c2efcfdc480b956e39d9ce887d tip.block_height=4981012 max_block_height=4981906 suppressed=202
2026-08-05T22:21:49.342701Z  INFO rewards.compute: amaru::ledger: rewards.summarize efficiency=19123/21600 incentives=34530570879086 treasury_tax=6910268478399 total_rewards=34551342391998 available_rewards=27641073913599 effective_rewards=401347345321 pots_reserves=13001103923517552 pots_treasury=1938281224706953 pots_fees=20771512912 [3mfor_epoch[0m[2m=[0m303
2026-08-05T22:21:54.362207Z  INFO stake_distribution.compute: amaru::ledger: stake_distribution.snapshot accounts=33169 dreps=279 pools=554 active_stake=1627117252973386 pools_voting_stake=1627119252973386 dreps_voting_stake=926885521633368 [3mepoch[0m[2m=[0m302
2026-08-05T22:21:54.362258Z  INFO stake_distribution.compute: amaru::ledger: close time.busy=5.02s time.idle=7.71µs [3mepoch[0m[2m=[0m302
2026-08-05T22:21:54.362278Z  INFO amaru::ledger: stake_distribution.rotate available_stake_distributions=302, 301
2026-08-05T22:21:54.372886Z  INFO rewards.compute: amaru::ledger: close time.busy=166ms time.idle=5.03s [3mfor_epoch[0m[2m=[0m303
2026-08-05T22:21:54.376891Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129384056 tip.hash=b6a0c5c98d803118fbe01aee1df395eced4c8bfcadee29a1a7ae58fad3787e97 tip.block_height=4981068 max_block_height=4981989 suppressed=55
2026-08-05T22:21:54.377427Z  WARN amaru_consensus::stages::fetch_blocks: timeout fetching blocks req_id=44871
2026-08-05T22:21:55.378543Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129387894 tip.hash=67913fa8c829f9e110b62c79e18ec92a5cb0162cc7741b497768d20e068ca30c tip.block_height=4981237 max_block_height=4982192 suppressed=168
2026-08-05T22:21:56.384648Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129392255 tip.hash=09031753b166b2b51564981d64f6d43a0cd79dfa7983cc2756016a0f14e9f3d9 tip.block_height=4981448 max_block_height=4982324 suppressed=210
2026-08-05T22:21:57.394262Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129396548 tip.hash=78400ca954d71f2a894b0e85aa5ecce7c562112ae75d976bb36d0768ad815429 tip.block_height=4981649 max_block_height=4982535 suppressed=200
2026-08-05T22:21:58.398091Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129400890 tip.hash=20a4053d24c828f4bcfdc1f64ceb1d683b9d5863b3564255865b091530ea37b6 tip.block_height=4981858 max_block_height=4982757 suppressed=208
2026-08-05T22:21:59.424569Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129408569 tip.hash=1d178619412bffbc19cba04904b50720a051ca26e4401c33106ad4531fef35b5 tip.block_height=4982215 max_block_height=4982959 suppressed=356
2026-08-05T22:22:00.429520Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129411967 tip.hash=68da91271f6a1019ed2528441fe495cdf226a1b2cccc632df5deb1213be72e9a tip.block_height=4982354 max_block_height=4983160 suppressed=138
2026-08-05T22:22:01.431517Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129415353 tip.hash=5129a95c6b8e7763ec36386652e6de455559039e24528a7ff242559c972e1f30 tip.block_height=4982505 max_block_height=4983412 suppressed=150
2026-08-05T22:22:02.437626Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129418975 tip.hash=f67ca2a46d35763a30dfbdff8a5e0ba25724637d0ba08fb5b262bce66d457682 tip.block_height=4982671 max_block_height=4983594 suppressed=165
2026-08-05T22:22:03.439717Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129422708 tip.hash=94ea4ba0ac92c985d92e7747441e166808c00d50de83544f55c29fea3a7484a1 tip.block_height=4982823 max_block_height=4983766 suppressed=151
2026-08-05T22:22:04.470877Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129425565 tip.hash=75f96810d0882cd1c8ebb521636c1c78f31c25682ffd16eee0f7b3bbb0ce1854 tip.block_height=4982946 max_block_height=4983907 suppressed=122
2026-08-05T22:22:05.477332Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129428619 tip.hash=be94acfa2d86fde77144cd929eccb288709835e443f101b762ff9e91073bb7ce tip.block_height=4983081 max_block_height=4984039 suppressed=134
2026-08-05T22:22:06.478603Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129431893 tip.hash=68ce0340f3fa2964bd54433cdbcd09c27ed76125c3a17bb1d1093a6564020073 tip.block_height=4983221 max_block_height=4984140 suppressed=139
2026-08-05T22:22:06.632110Z  INFO relays.collect: amaru::ledger: close time.busy=951µs time.idle=4.36µs [3mcount[0m[2m=[0m155
2026-08-05T22:22:07.486975Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129434477 tip.hash=8ab5377dc1d1e2175056d840f68a6bc2fd1a43acc53c383b6daee4c400051455 tip.block_height=4983332 max_block_height=4984303 suppressed=110
2026-08-05T22:22:08.499761Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129436938 tip.hash=26eb52e330f3316eb2950ba8901e46bacaf7e95f891eed658cb7077b556871a2 tip.block_height=4983416 max_block_height=4984384 suppressed=83
2026-08-05T22:22:09.501570Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129439742 tip.hash=96a674db979e5a5f6854d25498d717e90987ebdcafa4a8242db98ab0b802fde5 tip.block_height=4983566 max_block_height=4984524 suppressed=149
2026-08-05T22:22:10.509519Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129442685 tip.hash=5276f1f613760e95c8dcbc59abd4cf3ff1dea07db6acc64560be4a6fb2cfb1a1 tip.block_height=4983695 max_block_height=4984641 suppressed=128
2026-08-05T22:22:11.521712Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129445518 tip.hash=4b0a78a196d78c90550cda1a089575417c7fb2e9ce99d330c15f8086147cd06e tip.block_height=4983835 max_block_height=4984799 suppressed=139
2026-08-05T22:22:12.553878Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129450476 tip.hash=cfd7a9c81fd7e44424d137a0ed393daf4760a46bf9f3a6aad035fde983856374 tip.block_height=4984055 max_block_height=4984929 suppressed=219
2026-08-05T22:22:13.570975Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129458021 tip.hash=cf613a31880f31239c6b07a67baf0fe707b02a537590dbcdccb56ba0e33c7b80 tip.block_height=4984355 max_block_height=4985187 suppressed=299
2026-08-05T22:22:14.576113Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129464066 tip.hash=792b5429aafbb2b002a9abdbada788137f1af9f8cda557c464a988a16643dba2 tip.block_height=4984640 max_block_height=4985439 suppressed=284
2026-08-05T22:22:15.610347Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129470359 tip.hash=f14fcd7678d3b85646ab3864e8aa425d8a273b1f9c30385fb36e60ef4afdcab1 tip.block_height=4984955 max_block_height=4985711 suppressed=314
2026-08-05T22:22:16.612478Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129478251 tip.hash=dafd880e3e1b5647714351791dfdf99002e49f7046fcefa1610934c22457aa86 tip.block_height=4985308 max_block_height=4985967 suppressed=352
2026-08-05T22:22:17.621969Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129485903 tip.hash=69d1353459ffaca34eceb351df081e682edc6d20dea35c18e1fda373b2515667 tip.block_height=4985659 max_block_height=4986218 suppressed=350
2026-08-05T22:22:18.646233Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129492949 tip.hash=bae4ee134538dd207532161ad6cef993d1d3ce784aa93b1af6b99dac0fad6387 tip.block_height=4985980 max_block_height=4986497 suppressed=320
2026-08-05T22:22:19.647616Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129500247 tip.hash=de3bbde2ec5fa656858248e6aefa1dd5eeb9b61677281f530f86a236b542957e tip.block_height=4986294 max_block_height=4986721 suppressed=313
2026-08-05T22:22:20.661339Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129505923 tip.hash=055a32f0d6edd90c006eeb9a8b01e41eec441ef38e56cf02019580d79d65a1a3 tip.block_height=4986530 max_block_height=4987005 suppressed=235
2026-08-05T22:22:21.668148Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129512437 tip.hash=fe1e66bd22160c27996de2cc8808e2ae2fc3caf241164a10413ed716f997dd0c tip.block_height=4986805 max_block_height=4987262 suppressed=274
2026-08-05T22:22:22.670393Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129521375 tip.hash=f2bab1b3dcaea5e08c8267bcc0b362e2bf64395c21c2ea656360cf2103ed4e94 tip.block_height=4987190 max_block_height=4987502 suppressed=384
2026-08-05T22:22:23.671914Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129529136 tip.hash=f86389f9bcd82c33f90f9b62f5ad6b7d281f862431931f0cbd77d47f0abc0fa1 tip.block_height=4987556 max_block_height=4987758 suppressed=365
2026-08-05T22:22:24.672675Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129538169 tip.hash=ec61c578b5d0f8ac697d52153d1d3c5e6121076c5cad1bf37210cc96effb3df1 tip.block_height=4987952 max_block_height=4988010 suppressed=395
2026-08-05T22:22:25.676422Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129544350 tip.hash=405ee4965ce362fc1bc386e96dfc4a0021e767b52d18fa15675e57b61ec1752c tip.block_height=4988244 max_block_height=4988250 suppressed=291
2026-08-05T22:22:26.681244Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129550181 tip.hash=27ef976b6d3279c43926fa5e1fe7933843860a085e6d5e4c07a1591ca4fc4302 tip.block_height=4988521 max_block_height=4988533 suppressed=276
2026-08-05T22:22:27.710635Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129555397 tip.hash=7867b2a03eba4c3083e237969273a7d126009c463797faf7288d1d686cdca5c3 tip.block_height=4988766 max_block_height=4988766 suppressed=244
2026-08-05T22:22:28.719913Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129560643 tip.hash=223bb075c0ee6ba28120a1c4a576271d2cb7d02590ed153943d2562c01c0d695 tip.block_height=4989016 max_block_height=4989053 suppressed=249
2026-08-05T22:22:29.720666Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129566178 tip.hash=1a34d6de4383c56abd738e0af4b5df4e427b9c5ea5dfb97520145c5e0b79657f tip.block_height=4989280 max_block_height=4989309 suppressed=263
2026-08-05T22:22:30.721947Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129571461 tip.hash=efb5016d54c715e276af483e86232b6f887eafd53a04d8c0835ac080167f0a3d tip.block_height=4989504 max_block_height=4989558 suppressed=223
2026-08-05T22:22:31.722686Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129576899 tip.hash=a8a237d694c3305b5028720c581b4b5c9dc65d8646177cfc642f709730e19754 tip.block_height=4989754 max_block_height=4989824 suppressed=249
2026-08-05T22:22:32.735891Z  INFO amaru_consensus::stages::adopt_chain: adopted tip tip.slot=129582749 tip.hash=42bb9a0a17b361c71ae29bcaf4b2c73039e9fab1497a090868ed66a033d5ce98 tip.block_height=4990041 max_block_height=4990084 suppressed=286
2026-08-05T22:22:33.431984Z  WARN amaru_consensus::stages::validate_block: failed to fork the ledger to a new tip error=BlockValidationError: Invalid block: Transaction 96ae78f724a27b0d76c3d6a861857af3a644de971fe0c7fcbefe4e45811e5687 at index 0 is invalid: transaction failed phase one validation: invalid outputs: invalid transaction outputs: [#1: output value is too large: maximum: 5000, actual: 5001] parent=129586446.84e97a98f5ba4a023289ad9c2e06c0ebd93f4c712ed2c3e7453e888da0853213
2026-08-05T22:22:33.432042Z  INFO amaru_consensus::stages::validate_block: switching the ledger to a new fork parent=129586448.b2c00c166065cf92966677c315525ce7feab12378169f06a6cb6d8041197ff72 current=129586446.84e97a98f5ba4a023289ad9c2e06c0ebd93f4c712ed2c3e7453e888da0853213
2026-08-05T22:22:33.432391Z  INFO amaru_consensus::stages::select_chain: best tip candidate invalidated removed=1
2026-08-05T22:22:33.446083Z  INFO state.switch_to_fork:state.roll_backward: amaru::ledger: close time.busy=13.9ms time.idle=1.77µs [3mfork_point[0m[2m=[0m129586432.001226394b2304ae8a5d21f5c2258fd2c1ca5b49601db353a2a512b86ef67d66 [3mfork_length[0m[2m=[0m3

   amaru::fatal::error
   
   This is almost certainly a bug, and we'd appreciate a report so we can improve Amaru.
   
   Please report this error at https://github.com/pragma-org/amaru/issues/new.
   
   In your bug report please provide the information below and if possible the code
   that produced it.
   
   Operating System: linux
   Architecture:     x86_64
   Version:          v10.11.20260730+b7b2c0a
   

thread 'amaru-node' (1058955) panicked at crates/amaru-ledger/src/state.rs:932:9:
fork-switch replay would flush a still-rollback-able block to the stable store: after rollback the volatile holds 2159 block(s) and replaying 3 would push 2161 past the security parameter k=2160 before reaching the committing block
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

etorreborre:
Fixes https://github.com/pragma-org/amaru/issues/1162.

This PR fixes a panic that happened during a switch to fork. It was previously assumed that a fork replay blocks that would overflow the volatile window but this is not true. #1162 shows that it is possible to have a block failing to apply to the ledger then have the `select_chain` stage decide on a fork, with enough received blocks that we overflow the volatile window.

The fix consists in:

- Removing the incorrect assertion
- Deferring the action of stabilizing the volatile window when replaying the fork blocks. We need to do this because if a fork fails to be fully applied we want to be able to recover to the previous state, which is not possible if some blocks have been pushed to the stable db.

Note: this means that, in principle, we could have, temporarily, a volatile ledger that would consume twice it memory amount until stabilized.

## Summary by CodeRabbit

- **New Features**
  - Added clear outcomes for completed, partially applied, and rolled-back fork switches.
  - Added progress, failure details, and invalid-block information to ledger reporting.
  - Added optional fork-switch outcome details to observability data.

- **Bug Fixes**
  - Improved recovery when fork application fails.
  - Invalid blocks and their descendants are now consistently rejected.
  - Improved chain selection when the preferred branch becomes invalid.

- **Tests**
  - Expanded coverage for rollback, partial application, invalid forks, and recovery scenarios.


CHANGELOG.md
```diff
@@ -81,6 +81,7 @@ Other guiding principles:
 - **amaru-ledger**: validate provided treasury value matches actual treasury value (ConwayTreasuryValueMismatch) ([#1025][], [#888][])
 - **amaru-ledger**: reject governance proposals whose deposit return account is not registered. ([#928][])
 - **amaru-uplc**: fix the validation of Plutus scripts containing lists of BLS elements when they are empty, since the Haskell node accepts them ([#1159](https://github.com/pragma-org/amaru/issues/1159))
+- **amaru-ledger**: switching to a fork must be possible with forks temporarily exceeding the volatile window ([#1190][https://github.com/pragma-org/amaru/issues/1190]).
```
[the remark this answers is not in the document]
etorreborre: Fixed

crates/amaru-ledger/src/state.rs
```diff
@@ -532,36 +532,43 @@ impl<S: Store, HS: HistoricalStores + Send + Sync + 'static> State<S, HS> {
...
+    }
 
-                Some(now_stable)
-            } else {
-                trace!(ledger::volatile::WARM_UP, size = self.volatile.len());
-                None
-            };
+    /// Persist every fragment that has fallen out of the `k`-block volatile window.
+    ///
+    /// A fork replay defers the stabilization until the whole fork has been validated (because
```
KtorZ: We cannot do that; deferring an epoch transition would mean potentially validating a block with an invalid state of the ledger (protocol parameters may have changed, pools or proposals may no longer exists, etc..)

KtorZ:
I do not agree with the fix; though I agree we might have been too strict on the invariant? If we think about it, there's indeed nothing that really tells us a fork has to be exactly one block longer. Messages may be delayed and we may only hear about a longer chain once it's already longer than one block. 

At the same time; it should be _fine_ to persist blocks that are pushed out of the volatile during a rollback. 

Suppose we have a fork of length $m + n \leq k$, such that $m$ is the rollback length, and $n$ the suffix after that. Currently, we forbid $n \gt 1$ because the $(m+1)\text{th}$  block would indeed push an old block out of the volatile to persistence. This happens only if it is valid. 

And technically, at this point, we are *already* on a longer chain. So the validity of the blocks from $m+1..n$ does not really matter. If one is valid, then it'll push the next fragment out of the volatile. 

But that means, we must be able to report back to the consensus layer that a chain switch was _partially executed_. We may switch chain up-to the first invalid block of the new candidate chain; which only has permanent effects on the ledger if it's a longer chain -- even if it ends up shorter than announced.

etorreborre:
@KtorZ that makes sense. I propose the following then (or an attempt to reformulate what you wrote):

- During a switch to fork, we try to apply the fork blocks up to `max(current_volatile_height, k)` (that's `<= k`). If one block fails to apply:
   - If it's before `current_volatile_height`, we use the current recovery (because we ended up on a shorter chain).
   - If it's after `current_volatile_height` we have a longer chain anyway.
- If we still have some blocks to apply after `k`, we try to apply them using regular roll forwards.

In any case, a failure to apply reports:
  - The point that failed to apply.
  - The current ledger height
  
From this return value, we inform the `select_chain` so that it can select again which chain is the current best.
That will possibly bring to me change a few things in the `select_chain/validate_blocks` interaction because I think that it is slightly wrong at the moment (we may detect a fork when there is none if we get a batch of blocks with one invalid block in the middle).

KtorZ:
@etorreborre 👍

crates/amaru-ledger/src/state.rs
```diff
@@ -833,107 +842,116 @@ impl<S: Store, HS: HistoricalStores + Send + Sync + 'static> State<S, HS> {
...
+    /// Try to rollback the volatile state to a given point and roll the ledger forward with the new blocks.
+    /// Blocks are applied with the regular [`Self::roll_forward`], so epoch
+    /// transitions, rewards and snapshots are processed exactly as during normal chain extension.
+    ///
+    /// A block failing to apply is reported in the returned [`ForkSwitchOutcome`] value:
+    ///
+    /// - If the applied prefix is shorter than the rolled-back chain and nothing was written
+    ///   to the stable store, the pre-switch state is fully restored
+    ///   ([`ForkSwitchOutcome::RolledBack`]).
+    /// - If the fork is strictly longer than the chain it replaces, or if some blocks have been
```
```suggestion
    /// - If the fork is strictly longer than the chain it replaces, i.e. if some blocks have been
```

Those should hopefully be equivalent 😶

crates/amaru-ledger/src/state.rs
```diff
@@ -122,6 +122,11 @@ where
     rewards_join_handle: Option<JoinHandle<Result<RewardsSummary, StateError>>>,
 
     tip_update_emitter: TipUpdateEmitter,
+
+    /// Monotonic count of irreversible writes to the stable store: block evictions
+    /// ([`Self::apply_block`]) and epoch-transition flushes ([`Self::apply_transition`]).
+    /// A fork switch can only recover its pre-rollback state while this count is unchanged.
+    stable_writes: u64,
```
KtorZ: Why do we have to keep this globally to the ledger? Surely this can remain local to `switch_to_fork`, no?

crates/amaru-ledger/src/state.rs
```diff
@@ -833,107 +842,116 @@ impl<S: Store, HS: HistoricalStores + Send + Sync + 'static> State<S, HS> {
...
+            match self.roll_forward(block, arena_pool) {
+                BlockValidation::Valid(new_metrics) => {
+                    applied += 1;
+                    applied_tip = Some(block_tip);
+                    metrics = new_metrics;
+                }
+                BlockValidation::Invalid(tip, details) => {
+                    let failure = InvalidBlock { tip, reason: details.to_string() };
+                    let beats_replaced_chain = applied > rollback_length;
+                    let recoverable = self.stable_writes == stable_writes;
```
KtorZ: I don't think we need a global `stable_writes` counter that's a little awkward to keep in the state IMO; what about: 

1. We record the tip of the immutable prior to rolling back. 
2. At the first failure, we check if the tip has changed 
3. We assert that the tip has changed if-and-only-if we have applied strictly more than `rollback_length` blocks.

Then, `beats_replaced_chain` or `tip_before != tip_after` are somewhat equivalent in deciding to undo the rollback or not.

crates/amaru-ledger/src/state.rs
```diff
@@ -833,107 +842,116 @@ impl<S: Store, HS: HistoricalStores + Send + Sync + 'static> State<S, HS> {
...
+
+                    // If the fork is no longer than the chain it replaces,
+                    // and nothing was written to the stable store, recover the previous chain
+                    if !beats_replaced_chain && recoverable {
+                        info_record!(ledger::state::SWITCH_TO_FORK, outcome = "rolled_back");
```
```suggestion
                        info_record!(ledger::state::SWITCH_TO_FORK, outcome = "partial", is_longer_chain = false);
```

crates/amaru-ledger/src/state.rs
```diff
@@ -833,107 +842,116 @@ impl<S: Store, HS: HistoricalStores + Send + Sync + 'static> State<S, HS> {
...
+                        self.recover(state_recovery);
+                        return Ok(ForkSwitchOutcome::RolledBack { failure });
                     }
-                    BlockValidation::Err(error) => {
-                        recover(self);
-                        return BlockValidation::Err(error);
+
+                    info_record!(ledger::state::SWITCH_TO_FORK, outcome = "partial");
```
```suggestion
                    info_record!(ledger::state::SWITCH_TO_FORK, outcome = "partial", is_longer_chain = true);
```

crates/amaru-ledger/src/state.rs
```diff
@@ -833,107 +842,116 @@ impl<S: Store, HS: HistoricalStores + Send + Sync + 'static> State<S, HS> {
...
+                }
+                BlockValidation::Err(error) => {
+                    info_record!(ledger::state::SWITCH_TO_FORK, outcome = "failed");
+                    // Try to recover as much as possible if nothing was written to the stable store.
+                    if self.stable_writes == stable_writes {
+                        self.recover(state_recovery);
                     }
+                    return Err(error);
                 }
             }
```
KtorZ, line 977: I am not sure the handling of an Err should be any different than for an invalid block. If we have already pushed blocks to the stable store, we cannot undo the rollback.

crates/amaru-ledger/src/state.rs
KtorZ: turns out (3) is already naturally happening in `.recover` since we capture the tip in the rollback guard. So we can simplify a bit the logic here and simply try to recover when the length is smaller or equal. By construction, the immutable tip should not have moved.

crates/amaru-consensus/src/block_validator.rs
```diff
@@ -95,28 +95,25 @@ impl<S: Store + Send + Sync, HS: HistoricalStores + Send + Sync + 'static> CanVa
...
-                        .ok_or_else(|| anyhow!("block not found"))?
+                        .map_err(|e| BlockValidationError::new(anyhow!(e)))?
+                        .ok_or_else(|| BlockValidationError::new(anyhow!("block not found")))?
                         .decode()
-                        .map_err(|e| anyhow!(e))?;
-                    Ok((*point, block))
-                });
+                        .map_err(|e| BlockValidationError::new(anyhow!(e)))?;
+                    forward_blocks.push(block);
+                }
```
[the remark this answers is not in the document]
etorreborre: Done.

crates/amaru-ledger/src/state.rs
etorreborre: This one leads to a termination of the node anyway, so I propose (no need to even try to recover the volatile state):
```
BlockValidation::Err(error) => {
    info_record!(ledger::state::SWITCH_TO_FORK, outcome = "error");
    return Err(error);
}
```

crates/amaru-ledger/src/state.rs
etorreborre: Yes, this can completely stay local.

crates/amaru-ledger/src/state.rs
etorreborre: It could also be shorter but we already persisted some stable state. Now I wonder what that means in terms of correctness. I still need to think about it.

crates/amaru-ledger/src/state.rs
KtorZ: Might be good to add the logged event whether or not the immutable was touched here? What do you think? It might be surprising on restart otherwise.

crates/amaru-ledger/src/state/volatile/db.rs
```diff
@@ -1253,11 +1263,16 @@ mod tests {
...
+    #[test_case(None, Some(CommitteeAct::Resign) => Existence::Exists(false); "resigned in current"
+    )]
+    #[test_case(Some(CommitteeAct::Auth), None => Existence::Exists(true); "hot-auth in draining, untouched in current"
+    )]
+    #[test_case(Some(CommitteeAct::Resign), None => Existence::Exists(false); "resigned in draining shadows the stable store"
+    )]
+    #[test_case(Some(CommitteeAct::Auth), Some(CommitteeAct::Resign) => Existence::Exists(false); "current resignation overrides draining"
+    )]
+    #[test_case(None, None => Existence::Unknown; "untouched everywhere defers to the stable store"
+    )]
```
KtorZ: Weird changes.

crates/amaru-ledger/src/state.rs
```diff
@@ -833,107 +833,117 @@ impl<S: Store, HS: HistoricalStores + Send + Sync + 'static> State<S, HS> {
...
-                        return BlockValidation::Err(error);
-                    }
-                };
-                match self.roll_forward(&point, block, arena_pool) {
-                    BlockValidation::Valid(new_metrics) => metrics = new_metrics,
-                    BlockValidation::Invalid(slot, hash, details) => {
-                        recover(self);
-                        return BlockValidation::Invalid(slot, hash, details);
+        let mut applied = 0usize;
+        let mut applied_tip = None;
```
KtorZ: Maybe initialize to `initial_immutable_tip`? It cannot really remains `None` anyway. That would avoid the little panic below.

crates/amaru-ouroboros-traits/src/validators/blocks/can_validate_blocks.rs
```diff
@@ -23,18 +23,37 @@ use thiserror::Error;
...
+/// Outcome of asking the ledger to switch to a fork.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub enum ForkSwitchOutcome {
+    /// Every fork block was applied; the ledger tip is the fork tip.
+    Completed { metrics: LedgerMetrics },
+    /// A block failed to apply but the applied prefix was kept because it already beats the chain
+    /// it replaced; `applied_tip` is the first actual ledger tip after the switch.
+    Partial { applied_tip: Tip, metrics: LedgerMetrics, failure: InvalidBlock },
+    /// A block failed to apply and the ledger fully restored its pre-switch state.
+    RolledBack { failure: InvalidBlock },
```
KtorZ:
```suggestion
    Failed { failure: InvalidBlock },
```

It gets really confusing otherwise 😅

crates/amaru-stores/tests/state_rollback.rs
```diff
@@ -166,56 +124,326 @@ fn rollback_after_volatile_front_is_rejected() {
...
+        Err(err) => err,
+    };
+    assert!(
+        matches!(err.downcast_ref::<StateError>(), Some(StateError::RewardsSummaryNotReady)),
+        "unexpected error: {err:#}"
+    );
+
+    assert_eq!(*state.tip(), initial_tip.point(), "the pre-switch tip is restored after a failed transition");
+    assert!(flushed.lock().unwrap().is_empty(), "nothing reached the stable store");
+}
```
KtorZ: This scenario is a little weird because we should not, in principle, hit that error in a "normally constructed forward sequence". 

A better test could be to start in the previous epoch, roll forward a few blocks to cross the boundary (and trigger the rewards calculations and epoch transition logic); and then, rollback and re-execute the transition. 

No?

crates/amaru-stores/tests/state_rollback.rs
```diff
@@ -166,56 +124,326 @@ fn rollback_after_volatile_front_is_rejected() {
...
+
+#[test]
+fn a_successful_switch_to_fork_trims_the_volatile_window() {
+    let era_history: EraHistory = PREPROD_ERA_HISTORY.clone();
+    let k = PREPROD_GLOBAL_PARAMETERS.consensus_security_param;
+
+    // We make sure that the whole test stays inside a single epoch so that we don't have to worry
+    // about epoch transitions and the stable store.
+    let first_slot = 43_200;
+    let last_slot = first_slot + k + 2;
```
KtorZ: Awkward. That does not compute the dimensional analysis: `k` unit is `block`, and we sum it with `slots`. I get the idea, but seeing `k` here and a magic `43_200` when `k` is pulled just above it really awkward. 

Let's call f = `1 / active_slot_coeff_inverse` (so `/ f` is equivalent to `* active_slot_coeff_inverse`)

An epoch is `10 * k / f` long. The rewards calculation starts _at latest_ at `3 * k / f`; which on Mainnet and preprod is indeed `43_200`. 

I understand the `k` here refers to a number of blocks / fragments in the volatile and we simply go with the approximation that every slot is active. So.. I don't know. It's just a little awkward to mix `k`, hard coded magic numbers and slots.

crates/amaru-ledger/src/state/volatile/db.rs
etorreborre: I updated my editor and it sometimes reformats differently than `cargo fmt`, I still don't know why. It's reverted now.

crates/amaru-ledger/src/state.rs
etorreborre: Yes, that's better.

crates/amaru-ouroboros-traits/src/validators/blocks/can_validate_blocks.rs
etorreborre: Done.

rkuhn:
Before starting the review: the discussion between @KtorZ and @etorreborre seems to suggest a change only when the target fork ends in some invalid block(s), in which case sending the negative validation result(s) to `select_chain` should already trigger all the right consequences, no new interactions needed.

That being said: cover my six, I’m going in.

crates/amaru-consensus/src/effects/ledger_effects.rs
```diff
@@ -41,10 +42,7 @@ pub trait LedgerOps: Send + Sync {
         point: &Point,
     ) -> BoxFuture<'static, Result<Result<LedgerMetrics, BlockValidationError>, BlockValidationError>>;
 
-    fn switch_to_fork(
-        &self,
-        point: &Point,
-    ) -> BoxFuture<'static, anyhow::Result<Result<LedgerMetrics, BlockValidationError>, BlockValidationError>>;
+    fn switch_to_fork(&self, tip: &Tip) -> BoxFuture<'static, Result<ForkSwitchOutcome, BlockValidationError>>;
```
rkuhn, line 45: In general, I want to get rid of the `Tip` type we currently have because I find
- the name wrong
- the usage vs. `Point` unclear
- and we always have `BlockHeight` available when we know `Slot`
So I propose to add `BlockHeight` to `Point` and be done with it.

crates/amaru-consensus/src/stages/select_chain/tests.rs
```diff
@@ -812,6 +749,185 @@ fn test_last_best_tip_invalidated_falls_back_to_origin() {
...
+    prep.store_headers(&prep.headers.all());
+    prep.set_anchor(prep.headers.h0.hash());
+    // h3a is *not* the first pending element of its chain: h2a was replayed inside a fork switch
+    // and never got an individual validation result.
+    let tip = prep.headers.h3a.tip();
+    let msg = SelectChainMsg::BlockValidationResult(tip, false, BlockHeight::from(0));
+
+    let expected = SelectChain {
+        best_tip: Some(prep.headers.h3.clone()),
+        tips: BTreeMap::from_iter([(prep.headers.h3.hash(), vec![prep.headers.h2.hash(), prep.headers.h3.hash()])]),
```
rkuhn, line 832: Not falling back to h2a is a design decision, we could go either way. Removing it (as is done here) then needs to be handled correctly when a new header with parent h2a comes along, for which a test is missing (I think — not sure whether this is covered elsewhere).

crates/amaru-consensus/src/stages/select_chain/tests.rs
```diff
@@ -812,6 +749,185 @@ fn test_last_best_tip_invalidated_falls_back_to_origin() {
...
+    prep.set_validity(prep.headers.h1.hash(), true);
+    prep.set_best_chain(prep.headers.h1.hash());
+    // The failing block reported by a fork switch is the best tip itself.
+    // h2 stays pending.
+    let tip = prep.headers.h3.tip();
+    let msg = SelectChainMsg::BlockValidationResult(tip, false, BlockHeight::from(0));
+
+    let expected = SelectChain {
+        best_tip: Some(prep.headers.h2.clone()),
+        tips: BTreeMap::from_iter([(prep.headers.h2.hash(), vec![prep.headers.h2.hash()])]),
```
rkuhn: This is inconsistent with the above: if we nuke the branch leading to an invalid tip then we’ll need to always do that.

crates/amaru-consensus/src/stages/select_chain/tests.rs
```diff
@@ -812,6 +749,185 @@ fn test_last_best_tip_invalidated_falls_back_to_origin() {
...
+
+    let applied = SelectChainMsg::BlockValidationResult(applied_tip, true, BlockHeight::from(0));
+    let failed = SelectChainMsg::BlockValidationResult(failed_tip, false, BlockHeight::from(0));
+
+    let after_applied =
+        SelectChain { tips: BTreeMap::from_iter([(failed_tip.hash(), vec![failed_tip.hash()])]), ..prep.state.clone() };
+
+    let expected = SelectChain {
+        best_tip: Some(prep.headers.h2a.clone()),
+        tips: BTreeMap::from_iter([(applied_tip.hash(), vec![])]),
```
rkuhn: same here — feels inconsistent to me

crates/amaru-consensus/src/stages/validate_block/mod.rs
```diff
@@ -108,46 +134,99 @@ pub async fn stage(mut state: ValidateBlock, msg: ValidateBlockMsg, eff: Effects
...
             }
-        }
+        };
         state
     }
     .instrument(span)
     .await
 }
 
+async fn completed(state: &mut ValidateBlock, tip: Tip, eff: &Effects<ValidateBlockMsg>, metrics: LedgerMetrics) {
```
rkuhn: These should be methods on `ValidateBlock`.

crates/amaru-consensus/src/block_validator.rs
```diff
@@ -71,54 +71,51 @@ impl<S: Store + Send + Sync, HS: HistoricalStores + Send + Sync + 'static> CanVa
...
+        for forward_tip in forward_tips.iter() {
+            let block = self
+                .chain_store
+                .load_block(&forward_tip.hash())
+                .map_err(|e| BlockValidationError::new(anyhow!(e)))?
+                .ok_or_else(|| BlockValidationError::new(anyhow!("block not found")))?
+                .decode()
+                .map_err(|e| BlockValidationError::new(anyhow!(e)))?;
+            forward_blocks.push(block);
         }
```
rkuhn: `.collect::<Vec<_>>()` does the exact same thing as `Vec::with_capacity`+loop

crates/amaru-ledger/src/state.rs
```diff
@@ -833,107 +833,113 @@ impl<S: Store, HS: HistoricalStores + Send + Sync + 'static> State<S, HS> {
...
-                                st.volatile = *volatile;
-                            }
-                            StateRecovery::RecoverVolatileDBPart { recovery } => {
-                                st.volatile.undo_rollback(*recovery);
-                            }
-                        }
-                    }
-                }
+        );
+        let _guard = span.enter();
```
rkuhn: shorter is `let _guard = info_span!(...).entered()`

rkuhn (commented):
LGTM (modulo small points — should’ve said so) for the consensus parts, I only skimmed the ledger parts.

crates/amaru-ledger/src/state.rs
```diff
@@ -833,107 +833,113 @@ impl<S: Store, HS: HistoricalStores + Send + Sync + 'static> State<S, HS> {
...
+                    metrics = new_metrics;
+                }
+                BlockValidation::Invalid(tip, details) => {
+                    let failure = InvalidBlock { tip, reason: details.to_string() };
+                    let current_immutable_tip = self.immutable_tip();
+                    let fork_is_longer = applied > rollback_length;
+
+                    // If the fork is not longer than the chain it replaces,
+                    // and nothing was written to the stable store, recover the previous chain
+                    if !fork_is_longer && current_immutable_tip == initial_immutable_tip {
```
yHSJ: This `&&` makes me very nervous, especially considering the fact that we do NOT honor the chain growth property. I think I thought of a case in which this causes us to *knowingly* switch to a shorter chain. I think it is *very* unlikely to happen, it requires very specific preconditions, but it's worth considering.

Imagine there is some catastrophic network failure (similar to the "poisoned piggy incident") leading to blocks being very sparse, on every branch.

Let `C` be the last block of epoch `E`. The "canonical" chain extends the chain so that it becomes:
`C-h1-h2-h3-h4`. All 4 blocks land in the first `3k/f` slots of `E+1`. Then we stop receiving blocks.

Now, a malicious chain (intentionally or not, perhaps the better term is faulty) is served to us, sharing common ancestor `C`:

`C-m1-m2-m3-m4-m5`

Importantly, `m1` was *more* than `3k/f` into `E+1`. This *can* happen since the chain is very sparse.  Since it's longer, we switch to this fork, rolling back the chain containing the `h` blocks. That also undoes the *in-memory* transition sitting in our `volatile` (the `overlay`).

We then apply `m1`, triggering the transition again (which is expected since it's the first block of a new epoch). This time, however, we *also* trigger the flush of `epoch_tail()` because `m1`'s slot is past the the stability window. Amaru *intentionally* allows violations of the chain growth property and so we flush everything up to C and the transition to the stable store.

`m2` then applies fine, but it turns out that say `m3` is invalid.` Now our chain is only 2 blocks while the honest chain is 4 blocks. However we cannot recover the rolled back blocks because `current_immutable_tip == initial_immutable_tip` is false. Therefore, we knowingly accept a shorter chain.

This scenario is *very* specific and likely will never happen in reality (I do not believe it can be forced by a bad actor). That being said, we have already been surprised by things such as the chain growth property not being required in the Haskell node (the exact details of that claim depend on who you ask).


We should at the very least add a `warn!` that announces when we accept a shorter chain.

yHSJ: The thing I am not *certain* about is whether or not `m1` can even be `3k/f + N` slots after `C`. If I remember correctly, the Haskell team told us that the `cardano-node` requires at least 1 block every `3k/f` slots. In which case this scenario is impossible in the Haskell node (worth double checking). Unless Amaru's consensus has the same requirement, Amaru *is* vulnerable to this scenario.

KtorZ: > The thing I am not certain about is whether or not m1 can even be 3k/f + N slots after C

No it cannot; the key invariant the Haskell node maintains is that there's AT LEAST ONE block for every 3*k/f window. Things break beyond that. 

But that does not disqualify your scenario. You just add more block and make them very sparse so that there's some infix before 3*k/f and a suffix after that contains invalid block and we do indeed have a problem.

However! That is a separate concern from this PR. We have to solve the "force flush" at 3*k/f _anyway_ by having a vector of draining sequences in the volatile (at least, that seems to be the obvious design choice). Which would solve the problem here too. We would simply create a new draining sequence without touching the immutable store here too. 

So for the sake of this PR, it's fine. But we really need to address the multi-draining question.

yHSJ: > No it cannot; the key invariant the Haskell node maintains is that there's AT LEAST ONE block for every 3*k/f window. Things break beyond that.

That is what I remembered being discussed (but didn't verify myself). I don't believe we enforce that here in Amaru, so this case is still applicable to *us*, no? Granted, it would require a significant number of Amaru nodes in the network...

> But that does not disqualify your scenario. You just add more block and make them very sparse so that there's some infix before 3*k/f and a suffix after that contains invalid block and we do indeed have a problem.

Ah, fair enough. I didn't even consider just *adding more blocks*.

> So for the sake of this PR, it's fine. But we really need to address the multi-draining question.
I have only considered one case that could trigger this behavior, is it not possible there are others? I still think it is worth at least adding a `warn!()` just to be explicit about that behavior, even if we are relatively confident it won't happen.

crates/amaru-consensus/src/block_validator.rs
```diff
@@ -71,54 +71,51 @@ impl<S: Store + Send + Sync, HS: HistoricalStores + Send + Sync + 'static> CanVa
...
-                });
+    fn switch_to_fork(&self, fork_point: &Point, to: &Tip) -> Result<ForkSwitchOutcome, BlockValidationError> {
+        // Get all the headers of the block to apply between the fork point and the expected new tip
+        let Some(forward_tips) = self.chain_store.ancestors_between(fork_point, to.hash()) else {
+            return Err(BlockValidationError::new(anyhow!(
+                "the stored headers do not form a chain from {fork_point} to {}",
+                to.point()
+            )));
+        };
+        if forward_tips.is_empty() {
```
yHSJ: If we receive the same message twice, wouldn't we kill the stage here? What happens if we kill the stage? Do we want this to be a noop instead?

crates/amaru-consensus/src/block_validator.rs
```diff
@@ -71,54 +71,51 @@ impl<S: Store + Send + Sync, HS: HistoricalStores + Send + Sync + 'static> CanVa
...
-                match state.switch_to_fork(&fork_point, forward_blocks, &self.vm_eval_pool) {
-                    BlockValidation::Valid(metrics) => Ok(Ok(metrics)),
-                    BlockValidation::Invalid(_, _, details) => {
-                        Ok(Err(BlockValidationError::new(anyhow!("Invalid block: {details}"))))
-                    }
-                    BlockValidation::Err(err) => Err(BlockValidationError::new(anyhow!(err))),
-                }
-            }
+        // Load the blocks corresponding to the headers to apply, in order, from the chain store.
+        let mut forward_blocks = Vec::with_capacity(forward_tips.len());
```
yHSJ: This used to be lazy, now it's eagerly evaluated (I assume needed because of the change to the function signature where the iterator item is `Block` instead a `Result`). 

That is not a big deal, but it's worth keeping in mind that it *will* blow up memory in the case of a long fork. Noting so that we are sure we cover this in our benchmarks.

I guess it can't be a huge increase, since it is at *most* 2160 blocks * 88 KB which is like 190 MB. But if that *just* pushes over our stated memory requirements...

crates/amaru-consensus/src/effects/ledger_effects.rs
KtorZ: > and we always have BlockHeight available when we know Slot

You mean, contextually, or... because it can be derived from the `Slot` (because it cannot 😅 ...)?


Otherwise, that sounds reasonable. This is mostly inherited from the Haskell codebase, yet without much benefit.

rkuhn: Strictly speaking (since I haven’t done the precise research) my last bullet point was a conjecture: Either we have the HeaderHash, or we have the whole Header. To my knowledge there is no situation where we get the Slot without having the Header lying around somewhere. So we might as well just grab the BlockHeight as well.

crates/amaru-consensus/src/block_validator.rs
rkuhn: This is one of those points where it is really difficult to judge the benefits and downsides without real-world pain as background information.

crates/amaru-consensus/src/effects/ledger_effects.rs
etorreborre: Yes I would gladly do that!

crates/amaru-ledger/src/state.rs
etorreborre: Fixed

crates/amaru-ledger/src/state.rs
etorreborre: Now a partial fork cannot be longer than the previous chain

crates/amaru-ledger/src/state.rs
etorreborre: Good idea

crates/amaru-stores/tests/state_rollback.rs
etorreborre: See the recent commits

crates/amaru-consensus/src/effects/ledger_effects.rs
rkuhn: And once we have completely cleaned the code tree of `struct Tip` we can actually start using that word to mean tree leaves :-)

crates/amaru-stores/tests/state_rollback.rs
etorreborre: Yes the idea is to approximate 1 slot = 1 block. In any case I hope that the additional diagrams in tests help.

crates/amaru-consensus/src/block_validator.rs
etorreborre: Done (why no `traverse` in Rust 😢?)

crates/amaru-consensus/src/block_validator.rs
etorreborre: Yes, I went for more simplicity in the API, knowing that most forks are short and should iterate on all the blocks anyway.

crates/amaru-consensus/src/block_validator.rs
etorreborre: Yes that's a good point. I made it a no-op at the ledger level with a warning so that the `validate_block` message should be idempotent.

crates/amaru-consensus/src/stages/validate_block/mod.rs
etorreborre: Yes, that's nicer.

crates/amaru-consensus/src/block_validator.rs
etorreborre: Actually, I think we should raise an error here because the `validate_blocks` stage is not doing its job properly in that case and we need to know it.

crates/amaru-consensus/src/stages/select_chain/tests.rs
etorreborre: I added a test for this case.

crates/amaru-consensus/src/stages/select_chain/tests.rs
```diff
@@ -812,6 +807,176 @@ fn test_last_best_tip_invalidated_falls_back_to_origin() {
...
+#[test]
+fn test_invalid_block_validation_result_prunes_the_whole_branch() {
+    let mut prep = test_prep();
+    prep.state.best_tip = Some(prep.headers.h3.clone());
+    prep.state.tips = BTreeMap::from_iter([
+        (prep.headers.h3.hash(), vec![prep.headers.h2.hash(), prep.headers.h3.hash()]),
+        (prep.headers.h3a.hash(), vec![prep.headers.h2a.hash(), prep.headers.h3a.hash()]),
+    ]);
+    prep.store_headers(&prep.headers.all());
+    prep.set_anchor(prep.headers.h0.hash());
```
rkuhn: I’d store that h0 and h1 are valid, just to have a consistent state that can happen in reality.

crates/amaru-consensus/src/stages/select_chain/tests.rs
```diff
@@ -812,6 +807,176 @@ fn test_last_best_tip_invalidated_falls_back_to_origin() {
...
+        Level::ERROR,
+    ]);
+}
+
+#[test]
+fn test_new_tip_after_pruning_restores_pending_block_validations() {
+    let mut prep = test_prep();
+    // Start with the same scenario as the test above: the invalid h3a first prunes the h2a/h3a branch.
+    // Then a new header h3b with parent h2a re-tracks it, proving that pruning does not
+    // leave any state behind that would prevent the branch from being adopted again.
```
rkuhn, line 815: this test now subsumes the test above, which can thus be removed

rkuhn (approved):
looking good, apart from two nitpicks

crates/amaru-consensus/src/stages/select_chain/tests.rs
etorreborre: Done.

crates/amaru-consensus/src/stages/select_chain/tests.rs
etorreborre: Indeed.

crates/amaru-ledger/src/state.rs
```diff
@@ -868,149 +868,164 @@ impl<S: Store, HS: HistoricalStores + Send + Sync + 'static> State<S, HS> {
...
-                match self.roll_forward(&point, &block, arena_pool) {
-                    BlockValidation::Valid(new_metrics) => {
-                        if keep_blocks {
-                            deferred_blocks.push((point, block));
-                        }
-                        metrics = new_metrics;
+        // The fork must replace the rolled-back chain at equal length or extend it by exactly one
+        // block. If this condition is violated, this means that there is an issue with chain selection.
+        // We return an error to let the consensus layer deal with it.
+        if fork_length < rollback_length || fork_length > rollback_length + 1 {
```
KtorZ, line 923: What's the rationale for allowing `fork_length == rollback_length`? Shouldn't the chain selection only switch for longer chains?

crates/amaru-ledger/src/state.rs
```diff
@@ -868,149 +868,164 @@ impl<S: Store, HS: HistoricalStores + Send + Sync + 'static> State<S, HS> {
...
+        let mut applied_tip = Tip::new(initial_immutable_tip, BlockHeight::new(0));
+        let mut metrics = LedgerMetrics::default();
+
+        // Try to apply each block in the fork, and stop at the first failure.
+        for block in blocks {
+            let block_tip = block.tip();
+            match self.roll_forward(&block, arena_pool) {
+                BlockValidation::Valid(new_metrics) => {
+                    if keep_blocks {
+                        deferred_blocks.push((block_tip.point(), block));
```
KtorZ: Minor but, it seems a bit pointless to bundle the point with the block, when the point is _already available_ from the block.

crates/amaru-ledger/src/state.rs
```diff
@@ -868,149 +868,164 @@ impl<S: Store, HS: HistoricalStores + Send + Sync + 'static> State<S, HS> {
...
+                        self.recover(state_recovery);
+                        return Ok(ForkSwitchOutcome::Failed { failure });
                     }
-                    BlockValidation::Err(error) => {
-                        self.observers.on_block = real_on_block;
-                        restore(self, state_recovery);
-                        return BlockValidation::Err(error);
+
+                    warn_record!(ledger::state::SWITCH_TO_FORK, outcome = "partial");
+                    return Ok(ForkSwitchOutcome::Partial { applied_tip, metrics, failure });
```
KtorZ, line 964: Can this still happen actually? If the forks are **at most**, one block longer. Then only the last block can change the immutable tip, but it only does so if the block is valid.

crates/amaru-ledger/src/state.rs
etorreborre: We can indeed end-up on a fork chain of the same length. See the comparison [here](https://github.com/pragma-org/amaru/blob/main/crates/amaru-consensus/src/stages/select_chain/mod.rs#L450).

crates/amaru-ledger/src/state.rs
etorreborre: There's unfortunately the case where we force the snapshot if we have a chain-growth violation. In this case, there's no possible recovery.

crates/amaru-consensus/src/block_validator.rs
```diff
@@ -71,54 +71,51 @@ impl<S: Store + Send + Sync, HS: HistoricalStores + Send + Sync + 'static> CanVa
...
+            .iter()
+            .map(|forward_tip| {
+                self.chain_store
+                    .load_block(&forward_tip.hash())
+                    .map_err(|e| BlockValidationError::new(anyhow!(e)))?
+                    .ok_or_else(|| BlockValidationError::new(anyhow!("block not found")))?
+                    .decode()
+                    .map_err(|e| BlockValidationError::new(anyhow!(e)))
+            })
+            .collect::<Result<Vec<_>, _>>()?;
```
KtorZ, line 107: Do we have to collect here?

crates/amaru-ledger/src/state.rs
etorreborre: Absolutely. Refactored.

crates/amaru-consensus/src/block_validator.rs
etorreborre: To `traverse` the list of results. Do you see any other way?

crates/amaru-consensus/src/block_validator.rs
```diff
@@ -71,54 +71,51 @@ impl<S: Store + Send + Sync, HS: HistoricalStores + Send + Sync + 'static> CanVa
...
-                    let block = self
-                        .chain_store
-                        .load_block(&point.hash())
-                        .map_err(|e| anyhow!(e))?
-                        .ok_or_else(|| anyhow!("block not found"))?
-                        .decode()
-                        .map_err(|e| anyhow!(e))?;
-                    Ok((*point, block))
-                });
+    fn switch_to_fork(&self, fork_point: &Point, to: &Tip) -> Result<ForkSwitchOutcome, BlockValidationError> {
```
KtorZ, line 87: Where do we enforce that the `Tip` given here is at most one block ahead? I couldn't find the code that is producing this guarantee?

crates/amaru-stores/tests/switch_to_fork.rs
```diff
@@ -0,0 +1,454 @@
...
+
+// ASSERTIONS
+// ----------------------------------------------------------------------------
+
+/// The exhaustive fork-point cases (before the window, unknown slot or hash, in the future) are
+/// covered in `rollback.rs`. This test covers a call to switch_to_fork with rejects an incorrect fork
+/// point the same way.
+#[test]
+fn the_fork_point_must_be_in_the_volatile_window() {
+    let first_slot = 69_200_000; // inside epoch 163
```
KtorZ, line 80: Oddly specific 😅

crates/amaru-consensus/src/block_validator.rs
etorreborre: It's not enforced here but on the ledger side with the resulting fork length.
This part here assumes that `select_chain` did the right thing.

crates/amaru-stores/tests/switch_to_fork.rs
KtorZ, line 1: Nice set of tests here.

crates/amaru-ledger/src/state.rs
KtorZ: I see. Thanks.

crates/amaru-consensus/src/block_validator.rs
KtorZ: Ho. I missed the `Result` in the collect; unfortunate indeed. That would push the error handling down the consumer :/ ...

crates/amaru-consensus/src/block_validator.rs
KtorZ: >  This part here assumes that select_chain did the right thing.

This is kind of my point; this ended up being violated before, which has led to this whole PR; but how is it fixed now?

crates/amaru-ledger/src/state.rs
KtorZ: Ah! The slot battles. I forgot about that one. Okay.

crates/amaru-consensus/src/block_validator.rs
rkuhn: wait, this sounds wrong: why and how would `select_chain` guarantee that the switch arriving at `validate_block` only ever leads to a single block fork switch? that can’t be its responsibility, with fetch_blocks messing up things in between. perhaps we need to have a larger discussion.

rkuhn: As far as my design contract goes, fetch_block ensures that each block it sends downstream has all its ancestors also in the store. The job of validate_block is to validate the whole chain up to this new point, no matter how many blocks are new and no matter on which fork the ledger previously was. There are no further guarantees.

PR #1196, opened 2026-08-11
Fix Haskell Conformance Test Runner

yHSJ:
During work on #1138, I discovered that the Haskell test runner doesn't actually work anymore, after recent changes to the fixture structure.

I also discovered some other bugs with the test runner, since it's drifted from the Amaru implementation. Finally, there are some changes to the `make` Command to fix issues I ran into while going through the flow.

NOTE: The Haskell code was Claude-ed, as I am not a Haskell developer. It has been tested locally to confirm it gives the expected results.

Skip-Changelog

## Summary by CodeRabbit

- **New Features**
  - Added support for conformance scenarios containing expected validation outcomes.
  - Added fixture support for seeded governance proposals, proposal roots, treasury, and reserve balances.
  - Expanded validation coverage for governance, UTxO, certificates, reference scripts, and additional transaction failures.

- **Bug Fixes**
  - Validation mismatch messages now clearly distinguish expected and actual results.
  - Missing test fixtures now produce actionable errors.

- **Documentation**
  - Updated setup guidance, examples, and toolchain information.

- **Tests**
  - Improved prerequisite checks with warnings for optional or locally vendored tools.

### DIFF
--- a/crates/amaru/tests/conformance/validation-rules/src/Command/ValidatePhaseOne/Run.hs
+++ b/crates/amaru/tests/conformance/validation-rules/src/Command/ValidatePhaseOne/Run.hs
@@ -30,6 +30,7 @@ import Cardano.Ledger.Conway.Rules
     , ConwayCertsPredFailure (..)
     , ConwayDelegPredFailure (..)
     , ConwayGovCertPredFailure (..)
+    , ConwayGovPredFailure (..)
     , ConwayLedgerPredFailure (..)
     , ConwayUtxoPredFailure (..)
     , ConwayUtxowPredFailure (..)
@@ -51,6 +52,9 @@ import Cardano.Ledger.Shelley.LedgerState
     , LedgerState (..)
     , NewEpochState (..)
     , UTxOState (utxosUtxo)
+    )
+import Cardano.Ledger.Shelley.Rules
+    ( ShelleyPoolPredFailure (..)
     )
 import Cardano.Ledger.Shelley.StabilityWindow
     ( computeRandomnessStabilisationWindow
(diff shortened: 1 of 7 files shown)


crates/amaru/tests/conformance/validation-rules/src/Command/ValidatePhaseOne/Run.hs
```diff
@@ -449,3 +484,29 @@ normalizeGovCertFailure = \case
...
+    TreasuryWithdrawalReturnAccountsDoNotExist{} ->
+        "TreasuryWithdrawalReturnAccountsDoNotExist"
+    InvalidPrevGovActionId{} ->
+        "InvalidPrevGovActionId"
+    ZeroTreasuryWithdrawals{} ->
+        "TreasuryWithdrawalsAllZeros"
+    VotersDoNotExist{} ->
+        "VotersDoNotExist"
+    otherFailure ->
+        "unsupported:" <> showText otherFailure
```
etorreborre, line 512: Possibly we should put all the cases with "unsupported in amaru"? And thenbe forced to wonder what to do if a new enum variant becomes supported in the future?

crates/amaru/tests/conformance/validation-rules/src/Data/Fixture/InitialState.hs
```diff
@@ -372,6 +475,44 @@ toLedgerDRepEntry RegisteredDRep{credential, deposit, validUntil} = do
...
+
+buildProposals
+    :: GovRelation StrictMaybe
+    -> [GovActionState ConwayEra]
+    -> Either Error (Proposals ConwayEra)
+buildProposals roots =
+    foldlM addAction (def & pRootsL .~ fromPrevGovActionIds roots)
+  where
+    addAction acc proposalState =
+        maybe (Left (unplaceable proposalState)) Right (proposalsAddAction proposalState acc)
```
etorreborre, line 508: This might be a bit more idiomatic
```haskell
import Data.Maybe (maybeToRight)

maybeToRight (unplaceable proposalState) (proposalsAddAction proposalState acc)
```

Another version, since `Relude` is imported is, "add action and turn `Nothing` into an error":
```haskell
proposalsAddAction proposalState acc & maybeToRight (unplaceable proposalState)
```

But you can ignore it, that's mostly me code-golfing with Haskell :-)

KtorZ (approved):
Looks okay; we should really consider running this as part of the CI.

etorreborre (approved):
🏇

PR #1197, opened 2026-08-12
rebuild only when needed, demote noisy log

rkuhn:
- ledger DB directories were triggering rebuilds of `amaru`, which I don’t think is useful -- please correct me if wrong
- peer snapshot download metadata was triggering rebuilds of `amaru-node`, which was incorrect (moved that file to $out_dir)
- printing span close for every single gov prop at INFO level seems inappropriate to me, the aggregate count is still there and I’ve seen INFO log when a gov prop is opened

Skip-changelog

---

## Summary by CodeRabbit

- **Developer Experience**
  - Added automatic snapshot sharing across Git worktrees.
  - Improved hook installation for standard repositories and linked worktrees.
  - Reused cached snapshots and avoided rewriting unchanged generated files.

- **Tests**
  - Stake-distribution and ledger tests now warn and continue when snapshots are missing.

- **Documentation**
  - Clarified peer snapshot setup, offline usage, caching, rebuild behavior, and test data refresh guidance.

- **Observability**
  - Reduced proposal-ratification log verbosity while preserving enactment details.

### DIFF
--- a/crates/amaru/tests/summary.rs
+++ b/crates/amaru/tests/summary.rs
@@ -19,7 +19,7 @@ use std::{
     fmt::Write,
     fs::File,
     io::Read,
-    path::PathBuf,
+    path::{Path, PathBuf},
     sync::{Arc, LazyLock, Mutex},
 };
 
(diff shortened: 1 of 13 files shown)


KtorZ:
> ledger DB directories were triggering rebuilds of amaru, which I don’t think is useful -- please correct me if wrong

It is useful; this is how we trigger the epoch boundary tests. In principle this only happens once: if the folders don't exists, they are created and then, the build is skipped. 

Of course, it also happens if the content of the ledger.db changes, because that means new/different e2e snapshot tests are available.

rkuhn:
@KtorZ thanks for the explanation! Could you be more specific about what exactly is required to make this useful? I’d like to avoid that `amaru` is always rebuilt even if nothing changed.

rkuhn:
Discussion notes:
- need to generate test scenarios, should happen automatically
- perhaps we can monitor the snapshots instead of the ledger DB, then have the test routines dynamically decide whether to test or just pass

rkuhn:
@KtorZ does this now look sane to you?

crates/amaru/tests/summary.rs
```diff
@@ -80,6 +83,18 @@ fn compare_stake_distribution_with_haskell_node(
...
+    if !has_ledger_snapshot(&ledger_dir, epoch) {
+        // Soft-skip: missing snapshots used to be `#[ignore]`d at build time. Print a warning so
+        // `--nocapture` (or harness failure logs) still surfaces the gap.
+        eprintln!(
+            "warning: skipping stake distribution comparison for {network} epoch {epoch}; \
+             local ledger snapshot missing at {}",
+            ledger_dir.join(epoch.to_string()).display()
+        );
+        return Ok(());
+    }
```
KtorZ, line 96: 👍

KtorZ (approved):
LGTM

PR #1200, opened 2026-08-12
feat: filter tags in json output

jeluard:
Make sure `amaru.tag` fields end up in json traces

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tracing output across default and JSON formats.
  * Internal tag fields are now omitted from serialized JSON while standard tracing fields remain available.
  * JSON output remains valid and preserves field merging behavior.

### DIFF
--- a/crates/amaru/src/observability.rs
+++ b/crates/amaru/src/observability.rs
@@ -42,11 +42,11 @@ use tracing::{
 };
 use tracing_subscriber::{
     EnvFilter, Registry,
-    field::{MakeVisitor, VisitFmt, VisitOutput},
+    field::{MakeVisitor, RecordFields, VisitOutput},
     filter::Filtered,
     fmt::{
         FmtContext, FormatEvent, FormatFields, FormattedFields, Layer,
-        format::{DefaultFields, FmtSpan, Format, Json, JsonFields, Writer},
+        format::{DefaultFields, FmtSpan, Format, Json, JsonFields, JsonVisitor, Writer},
     },
     layer::{Context, Filter, Layered, SubscriberExt},
     prelude::*,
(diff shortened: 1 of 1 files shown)


KtorZ (changes requested):
No; why? We want them there.

jeluard:
Why do we want them here? Aren't they only useful at the oltp level for filtering?

crates/amaru/src/observability.rs
```diff
@@ -207,14 +193,102 @@ impl<Out, V: VisitOutput<Out>> VisitOutput<Out> for HideTagVisitor<V> {
...
+    visitor.finish()
+}
+
+fn remove_json_tag_fields(formatted: &mut String) -> fmt::Result {
+    if !formatted.contains(TAG_FIELD_PREFIX) {
+        return Ok(());
     }
+
+    let mut fields: serde_json::Map<String, serde_json::Value> =
+        serde_json::from_str(formatted).map_err(|_| fmt::Error)?;
```
rkuhn, line 225: ah, this approach should no longer be available on `main`, I replaced the deser/mod/ser dance with a proper map

rkuhn (commented):
No matter how the discussion with @KtorZ ends, it would be good to add some tests that document the intended full contents of example JSON traces, also with nested spans in scope, so that we all know what we are speaking of. I’m also confused by the subtle differences between the various export stacks (console, JSON, OTLP, TUI — everything is slightly different).

KtorZ:
> Why do we want them here? Aren't they only useful at the oltp level for filtering?

They are useful for filtering, to any consumer. The JSON traces are yet-another-way to get traces out of Amaru for those who don't want to use OTLP. But it has to provides the same level of observability.

jeluard:
cc @etorreborre

etorreborre:
@jeluard @KtorZ ok. One issue originally I had was duplicated tags because spans were inheriting tags from their parent when there were nested tags. Now I don't see any `amaru.tag` anymore with `--with-json-traces`, what's up with that?

@rkuhn we used to have a test for the traces output when the simulation was executed. Once it's back we will be able to resurrect that test.

rkuhn:
@etorreborre : #1210 fixes the regression you noted, includes the deduplication (discussed with @KtorZ to only keep fields from the innermost wrapping span), and adds unit tests for the shape of the traces of all subscriber stacks we have.

PR #1203, opened 2026-08-12
Reduce stake distribution / rewards churn allocation

KtorZ:
These little changes have noticeable effects on the transient
  allocations and pointless work that is done creating temporary
  structures; the reward calculation and stake distribution are
  significantly faster:

  - distribution build: 62.704s -> 37.453s
  - Rewards stake summary: 63.991s -> 36.205s
  - Rewards calculation: 102.687s -> 70.352s

  This effect is also visible throughout the whole sync, and provides
  ~10-12% performance increase for epoch sync:

  - Wall time: 2279.956s -> 1993.132s (-12.6%)
  - Mean epoch sync: 557.035s -> 489.954s (-12.0%)

  It also helps with the overall peak heap: 696.69 MiB -> 636.80 MiB
  (-8.6%), though has no observable effect on the RSS (mean or max);
  which was kind of the main driver for this change.

  So, probably good to keep around due to the other benefits without
  much counterpart (if only, a bit more code...).

## New Features
- Added support for reading stake distribution entries in modern and legacy formats.
- Improved handling of Cardano address types, stake credentials, and lovelace values.
- Enhanced reward and account-state processing.

## Performance
- Reduced allocation overhead during stake distribution and rewards calculations.

## Bug Fixes
- Improved validation and error handling for malformed stake distribution data.
- Preserved compatibility with supported legacy data encodings.
- Improved the timeliness of state updates after successful block processing.

### DIFF
--- a/crates/amaru-kernel/src/cardano/address/address_type.rs
+++ b/crates/amaru-kernel/src/cardano/address/address_type.rs
@@ -1,0 +1,59 @@
+// Copyright 2026 PRAGMA
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+#[repr(u8)]
+pub enum AddressType {
+    /// Base Key/Key Shelley address
+    Type0 = 0,
+    /// Base Script/Key Shelley address
+    Type1 = 1,
+    /// Base Key/Script Shelley address
+    Type2 = 2,
+    /// Base Script/Script Shelley address
+    Type3 = 3,
+    /// Pointer Key Shelley address deprecated since Conway
+    Type4 = 4,
+    /// Pointer Script Shelley address deprecated since Conway
+    Type5 = 5,
+    /// Payment Key Shelley address (a.k.a. Key Enterprise address)
+    Type6 = 6,
+    /// Payment Script Shelley address (a.k.a Script Enterprise address)
+    Type7 = 7,
+    /// Byron / Bootstrap address
+    Type8 = 8,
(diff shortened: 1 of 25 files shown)


crates/amaru-kernel/src/cardano/address/address_type.rs
```diff
@@ -0,0 +1,59 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+#[repr(u8)]
+pub enum AddressType {
```
yHSJ, line 17: I kind of hate these names, but it's understandable. I would prefer the descriptive comments be the actual variant names...

crates/amaru-kernel/src/cardano/stake_entry.rs
```diff
@@ -0,0 +1,108 @@
...
+}
+
+fn decode_modern<C>(d: &mut cbor::Decoder<'_>, ctx: &mut C) -> Result<StakeEntry, cbor::decode::Error> {
+    let (credential, lovelace) = cbor::heterogeneous_map(
+        d,
+        (None, None),
+        |d| d.u8(),
+        |d, state, field| {
+            match field {
+                0 => state.0 = Some(StakeCredential::from_raw_address(d.bytes()?)),
```
yHSJ: `d.bytes()` does not handle indefinite length bytes. #1172 introduced `decode_bytes`, which we should use instead.

yHSJ (approved):
Looks good, except for the `d.bytes` call (which we discussed in Discord).

crates/amaru-kernel/src/cardano/stake_entry.rs
etorreborre: See https://github.com/pragma-org/amaru/pull/1217

crates/amaru-kernel/src/cardano/address/address_type.rs
KtorZ: I hate every name that includes the word `State`, `Status`,`Type`, `Kind`, ... but it's just so often necessary :(

yHSJ: I meant more the `Type0`, `Type1`, etc... (GitHub didn't show that in the conversation for some reason)

Could be named more descriptively, but not the end of the world either.

PR #1204, opened 2026-08-13
Tweak jemalloc configuration

KtorZ:
This magic line is the result of multiple benchmarks evaluating different configurations for isolated mainnet runs over 10 epochs. This gives a good balance between allocation time and memory management overall:

#### vanilla

```
- RSS max: `1.29 GiB`
- RSS mean: `797.84 MiB`
- Heap live max: `621.30 MiB`
- Heap live mean: `251.98 MiB`
- Heap peak: `696.69 MiB`
- Mean epoch sync time: `557.035 s`
- Median epoch sync time: `588.252 s`
- P95 epoch sync time: `595.890 s`
```

#### background_thread:true,narenas:2,dirty_decay_ms:250,muzzy_decay_ms:0,tcache_max:512

```
- RSS max: `1.16 GiB`
- RSS mean: `717.63 MiB`
- Heap live max: `620.61 MiB`
- Heap live mean: `252.21 MiB`
- Heap peak: `664.54 MiB`
- Mean epoch sync time: `581.295 s`
- Median epoch sync time: `597.157 s`
- P95 epoch sync time: `627.741 s`
```

(note that some of sync time is recovered with #1203)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -108,7 +108,7 @@ tar = "0.4.46"
 tempfile = "3.27.0"
 test-case = "3.3.1"
 thiserror = "2.0.19"
-tikv-jemallocator = { version = "0.7.0" }
+tikv-jemallocator = { version = "0.6.1" }
 tokio = { version = "1.53.0", features = ["sync"] }
 tokio-util = "0.7.18"
 toml = "1.1.3"
--- a/.github/workflows/release-begin.yml
+++ b/.github/workflows/release-begin.yml
@@ -179,6 +179,7 @@ jobs:
         shell: bash
         env:
           BUILT_OVERRIDE_amaru_PKG_VERSION_PATCH: ${{ needs.build-metadata.outputs.release_patch }}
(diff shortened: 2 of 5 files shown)


Cargo.toml
```diff
@@ -108,7 +108,7 @@ tar = "0.4.46"
 tempfile = "3.27.0"
 test-case = "3.3.1"
 thiserror = "2.0.19"
-tikv-jemallocator = { version = "0.7.0" }
+tikv-jemallocator = { version = "0.6.1" }
```
jeluard, line 111: Any reason not to use the latest version? Could we add a comment so that it's not updated by accident?

KtorZ: It's only a dependency issue. Rocksdb "jemalloc" feature only works with 0.6.1; and somehow the solver cannot resolve this and the build just throw an error stating the divergence. So no need for a comment I believe; we should update if we can, but that depends on the rocksdb dependency. 

I've also inspected the changelog and there's not much we're missing from 0.7.* anyway.

PR #1206, opened 2026-08-13
chore: unify BlockHeader and Header

KtorZ:
The main motivation for this is not only the code cleanup that comes with it, but also an attempt at getting rid of the constant re-serialisation of headers for hash computation. This is not only inefficient but also potentially wrong should the header be constructed using a non 'canonical' cbor.
    
The `recompute_hash` function was identified as one of the main culprit using DHat; and instead of patching over it I thought we might as well kill two birds on stone and unify these two types and take the opportunity to memoized the hash during decode; entirely avoiding the need for `recompute_hash`.

Doing that, I discovered another location where we re-serialize header body: the KES signature verification. For now, I've thus chosen to memoized the bytes during the decoding but this is slightly unsatisfactory as the header is quite large (hundred of bytes / kilobyte). It'll get bigger with Leios and Peras too. However, KES signature is verified against a hash digest of the body; a better approach would therefore be to pre-compute and store that digest instead. 

## Summary by CodeRabbit

* **New Features**
  * Added richer header access, including block metadata, hashes, signatures, and chain points.
  * Added a direct conversion from fixed-size byte arrays and a method to retrieve block heights as numbers.
* **Bug Fixes**
  * Improved block and header decoding, validation, storage, and chain processing consistency.
* **Documentation**
  * Updated trace documentation and schemas to report `block_id` instead of `block_body_hash`.
* **Refactor**
  * Standardized block processing and public interfaces around the unified `Header` type.


crates/amaru-kernel/src/cardano/header.rs
```diff
@@ -12,23 +12,320 @@
...
+                let signature = d.decode_with(ctx)?;
+                Ok((body, body_original_bytes, signature))
+            })
+        })?;
+
+        let hash = Hasher::<{ HEADER * 8 }>::hash(header_bytes);
+
+        Ok(Self { body, body_original_bytes: body_original_bytes.to_vec(), signature, hash })
+    }
+}
```
KtorZ, line 210: Core of the changes are here ☝️

crates/amaru-kernel/src/cardano/header.rs
```diff
@@ -12,23 +12,320 @@
...
+
+impl fmt::Display for Header {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str(&format!(
+            "{}. {}{}",
+            self.slot(),
+            self.hash,
+            self.parent_hash().map(|p| format!(" ({p})")).unwrap_or_default()
+        ))?;
```
rkuhn, line 53: This should `write!(...)` instead of building multiple short-lived String allocations.

crates/amaru-kernel/src/cardano/header.rs
```diff
@@ -12,23 +12,320 @@
...
+    //   <https://docs.rs/ed25519-dalek/3.0.0/ed25519_dalek/struct.VerifyingKey.html#method.verify_prehashed_strict>
+    body_original_bytes: Vec<u8>,
+
+    signature: Bytes,
+}
+
+impl fmt::Display for Header {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str(&format!(
+            "{}. {}{}",
```
rkuhn, line 49: Why the space?

crates/amaru-kernel/src/cardano/header.rs
```diff
@@ -12,23 +12,320 @@
...
+        ))?;
+        Ok(())
+    }
+}
+
+impl fmt::Debug for Header {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("Header")
+            .field("hash", &hex::encode(self.hash()))
+            .field("slot", &self.slot().as_u64())
```
rkuhn, line 62: This should also print the block height.

crates/amaru-kernel/src/cardano/header.rs
```diff
@@ -12,23 +12,320 @@
...
+    ///
+    // TODO: replace with its sha512 hash by using dalek's `verify_prehashed_strict` down the line:
+    //
+    //   <https://docs.rs/ed25519-dalek/3.0.0/ed25519_dalek/struct.VerifyingKey.html#method.verify_prehashed_strict>
+    body_original_bytes: Vec<u8>,
+
+    signature: Bytes,
+}
+
+impl fmt::Display for Header {
```
rkuhn, line 46: What is the design goal of this representation? As it stands, it basically is `Point` plus a parent hash, which I think should be distinguished better. If I want to display the point I can just get that; do we even need Display on Header?

crates/amaru-kernel/src/cardano/header.rs
```diff
@@ -12,23 +12,320 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str(&format!(
+            "{}. {}{}",
+            self.slot(),
+            self.hash,
+            self.parent_hash().map(|p| format!(" ({p})")).unwrap_or_default()
```
rkuhn, line 52: This part shouldn’t be optional, the genesis header should be marked clearly as such.

crates/amaru-kernel/src/cardano/header.rs
```diff
@@ -12,23 +12,320 @@
...
+
+    /// Original bytes kept around for KES signature verification. This is a bit unfortunate as it
+    /// makes header ~1KB larger to store in memory and clone.
+    ///
+    // TODO: replace with its sha512 hash by using dalek's `verify_prehashed_strict` down the line:
+    //
+    //   <https://docs.rs/ed25519-dalek/3.0.0/ed25519_dalek/struct.VerifyingKey.html#method.verify_prehashed_strict>
+    body_original_bytes: Vec<u8>,
+
+    signature: Bytes,
```
rkuhn, line 35: Which signature exactly is this? (I could probably find out, but it would be easier for new or future readers to see a doc comment)

crates/amaru-kernel/src/cardano/header.rs
```diff
@@ -12,23 +12,320 @@
...
+        let ((body, body_original_bytes, signature), header_bytes) = cbor::tee(d, |d| {
+            cbor::heterogeneous_array(d, |d, assert_len| {
+                assert_len(2)?;
+                let (body, body_original_bytes) = cbor::tee(d, |d| d.decode_with(ctx))?;
+                let signature = d.decode_with(ctx)?;
+                Ok((body, body_original_bytes, signature))
+            })
+        })?;
+
+        let hash = Hasher::<{ HEADER * 8 }>::hash(header_bytes);
```
rkuhn, line 206: It should probably be highlighted in the `Header::new()` doc comment that that method is exclusively to be used for headers created by the local node — re-creating a Header from body and signature is simply not possible. Correct?

crates/amaru-observability/src/schemas.rs
```diff
@@ -487,7 +487,7 @@ define_schemas! {
             block_validation_context {
                 /// Create validation context for a block
                 public CREATE {
-                    required block_body_hash: amaru_kernel::HeaderHash
+                    required block_id: amaru_kernel::HeaderHash
```
rkuhn, line 493: this is a bugfix, right?

crates/amaru-kernel/src/cardano/header.rs
```diff
@@ -12,23 +12,320 @@
...
+    size::{HEADER, POOL_COLD_KEY},
+};
+
+#[derive(Clone, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
 pub struct Header {
-    #[n(0)]
-    pub header_body: HeaderBody,
+    /// Header hash computed upon deserialisation and kept in memory to avoid needless
+    /// re-serialisations.
+    hash: HeaderHash,
```
rkuhn, line 30: perhaps document that we keep this in serde for replay to work correctly (in case serializing HeaderBody yields different bytes than `body_original_bytes`)

crates/amaru-kernel/src/cardano/header.rs
```diff
@@ -12,23 +12,320 @@
...
+
+    body: HeaderBody,
+
+    /// Original bytes kept around for KES signature verification. This is a bit unfortunate as it
+    /// makes header ~1KB larger to store in memory and clone.
+    ///
+    // TODO: replace with its sha512 hash by using dalek's `verify_prehashed_strict` down the line:
+    //
+    //   <https://docs.rs/ed25519-dalek/3.0.0/ed25519_dalek/struct.VerifyingKey.html#method.verify_prehashed_strict>
+    body_original_bytes: Vec<u8>,
```
rkuhn: (de)serializing these bytes in serde is even more costly; could we do this KES improvement right away?

rkuhn (commented):
Great work! Some discussions remaining, though.

crates/amaru-kernel/src/cardano/header.rs
KtorZ: No idea. This is preexisting code that I simply ported over and preserved as-is. We can change; it might break some tests.

crates/amaru-kernel/src/cardano/header.rs
KtorZ: There's no such thing as a "genesis header". The genesis is a configuration file, not a block. I know we've been making this approximation and followed a convention in some places of the code where genesis hash = null hash; but that has always bothered me. 

Also; like the other, that's just existing code from `BlockHeader` that's been ported over here.

crates/amaru-kernel/src/cardano/header.rs
KtorZ: Correct; the only constructor is only available for tests. So, when we get into block forging, this will likely need to change.

crates/amaru-kernel/src/cardano/header.rs
rkuhn: Okay, then I vote for removing the Display instance until we have a clear picture what it should do.

crates/amaru-kernel/src/cardano/header.rs
rkuhn: Yes, that was sloppy wording on my part. What I meant is that the printed text should say “there is no parent, you have reached the end of the road”.

crates/amaru-observability/src/schemas.rs
KtorZ: Technically, yes.

crates/amaru-kernel/src/cardano/header.rs
KtorZ: The header's! That's why it's in the `Header` type 😄 ...

crates/amaru-kernel/src/cardano/header.rs
KtorZ: I can try.

KtorZ: So, I ended up reverting this because:

1. Using `verify_prehashed_strict` turned out to be a dead-end. Unfortunately, when using the pre-hashed version, the signature algorithm changes from Ed25519 to Ed25519ph which adds an extra padding to the hashed bytes; and so unless the signature was also produced with that padding, the verification would obviously not compute. There's a world where we could fork the library and make it work but ... that seemed to be the wrong rabbit to follow. 

2. The header bytes are (surprisingly) mostly _canonical_. From what I've checked, there should be only one way to encode and decode headers; which makes the round-tripping less of a problem and we only do it in one place: when verifying the KES signature. 

3. So I contemplated for a moment having two types: RawHeader (with the original bytes) and Header (without them); with the idea to carry RawHeader up to the signature verification, and ditch Header after that. But, the changes were impacted much more parts of the consensus than I expected and it felt like everything was storing RawHeader in the end, loosing the benefit a little. 

Beside, I've been surprised by the number of calls to `validate_header` which happened in multiple places and looked surprising to me. I would be glad if we could look into this in a pairing session one day.

PR #1207, opened 2026-08-13
VRF Key Hash Validations

### LINKED ISSUE #910: Ledger predicate: VRFKeyHashAlreadyRegistered
### Abstract

Implement predicate `VRFKeyHashAlreadyRegistered`

> [!IMPORTANT]
> This predicate is only enabled post Van Rossem hard fork in pv 11

### Why?

Conformance

### How?

[Reference Implementation](https://github.com/IntersectMBO/cardano-ledger/blob/0cfbf861cfb456660a7b73281c6fb714a53d40f9/eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Pool.hs#L263-L268) (new pool)

[Reference Implementation](https://github.com/IntersectMBO/cardano-ledger/blob/0cfbf861cfb456660a7b73281c6fb714a53d40f9/eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/Pool.hs#L279-L283) (re-registered pool)

### Discussion points

_No response_

_No response_

### Checklist

yHSJ:
Close #910 

This PR introduces logic to validate that VRF key hashes are not used by multiple pools. In order to do this, there had to be cascading changes to the database schema, the volatile structs, and the context.

## Summary by CodeRabbit

* **New Features**
  * Added VRF key-hash occupancy tracking for stake pools.
  * Prevented duplicate VRF key-hash registrations from protocol version 11 onward, while allowing pools to retain their current key.
  * Added support for tracking VRF key changes, releases, retirements, and pending registrations across epoch transitions.
  * Bootstrap imports now preserve VRF key-hash occupancy data.

* **Bug Fixes**
  * Added validation coverage for conflicting, retained, fresh, and legacy-protocol VRF key registrations.

* **Documentation**
  * Updated changelog, trace documentation, schemas, and transaction fixture formats.


KtorZ:
@yHSJ I am holding off a bit on this one because we currently have a sync issue on mainnet due to previously merged rules (unknown voter).

crates/amaru-ledger/src/context/default/preparation.rs
```diff
@@ -203,47 +212,126 @@ fn resolve_inputs<'block, 'volatile>(
...
 fn resolve_pools(
-    volatile: &impl VolatileState<Pool = <VolatileDB as VolatileState>::Pool>,
+    volatile: &impl VolatileState<
+        Pool = <VolatileDB as VolatileState>::Pool,
+        PoolVrfs = <VolatileDB as VolatileState>::PoolVrfs,
+    >,
     db: &impl ReadStore,
     mut keys: impl Iterator<Item = PoolId>,
-) -> Result<BTreeSet<PoolId>, ContextHydratationError> {
+) -> Result<BTreeMap<PoolId, PoolVrfs>, ContextHydratationError> {
```
KtorZ, line 230: That is weird; the VRF should not be indexed by pool ids at this point? Otherwise we need to traverse the entire map every time we must know whether the VRF is used.

crates/amaru-ledger/src/context/default/preparation.rs
```diff
@@ -203,47 +212,126 @@ fn resolve_inputs<'block, 'volatile>(
...
+}
+
+/// The in-use subset of the provided candidate VRF key hashes, layering the volatile window's
+/// occupancy verdicts over the stable counters: a volatile claim or release settles a key, and a
+/// deferred verdict weighs the stable count against the pending boundary decrements it carries.
+fn resolve_vrf_key_hashes(
+    volatile: &impl VolatileState<VrfKeyHash = VrfOccupancy>,
+    db: &impl ReadStore,
+    mut keys: impl Iterator<Item = pools_vrf::Key>,
+) -> Result<BTreeSet<pools_vrf::Key>, ContextHydratationError> {
```
KtorZ, line 298: I don't think a set can work due to duplicate pre-dating the hard fork that introduces this restriction? It should be a map to a counter -- unless you've checked separately that there was _actually_ no duplicate on any of the network?

crates/amaru-ledger/src/context/default/preparation.rs
```diff
@@ -203,47 +212,126 @@ fn resolve_inputs<'block, 'volatile>(
...
+
+                VrfOccupancy::Released => {
+                    from_volatile += 1;
+                }
+
+                VrfOccupancy::Deferred(decrements) => {
+                    let count = db
+                        .vrf_key_hash(&vrf)
+                        .map_err(ContextHydratationError::ResolveVrfKeyHashes)?
+                        .unwrap_or_default();
```
KtorZ, line 318: I don't think I am fond of this model. It diverges quite a lot from the Haskell node and .. given that the current VRF check implementation seems buggy, it would be better if we sticked with an implementation that's closer to Haskell's; because it will make it easier to re-implement bugs.

yHSJ:
Closing this PR in favor of #1247. I am not yet deleting the branch, as this PR has some of the Haskell bugs implemented correctly, which #1247 does not yet, but the model in #1247 is much nicer.

PR #1209, opened 2026-08-13
fix(amaru-pure-stage): avoid unnecessary string allocation in pure-stage's try_cast and cast_ref

KtorZ:
This commits does two things really:

  1. It avoids debug-serialiasing the entire type in the error message unless we are at the TRACE level
  2. It defines a proper CastError that carries type names instead of a string message; the message is only produced when the error is displayed (if displayed).

This was identified as the number 1 churn allocation cost center from a dhat's run.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error messages when message type casting fails.
  * Error details now identify the expected and received types by default.
  * Full source-value details are included only when trace logging is enabled.
  * Reduced unnecessary error-detail generation for shorter, more efficient failures.

* **Documentation**
  * Added a changelog entry describing the updated casting error behavior.

### DIFF
--- a/crates/amaru-pure-stage/src/types.rs
+++ b/crates/amaru-pure-stage/src/types.rs
@@ -24,6 +24,7 @@ use std::{
 
 use anyhow::Context;
 use cbor4ii::serde::from_slice;
+use poolshark::local::LPooled;
 use tokio::sync::mpsc;
 
 use crate::{
@@ -105,7 +106,7 @@ impl dyn SendData {
 
     /// Cast a message to a given concrete type.
     pub fn cast_ref<T: SendData>(&self) -> anyhow::Result<&T> {
-        (self as &dyn Any).downcast_ref::<T>().ok_or_else(|| CastError::anyhow::<T>(self))
+        (self as &dyn Any).downcast_ref::<T>().ok_or_else(|| CastError::new::<T>(self).into())
     }
 
     pub fn try_cast<T: SendData>(self: Box<Self>) -> Result<Box<T>, Box<Self>> {
@@ -119,7 +120,7 @@ impl dyn SendData {
 
     /// Cast a message to a given concrete type, yielding an informative error otherwise
     pub fn cast<T: SendData>(self: Box<Self>) -> anyhow::Result<Box<T>> {
-        self.try_cast::<T>().map_err(|b| CastError::anyhow::<T>(&b))
+        self.try_cast::<T>().map_err(|b| CastError::new::<T>(&b).into())
     }
 
     pub fn cast_deserialize<T>(self: Box<Self>) -> anyhow::Result<T>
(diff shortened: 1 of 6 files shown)


crates/amaru-pure-stage/src/types.rs
```diff
@@ -182,6 +173,28 @@ impl PartialEq for dyn SendData {
     }
 }
 
+#[derive(Debug, thiserror::Error)]
+#[error("message type error: expected {expected}, got {got}")]
+pub struct CastError {
+    expected: &'static str,
+    got: &'static str,
```
rkuhn, line 181: How about adding the debug format for the `source` here as well, written into a String taken from something like [poolshark](https://docs.rs/poolshark/latest/poolshark/). The error is travelling up the stack and should probably not leave the thread anyway, so a tiny local pool would be enough and we still get full customer service. We can keep it empty when we don’t want to spend the CPU cycles for the debug print.

crates/amaru-pure-stage/src/types.rs
```diff
@@ -182,6 +173,28 @@ impl PartialEq for dyn SendData {
...
+}
+
+impl CastError {
+    pub fn anyhow<T>(source: &dyn SendData) -> anyhow::Error {
+        if tracing::enabled!(target: "amaru_pure_stage", tracing::Level::TRACE) {
```
rkuhn: I’d argue that such errors are DEBUG material, meaning we’d normally avoid it (because default is `amaru_pure_stage=warn` or usual user override `AMARU_LOG=info`). With the proposal above this should still be bearable.

rkuhn (commented):
Good find! I’d defer the full string build even further.

crates/amaru-pure-stage/src/types.rs
KtorZ: We still pay the cost of debug-serialization for stuff that isn't super useful IMO. From what I understand, the casting here is used as a control flow mechanism and so, "errors" are frequent and _most of the time_ we don't care about the source value?

crates/amaru-pure-stage/src/types.rs
KtorZ: We would avoid logging them yes, but the problem here is that we still pay the cost of constructing the error message, irrespective of it being logged or not. And this is because this happens outside of any tracing macro.

crates/amaru-pure-stage/src/types.rs
rkuhn: Well, yes, until you have to debug why a certain failure happened down the line. I haven’t had issues in a long while, but when I had them this debug print was extremely helpful. Most people shouldn’t enable DEBUG on pure-stage anyway, so most people won’t pay the overhead.

crates/amaru-pure-stage/src/types.rs
rkuhn: I’m arguing _for_ your change, just with `tracing::Level::DEBUG` instead.

KtorZ: I see. And, that sounds reasonable. 

I am not fond of the other comment to be honest. For the sake of fixing this _now_, I rather just proceed with the proposed changes (modulo the severity adjustment) if you're okay.

rkuhn: Okay, let’s do that for now. Just to be clear: my other suggestion strictly optimises resource usage, but it can be done later.

PR #1212, opened 2026-08-13
chore: leaner bootstrap

jeluard:
Various improvements to speedup bootstrap and reduce memory usage:
* better tx usage when loading utxos
* more efficient map decoding
* stream node state snapshots

Correctness validated with `AMARU_NETWORK=mainnet ./scripts/run-until-then-test --amaru 600 610`

Memory usage is now bellow 500MB.

Time to run `AMARU_NETWORK=mainnet amaru bootstrap` (with snapshots locally cached)

main: 3min12s
after batch utxo write: 2min34s
after skip_map: 1min49s
after stream node state: 2min50s
after more stream stuff: 1m11s

## Summary by CodeRabbit

* **Performance**
  * Improved bootstrap speed and memory efficiency through streaming snapshot, account, and UTxO imports.
  * Reduced peak resource usage when processing large snapshots.

* **Monitoring**
  * Added automated profiling for memory, CPU, swap, throttling, and out-of-memory events.
  * Reports include graphs, summaries, and downloadable performance data.

* **Reliability**
  * Improved validation and handling of snapshot state and UTxO data.
  * Added nightly checks for bootstrap results and resource usage.
  * Improved incremental persistence during large snapshot imports.


KtorZ:
Sweet. Any comparison from the memory usage?

jeluard:
@KtorZ see my updated PR summary

.github/workflows/nightly-non-functional-tests.yml
```diff
@@ -0,0 +1,81 @@
...
+        with:
+          persist-credentials: false
+
+      - uses: ./.github/actions/stage-peer-snapshots
+
+      - uses: ./.github/actions/cache
+
+      - name: Build Amaru
+        shell: bash
+        run: cargo build --release --bin amaru --locked
```
KtorZ: Not sure why you insist with those 😛 ... we don't have any cache for `--release`! If this runs nightly, it could as well download the latest release artifact, the URL is fixed (e.g. https://nightly.amaru.global/downloads/linux-x86_64.deb )

For this kind of test though, a dev build is fine. It's mostly about noticing regressions, not measuring production times.

.github/workflows/nightly-non-functional-tests.yml
KtorZ, line 1: Can we park this one in a separate PR? I am not sure about the 400 lines script in CI which will be quite hard to maintain.

CHANGELOG.md
```diff
@@ -74,6 +74,10 @@ Other guiding principles:
...
+- **amaru-bootstrap**: speed up node bootstrap by avoiding unnecessary optimistic-transaction conflict tracking when importing UTxO batches into a fresh database.
+- **amaru-bootstrap**: speed up node bootstrap by decoding large snapshot maps incrementally and avoiding unnecessary optimistic-transaction conflict tracking when importing UTxO batches into a fresh database.
+- **amaru-bootstrap**: speed up node bootstrap by streaming state archives, decoding large snapshot maps incrementally, and avoiding unnecessary optimistic-transaction conflict tracking when importing UTxO batches into a fresh database.
+- **amaru-bootstrap**: speed up node bootstrap by streaming state archives and account imports, decoding large snapshot maps incrementally, and avoiding unnecessary optimistic-transaction conflict tracking when importing fresh database batches.
```
KtorZ: These should really be one line I believe,

crates/amaru-ledger/src/bootstrap.rs
```diff
@@ -130,23 +130,380 @@ struct InitialSnapshot {
...
+    decode_map_with_decoder(
+        decoder,
+        |d| Ok((d.decode()?, d.decode()?)),
+        BTreeMap::new(),
+        |mut entries, (key, value)| {
+            entries.insert(key, value);
+            entries
+        },
+    )
+}
```
KtorZ, line 191: Please move what's reusable and generic under `amaru-minicbor-extra`, with dedicated tests and docs comments 🙏

.github/workflows/nightly-non-functional-tests.yml
jeluard: Makes sense, could as well be some other scripts to ensure no mem/cpu regression happens as I believe you have some too? Would be useful to have a common set of such that could be used both on CI and locally.

crates/amaru-ledger/src/store.rs
```diff
@@ -226,6 +226,47 @@ pub trait Store: ReadStore {
...
+    }
+
+    /// Save one batch of account rows while importing a bootstrap snapshot.
+    ///
+    /// Snapshot import calls this before the snapshot's protocol parameters have been decoded,
+    /// so stores must persist the already-normalized rows directly.
+    fn save_bootstrap_accounts(&self, _accounts: impl Iterator<Item = (accounts::Key, accounts::Row)>) -> Result<()> {
+        Err(StoreError::Internal("bootstrap account batches are not supported by this store".into()))
+    }
 }
```
KtorZ: I don't think these should exists. I acknowledge the `save` function is not super practical and comes with a lot of boilerplate at the moment, but we should not extend the store interface for helpers that rewrap existing stuff.

crates/amaru-minicbor-extra/src/decode/lazy.rs
```diff
@@ -59,9 +65,16 @@ impl<'a> LazyDecoder<'a> {
...
+                let offset = self.bytes.len();
+                self.bytes.resize(offset + Self::CHUNK_SIZE, 0);
+                let read = match self.reader.read(&mut self.bytes[offset..]) {
+                    Ok(read) => read,
+                    Err(error) => {
+                        self.bytes.truncate(offset);
+                        return Err(cbor::decode::Error::custom(error).into());
+                    }
+                };
+                self.bytes.truncate(offset + read);
```
KtorZ, line 117: Interesting 👍

crates/amaru-minicbor-extra/src/decode/lazy.rs
```diff
@@ -93,4 +106,186 @@ impl<'a> LazyDecoder<'a> {
...
+        self.stream_map(
+            |d| {
+                d.skip()?;
+                d.skip()
+            },
+            |_| (),
+            |_, _| Ok(()),
+        )
+    }
+}
```
KtorZ, line 180: Alternatively, we could define an alternative `skip()` function that is more holistic: 

- decode the `.datatype`
- lazy skip array and maps using the streaming approach
- normal skip for the other.

crates/amaru-ledger/src/bootstrap.rs
```diff
@@ -130,100 +130,315 @@ struct InitialSnapshot {
...
+            let amount = rewards.into_iter().fold(0, |total, reward| total + reward.amount);
+            Ok((credential, amount))
+        },
+        |length| {
+            (
+                with_progress(
+                    length.unwrap_or_default() as usize,
+                    "{spinner:.green} Applying reward updates {bar:40.green} [{pos:>7}/{len:7}] ({eta} remaining)",
+                ),
+                0_u64,
```
KtorZ, line 212: This should use `length` when available and default to the account size here as a fallback estimate for the denominator instead of `0`; the reward updates will be slightly smaller than the number of accounts in general, but that's a good approximate.

KtorZ (commented):
Locally, I am getting: 

```
ERROR amaru::cli: error description="decode ocert counters: end of input bytes"
```

jeluard:
@KtorZ I pushed a new commit recently so you might have to update your branch.

crates/amaru-ledger/src/store.rs
jeluard: Removed, at the cost of slower import (30s more)

KtorZ:
Tested with mainnet 600 & 640; preprod 270 and currently testing ALL preview snapshots between 1000 and 1100... will merge if they all pass.

PR #1213, opened 2026-08-13
feat: make sure missing snapshots fail tests

jeluard:
To avoid false positives, make sure stake tests fails if associated snapshot is missing.

### DIFF
--- a/crates/amaru/build/stake_distribution.rs
+++ b/crates/amaru/build/stake_distribution.rs
@@ -27,7 +27,8 @@ use crate::{emit_rerun_if_changed, write_if_changed};
 ///
 /// Availability of the local `ledger.<network>.db` snapshot is **not** checked here: that path is
 /// a live RocksDB for node runs, so watching it would rebuild `amaru` on every SST/LOG write.
-/// The generated tests check snapshot presence at runtime and soft-skip when missing.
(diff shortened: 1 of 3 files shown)


KtorZ (changes requested):
Nope; this was done on purpose. Before, we would take the intersection of what was available locally and the available conformance files. The build.rs was then generating corresponding tests and ignored tests.  

However, checking what's available locally would make the build "impure" and cause rebuild constant rebuilds. So instead, we perform this check dynamically now. And we cannot mark a test as "ignored" dynamically without instrumenting the whole test runner. So the alternative was to leave a warning. 

Either way, those tests have only run when the files were available to begin with. We don't want to require everyone to carry multiple hundred gigabytes of conformance files around.

jeluard:
Current behavior makes it easy to fool yourself believing some test went file while under the hood some environment issue happened. Didn't see any warning btw.

KtorZ:
> Current behavior makes it easy to fool yourself believing some test went file while under the hood some environment issue happened. Didn't see any warning btw.

That's why the `--no-capture` is for; the rust test runner only show println! one errors. The whole thing is a bit annoying but we didn't find any good way to trigger a clear warning _at runtime_ for tests; if you have suggestions, I am all ears 😅 ...

PR #1217, opened 2026-08-14
chore: add a clippy rule for using decode_bytes

etorreborre:
This PR adds a lint to clippy in order to warn for the use of `d.bytes()` when decoding bytes that can potentially accept definite and indefinite length bytes.

I adjusted some call sites to what the Haskell node is doing pre-version 12 (some places aren't strict anymore in v12) but this will have to be revisited when we support v12 in the future.

Skip-Changelog

## Summary by CodeRabbit

* **Improvements**
  * Standardized CBOR byte decoding across import, protocol, script, and benchmark workflows.
  * Byte data is handled with owned buffers, improving consistency and reducing unnecessary copying.
  * Clarified conformance behavior for unsupported indefinite-length CBOR byte strings.

* **Documentation**
  * Added guidance and examples for the recommended byte-decoding approach.
  * Documented existing validation behavior across supported decoding paths.

### DIFF
--- a/.cargo/clippy.toml
+++ b/.cargo/clippy.toml
@@ -4,13 +4,14 @@ allow-unwrap-in-tests = true
 type-complexity-threshold = 600
 enum-variant-size-threshold = 1024
 disallowed-types = [
-    "std::collections::HashSet",
-    "std::collections::HashMap"
+  "std::collections::HashSet",
+  "std::collections::HashMap"
 ]
 disallowed-methods = [
-    "slot_arithmetic::EraHistory::slot_to_relative_time_unchecked_horizon",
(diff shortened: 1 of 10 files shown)


jeluard:
I like the idea now it looks like we would want to ensure all other usages of `decode_bytes` are expected.
Also, we could instead be explicit and have a dedicated `decode_bytes_conformance` (find better name) instead of having the scary `#[allow(clippy::disallowed_methods)]` and comment repeated.

KtorZ (approved):

etorreborre:
@jeluard I like the idea that we have a scary warning on the cases that deviate from the norm but maybe the comment could be slimmer, yes.

PR #1219, opened 2026-08-14
fix: cost models encoding + value size calculation

etorreborre:
I found those 2 issues while synchronizing with `preprod`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility with Cardano node encoding for ledger value sizes, including non-minimal and large multi-asset values.
  * Corrected Plutus `CostModels` serialization to use language-keyed maps for V1–V3 models.
  * Updated transaction output handling to preserve value data consistently across ledger, bootstrap, and Plutus processing.

* **Changelog**
  * Documented the updated value-size and `CostModels` encoding behavior.


crates/amaru-ledger/src/rules/transaction/phase_one/outputs/value_size.rs
```diff
@@ -0,0 +1,91 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Value serialization following the Haskell cardano-ledger's encoding conventions.
+//!
+//! The `OutputTooBigUTxO` rule does not measure the bytes found on the wire: cardano-ledger
+//! re-serializes the value with its own encoder and compares that length against `maxValueSize`.
+//! That encoder writes maps of more than 23 entries with an indefinite-length header and a break byte,
+//! which is one byte shorter than the definite-length encoding used once a map holds 256 entries or more.
```
KtorZ: > which is one byte shorter than the definite-length encoding used once a map holds 256 entries or more.

FFS

crates/amaru-ledger/src/rules/transaction/phase_one/outputs/inherent_value.rs
```diff
@@ -29,7 +29,7 @@ pub fn execute(
     }
 
     let max_value_size = protocol_parameters.max_value_size;
-    let given_val_size = output.value.original_bytes().len();
```
KtorZ, line 32: That means we can also ditch those original bytes from the values instead of keeping them around.

etorreborre: Oh right! Removed

KtorZ (approved):

PR #1220, opened 2026-08-14
fold BlockHeight into Point and remove Tip

rkuhn:
- **fold Tip into Point**, network layer and other external interfaces (like mithril) use
  NetworkPoint for the `slot.hash` format
- **switch chain store to Point where beneficial**
- **bump chain store to v6 and add migrations**


KtorZ:
Reviewing the changes here; it seems that we are mostly trading `Tip` for `Point`, but also `Point` for `NetworkPoint`...

So I am not quite sur it's worth the trade and it seems that the distinction between `Tip` and `Point` existed for a good reason.

rkuhn:
You’re correct regarding the shape, but incorrect about the rest:

- Tip was always the wrong name, utterly confusing, because it is a “useful block identity”, i.e. a Point in our fractal space; it is by no means indicative of the referenced point being an actual tip!
- NetworkPoint has been banned from amaru-consensus apart from one place in track_peers where the information arrives from the network — within Amaru we only use the useful Point or the minimal Hash, which is much simpler to understand; other uses are in externally defined file formats like mithril, but this is properly contained

I’ll try to remove NetworkPoint completely from consensus to make it even clearer.

KtorZ:
What about we go for `SlotRef` and `BlockRef` then 😶 ?

crates/amaru-consensus/Cargo.toml
```diff
@@ -70,11 +70,6 @@ amaru-stores.workspace = true
 [target.'cfg(unix)'.dependencies]
 pprof = { version = "0.15", features = ["flamegraph"], optional = true }
 
-[[bench]]
-name = "headers_tree"
-harness = false
-required-features = ["test-utils"]
```
KtorZ, line 76: This one runs in CI currently: https://github.com/pragma-org/amaru/blob/main/.github/workflows/ci-build-and-test.yml#L113-L115

We probably want to remove it from there too, and perhaps add the actually relevant benches.

KtorZ (approved):
As far as it is possible to humanly review this, I've tried. I've also iterated a couple of times with the clanker without finding anything materially worrying.

rkuhn:
Thanks a lot @KtorZ , these were some cleanups that I have wanted to do for a long time.

PR #1221, opened 2026-08-14
Update Haskell Harness 

yHSJ:
Many tests were failing the Haskell evaluation because of drift. This PR fixes *almost* all the failing cases, except for `00044`, which is fixed in #1138.

This change also exposed a test that was passing in Amaru that shouldn't. I've fixed that bug, and fixed the test case to *actually* test what we want.

As always, the Haskell changes are claude-ed. They have been tested locally, but some feedback or contribution from a Haskell developer would be greatly appreciated.

## Summary by CodeRabbit

* **New Features**
  * Added richer governance proposal lineage tracking, including hard-fork protocol versions.
  * Added support for guardrail script hashes and proposal expiry details in validation scenarios.
  * Added parsing and display support for protocol versions and proposal lineage categories.

* **Bug Fixes**
  * Improved hard-fork sequencing validation for proposals following an in-flight hard fork.
  * Governance validation now reports clearer errors when proposals cannot follow the required sequence.

* **Tests**
  * Expanded conformance coverage for governance actions, proposal sequencing, guardrails, and script-hash validation.


crates/amaru-ledger/src/rules/transaction/phase_one/proposals.rs
```diff
@@ -246,6 +252,32 @@ where
...
+    context: &C,
+    parent: Option<&ProposalId>,
+    new_version: ProtocolVersion,
+    current_version: ProtocolVersion,
+) -> ProtocolVersion
+where
+    C: ProposalsSlice,
+{
+    let follows_root = parent == context.roots().hard_fork.as_ref();
+    let skips_major = new_version.major() > current_version.major() + 1;
```
KtorZ: I don't get where this `skips_major` is coming from.

PR #1225, opened 2026-08-16
pure-stage contramap no longer leaks memory

### LINKED ISSUE #762: Garbage collect contramapped stages
### Abstract

See https://github.com/pragma-org/amaru/pull/745/changes#r3065077470

### Why?

The current implementation of `pure-stage` `contramap` creates a pseudo-stage that won’t be removed when the last reference is dropped. This opens a possibility to create memory leaks.

### How?

The `StageRef::extra` field can be used to store the injection function from this ref’s message type to the original ref’s message type. The `name` would then remain the original stage name, no new names are allocated for contramapped refs. This will lead to significant internal cleanup of the StageOrAdapter enum that won’t be needed anymore.

The trace buffer shall then see only the transformed message type, meaning that a contramapped ref just runs the message type injection in the sending stage.

Unit tests.

### Discussion points

_No response_

_No response_

### Checklist

rkuhn:
The transformation function is no longer registered under its name, meaning that contramap can be
used without worrying about memory leaks. This also is one more reason why resuming a simulation
from a replay won’t work (since the mapped StageRef will not be rehydrated with the transformation
function), but that was already broken earlier (for `.call`) and is not currently a design goal.

fixes #762 

## Summary by CodeRabbit

- **New Features**
  - Added flexible message transformation support with preserved routing and composed transformations.
  - Improved delivery of transformed messages across protocol and node workflows.
  - Added clearer delivery failure information, including the unavailable destination.

- **Bug Fixes**
  - Improved chain synchronization initialization detection.
  - Strengthened message delivery handling when destinations are unavailable or terminated.

- **Tests**
  - Added coverage for transformations, payload delivery, dropped destinations, and handshake behavior.

- **Documentation**
  - Updated changelog entries for messaging and delivery behavior.


crates/amaru-pure-stage/src/tokio.rs
```diff
@@ -381,16 +373,26 @@ async fn run_stage_boxed(
...
+            Some(transform) => transform(Box::new(msg)),
+            None => Box::new(msg),
+        };
         Box::pin(async move {
-            tx.send(Box::new(msg)).await.map_err(|msg| *msg.0.cast::<Msg>().expect("message was just boxed"))
+            tx.send(payload).await.map_err(|msg| match msg.0.cast::<Msg>() {
+                Ok(msg) => *msg,
+                Err(_) => panic!("contramapped input send failed (stage terminated)"),
+            })
         })
```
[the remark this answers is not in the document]
rkuhn: correct, I removed the Msg from the error case instead because it wasn’t needed

crates/amaru-pure-stage/src/stage_ref.rs
```diff
@@ -30,7 +30,26 @@ impl<Msg, State, RefAux> StageBuildRef<Msg, State, RefAux> {
...
+/// than the destination stage.
+///
+/// `next` is only a leftover call-reply payload (`StageRefExtra` / `ScheduleId`), never another
+/// [`Injection`]: [`StageRef::contramap`] composes transforms into a single function.
+pub(crate) struct Injection {
+    pub transform: Arc<dyn Fn(Box<dyn SendData>) -> Box<dyn SendData> + Send + Sync>,
+    pub next: Option<Arc<dyn Any + Send + Sync>>,
+}
+
+impl Injection {
```
etorreborre, line 43: Why don't we store the `Injection` separately from the `extra` in a `StageRef`:
```
 struct StageRef<Msg> {
      name: Name,
      injection: Option<Arc<Injection>>,
      runtime_extra: Option<Arc<dyn Any + Send + Sync>>,
      _ph: PhantomData<Msg>,
  }
```
This way, we don't have to cast it.

crates/amaru-pure-stage/src/tokio.rs
```diff
@@ -381,17 +373,22 @@ async fn run_stage_boxed(
...
         let tx = tx.clone();
-        Box::pin(async move {
-            tx.send(Box::new(msg)).await.map_err(|msg| *msg.0.cast::<Msg>().expect("message was just boxed"))
-        })
+        let transform = transform.clone();
+        let payload = match transform {
+            Some(transform) => transform(Box::new(msg)),
+            None => Box::new(msg),
+        };
+        Box::pin(async move { tx.send(payload).await.map_err(|_| crate::SendError) })
```
etorreborre: Can / should the `SendError` be more specific and at least specify the target stage?

crates/amaru-pure-stage/src/stage_ref.rs
rkuhn: The honest answer is that I’m slightly paranoid regarding the memory footprint of a stage ref — even though that might not really be all that important. Plus cloning/dropping two arcs costs twice the atomic ops.

crates/amaru-pure-stage/src/tokio.rs
rkuhn: That is not typically the case for other libraries’ SendError types, but then again their channels don’t have names. It might help debugging, you’re right, will add it.

crates/amaru-pure-stage/src/stage_ref.rs
etorreborre: I think it would be worth the added clarity in that case, unless proven that it's detrimental.

PR #1226, opened 2026-08-17
feat: make sure CI runs snapshots tests too

jeluard:
## Summary by CodeRabbit

* **Tests**
  * Updated epoch snapshot testing to run through the target epoch and validate snapshots automatically.
  * Improved test observability by extending metric retention during test runs.
  * Standardized peer connectivity settings for more consistent snapshot test execution.

* **Chores**
  * Removed the obsolete build command and related configuration.
  * Simplified the snapshot testing workflow for more reliable automated validation.
  * Refreshed the database bootstrap cache to improve workflow consistency.


.github/workflows/template-ledger-epoch-snapshots.yml
```diff
@@ -158,10 +160,10 @@ jobs:
           # it takes many seconds to empty the log buffer after `run-until` exists, at which point the
           # metrics are no longer retained within the collector.
           AMARU_TRACE: "amaru=info"
-          AMARU_MAX_EXTRA_LEDGER_SNAPSHOTS: "all"
```
KtorZ, line 161: Tests won't run without that

jeluard: `run-until` now [hardcodes it](https://github.com/pragma-org/amaru/blob/main/crates/amaru-node/examples/run_until.rs#L57)

rkuhn (approved):
thanks for the catch and fix, LGTM as far as I understand

KtorZ (approved):

jeluard:
See https://github.com/pragma-org/amaru/pull/1231#issuecomment-5332434140

yHSJ:
@jeluard Fixed the CI problem by bumping the cache. The cache must've been constructed specifically during an epoch where there was a non-zero `consecutive_dormant_epochs` value, hence the fact it was still passing on the old cache on preview. That probably is the same issue that happened to you when running the script locally. I will confirm myself before merging, though.

yHSJ:
Confirmed it works locally.

PR #1227, opened 2026-08-17
Fix CC members preparations

KtorZ:
Two fixes; the first one was causing a sync issue on Mainnet epoch 602 due to the preparation step failing to correctly resolve the volatile hot delegation of a newly added member.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved committee membership resolution for certificates and votes.
  * Recent hot-credential delegations and newly added members are now recognized during vote processing.
  * Improved handling of inactive committee members in certificate processing.
  * Multi-member committee updates now correctly include all matching additions.
  * Expanded coverage for committee filtering, delegations, unelected members, and proposal updates.

### DIFF
--- a/crates/amaru-ledger/src/context/default/preparation.rs
+++ b/crates/amaru-ledger/src/context/default/preparation.rs
@@ -18,8 +18,8 @@ use std::{
 };
 
 use amaru_kernel::{
-    DRep, DRepRegistration, GovernanceAction, MemoizedTransactionOutput, PoolId, ProposalId, ProposalSlim,
-    ProposalsRoots, StakeCredential, TransactionInput, drep,
+    ConstitutionalCommitteeMemberStatus, DRep, DRepRegistration, GovernanceAction, MemoizedTransactionOutput, PoolId,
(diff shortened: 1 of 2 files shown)


crates/amaru-ledger/src/context/default/preparation.rs
```diff
@@ -397,38 +397,34 @@ fn resolve_committee<'block, 'volatile>(
...
+                Some(Existence::Unknown) | None => {
+                    if for_certificates || for_votes(row.status.as_ref()) {
                         cc_members.insert(cold_credential, row);
                     }
+                }
 
-                    Some(Existence::Exists(Bind { left: status, right: valid_until, .. })) => {
+                Some(Existence::Exists(Bind { left: volatile_status, right: valid_until, .. })) => {
+                    let status = volatile_status.to_option(row.status.as_ref());
+                    if for_certificates || for_votes(status.as_ref()) {
```
KtorZ, line 415: The key change is here; we must consider whether to include the CC based on the most recent status we know, not the one we pull from the stable store.

Issue #1174, opened 2026-08-07
Fix the decoding FIXMEs


etorreborre:
### Abstract

The following FIXMEs should be investigated and fixed:

| source | fixme |
|--------|--------|
|witness_set | accidentally not a set |
|protocol_parameters | update in Pallas; should be a u16 |
|metadatum| multi-era|
|metadatum| use stream encoding for length > 64|
|address/byron| worry about definite vs indefinite length here?|
|block|constraints & multi-era decoding|
|block|duplicate keys in aux data top-level map?|
|key_value_pairs|strict ordering/duplicate keys|
|decode| allow callers to check that the length is not static|
|decode|suspicious check_tagged_array_length|
|governance_rule|CDDL says just "6", not group/array "(6)"?|
|bounded_bytes|BoundedBytes should not exists|
|rational_number|Enforce tag == 30|
|transaction|We should move this validation to the Witness decoding|
|bootstrap_witness|this is era-dependent. V11 is unbounded but V12 is bounded to 32 bytes|

For each of them we should have a test case in the form of a generated fixture.

### Why?

To be conformant with the specification and the Haskell node.

### How?

By fixing them one by one :-).

### Testing Strategy / Acceptance Criteria

Possibly we need to add the possibility to keep some fixtures from being re-generated in order to make sure that we really cover all those cases.

### Checklist

- [x] I understand that feature requests and unrefined work item should be open as [GitHub Discussions](https://github.com/pragma-org/amaru/discussions) instead.
- [x] I have assigned this item to an existing milestone from the roadmap
- [x] I have added a label capturing the impact of this item (i.e. value for users/stakeholders if successful)
- [x] I have added a label capturing the delivery risk of this item (i.e. how likely is it that this task will succeed as planned)
- [x] I have added a label capturing the effort of this item (i.e. how large is the task?)

yHSJ:
In #1330, I've addressed most of these, along with some discrepancies that are not listed here that I discovered along the way. I have not touched any of the "multi-era" decoding logic, nor did I change the witness set, since it's correct now (and will change in pv12, which doesn't exist yet).

Some of the FIXMEs were incorrect or not needed, so some of them were removed with no code changes. Where that is applicable, I introduced tests to cover those cases.

yHSJ:
I've created issues #1351, #1352 #1353 for the remaining unaddressed FIXMEs. Everything else is resolved in #1330.

PR #1229, opened 2026-08-17
fix: decode bytes with indefinite length only in protocol version 12

etorreborre:
This PR mostly fixes the decoding of bytes so that Conway data only accepts definite-length encoded bytes:

- The protocol version has been added to the decoding context to support both definite and indefinite length coming up in protocol version 12 (in some places, not everywhere)
- The fixtures have been regenerated to match the Conway behavior.
- The `operational_cert_sigma` field has been changed to `FixedBytes`.


jeluard:
Do we have some links detailing this support for V12?

etorreborre:
@jeluard not that I know of, except for the [Dijkstra CDDL file](https://github.com/IntersectMBO/cardano-ledger/blob/9f6b6f1ab10d7cc730dae3328f4003e7fa55afe2/eras/dijkstra/impl/cddl/data/dijkstra.cddl).

jeluard:
Scary

crates/amaru-kernel/src/cardano/address/byron.rs
```diff
@@ -113,7 +112,7 @@ impl ByronAddress {
     }
 }
 
-impl<C> cbor::Encode<C> for ByronAddress {
+impl<C: cbor::HasProtocolVersion> cbor::Encode<C> for ByronAddress {
```
KtorZ: Do any of the Byron decoders actually need this constraint?

crates/amaru-kernel/src/cardano/memoized/native_script.rs
```diff
@@ -52,14 +52,14 @@ impl TryFrom<String> for MemoizedNativeScript {
...
-impl<'b, C> cbor::Decode<'b, C> for MemoizedNativeScript {
+impl<'b, C: cbor::HasProtocolVersion> cbor::Decode<'b, C> for MemoizedNativeScript {
     fn decode(d: &mut cbor::Decoder<'b>, ctx: &mut C) -> Result<Self, cbor::decode::Error> {
         let (expr, original_bytes) = cbor::tee(d, |d| d.decode_with(ctx))?;
         Ok(Self { original_bytes: original_bytes.to_vec(), expr })
     }
 }
 
-impl<C> cbor::Encode<C> for MemoizedNativeScript {
+impl<C: cbor::HasProtocolVersion> cbor::Encode<C> for MemoizedNativeScript {
```
KtorZ:
```suggestion
impl<C> cbor::Encode<C> for MemoizedNativeScript {
```

crates/amaru-kernel/src/cardano/memoized/plutus_data.rs
```diff
@@ -70,14 +70,14 @@ impl TryFrom<String> for MemoizedPlutusData {
...
-impl<'b, C> cbor::Decode<'b, C> for MemoizedPlutusData {
+impl<'b, C: cbor::HasProtocolVersion> cbor::Decode<'b, C> for MemoizedPlutusData {
     fn decode(d: &mut cbor::Decoder<'b>, ctx: &mut C) -> Result<Self, cbor::decode::Error> {
         let (data, original_bytes) = cbor::tee(d, |d| d.decode_with(ctx))?;
         Ok(Self { data, original_bytes: original_bytes.to_vec() })
     }
 }
 
-impl<C> cbor::Encode<C> for MemoizedPlutusData {
+impl<C: cbor::HasProtocolVersion> cbor::Encode<C> for MemoizedPlutusData {
```
KtorZ:
```suggestion
impl<C> cbor::Encode<C> for MemoizedPlutusData {
```

crates/amaru-kernel/src/cardano/network_name.rs
```diff
@@ -27,6 +28,14 @@ pub enum NetworkName {
     Testnet(u32),
 }
 
+/// Snapshots imported from the Haskell node are decoded with the network as context.
+/// They are produced at (or below) the current protocol version.
+impl HasProtocolVersion for NetworkName {
+    fn protocol_version(&self) -> ProtocolVersion {
+        protocol_version::DEFAULT
+    }
+}
```
KtorZ: Seems quite risky as an instance; we should rather derive that from `ProtocolParameters` which already carries the protocol version.

crates/amaru-kernel/src/cardano/reward_account.rs
```diff
@@ -18,6 +18,13 @@ use thiserror::Error;
...
 use crate::{Address, Bytes, Lovelace, NonEmptyKeyValuePairs, PlutusStakeAddress, StakeCredential, StakePayload};
 
+/// FIXME(cbor): RewardAccount decodes as raw bytes, without validation.
+///
+/// The Haskell node decodes this position (e.g. withdrawal map keys) as a structured
+/// `RewardAccount` — the same bytes primitive underneath, but followed by parsing and validating
+/// the address header, network tag, and stake credential (`fromCborRewardAcnt` in
+/// Cardano.Ledger.Address). By keeping the raw bytes, amaru accepts malformed reward accounts
+/// that the Haskell node rejects at deserialisation time.
 pub type RewardAccount = Bytes;
```
KtorZ, line 28: Could be a TODO since we catch this one as explicit validation rules at the moment; but it's inconvenient. The same is true for `Address` at moment, kept as plain bytes in output when they should really be parsed immediately.

KtorZ (approved):
Nice, thanks for double(triple?)-checking this.

crates/amaru-kernel/src/cardano/address/byron.rs
rkuhn: Since we’re adding such a constraint, shouldn’t we make it more generic from the beginning? We’ll need the ability in the context to signal “don’t decode these bytes, just gimme the slice”, for example. I wouldn’t want to add a bunch of trait bounds for different aspects, when most Amaru use will settle on a certain combination. Perhaps have `trait AmaruCtx: HasProtocolVersion`?

KtorZ: Some extra context here:

- We already need the ability to decode with 3 different heuristic:
  - Network/on-chain bytes; scrupulously matching CDDL & Haskell implementation
  - Compact ledger form; which is another format the Haskell ledger team has used for some types stored in the on-disk snapshots; many types are the same as the on-chain bytes but many differs too
  - RocksDB/Store bytes: which currently happens to conveniently align with the network/on-chain representation.

- Eventually, we also want that context to construct a PreparationContext when decoding from network bytes, so that we don't need to traverse the block a second time before validation, we can produce it while parsing. Yet, we still want to permit decoding without that preparation context. 

- Also, we don't want to dynamically select decoders based on the context; so we need different concrete types for the network and compact form for type that have different decoding rules there. Others may use a generic context that can satisfy both. 

So we'll definitely need some combination of traits here for types that share encoding rules, but we'll resort to concrete context types for few specific types. I would avoid abstracting too early on this, because we might end up undoing a lot later on.

etorreborre: Right, that's unnecessary.

crates/amaru-kernel/src/cardano/network_name.rs
etorreborre: Done, I needed to do a bit of parameter passing in `bootstrap.rs` but that's the right call.

crates/amaru-kernel/src/cardano/reward_account.rs
etorreborre: It's a `FIXME` , I'll get back to it with https://github.com/pragma-org/amaru/issues/1174

PR #1230, opened 2026-08-17
fix: switch to fork on the ledger only with a better tip

etorreborre:
This fix is an addition to #1190. That previous PR was assuming that the `fetch_blocks` stage was sending only the tip of a given branch to the `validate_block` stage. This made it safe to call `switch_to_fork` at the ledger level since the ledger only expects a fork that has the same length or one more block than the current chain.

The fix consists in comparing the current tip header to the incoming message header in the `validate_block` stage, using the same function that is used in the `select_chain` stage and only switch when the new tip is greater than the previous one.

During the test I also noticed that the output of of stage tests could contain raw CBOR for some fields and I fixed that too.

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented adoption of shorter or otherwise inferior forks, avoiding unnecessary ledger and downstream changes.
  - Added safeguards for invalid fork height transitions.

- **Observability**
  - Added trace events for skipped fork switches, including current and candidate chain tips.
  - Improved trace output with readable, typed event fields.

- **Tests**
  - Expanded coverage for fork switching, rollback failures, partial application, equal-height and longer-fork adoption, and rejected shorter forks.

- **Documentation**
  - Documented the new block-skip trace event and its required fields.

### DIFF
--- a/crates/amaru-consensus/src/stages/validate_block/mod.rs
+++ b/crates/amaru-consensus/src/stages/validate_block/mod.rs
@@ -204,6 +204,11 @@ pub async fn stage(
             // fetch_blocks streams the blocks of a new best candidate one by one, each with its own
             // tip. Only a tip that is strictly better than the current one (per the chain-selection
             // order) can be accepted as a candidate for a fork switch on the ledger.
+            //
+            // NOTE: the headers are loaded from the store on demand rather than kept in the stage
+            // state. This branch is neither on the sync hot path nor on the caught-up common path
+            // (both extend `current` and take the branch above), while a header held in the state
+            // would add ~1.5kB to every stage state snapshot serialized into the TraceBuffer.
             let store = Store::new(eff.clone());
             let message_header = store.load_header(&tip.hash()).await;
             let current_header = store.load_header(&state.current.hash()).await;
--- a/crates/amaru-consensus/src/stages/validate_block/mod.rs
+++ b/crates/amaru-consensus/src/stages/validate_block/mod.rs
@@ -16,14 +16,19 @@ use std::collections::BTreeMap;
 
 use amaru_kernel::{BlockHeight, HeaderHash, Point};
 use amaru_metrics::LedgerMetrics;
-use amaru_observability::{TraceContext, debug_span};
+use amaru_observability::{TraceContext, debug, debug_span};
 use amaru_ouroboros_traits::ForkSwitchOutcome;
+use amaru_protocols::store_effects::Store;
 use amaru_pure_stage::{Effects, OrTerminateWith, StageRef};
 use tracing::Instrument;
 
 use crate::{
     effects::{Ledger, LedgerOps, Metrics, MetricsOps},
-    stages::{adopt_chain::AdoptChainMsg, block_source::BlockSourceMsg, select_chain::SelectChainMsg},
+    stages::{
+        adopt_chain::AdoptChainMsg,
+        block_source::BlockSourceMsg,
+        select_chain::{SelectChainMsg, cmp_tip},
+    },
 };
 
 /// ValidateBlock stage: thin validation dispatcher + result router for the consensus pipeline.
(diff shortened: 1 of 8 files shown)


crates/amaru-consensus/src/stages/validate_block/tests.rs
```diff
@@ -86,96 +86,98 @@ fn test_parent_equals_current_validates_tip_only() {
...
+    let tip = prep.headers.h2.point();
+    let parent = prep.headers.h1.point();
     let msg = ValidateBlockMsg::new(tip, parent, BlockHeight::from(0));
 
     let (running, _guards, mut logs) = setup(&prep, msg.clone());
     assert_trace_contains(
         &running,
         &[
             te_input("vb-1", &msg).into(),
+            te_rollback_ledger("vb-1", &tip).into(),
```
rkuhn: `assert_trace_contains` should be used carefully because the match can become anaemic: here I’d prefer to assert that in order to make the decision h1 needs to be loaded, maybe there’s even more to it.

crates/amaru-consensus/src/stages/validate_block/tests.rs
```diff
@@ -86,96 +86,98 @@ fn test_parent_equals_current_validates_tip_only() {
...
-    assert_trace(
+    assert_trace_contains(
         &running,
         &[
-            te_state("vb-1", &prep.state),
-            te_input("vb-1", &msg),
-            te_rollback_ledger("vb-1", &tip),
-            te_terminate("vb-1"),
-            te_terminated("vb-1", TerminationReason::Voluntary),
+            te_input("vb-1", &msg).into(),
```
rkuhn: This should use a switch_fork effect, right? I’d like to see that here.

crates/amaru-consensus/src/stages/validate_block/mod.rs
```diff
@@ -194,6 +201,17 @@ pub async fn stage(
                 }
             }
         } else {
+            // fetch_blocks streams the blocks of a new best candidate one by one, each with its own
+            // tip. Only a tip that is strictly better than the current one (per the chain-selection
+            // order) can be accepted as a candidate for a fork switch on the ledger.
+            let store = Store::new(eff.clone());
+            let message_header = store.load_header(&tip.hash()).await;
+            let current_header = store.load_header(&state.current.hash()).await;
```
rkuhn, line 214: I’d put a `NOTE:` comment explaining that this is not on the sync hot path and not on the caught-up common path either, so we save the 1.5kB stage state that would need to be serialized for the TraceBuffer otherwise.

rkuhn (approved):
Yes, still makes sense to me.

We know why we’re doing it, even though it delays the application of the fork switch until we got all the blocks downloaded.

crates/amaru-consensus/src/stages/validate_block/mod.rs
```diff
@@ -194,6 +201,17 @@ pub async fn stage(
...
+            if cmp_tip(message_header.as_ref(), current_header.as_ref()) != std::cmp::Ordering::Greater {
```
KtorZ, line 215: Any reason we don't define an `Ord` instance on the header instead?

KtorZ (approved):
LGTM

crates/amaru-consensus/src/stages/validate_block/mod.rs
rkuhn: We might, and it would even work for `Option<Header>`, which is what we’d actually need. But `Ord` tends to be used for sorting in-memory data structures to put them into trees while this function has manifest Ouroboros protocol relevance, it compares chains and not headers. My gut feeling tells me that we are better off giving it a separate name and docs.

Ah, that just reminds me: `cmp_tip` is not an `Ord` because it isn’t transitive, the VRF comparison is only performed if the slot numbers are close enough.

PR #1233, opened 2026-08-18
fix(amaru-ledger): reduce peak allocation during stake distribution calculation

KtorZ:
Elements are already *mostly* ordered and we can rotate vectors in
  place and reduce unnecessary allocations.

  ## Before

  ```
  =============== STAKE SUMMARY MEMORY USAGE REPORT ===============
  network              mainnet
  summary              rss=407MiB     alloc=160MiB   peak=333MiB
  ```

  ## After

  ```
  =============== STAKE SUMMARY MEMORY USAGE REPORT ===============
  network              mainnet
  summary              rss=264MiB     alloc=160MiB   peak=160MiB
  ```

No-changelog

## Summary by CodeRabbit

* **Improvements**
  * Improved account data handling with consistent ordering across credential types.
  * Added support for normalizing collections into key order during processing.
  * Simplified account capacity tracking and improved resource estimation with additional headroom.

* **Bug Fixes**
  * Improved capacity calculations when loading account data.

* **Tests**
  * Updated a memory measurement test to run only when explicitly requested.

### DIFF
--- a/crates/amaru-kernel/src/data_structures/sorted_pairs.rs
+++ b/crates/amaru-kernel/src/data_structures/sorted_pairs.rs
@@ -44,29 +44,6 @@ impl<K: Ord, V> SortedPairs<K, V> {
         self
     }
 
-    /// Concatenate two pairs together.
-    ///
-    /// Panics if the first key of the right-hand side is not strictly after the last key of the
-    /// left-hand side.
-    pub fn append(mut self, mut rhs: Self) -> Self
-    where
-        K: fmt::Debug,
-    {
-        if let Some((lhs_last, _)) = self.0.last()
-            && let Some((rhs_first, _)) = rhs.0.first()
-        {
-            assert!(
-                rhs_first > lhs_last,
-                "invariant violation (SortedPairs.append): rhs' first element ({:?}) is not after lhs' last element ({:?})",
-                rhs_first,
-                lhs_last,
-            );
-        }
-
-        self.0.append(&mut rhs.0);
-        self
-    }
-
     /// Returns true if the pairs contains a value for the specified key.
     ///
     /// The key may be any borrowed form of the pairs’ key type, but the ordering on the borrowed form must match the ordering on the key type.
@@ -146,6 +123,8 @@ impl<K: Ord, V> SortedPairs<K, V> {
 
     /// Add a new key/value pair after existing keys.
     ///
+    /// # Panics
+    ///
     /// Panics if the key is not strictly after the last inserted one.
     pub fn push(&mut self, k: K, v: V)
     where
@@ -157,7 +136,6 @@ impl<K: Ord, V> SortedPairs<K, V> {
                 "invariant violation (SortedPairs.push): pushed element ({k:?}) is not after last element ({last:?})",
             );
         }
-
         self.0.push((k, v));
     }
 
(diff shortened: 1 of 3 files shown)


crates/amaru-kernel/src/data_structures/sorted_pairs.rs
```diff
@@ -46,6 +46,8 @@ impl<K: Ord, V> SortedPairs<K, V> {
 
     /// Concatenate two pairs together.
     ///
+    /// # Panics
+    ///
     /// Panics if the first key of the right-hand side is not strictly after the last key of the
     /// left-hand side.
     pub fn append(mut self, mut rhs: Self) -> Self
```
rkuhn: I don’t like panics like this, why not provide “merge” semantics only? The allocation is the same, the check is the same for already ordered append, handling the merge is cheap because both sequences are ordered.

crates/amaru-kernel/src/data_structures/sorted_pairs.rs
```diff
@@ -170,6 +182,16 @@ impl<K: Ord, V> SortedPairs<K, V> {
...
         Self(Vec::with_capacity(capacity))
     }
+
+    /// Rotates the slice in-place such that the first mid elements of the slice move to the end while the last self.len() - mid elements move to the front.
+    ///
+    /// # Safety
+    ///
+    /// This function allows to break the original sorting invariant maintained by the `SortedPairs`
+    /// API. Only use if the order can be controlled.
+    pub unsafe fn rotate_left(&mut self, mid: usize) {
```
rkuhn: This method does not make sense on this type because there is no conceivable operation for restoring the invariant apart from rotating back. Independent of the purpose (I haven’t yet read the rest of the PR), this is not how `unsafe` is supposed to be used.

crates/amaru-ledger/src/summary/stake_distribution.rs
```diff
@@ -169,27 +169,36 @@ impl StakeSummary {
...
-                StakeCredential::ScriptHash(..) => &mut script_accounts,
+
+            if let StakeCredential::AddrKeyhash(..) = credential {
+                count_keys += 1;
             }
-            .push(credential, state)
-        }
 
-        Capacity::update(key_accounts.len(), script_accounts.len());
+            accounts.push_with(credential, state, |lhs, rhs| match (lhs, rhs) {
```
rkuhn: So you’re doing an insert sort here, followed by a rotation later. Wouldn’t it be just as efficient and less obscure to just push without order, then sort? Tim-sort should deal with this nicely. If you have benchmarked this and found the shenanigans necessary then please add a NOTE.

rkuhn (commented):
I’d rather add an API that borrows the Vec without constraints and demands that when the borrow is returned, the Ord is respected again. Breaking the type guarantees defeats the whole purpose (and of course `push_with` should also be marked `unsafe`).

crates/amaru-kernel/src/data_structures/sorted_pairs.rs
KtorZ: That function is not even needed anymore, I'll just remove it.

crates/amaru-ledger/src/summary/stake_distribution.rs
KtorZ: > Wouldn’t it be just as efficient and less obscure to just push without order, then sort?

I have honestly no idea. It's a 1.5M+ vector, so I _assumed_ sorting would be not great (n * log n at best?) but I did not benchmark it. An ordered insert is totally out of the picture though because we are in the worse situation after all the key credentials have been processed and every insertion of a script causes the entire vector to shift. 

I just found the rotate_left somewhat elegant 🤓! But it's true that a mere vector that we sort in-place into a SortedPairs may do the trick just fine.

KtorZ: This is the baseline (i.e. current `main`)

```
rss=398MiB     alloc=159MiB   peak=331MiB

❯ hyperfine "cargo test --test summary -p amaru -- --ignored --no-capture"
  Time (mean ± σ):     13.553 s ±  0.478 s    [User: 12.671 s, System: 0.776 s]
  Range (min … max):   12.809 s … 14.330 s    10 runs
```

So, this is with `rotate_left` + `push_with`

```
rss=229MiB     alloc=174MiB   peak=174MiB

❯ hyperfine "cargo test --test summary -p amaru -- --ignored --no-capture"
  Time (mean ± σ):     13.067 s ±  0.606 s    [User: 12.200 s, System: 0.778 s]
  Range (min … max):   12.370 s … 14.187 s    10 runs
```

And just accumulating into a vector then `sort_unstable` (which is fine here since key don't clash, and it's in-place):

```
rss=241MiB     alloc=174MiB   peak=174MiB

❯ hyperfine "cargo test --test summary -p amaru -- --ignored --no-capture"
  Time (mean ± σ):     12.701 s ±  0.391 s    [User: 11.883 s, System: 0.752 s]
  Range (min … max):   12.161 s … 13.507 s    10 runs
```

The vec + sort is much cleaner, and simpler. So that's a clear winner :), thanks.

rkuhn (approved):

PR #1234, opened 2026-08-18
fix: normalize types around conn_id / peer + ensure various simple types are traced as strings

KtorZ:
This commit also introduce better parsing for the tracing macros, so that %,@,? and full expressions can be used for span and event fields.

I noticed this when looking at the TUI now showing peer's as `{ "address": "foo.bar:1234" }` instead of just the address. I started replacing some field instanciation with `peer = %peer` and figured it would be a bit too brittle, and that it's better to treat the type directly. 

@etorreborre I purposely didn't touch any of the non-schema-based traces to minimize conflicts since I believe you've been looking into transforming them into new schema-based ones?

## Summary by CodeRabbit

* **Improvements**
  * Standardized trace fields for peers, connection IDs, hashes, and other displayed values.
  * Reported connection identifiers as numeric values for clearer trace analysis.
  * Added support for display, debug, and preformatted trace values.
  * Simplified consensus rollback tracing and removed redundant header-hash fields.

* **Documentation**
  * Updated trace documentation and schemas to reflect revised field types and names.

* **Bug Fixes**
  * Corrected trace output so peer and hash values are consistently represented as readable strings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,7 @@ Other guiding principles:
 - **amaru-stores**: prune obsolete votes when removing expired/ratified/evicted proposals.
 - **amaru**: missing epoch stake distribution snapshots for mainnet, and fixed stake distribution management Makefile.
 - **amaru-node**: `Telemetry` (used by `run_until`) now uses the product CBOR-aware tracing layers so structured log properties decode to numbers and strings instead of diagnostic byte strings.
+- **amaru**: more homogenous traces fields representations across peers, connection id and hashes.
 
 ## v10.11.20260813 _[unreleased; planned for 2026-08-13]_
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -509,6 +509,7 @@ dependencies = [
 name = "amaru-observability"
 version = "0.1.2"
 dependencies = [
+ "amaru-kernel",
  "amaru-observability-macros",
  "cbor-data",
  "cbor4ii",
--- a/crates/amaru-consensus/src/stages/fetch_blocks/mod.rs
+++ b/crates/amaru-consensus/src/stages/fetch_blocks/mod.rs
@@ -400,7 +400,7 @@ impl FetchBlocks {
 
         // check that body belongs to header
         if block.header.body().block_body_hash != block.body_hash() {
-            let span = debug_span!(consensus::block::MISMATCHED_HASH, peer = peer.clone(), header_hash = point.hash());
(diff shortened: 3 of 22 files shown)


etorreborre (approved):

rkuhn:
This is what I wanted to talk to you about “when there is time after that important release” ;-)

To my mind, the current schema is quite meaningless because it only specifies which type of value we put into the macro at the call site, it doesn’t specify which type will appear in the JSON / CBOR / whatever. I think we should regulate both, which means removing the choice at the call site and defining the rendering within the schema definition instead. Then we are in a position to emit the actual JSON schema.

```
                FETCH_NEXT {
                    tags: cpu
                    required point: amaru_kernel::Point
                    required header_hash: %amaru_kernel::HeaderHash
                }
```

This would require the correct types as arguments and render `Point` as a structured object (an array containing slot and hash, the latter currently serializes as a String — which I don’t agree with since we do use CBOR that has byte strings) while the `HeaderHash` would be serialized via `Display::fmt` as a `String`, regardless of its serde instance. `?` could be used for `Debug`.

We should add `schemars` and derive or implement `JsonSchema` for all relevant types, i.e. those arguments that are used without Display or Debug modifier.

KtorZ:
> @rkuhn To my mind, the current schema is quite meaningless 

Strong words 😅 ... I don't think it's meaningless; but it is certainly opens to _interpretation_ when it comes to encoding. 

> @rkuhn which means removing the choice at the call site and defining the rendering within the schema definition instead.

Agreed; and this PR is sort of a very minimalist step towards that; there's just no choice given and some types have one "canonical" scalar representations. Not ideal, but I am just chasing the obvious and low-hanging fruit to get something presentable. 

Something I miss from golang actually (yes yes, I have once qualified as a golang developer) was the ability to easily annotate structs for marshalling targets; like:

```go
type Transaction struct {
    Inputs  []TxInput  `json:"inputs"  cbor:"0,keyasint"`
    Outputs []TxOutput `json:"outputs" cbor:"1,keyasint"`
    Fee     uint64     `json:"fee"     cbor:"2,keyasint"`
    TTL     *uint64    `json:"ttl,omitempty" cbor:"3,keyasint,omitempty"`
}
``` 

This was great; all in one place and somewhat flexible.

rkuhn:
> Strong words […]

I don’t want to harp on it, your response points out that I was probably not clear enough: today, the schema might say `u64`, the JSON schema will say `number`, but the call-site says `%i_know_better` and the actual log contains a string. That’s what I meant.

Capacity permitting, I might try the declaration-site formatting on the weekend.

PR #1235, opened 2026-08-18
chore: reduce traces noise around transaction validation: emit a single trace per phase-one, and phase-two.

KtorZ:
## Summary by CodeRabbit

- **New Features**
  - Added detailed timing information for block validation, transaction phase-one checks, and phase-two script execution.
  - Added trace data for script-processing stages, including decoding, program construction, and evaluation.

- **Documentation**
  - Updated trace documentation and schemas to reflect the consolidated ledger validation and script-execution traces.
  - Documented the latest bootstrap import and measurement-trace improvements in the changelog.

- **Refactor**
  - Simplified ledger tracing by consolidating numerous stage-specific spans into clearer, timing-focused trace entries.

### DIFF
--- a/crates/amaru-kernel/src/utils/duration.rs
+++ b/crates/amaru-kernel/src/utils/duration.rs
@@ -12,7 +12,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::time::Duration;
+use std::time::{Duration, Instant};
+
+/// Measure the time in μs since a previous checkpoint while refreshing the checkpoint.
+pub fn elapsed_and_reset(meter: &mut Instant) -> u64 {
+    let now = Instant::now();
+    let us = now.saturating_duration_since(*meter).as_micros() as u64;
+    *meter = now;
+    us
+}
 
 pub fn parse_duration(raw: &str) -> Result<Duration, String> {
     let split = raw.find(|c: char| !c.is_ascii_digit()).unwrap_or(raw.len());
--- a/crates/amaru-ledger/src/rules/transaction/phase_one/mod.rs
+++ b/crates/amaru-ledger/src/rules/transaction/phase_one/mod.rs
@@ -12,11 +12,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::{fmt, mem, ops::Deref};
+use std::{fmt, mem, ops::Deref, time::Instant};
 
 use amaru_kernel::{
     AuxiliaryData, EraHistory, HasTransactionId, Hash, Lovelace, Network, NetworkName, ProtocolParameters,
     TransactionBody, TransactionInput, TransactionPointer, WitnessSet, cardano::value::Balance, size::SCRIPT,
+    utils::duration::elapsed_and_reset,
 };
 use amaru_observability::debug_span;
 use amaru_plutus::arena_pool::ArenaPool;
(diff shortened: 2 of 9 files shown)


crates/amaru-kernel/src/utils/duration.rs
```diff
@@ -12,7 +12,14 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::time::Duration;
+use std::time::{Duration, Instant};
+
+/// Measure the time in μs since a previous checkpoint while refreshing the checkpoint.
+pub fn elapsed(meter: &mut Instant) -> u64 {
+    let μs = meter.elapsed().as_micros() as u64;
```
rkuhn: Oh oh, please don’t ever use `std::time::Instant::elapsed()` — it panics on hardware that doesn’t perfectly synchronise clock reads across cores or hardware threads (well, supposedly it _currently_ doesn’t, but it may reintroduce that behaviour). Please always use `meter.saturating_duration_since(Instant::now())`.

This is BTW the only reason for the existence of `tokio::time::Instant` as far as I’m aware.

crates/amaru-ledger/src/rules/transaction/phase_one/mod.rs
```diff
@@ -136,46 +137,47 @@ pub fn execute<C>(
...
-        fees::execute(
-            context,
-            transaction_body.fee,
-            tx_size,
-            transaction_witness_set,
-            ref_scripts_size,
-            protocol_parameters,
-        )
-    })?;
+    span.record(VALIDATE::FIELD_PREFLIGHT_MICROS, elapsed(&mut meter));
```
rkuhn: It is unclear what `span` is here: how can we be sure that its schema includes this property?

crates/amaru-kernel/src/utils/duration.rs
```diff
@@ -12,7 +12,14 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::time::Duration;
+use std::time::{Duration, Instant};
+
+/// Measure the time in μs since a previous checkpoint while refreshing the checkpoint.
+pub fn elapsed(meter: &mut Instant) -> u64 {
```
rkuhn: This should be called `elapsed_and_reset`.

rkuhn (commented):
We might also want to return a properly typed `Span` that knows which properties may be recorded on it according to the schema.

crates/amaru-kernel/src/utils/duration.rs
KtorZ: Surprising. Why is this still in the standard lib :| ... it's not like the API surface for Instant is large.

crates/amaru-ledger/src/rules/transaction/phase_one/mod.rs
KtorZ: So, we can be sure that this schema includes this property because it comes directly from the generated schema (`VALIDATE::FIELD_PREFLIGHT_MICROS` only exists for _that_ schema). Now, we cannot be sure that the `span` is indeed a "transaction validate" span. Ideally we should move the span creation from the caller to the callee here.

crates/amaru-kernel/src/utils/duration.rs
```diff
@@ -12,7 +12,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::time::Duration;
+use std::time::{Duration, Instant};
+
+/// Measure the time in μs since a previous checkpoint while refreshing the checkpoint.
+pub fn elapsed_and_reset(meter: &mut Instant) -> u64 {
+    let now = Instant::now();
+    let μs = meter.saturating_duration_since(now).as_micros() as u64;
```
yHSJ: oh my god, *why*

yHSJ (approved):
Seems reasonable, once you fix that variable name

crates/amaru-kernel/src/utils/duration.rs
KtorZ: I originally had that in the schema field names too, and thought people will hate me then when trying to build `jq` or Grafana filters matching on those greek letters. So I reversed to `us` but here; it's local to the function so no harm done, right 😶 ... ?

crates/amaru-kernel/src/utils/duration.rs
```diff
@@ -12,7 +12,15 @@
...
+    *meter = now;
```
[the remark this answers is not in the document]
KtorZ: Good point rabbit.

KtorZ:
Do not merge until https://github.com/pragma-org/amaru/pull/1235#discussion_r3813986159 is addressed

PR #1236, opened 2026-08-18
fix: show the most recent info logs

etorreborre:
@KtorZ quick fix done with the help of Claude but I checked manually. 

The `INFO` logs were truncated and, for example, not showing the latest adopted tip, while the corresponding log line was shown in the `DEBUG` tab. It looks like lines wrapping was pushing new log messages out.

## Summary by CodeRabbit

- **Bug Fixes**
  - The TUI now displays the latest `INFO` log messages reliably.
  - Improved log scrolling keeps the newest entries visible, including when messages wrap across multiple lines.
- **Tests**
  - Added coverage for scrolling behavior with wrapped log entries.
- **Documentation**
  - Updated the changelog with the log display fix.

### DIFF
--- a/crates/amaru-tui/src/ui/components/logs.rs
+++ b/crates/amaru-tui/src/ui/components/logs.rs
@@ -73,16 +73,24 @@ pub(in crate::ui) fn render_logs(frame: &mut Frame<'_>, area: Rect, model: &Mode
     render_log_controls(frame, layout[0], model, views);
     render_horizontal_separator(frame, layout[1], model.interaction_mode, focused);
 
-    let logs = model.filtered_logs();
-    let visible = layout[2].height as usize;
-    let scroll = model.log_scroll.min(logs.len().saturating_sub(visible));
-    let end = logs.len().saturating_sub(scroll);
-    let start = end.saturating_sub(visible);
-    let lines = logs[start..end].iter().map(|record| log_record_line(record.as_ref())).collect::<Vec<_>>();
-
-    let paragraph = Paragraph::new(lines).wrap(Wrap { trim: false });
+    let lines = model.filtered_logs().iter().map(|record| log_record_line(record.as_ref())).collect::<Vec<_>>();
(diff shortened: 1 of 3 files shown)


crates/amaru-tui/src/ui/components/logs.rs
```diff
@@ -73,16 +73,24 @@ pub(in crate::ui) fn render_logs(frame: &mut Frame<'_>, area: Rect, model: &Mode
...
+    lines: Vec<Line<'static>>,
+    area: Rect,
+    scroll_from_bottom: usize,
+) -> (Paragraph<'static>, usize, usize) {
     let paragraph = Paragraph::new(lines).wrap(Wrap { trim: false });
-    frame.render_widget(paragraph, layout[2]);
-    render_scrollbar(frame, layout[2], logs.len(), visible, start, model.interaction_mode);
+    let total = paragraph.line_count(area.width);
+    let max_position = total.saturating_sub(area.height as usize);
+    let position = max_position.saturating_sub(scroll_from_bottom.min(max_position));
```
rkuhn, line 90:
```suggestion
    let position = max_position.saturating_sub(scroll_from_bottom);
```
`saturating_sub` should already take care of this, no?

rkuhn (commented):
Do I understand correctly that previously scrolling didn’t work? It isn’t really obvious what the error was.

KtorZ:
I haven't really experienced the problem myself @etorreborre ? The log lines are wrapped correctly and without overlap on my end 🤔 ?

etorreborre:
The problem is that the `INFO` tab is lagging compared to the `DEBUG` tab. I was reading an `adopted tip` log message in the `DEBUG` tab while it was not visible in the `INFO` tab. You don't see this?

KtorZ:
@etorreborre I had never noticed 😶 .... but yes, there's a lag indeed.

KtorZ:
Merged manually after addressing Roland's remark and testing locally.

PR #1237, opened 2026-08-18
fix: amaru binary fails to create snapshot

jeluard:
Make sure amaru doesn't rely on build time constant. Improve error messages when directory can't be written to.

## Summary by CodeRabbit

* **Bug Fixes**
  * Snapshot creation and publishing now resolve default directories relative to the running `amaru` executable.
  * Improved reporting when required directories cannot be accessed or created.
  * Snapshot commands automatically download configuration when no explicit directory is provided.

* **Documentation**
  * Updated command-line documentation to reflect the new default locations.
  * Removed outdated configuration download and usage trace documentation.

### DIFF
--- a/crates/amaru/src/bin/amaru/cmd/snapshot/create/config.rs
+++ b/crates/amaru/src/bin/amaru/cmd/snapshot/create/config.rs
@@ -18,11 +18,9 @@ use std::{
     path::{Path, PathBuf},
 };
 
-use amaru_kernel::{NetworkName, utils::path::relative_path};
-use amaru_observability::info;
+use amaru_kernel::NetworkName;
 use serde::Deserialize;
 
-use super::repo_root;
 const OFFICIAL_CARDANO_NODE_CONFIG_BASE_URL: &str = "https://book.world.dev.cardano.org/environments";
 
 #[derive(Debug, Deserialize)]
@@ -76,19 +74,6 @@ pub(super) async fn resolve_config_dir(
         return Ok(config_dir);
     }
 
-    if let Some(config_dir) = bundled_config_dir(network) {
-        match validate_config_dir(&config_dir) {
-            Ok(()) => {
-                info!(cli::cardano_node_config::USE, config_dir = %relative_path(&config_dir)?.display(), network = %network);
-                return Ok(config_dir);
-            }
-            Err(_) => {
-                info!(cli::cardano_node_config::DOWNLOAD, config_dir = %relative_path(&config_dir)?.display(), network = %network);
-                return download_official_config_bundle(client, network, &config_dir).await;
-            }
-        }
-    }
-
     download_official_config_bundle(client, network, &cached_config_dir(work_dir, network)).await
 }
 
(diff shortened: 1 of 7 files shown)


crates/amaru/src/bin/amaru/cmd/snapshot/create/mod.rs
```diff
@@ -193,12 +195,30 @@ impl EpochTarget {
...
+}
+
+pub(super) fn default_snapshot_output_dir(network: NetworkName) -> io::Result<PathBuf> {
+    Ok(executable_dir()?.join(default_snapshots_dir(network)))
 }
 
-pub(super) fn default_snapshot_output_dir(network: NetworkName) -> PathBuf {
-    repo_root().join(default_snapshots_dir(network))
+fn create_directory(path: &Path, purpose: &str) -> io::Result<()> {
+    fs::create_dir_all(path).map_err(|err| {
```
yHSJ, line 219: This will return `Ok` when the directory already exists, regardless of permissions. So I can point this at `/`, which fails further along the snapshot create process (which can take a while).

I don't know the `fs` API very well, but we could always try writing a tmp file to the directory, to ensure we have write permission there.

crates/amaru/src/bin/amaru/cmd/snapshot/create/config.rs
yHSJ, line 113: This will *also* fail with an OS error, should probably just do the same wrapper and provide a better error message.

crates/amaru/src/bin/amaru/cmd/snapshot/publish.rs
yHSJ, line 116: This swallows errors, I pointed it at `/dev/null` and got an "unexpected" error:
"no archives found".

yHSJ (approved):
I tested this with a bunch of strange scenarios, trying to mimic as many of the mistakes users could make as I could think of. Good news, the specific changes here are good to go (they fail fast with good error messages). There are some orthogonal issues/questions I discovered though:

1. The checks are only in the one file. `config.rs` still has a few raw `?`s that provide opaque error messages that should probably follow the same pattern.

2. Can we decide on a standard data directory instead of the executable's directory? There are many ways to run this, and it can be strange. For example, running with `cargo`, means it goes into `target/` which means the snapshots are deleted if you ever run `cargo clean`. If a user installs it in `/usr/bin`, they can't use the default at all due to permission issues. Related to that, it seems the CLI uses a few different places for data, it would be *great* to unify that. I also worry that `current_exe()` may produce different behavior on different systems.

3. `amaru dev traces dump` has the same behavior if running outside the repo. Doesn't have to be fixed in this PR, but wanted to note it as well.


So, all in all, this PR does *exactly* the minimum of what it claims to do. So it can be merged, but I wanted to make sure the above things were noted. They could be fixed in this PR, or be fixed in follow ups.

KtorZ:
@yHSJ thanks for looking into this 🫡 

Regarding:

2. Yes; we should likely set for `$HOME/.cache/amaru` on Unix system. That seems _meant for that_. 
3. That is "on purpose". This command must be run from the repository's root; an it's only meant for us (hence the `dev`). We could optionally take a filepath to the repository, but it's a command that literally reads the source files.

PR #1239, opened 2026-08-19
Fix committee acknowledgment to NOT modify the final/resulting state 

KtorZ:
## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected constitutional committee updates so newly acknowledged members are available at the start of the block.
  - Prevented unnecessary empty volatile committee entries during committee updates.
  - Improved handling of committee membership additions and removals while preserving existing committee state.

- **Tests**
  - Added regression coverage verifying newly introduced committee members can be found correctly.

### DIFF
--- a/crates/amaru-ledger/src/context/default/validation.rs
+++ b/crates/amaru-ledger/src/context/default/validation.rs
@@ -394,10 +394,7 @@ impl ProposalsSlice for DefaultValidationContext {
         if let GovernanceAction::UpdateCommittee(_, _, added, _) = &state.proposal.gov_action {
             for (cold_credential, _) in added.iter() {
                 if !matches!(self.state.committee.get(cold_credential), Existence::Exists(..)) {
-                    self.state
-                        .committee
-                        .bind_left(*cold_credential, None)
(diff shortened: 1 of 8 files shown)


crates/amaru-ledger/src/context/default/validation.rs
```diff
@@ -394,10 +394,7 @@ impl ProposalsSlice for DefaultValidationContext {
         if let GovernanceAction::UpdateCommittee(_, _, added, _) = &state.proposal.gov_action {
             for (cold_credential, _) in added.iter() {
                 if !matches!(self.state.committee.get(cold_credential), Existence::Exists(..)) {
-                    self.state
-                        .committee
-                        .bind_left(*cold_credential, None)
-                        .unwrap_or_else(|_| unreachable!("committee members are never 'unregistered'"));
```
KtorZ, line 400: This was effectively creating an update on the CC resetting the left-bind (so the hot delegation) as soon as a update-cc proposal is stabilizing (but way before it's actually ratified) 😬 ... 

Of course, this should only modify the validation context and not the state diff!

etorreborre (approved):

PR #1253, opened 2026-08-20
Fix the Min UTxO Value Calculation

yHSJ:
Our min UTxO value calculation was incorrectly missing the constant of `160` that is added to the size of the UTxO. This was not caught anywhere since fixture 49 set `coins_per_utxo_byte` to 100 billion.

This PR fixes the problem, and adds tests around the boundary. It's worth noting, that if the value of `coins_per_utxo_byte` changes, the tests will have to change as well.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected minimum lovelace calculation for ledger outputs by accounting for the fixed UTxO entry overhead.
  * Outputs containing exactly the required minimum lovelace are now accepted.
  * Outputs one lovelace below the minimum are correctly rejected.

* **Tests**
  * Added coverage for both boundary cases to verify minimum-value output validation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,10 @@ Other guiding principles:
 
 ## v10.11.20260827 _[unreleased; planned for 2026-08-27]_
 
+### Fixed
+
+- **amaru-ledger**: Correctly calculate an output's minimum lovelace value.
+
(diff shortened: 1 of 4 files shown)


KtorZ (commented):
Embarrassing. 

Although to be fair, we don't have any in-memory UTxO overhead, so this could qualify as Amaru's pricing 😶 ...

KtorZ (approved):

PR #1254, opened 2026-08-20
 align conformance fixtures on canonical snake_case

### LINKED ISSUE #1192: Cleanup serde-json Proxy instances
### Abstract

Remove the remaining Proxy types used throughout the ledger conformance tests for JSON serialization/deserialization, and define the required JSON instances directly on the corresponding Amaru types instead.

These proxies were originally necessary because several ledger types came from the external Pallas library, which prevented us from implementing custom serialization instances directly due to Rust's orphan rules. Now that the relevant types are owned by Amaru, this indirection is no longer necessary and mostly adds boilerplate and cognitive overhead.

### Why?

This constraint no longer applies to Pallas types that have since moved into Amaru.

Keeping the proxies around now has several downsides:

- Additional wrapper types need to be defined and maintained.
- Conversion to and from proxy types obscures otherwise straightforward serialization logic.
- Test code becomes harder to read because the type represented in JSON is not necessarily the type appearing in the serialization implementation.
- Adding or modifying ledger types often requires touching an unnecessary extra layer.
- The historical reason for the abstraction is no longer obvious, making the codebase harder to navigate.

The conformance test serialization layer should reflect the current ownership model rather than preserving an abstraction introduced solely to work around an old external-library constraint.

### How?

Review the Proxy types used by the ledger conformance tests and identify those that only exist to provide JSON serialization/deserialization for types now owned by Amaru.

For each applicable proxy:

- Move the corresponding JSON serialization/deserialization implementation directly onto the underlying Amaru type in the kernel
- Replace usages of the proxy with the actual ledger type.
- Remove conversion code that only exists to cross the proxy boundary.
- Remove the proxy type once it has no remaining use.
- Preserve the exact JSON representation expected by the conformance test vectors. Note, there might be a clash with the JSON instances we use for epoch state comparison. Those tests are more annoying to modify, so in case of conflict, we should preferably align the rules conformance serialisation on the epoch snapshot ones.

Some proxies may still be justified where the underlying type remains external or where the proxy provides meaningful behaviour beyond working around trait coherence. Those should be kept as type wrappers in the kernel rather than removed mechanically and/or specialised to a specific type. The wrapper should capture a behavior. 

The intent is not necessarily to eliminate every type named Proxy, but to eliminate proxy types whose original justification no longer exists.

- Existing ledger conformance tests continue to pass unchanged against the existing test vectors.
- JSON representations remain backwards-compatible with the current conformance test format. 
- Proxy types used solely for serialization of Amaru-owned ledger types are removed.
- Serialization/deserialization implementations live directly alongside, or are directly implemented for, the corresponding Amaru ledger types.
- Unnecessary From/Into conversions introduced solely for proxy serialization are removed.

### Discussion points

_No response_

_No response_

### Checklist

bomanaps:
Closes #1192. Following KtorZ's guidance, this removes the proxy types and the HasProxy machinery from amaru-kernel, moves JSON parsing directly onto the types themselves, and switches all 242 test fixtures plus the schema, and Haskell harness to snake_case matching the format the rest of the codebase already uses everything is verified locally all conformance tests pass with the exact same counts as before, make all-ci-checks runs fully green, and the updated Haskell harness passes the full sweep, 240/240 scenarios validate correctly against the reference ledger. Per review feedback: the changelog entry is removed, the shared fixture directories are kebab-case (common/era-history/, common/protocol-parameters/), and the Haskell records keep camelCase fields mapped to the snake_case keys via a shared aeson helper the branch is rebased on latest main including the two fixtures added in #1253.
no-changelog


CHANGELOG.md
```diff
@@ -35,26 +35,21 @@ Other guiding principles:
...
+## v10.11.20260820 _[unreleased; planned for 2026-08-20]_
 
 ### Added
 
 - **amaru-pure-stage**: simulate how long an `ExternalEffect` occupies time (`DurationDist`: zero, constant, uniform, or until the effect future resolves). Sampled durations are scheduled when the effect is issued; `SimulationBuilder::run` now takes a Tokio `Handle` so a still-pending `run()` can be forced at that deadline. ([#1224](https://github.com/pragma-org/amaru/pull/1224))
-- **amaru-consensus**: log the transaction ids evicted from the mempool when a block is adopted, and specify if those transactions were included in a block (#1243).
 
 ### Changed
 
+- **amaru-kernel, amaru-ledger**: ledger conformance fixtures now use snake_case keys aligned with the canonical serde instances of the underlying types; the serialization-only JSON proxy types and the `HasProxy` machinery are removed, including the previously exported `EraHistoryProxy`. Voting-threshold and execution-unit fixture shapes now match the in-memory types, and the Haskell conformance harness parses the new format. ([#1192](https://github.com/pragma-org/amaru/issues/1192))
```
yHSJ: This kind of a change doesn't need a changelog entry, since it's not "public facing".

I would recommend adding `no-changelog` to the PR description or to the commit, instead of adding an entry.

crates/amaru-ledger/tests/data/transaction/scenarios/00072-fail-stake-registration-cert-with-incorrect-deposit.json
```diff
@@ -2,26 +2,26 @@
   "title": "stake registration cert with incorrect deposit",
   "description": "A Conway `Reg` cert supplies a deposit that disagrees with protocolParameters.stakeCredentialDeposit.",
   "network": "preprod",
-  "eraHistory": {
-    "$ref": "common/eraHistory/preprod-conway.json"
+  "era_history": {
+    "$ref": "common/era_history/preprod-conway.json"
```
yHSJ: The directory names should be `kebab-case`, to match the rest of the repository.

yHSJ (commented):
Welcome! Thank you for your first contribution.

Two small pieces of feedback from an initial quick scan of the PR. I will approve the workflows to run, and give it a deeper review once the CI completes.

Whether this change is *needed*, is a different question (I am partial to camel case in JSON, but unfortunately, there is no required, or recommended casing), but I will not subject you to that. I know @KtorZ has his opinions on this one.

crates/amaru/tests/conformance/validation-rules/src/Data/Fixture/EraHistory.hs
```diff
@@ -63,7 +63,7 @@ import qualified Ouroboros.Consensus.HardFork.History.Summary as HardForkHistory
 import qualified Data.SOP.NonEmpty as SOPNonEmpty
 
 data EraHistory = EraHistory
-    { stabilityWindow :: !Word64
```
KtorZ, line 66: My bad on this one; I should've been clearer but the snake_casing only applies to Rust / Json; Haskell's conventions is to stick with camelCase.

KtorZ (commented):
Thanks for this; lots of deadcode indeed as anticipated 😅

CHANGELOG.md
bomanaps: Removed the entry and added no-changelog to the PR description instead, thanks for the pointer.

crates/amaru-ledger/tests/data/transaction/scenarios/00072-fail-stake-registration-cert-with-incorrect-deposit.json
bomanaps: renamed both directories to `common/era-history/` and `common/protocol-parameters/`, and updated the $refs and the fixtures README to match.

crates/amaru/tests/conformance/validation-rules/src/Data/Fixture/EraHistory.hs
bomanaps: reverted the Haskell record fields to camelCase, they map to the snake_case JSON keys via a shared `snakeCaseOptions` helper in `Data.Fixture.Common`.

crates/amaru-kernel/src/cardano/protocol_parameters.rs
```diff
@@ -183,12 +176,11 @@ mod proxy {
...
     /// [Ogmios's `ProtocolParameters` schema](https://github.com/CardanoSolutions/ogmios) but
-    /// the wire format intentionally differs: ratios are `{ numerator, denominator }` objects
-    /// rather than `"n/m"` strings, lovelace amounts and byte sizes are bare integers rather
-    /// than wrapped, and Plutus cost-model keys are camelCase (`plutusV1`, etc.).
+    /// snake_cased, and the wire format intentionally differs: ratios are
+    /// `{ numerator, denominator }` objects rather than `"n/m"` strings, and lovelace amounts
+    /// and byte sizes are bare integers rather than wrapped.
     #[derive(Deserialize)]
-    #[serde(rename_all = "camelCase")]
-    pub struct ProtocolParametersProxy {
```
KtorZ, line 191: I don't think the Proxy is even needed anymore here; we could define the instance on the ProtocolParameters type directly.

bomanaps: The proxy itself isn't a 1:1 mirror,the fixture uses Ogmios-style names `min_fee_coefficient` vs `min_fee_a` and groups `min_fee_reference_scripts` into a nested object that fans out into three flat fields, I can remove it entirely but do you prefer I rename the keys in the pparams fixtures to match the ProtocolParameters field names so a plain derive works, or keep the fixture vocabulary and hand-write the impl?

yHSJ: This choice was very intentional (as written in the fixtures' `README`), to use the Ogmios naming conventions. Please do *not* change the fixture. If we want to remove proxy, I would add a hand-written implementation of `serde::Deserialize` with a [`NOTE`](https://github.com/pragma-org/amaru/blob/main/engineering-decision-records/023-code-annotations.md#use-of-note) comment explaining why the implementation is written by hand.

Renaming the fields is a valid option, but I would argue it needs more thought and discussion. We must act as if there are other consumers of these fixtures (other node implementations, for example), and we don't want to be disruptive to them.

bomanaps: I removed the proxy  ProtocolParameters now has a hand-written Deserialize mapping the fixture keys directly, with a NOTE per EDR-023 explaining why.

PR #1255, opened 2026-08-20
Fix Disjoint Inputs Check(s)

yHSJ:
In pv11+, the reference inputs and the inputs do *not* need to be disjoint sets. That being said, script executions in PlutusV3 must have disjoint sets regardless of the protocol version.

This PR fixes that behavior, and adds a test for it.

## Summary by CodeRabbit

* **New Features**
  * Transactions on protocol version 11 and later can reuse an input as both spent and referenced.
  * Reference-script sizes now count unique inputs correctly.
  * Telemetry metrics can be configured, with updated metric headers in the TUI.

* **Bug Fixes**
  * Overlapping inputs are rejected when Plutus V3 scripts are used.
  * Improved consensus and protocol logging.
  * Added chain-store hash validation and preserved transaction bytes.
  * Updated ledger validation and transaction scenario coverage.

### DIFF
--- a/crates/amaru-ledger/src/rules/transaction/phase_one/inputs.rs
+++ b/crates/amaru-ledger/src/rules/transaction/phase_one/inputs.rs
@@ -14,7 +14,7 @@
 
 use amaru_kernel::{
     Address, HasScriptHash, MemoizedDatum, ProtocolParameters, RedeemerTag, RequiredScript, Set, TransactionInput,
-    address::byron::AddressType, utils::string::display_collection,
+    address::byron::AddressType, protocol_version::PROTOCOL_VERSION_11, utils::string::display_collection,
 };
 use thiserror::Error;
 
@@ -53,7 +53,7 @@ where
 
     let mut intersection = Vec::new();
     let mut ref_scripts_size: u64 = 0;
-    let check_disjoint_reference_inputs = protocol_parameters.protocol_version.major() < 11;
+    let check_disjoint_reference_inputs = protocol_parameters.protocol_version < PROTOCOL_VERSION_11;
 
     if let Some(reference_inputs) = reference_inputs {
         for reference_input in reference_inputs {
@@ -122,7 +122,10 @@ where
         }
 
         if let Some((script_hash, script_size)) = script_ref {
-            if !reference_inputs.is_some_and(|reference_inputs| reference_inputs.contains(input)) {
+            // Sizes are summed over the union of inputs and reference inputs; an input present
+            // in both, allowed in pv11+, was already counted above.
+            let already_counted = reference_inputs.is_some_and(|reference_inputs| reference_inputs.contains(input));
+            if !already_counted {
                 ref_scripts_size += script_size;
             }
             context.acknowledge_script(script_hash, *input);
--- a/crates/amaru-ledger/src/rules/transaction/phase_one/inputs.rs
+++ b/crates/amaru-ledger/src/rules/transaction/phase_one/inputs.rs
@@ -53,11 +53,11 @@ where
 
     let mut intersection = Vec::new();
     let mut ref_scripts_size: u64 = 0;
+    let check_disjoint_reference_inputs = protocol_parameters.protocol_version.major() < 11;
 
     if let Some(reference_inputs) = reference_inputs {
         for reference_input in reference_inputs {
-            // Non-disjoint reference inputs
-            if inputs.contains(reference_input) {
+            if check_disjoint_reference_inputs && inputs.contains(reference_input) {
                 intersection.push(*reference_input);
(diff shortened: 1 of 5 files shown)


crates/amaru-ledger/src/rules/transaction/phase_one/inputs.rs
```diff
@@ -53,11 +53,11 @@ where
 
     let mut intersection = Vec::new();
     let mut ref_scripts_size: u64 = 0;
+    let check_disjoint_reference_inputs = protocol_parameters.protocol_version.major() < 11;
```
KtorZ:
```suggestion
    let check_disjoint_reference_inputs = protocol_parameters.protocol_version < PROTOCOL_VERSION_11;
```

crates/amaru-ledger/src/rules/transaction/phase_two/mod.rs
```diff
@@ -136,6 +142,14 @@ where
...
 
     let scripts_to_execute = tx_info.to_script_contexts();
+    let non_disjoint_inputs = transaction_body
+        .reference_inputs
+        .as_deref()
+        .unwrap_or_default()
+        .iter()
+        .filter(|input| transaction_body.inputs.contains(input))
+        .copied()
+        .collect::<Vec<_>>();
```
KtorZ: We should only build this iterator and collect when necessary (i.e.  (1) for a V3 script, and (2) when it's non empty)

crates/amaru-ledger/src/rules/transaction/phase_one/inputs.rs
```diff
@@ -122,7 +122,9 @@ where
         }
 
         if let Some((script_hash, script_size)) = script_ref {
-            ref_scripts_size += script_size;
+            if !reference_inputs.is_some_and(|reference_inputs| reference_inputs.contains(input)) {
```
KtorZ: `!` ? This only counts the size of scripts that are NOT in reference inputs?

yHSJ: You're right it's confusing, but it's correct.

This logic only counts scripts that are in the inputs and NOT in the reference inputs. The reference inputs are summed above (not visible in the original diff due to GitHub). I tried to make the logic more clear, and added a comment as well.

nemo83:
We hit this on preview and can confirm this PR resolves it.

**The failure.** Amaru `v10.11.20260820` halts permanently at block **4,588,007** (slot 120569440):

```
BlockValidationError: Invalid block: Transaction 97df1aec…228dc4 at index 0 is invalid:
transaction failed phase one validation: invalid inputs:
inputs included in both reference inputs and spent inputs:
intersection [fbe65e04855b8f0592528f442a098e54b959228477b49752fe83b31e2a30611e#0]
parent=120569296.ee3e9a12…(4588006) failed_tip=120569440.42528f9d…(4588007)
```

The node doesn't exit — it stays up and refuses every later block as descending from an invalid one, so the process looks healthy while the height never moves again.

**The block is on chain.** Koios has tx `97df1aec28f34d1e77e480a7794ebda672dde39f0d36291892d3624c72228dc4` at that height in a 6-transaction block, with `fbe65e04…30611e#0` in **both** its `inputs` and its `reference_inputs`. Preview is at `protocol_major` 11.

**The PlutusV3 carve-out doesn't cover this one.** The transaction executes exactly one script and it is `plutusV2` (9 bytes), spending a *different* input; the overlapping UTxO is a plain key-locked 2 ADA output with no datum, no inline datum and no reference script. So there is no V3 execution involved under either reading of "must have disjoint sets". (Version via `script_info` — `tx_info` alone doesn't return it.)

**Reproduces in about 66 seconds** on the published image — 44 s bootstrap, 21 s to the halt:

```bash
docker volume create amaru-repro   # must be writable by uid 10000

docker run --rm -v amaru-repro:/data ghcr.io/pragma-org/amaru:v10.11.20260820 \
  node bootstrap --network preview \
  --ledger-dir /data/ledger.db --chain-dir /data/chain.db

docker run --rm -v amaru-repro:/data ghcr.io/pragma-org/amaru:v10.11.20260820 \
  node run --network preview \
  --ledger-dir /data/ledger.db --chain-dir /data/chain.db --no-tui
```

**With this PR, it syncs normally.** We built `db5961c` and swapped only the binary into that same published image, then ran the identical commands on a fresh volume. It stepped over the failing block without stopping — `…4587545 → 4588321 → 4588731…` — and reached block **4,602,721** with `max_block_height` equal to it, i.e. caught up to the tip. Zero `BlockValidationError`, and no occurrence of 4,588,007.

That's one case verified end to end, not a claim about every instance of the rule.

PR #1268, opened 2026-08-21
Remove grok artifacts

yHSJ:
As more developers begin to adopt AI tooling into their workflow, we will (and have) end up seeing artifacts committed to the repo. In general, these are not things that should be included in the repo, they're specific to the individual user's workflow and tooling. There may be exemptions from that rule, such as an AGENTS.md that includes some rules/guidelines we want to enforce globally.


.gitignore
```diff
@@ -76,6 +76,8 @@ scripts/demos/*/run/**
 scripts/demos/*/keys/**/*.skey
 
 /.opencode/
+.grok/
+.delta/
```
yHSJ, line 80: `Delta` is a new tool I now have beta access to try out (https://delta.dev). I am preemptively git ignoring the directory, just in case.

rkuhn:
Adding any such directories to `/.gitignore` is a good move to avoid committing things by accident. The skill you are removing, on the other hand, is specific to Amaru (it is specific to fixing the trace-based unit tests of consensus), so please don’t remove that. I think @etorreborre has also been using that with Claude.

yHSJ:
> The skill you are removing, on the other hand, is specific to Amaru (it is specific to fixing the trace-based unit tests of consensus), so please don’t remove that. I think @etorreborre has also been using that with Claude.

I still don't think those skills should be committed in the repo. If we decide we *do* want shared skills for Amaru, we should create an `amaru-skills` directory, or a new repo.

rkuhn:
I went ahead and merged my proposed alternative, removing the files remains an easy thing if we still want that. Putting the skills into a different repo defeats the purpose, so that’s not a solution, and it turns out there is a vendor-agnostic place to put these.

yHSJ:
> I went ahead and merged my proposed alternative, removing the files remains an easy thing if we still want that. Putting the skills into a different repo defeats the purpose, so that’s not a solution, and it turns out there is a vendor-agnostic place to put these.

 I'm not sure putting the skills in a different repo defeats the purpose at all, it would be quite easy to tell an agent to look in a different location for skills. It feels to me like skills are a bit like IDE configuration or local commit hooks. They're specific to the person (even if they're used by multiple people) and shouldn't be part of the repository. But, I'm very open to having a more involved conversation.
 
 Either way, thanks for at least moving it! Closing this.

rkuhn:
I'm with you almost all the way. The skill here would be quite useless in my user directory because it only applies to Amaru. And putting the information where to find this skill into the repo (because nothing else can be the anchor) requires a specific pointer, so I don't see a large difference to just inlining the skill definition.

Anyway, let's discuss this with the others after their respective vacations.

PR #1278, opened 2026-08-24
chore: improve ios support

jeluard:
Make sure `amaru-node` can be used to build ios based apps.

## Summary by CodeRabbit

* **New Features**
  * Added an option to disable system metrics collection when configuring telemetry.
  * System metrics collection remains enabled by default.
  * Exposed telemetry configuration options for integration.

* **Bug Fixes**
  * Improved iOS compatibility for RocksDB by adjusting memory allocator settings.

### DIFF
--- a/crates/amaru-node/src/telemetry.rs
+++ b/crates/amaru-node/src/telemetry.rs
@@ -15,8 +15,8 @@
 //! Embedder-facing observability install helpers.
 //!
 //! Used by thin programs such as the `run_until` example so OTLP metrics/traces
-//! can be enabled without the product CLI. When OTLP is on, also starts
-//! process/build gauges via [`crate::system_metrics`].
+//! can be enabled without the product CLI. When OTLP is on, process/build gauges
+//! are collected by default; [`TelemetryOptions`] can disable their sysinfo poller.
 //!
 //! Console, JSON, and OTEL log/trace layers are the same CBOR-aware formatters
 //! as the product binary (`CborConsoleEventFormat`, `CborJsonEventFormat`,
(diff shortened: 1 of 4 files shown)


crates/amaru-node/src/telemetry.rs
```diff
@@ -196,8 +217,12 @@ impl Telemetry {
         let meter = Arc::new(Meter::from(meter_provider.meter(METRICS_METER_NAME)));
         // Same process/build gauges as the product binary so e2e compare-metrics
         // sees the full supported set when OTLP is enabled.
-        let system_metrics = track_system_metrics(Arc::clone(&meter), BuildIdentity::default())
-            .map_err(|e| anyhow!("start system metrics: {e}"))?;
+        let system_metrics = if options.collect_system_metrics {
+            track_system_metrics(Arc::clone(&meter), BuildIdentity::default())
+                .map_err(|e| anyhow!("start system metrics: {e}"))?
```
rkuhn: I switched from `Box<dyn Error>` to `anyhow::Error` yesterday, which means that we can now provide the proper causal chain by using the following instead:

```suggestion
                .context("start system metrics")?
```

rkuhn (approved):
Nice! I take this to mean that you’re building an iOS app based on `amaru-node`? Very nice!

Just a tiny nitpick regarding error propagation (since you’re touching this campsite — boy-scout rule).

crates/amaru-node/src/telemetry.rs
jeluard: Thx, good to know!

PR #1279, opened 2026-08-24
move formatting choice into trace schema

### LINKED ISSUE #1263: ensure emitted trace conform to emitted schema
Currently, the schema declared in `amaru-observability` only specifies which type of value we put into the macro at the call site, it doesn’t specify which type will appear in the JSON / CBOR / whatever. I think we should regulate both, which means removing the choice at the call site and defining the rendering within the schema definition instead. Then we are in a position to emit the actual JSON schema.

```
                FETCH_NEXT {
                    tags: cpu
                    required point: amaru_kernel::Point
                    required header_hash: %amaru_kernel::HeaderHash
                }
```

This would require the correct types as arguments and render `Point` as a structured object (an array containing slot and hash, the latter currently serializes as a String — which I don’t agree with since we do use CBOR that has byte strings) while the `HeaderHash` would be serialized via `Display::fmt` as a `String`, regardless of its serde instance. `?` could be used for `Debug`.

We should add `schemars` and derive or implement `JsonSchema` for all relevant types, i.e. those arguments that are used without Display or Debug modifier.

## Implementation

- remove ability to specify formatter at observability macro use site (also: allow short-hand notation like `tracing`)
- add optional formatter specification to schema declaration (`%` for Display, `?` for Debug)
- constrain the property’s type to implement the required trait:
    - `%`: Display
    - `?`: Debug
    - otherwise: `serde::Serialize + JsonSchema`
- correctly emit the property’s serialized type schema with the tracing schema
- add unit tests ascertaining non-regression

rkuhn:
this allows emitting a reliable JSON schema that cannot be overridden
using % formatter at the use-site

fixes #1263

## Summary by CodeRabbit

* **New Features**
  * Trace schemas now provide richer metadata and accurate JSON Schema types.
  * Added schema-defined display/debug formatting and tracing-value passthrough.
  * Point values use compact tuple encoding with CBOR-compatible hashes.
  * Hash values serialize as readable hex or compact binary data.
  * Trace schema dumps now generate structured JSON schemas automatically.

* **Documentation**
  * Updated trace and observability guidance for field types, rendering, and encoding.

* **Chores**
  * Updated the pinned Rust nightly toolchain used by builds and tests.


crates/amaru-bootstrap/src/bootstrap/mod.rs
```diff
@@ -288,7 +288,7 @@ async fn fetch_headers_from_point(
                 }
             }
             NextResponse::RollBackward(point, tip) => {
-                info!(bootstrap::fetch::ROLLBACK, ?point, ?tip);
+                info!(bootstrap::fetch::ROLLBACK, point = format!("{point:?}"), tip = format!("{tip:?}"));
```
rkuhn: This should use structured logging:
```suggestion
                info!(bootstrap::fetch::ROLLBACK, point, tip);
```

crates/amaru-consensus/src/stages/track_peers/mod.rs
```diff
@@ -657,7 +657,7 @@ impl TrackPeers {
                     consensus::perf::header::LIFECYCLE,
                     peer = peer,
                     header_hash = header.hash(),
-                    error = %error,
+                    error = error.to_string(),
```
rkuhn, line 660: should test what the result looks like, perhaps add a unit test to ensure usefulness (similar for all error logging, but perhaps one solution works everywhere)

crates/amaru-kernel/src/cardano/epoch.rs
```diff
@@ -23,7 +23,20 @@ use num::{CheckedAdd, CheckedSub};
...
+    PartialEq,
+    PartialOrd,
+    Ord,
+    Eq,
+    serde::Serialize,
+    serde::Deserialize,
+    Default,
+    schemars::JsonSchema,
+)]
+#[schemars(transparent)]
```
rkuhn, line 39: unit test missing for serde representation (same for other transparent newtypes)

crates/amaru-ledger/src/governance/ratification/proposals_forest.rs
```diff
@@ -379,9 +379,9 @@ impl ProposalsForestCompass {
         if proposed_in > &forest.current_epoch {
             debug!(
                 ledger::proposal::SKIP,
-                id = %id,
-                %proposed_in,
-                ratifying_epoch = %forest.current_epoch,
+                id = id.to_string(),
+                proposed_in = proposed_in.to_string(),
```
rkuhn: should use structured logging
```suggestion
                proposed_in,
```

crates/amaru-ledger/src/governance/ratification/proposals_forest.rs
```diff
@@ -379,9 +379,9 @@ impl ProposalsForestCompass {
         if proposed_in > &forest.current_epoch {
             debug!(
                 ledger::proposal::SKIP,
-                id = %id,
-                %proposed_in,
-                ratifying_epoch = %forest.current_epoch,
+                id = id.to_string(),
```
rkuhn: should also use structured logging (everywhere)

crates/amaru-ledger/src/rules/transaction/phase_two/mod.rs
```diff
@@ -147,7 +147,7 @@ where
             let span_execute = debug_span!(
                 parent: &span,
                 ledger::transaction::script::EXECUTE,
-                purpose = %purpose,
+                purpose = purpose.to_string(),
```
rkuhn: Should probably also use the RedeemerTag type in the schema and render as String

crates/amaru-ledger/src/summary/rewards.rs
```diff
@@ -410,15 +410,15 @@ impl RewardsSummary {
...
-            %efficiency,
-            %incentives,
-            %treasury_tax,
-            %total_rewards,
-            %available_rewards,
-            %effective_rewards,
-            pots_reserves = %pots.reserves,
-            pots_treasury = %pots.treasury,
-            pots_fees = %pots.fees,
+            efficiency = efficiency.to_string(),
```
rkuhn, line 413: This is a question @yHSJ: would you rather see this as `"3/7"` or as `[3,7]`?

crates/amaru-observability/macros/src/traces.rs
```diff
@@ -505,15 +492,15 @@ pub fn expand_trace_record(input: TokenStream) -> TokenStream {
 ///
 /// ```text
 /// trace_event!(LEVEL, SCHEMA, field = value, ...);   // type-checked, typed or CBOR
-/// trace_event!(LEVEL, SCHEMA, field = %value, ...);  // pre-formatted, Display-rendered
-/// trace_event!(LEVEL, SCHEMA, field = ?value, ...);  // pre-formatted, Debug-rendered
-/// trace_event!(LEVEL, SCHEMA, value, %value, ?value) // shorthands for the above
+/// trace_event!(LEVEL, SCHEMA, field = value, ...);  // pre-formatted, Display-rendered
+/// trace_event!(LEVEL, SCHEMA, field = value, ...);  // pre-formatted, Debug-rendered
```
rkuhn: comment incorrect: structured logging, formatting according to schema definition

crates/amaru/src/bin/amaru/cmd/dev/traces/dump.rs
```diff
@@ -12,10 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::{collections::BTreeMap, env, io};
+use std::io;
 
 use amaru::lifecycle::{Runnable, RuntimeKind};
-use amaru_observability::{aliases, registry::SchemaEntry};
```
yHSJ, line 18: We can probably remove `aliases` now, right? It is totally unused after this change.

yHSJ (approved):

crates/amaru-ledger/src/summary/rewards.rs
yHSJ: As a human, I would rather see `"3/7"`. However, the span consumers could provide whatever formatting makes sense for them from `[3,7]`. This is the least helpful answer though 😛

rkuhn: nah, the first part of what you said tips the scales for me: the TUI should show `"3/7"`, so let’s leave it

crates/amaru/src/bin/amaru/cmd/dev/traces/dump.rs
rkuhn: yes, good catch!

PR #1281, opened 2026-08-24
fix: properly parse new point format

jeluard:
`yaci` nightly test relies on `amaru dev chain best-chain` to extract current tip. Update bash script to reflect for recent change in Point format.


rkuhn (approved):
Thanks for the fix-up! I keep seeing new interfaces that we have … sorry about missing this one.

PR #1282, opened 2026-08-24
`manual_async_fn` switch affected fn from `interpreter` to `run_stage_boxed`

### LINKED ISSUE #161658: new trait solver regression in `amaru-pure-stage` library
Compiling [amaru-pure-stage](https://github.com/pragma-org/amaru/blob/d0d7cf43008923dabae227996f120756c84627fd/crates/amaru-pure-stage/src/tokio.rs#L515) fails with the following error message on `nightly-2026-08-22`:

```
error[E0275]: overflow evaluating the requirement `impl futures_util::Future<Output = ()>: std::marker::Send`
   --> crates/amaru-pure-stage/src/tokio.rs:515:47
    |
515 |                     let handle = tokio::spawn(stage);
    |                                  ------------ ^^^^^
    |                                  |
    |                                  required by a bound introduced by this call
    |
note: required by a bound in `tokio::spawn`
   --> /Users/rkuhn/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tokio-1.53.1/src/task/spawn.rs:176:21
    |
174 |     pub fn spawn<F>(future: F) -> JoinHandle<F::Output>
    |            ----- required by a bound in this function
175 |     where
176 |         F: Future + Send + 'static,
    |                     ^^^^ required by this bound in `spawn`
```

It does not fail with `RUSTFLAGS=-Znext-solver=coherence` or on `nightly-2026-08-21` (no matter which solver).

lcnr:
This fixes https://github.com/rust-lang/rust/issues/161658. Best reviewed with whitespace changes hidden.

Changing `run_stage_boxed` to explicitly return a `Send` future is more effective than doing it for `interpreter`.

The reason that this is required at all is that these two functions recursively depend on each other. The minimal example would be
```rust
// This lifetime is what causes #161658, otherwise it's irrelevant
async fn interpreter<'a>() {
    async move { tokio::spawn(run_stage_boxed()); }
}

async fn run_stage_boxed() {
    interpreter().await
}
```

Type checking `interpreter` has to prove `run_stage_boxed::{opaque}: Send`. The future returned by `run_stage_boxed` needs to store the future returned by `interpreter` at an await-point. This means that this state machine is only `Send` if the state machine returned by `interpreter` is `Send`. We only know the state machine returned by `interpreter` after we've finished type checking that function.

This cyclic reasoning which goes through different functions is quite brittle and fairly easily results in query cycles or whatever causes the current error to happen which pushed you to initially change `interpreter` to explicitly return `impl Send`.

The reason https://github.com/rust-lang/rust/issues/161658 happens for your current setup with `-Znext-solver` is as follows. We have to prove `run_stage_boxed::{opaque}: Send` for the call to `tokio::spawn`. Doing so requires proving `for<'a> interpreter<'a>::{opaque}: Send`. This is then recursively requires `run_stage_boxed::{opaque}: Send` again, which is a cycle. This cycle should get accepted, however, for very technical reason, this cycle ends up *registering a new non-defining use of `interpreter<'!a>::{opaque}` where `'!a` is a unique placeholder region*. The fact that every time we prove `for<'a> interpreter<'a>::{opaque}: Send` emits a new non-defining use causes us to end up failing with overflow.

This is unfortunately a really technical issue and I don't know how to explain it better. This is a limitation of the current implementation and something we intend to fix long term. I don't expect us to fix this issue in the near future however, so working around it feels easier for now.

We avoid this issue by swapping which function explicitly returns a `Send` future. We only ever prove `run_stage_boxed::{opaque}: Send` in `interpreter` and not `interpreter::{opaque}: Send` when checking `run_stage_boxed`. This means by explicitly promising that `run_stage_boxed::{opaque}: Send` holds checking this when type checking `interpreter` never looks into that function, avoiding both any cycles and the overflow issue.

Thank you @rkuhn for reporting this issue and I am sorry that you're encountering this breakage.

## Summary by CodeRabbit

* **Refactor**
  * Updated internal asynchronous stage execution to improve compatibility with the Rust toolchain.
  * Preserved existing stage-loop behavior, message scheduling, supervision, tracing, and effect handling.

### DIFF
--- a/crates/amaru-pure-stage/src/tokio.rs
+++ b/crates/amaru-pure-stage/src/tokio.rs
@@ -272,104 +272,119 @@ enum PriorityMessage {
     Tombstone(Box<dyn SendData>),
 }
 
-async fn run_stage_boxed(
+// clippy is lying, changing to async fn does not work.
+#[expect(clippy::manual_async_fn)]
+fn run_stage_boxed(
     mut state: Box<dyn SendData>,
     mut rx: Receiver<Box<dyn SendData + 'static>>,
     transition: TransitionFactory,
     stage_name: Name,
     inner: Arc<TokioInner>,
-) {
-    tracing::debug!("running stage `{stage_name}`");
+) -> impl Future<Output = ()> + Send {
+    // Trying to write this as an async fn encounters a bug
+    // in the Rust compiler.
+    //
+    // cc <https://github.com/rust-lang/rust/issues/161658>.
+    async move {
+        tracing::debug!("running stage `{stage_name}`");
 
-    let effect = Arc::new(Mutex::new(None));
-    let mut transition = transition(effect.clone());
+        let effect = Arc::new(Mutex::new(None));
+        let mut transition = transition(effect.clone());
 
-    // this also contains tasks tracking the termination of spawned stages, which when dropped
-    // will terminate those spawned stages
-    let mut timers = FuturesUnordered::<BoxFuture<'static, PriorityMessage>>::new();
-    let mut cancel_senders = BTreeMap::<ScheduleId, watch::Sender<bool>>::new();
-    // Armed schedules not yet consumed by receive (matches simulation `scheduled_pending`).
-    let mut scheduled_pending: usize = 0;
+        // this also contains tasks tracking the termination of spawned stages, which when dropped
+        // will terminate those spawned stages
+        let mut timers = FuturesUnordered::<BoxFuture<'static, PriorityMessage>>::new();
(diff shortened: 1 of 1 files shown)


rkuhn:
Wow, thanks for the super-fast response and the great explanation! Let’s see whether we can get our rather picky CI to accept it (we haven’t yet figured out how to gracefully deal with outside PRs that can’t access the workflow secrets). The change looks good to me, though.

yHSJ:
Closed in favor of #1283, which moves the commit to a branch in the main repo, allowing the CI to run as expected.

PR #1285, opened 2026-08-25
New "Compact" Data Structures

### LINKED ISSUE #1147: Performance: define `CompactSet` & `CompactMap` for small b-tree structures
### Abstract

Introduce `CompactSet` and `CompactMap` data structures for Amaru as memory-efficient replacements for `BTreeSet` and `BTreeMap` in the common case of very small collections. The goal is to provide a near drop-in API (limited to the operations currently used within Amaru) while significantly reducing the memory footprint of collections containing only a handful of elements.

### Why?

Many data structures in Amaru contain sets or maps that are (potentially) instantiated millions of times, yet almost always contain only a few elements in practice. Examples include:

- Pool parameter owners.
- DiffBind and DiffSet instances in the volatile database.
- Similar repeated structures throughout the aggregate state.

While `BTreeSet` and `BTreeMap` are excellent general-purpose containers, they are [optimized around B-tree nodes that reserve space for multiple entries](https://doc.rust-lang.org/alloc/collections/btree_map/struct.BTreeMap.html#background) and provide stable references into those nodes. As a result, even a single-element collection carries the overhead of an allocated node whose capacity is largely unused. That number is [11 in the default implementation](https://doc.rust-lang.org/src/alloc/collections/btree/node.rs.html#43) which means that, a B-tree of a single element takes the space of 11 elements. 

Given the sheer number of these small collections in memory, this overhead becomes a meaningful contributor to Amaru's overall memory consumption. A compact representation can substantially reduce resident memory without affecting the behaviour of larger collections.

### How?

The implementation should store elements inline using a fixed-capacity vector while the collection remains below a configurable threshold (e.g. fewer than five elements). Once that threshold is exceeded, the collection should transparently migrate to a standard `BTreeSet` or `BTreeMap,` preserving their semantics and performance characteristics for larger collections.

Implement two new container types:

- `CompactSet<T>`
- `CompactMap<K, V>`

The implementation should:

- Expose only the subset of the `BTreeSet`/`BTreeMap` APIs that are currently required by Amaru, allowing them to be used as drop-in replacements where applicable.
- Store elements inline using a fixed-capacity vector while the number of elements remains below a threshold (initially < 5).
- Automatically migrate to an internally owned `BTreeSet` or `BTreeMap` once the threshold is exceeded.
- Preserve the expected ordering and semantics of the corresponding B-tree collections.
- Make the transition between representations transparent to callers.

The exact inline representation (e.g. ArrayVec, fixed array, or another small-vector implementation) is left to the implementation, provided it avoids heap allocation for the small-case representation.

- Implement unit tests covering all supported operations in both representations using `BTreeSet` / `BTreeMap` as model.
- Verify that promotion preserves ordering, contents, and observable behaviour.
- Add dedicated benchmarks to the existing volatile_db benchmark suite exercising workloads with a large number of small sets/maps, representative of real Amaru usage.
- Compare benchmark results against the existing implementation and demonstrate a measurable reduction in memory usage with no significant regression in execution time for the targeted workloads.
- At least one existing Amaru component (e.g. DiffSet, DiffBind, or pool owners) should be migrated to validate the API and demonstrate practical applicability.

### Discussion points

_No response_

_No response_

### Checklist

yHSJ:
Fixes #1147

`BTree` collections are not ideal for small collections, as they allocate space for 11 elements per node. That means if you have a single value, you've allocated 11x more memory than needed.

This PR introduces two new data structures; `CompactMap` and `CompactSet`, which behave like a `Vec` (without amortized growth) when the collection's cardinality is below some threshold, "promoting" to a `BTree` collection once it crosses that threshold. 

To measure this, I introduced two new benchmark scenarios:
- `Observed`: This creates a `VolatileDB` that emulates a real series of 2,160 blocks in epoch `638` on mainnet.
- `Singleton`: This creates a `VolatileDB` that stores only 1 element in every collection.

These benchmarks show this new data structure improved memory usage. The `observed` benchmark fell from a baseline of `24.775 MB` to `21.028 MB`. And it actually *sped up* the benchmarks: `roll_forward` 4.6ms -> 2.9ms, `hydrate_context` 4.4ms -> 3.1ms, `switch_to_fork` 28.1ms -> 23.9ms.

The singleton case (which is the "worst case scenario" for a `BTree` collection) went from `20.3MB` to `8.9MB`. with roughly 2x speed improvements across the board.

EDIT: I have replaced the `Observed` scenario with the pre-existing `Mixed` scenario, but I added the ability to specify weights for each field in a volatile DB, to generate mainnet-like VolatileDBs.

## Summary by CodeRabbit

* **New Features**
  * Added compact map and set collections optimized for small datasets, with automatic promotion for larger collections.
  * Added configurable benchmark scenarios, workload mixes, block-fill targets, and retained-memory reporting.
  * Added support for singleton and observed workload scenarios in volatile database benchmarks.
  * Improved public access to collection and UTxO difference types.

* **Performance**
  * Updated volatile ledger state handling to use compact collections, reducing overhead for small changesets.

* **Refactor**
  * Consolidated shared UTxO difference types and updated related collection interfaces.


crates/amaru-kernel/src/lib.rs
```diff
@@ -36,6 +36,8 @@ pub use traits::{
 
 mod data_structures;
 pub use data_structures::{
+    compact_collections,
+    compact_collections::{CompactMap, CompactSet},
```
rkuhn:
```suggestion
    compact_collections::{self, CompactMap, CompactSet, Entry},
```
is I think more idiomatic

crates/amaru-kernel/src/data_structures/compact_collections/small_buffer.rs
```diff
@@ -0,0 +1,60 @@
...
+// limitations under the License.
+
+use std::ops::Deref;
+
+/// A sorted vector for the small regime.
+///
+/// Derefs to a slice for reads; mutations go through the inherent methods, which preserve
+/// ordering and grow exactly rather than amortized growth.
+#[derive(Debug, Clone)]
+pub(crate) struct SmallBuffer<T> {
```
rkuhn: should probably be `pub(super)`

crates/amaru-kernel/src/data_structures/compact_collections/small_buffer.rs
```diff
@@ -0,0 +1,60 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::ops::Deref;
+
+/// A sorted vector for the small regime.
```
rkuhn: Nothing about the implementation guarantees or enforces any ordering, so this wording is misleading at best.

crates/amaru-kernel/src/data_structures/compact_collections/compact_set.rs
```diff
@@ -0,0 +1,282 @@
...
+}
+
+/// Content-based equality: a promoted set equals a small one holding the same values.
+impl<T: Ord, const N: usize> PartialEq for CompactSet<T, N> {
+    fn eq(&self, other: &Self) -> bool {
+        self.len() == other.len() && self.iter().zip(other.iter()).all(|(a, b)| a == b)
+    }
+}
+
+impl<T: Ord, const N: usize> Eq for CompactSet<T, N> {}
```
rkuhn: This should require `T: Eq` as well.

crates/amaru-kernel/src/data_structures/compact_collections/compact_set.rs
```diff
@@ -0,0 +1,282 @@
...
+        self.len() == 0
+    }
+
+    pub fn contains<Q>(&self, value: &Q) -> bool
+    where
+        T: Borrow<Q>,
+        Q: Ord + ?Sized,
+    {
+        match &self.storage {
+            SetStorage::Small(entries) => entries.binary_search_by(|candidate| candidate.borrow().cmp(value)).is_ok(),
```
rkuhn: If `SmallBuffer` were sorted, then the `.contains` should just forward to a `fn contains_by(...)` on it.

crates/amaru-kernel/src/data_structures/compact_collections/compact_set.rs
```diff
@@ -0,0 +1,282 @@
...
+            SetStorage::Tree(entries) => entries.contains(value),
+        }
+    }
+
+    pub fn insert(&mut self, value: T) -> bool {
+        let small_search = match &self.storage {
+            SetStorage::Small(entries) => Some(entries.binary_search_by(|candidate| candidate.cmp(&value))),
```
rkuhn: same as above

crates/amaru-kernel/src/data_structures/compact_collections/compact_set.rs
```diff
@@ -0,0 +1,282 @@
...
+        T: Borrow<Q>,
+        Q: Ord + ?Sized,
+    {
+        match &mut self.storage {
+            SetStorage::Small(entries) => entries
+                .binary_search_by(|candidate| candidate.borrow().cmp(value))
+                .ok()
+                .and_then(|index| entries.remove(index))
+                .is_some(),
+            SetStorage::Tree(entries) => entries.remove(value),
```
rkuhn, line 101: should this not demote when `entries.len() <= N` after the removal?

crates/amaru-kernel/src/data_structures/compact_collections/compact_set.rs
```diff
@@ -0,0 +1,282 @@
...
+        }
+    }
+
+    pub fn iter(&self) -> CompactSetIter<'_, T> {
```
rkuhn, line 105: I’d return `Box<dyn Iterator>`: the enum looks nice, but it introduces a conditional into each and every `.next()` call. The dynamic dispatch on the other hand is prefetched by the CPU without any choice.

crates/amaru-kernel/src/data_structures/compact_collections/compact_set.rs
```diff
@@ -0,0 +1,282 @@
...
+
+impl<T: Ord, const N: usize> Default for CompactSet<T, N> {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl<T: Ord, const N: usize> FromIterator<T> for CompactSet<T, N> {
+    fn from_iter<I: IntoIterator<Item = T>>(iter: I) -> Self {
+        let mut set = Self::new();
```
rkuhn: this should use the `iter.size_hint()` to preallocate where possible

crates/amaru-kernel/src/data_structures/compact_collections/compact_set.rs
```diff
@@ -0,0 +1,282 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::{borrow::Borrow, collections::BTreeSet, iter::FusedIterator, slice};
+
+use super::small_buffer::SmallBuffer;
+
+#[derive(Debug, Clone)]
+enum SetStorage<T: Ord, const N: usize> {
```
rkuhn: `N` isn’t needed here, correct? And the `T: Ord` bound should also not be on the data structure but only on its usage.

crates/amaru-kernel/src/data_structures/compact_collections/compact_map.rs
```diff
@@ -0,0 +1,449 @@
...
+
+use std::{
+    borrow::Borrow,
+    collections::{BTreeMap, btree_map},
+    iter::FusedIterator,
+    slice,
+};
+
+use super::small_buffer::SmallBuffer;
+
```
rkuhn, line 23: comments from CompactSet apply to CompactMap mutatis mutandis

crates/amaru-kernel/src/data_structures/compact_collections/compact_map.rs
```diff
@@ -0,0 +1,449 @@
...
+    Tree(btree_map::OccupiedEntry<'a, K, V>),
+}
+
+impl<'a, K: Ord, V, const N: usize> OccupiedEntry<'a, K, V, N> {
+    pub fn get(&self) -> &V {
+        match &self.0 {
+            OccupiedInner::Small { entries, index } => match entries.get(*index) {
+                Some((_, value)) => value,
+                None => unreachable!("occupied entries hold a valid index"),
+            },
```
rkuhn: This dance seems extraneous because it leads to the exact same outcome as saying `... => entries[*index]` (which should of course implement Index & IndexMut).

crates/amaru-kernel/src/data_structures/compact_collections/compact_map.rs
```diff
@@ -0,0 +1,449 @@
...
+                    map.promote();
+                }
+
+                match &mut map.storage {
+                    MapStorage::Small(entries) => {
+                        entries.insert(index, (key, value));
+                        match entries.get_mut(index) {
+                            Some((_, value)) => value,
+                            None => unreachable!("value was just inserted at this index"),
+                        }
```
rkuhn: here, too, I’d just return `&mut entries[index].1`

crates/amaru-ledger/src/state/volatile/fragment/diff_set.rs
```diff
@@ -21,21 +21,24 @@ use crate::state::volatile::Existence;
...
+impl<K: Ord, V, const PRODUCED: usize, const CONSUMED: usize> Default for DiffSet<K, V, PRODUCED, CONSUMED> {
     fn default() -> Self {
         Self { consumed: Default::default(), produced: Default::default() }
     }
 }
 
-impl<K: Ord, V> DiffSet<K, V> {
+impl<K: Ord, V, const PRODUCED: usize, const CONSUMED: usize> DiffSet<K, V, PRODUCED, CONSUMED> {
     /// Borrow all keys and values in this diff.
     pub fn as_refs(&self) -> DiffSet<&K, &V> {
```
rkuhn: This method probably shouldn’t exist since it looks innocent but may allocate huge trees. Is it even used?

crates/amaru-kernel/src/data_structures/compact_collections/compact_set.rs
yHSJ: Yeah, good point. Just have to be careful of duplicate values

crates/amaru-kernel/src/data_structures/compact_collections/small_buffer.rs
yHSJ: Yeah, you're right. I can either make it clear the users of this type own the sorted invariant (which is currently the case), or I could just change this to be a `SmallSortedBuffer` and enforce the invariant at this level. I think I like that more.

crates/amaru-kernel/src/data_structures/compact_collections/compact_map.rs
yHSJ: Yeah, I think this is not needed. And, if I enforce the sorted invariant at the `SmallBuffer` layer, this shouldn't even be needed...

crates/amaru-kernel/src/data_structures/compact_collections/compact_set.rs
yHSJ: No, I made it explicit that we do *not* demote to the smaller variant ever. I felt that the allocations that would happen around `N` would be worse than anything if we ever *did* do significant removals and insertions. Ultimately though, we don't really remove from these collections very frequently, from what I can tell.

crates/amaru-kernel/src/data_structures/compact_collections/compact_set.rs
yHSJ: So this one I am not certain about, and I trust that you know more than me. I went with this mostly *because* it looks nice and, often, Rust code that looks like is the most optimal (with many obvious exceptions).

I did consult Claude on this, which told me that because the enum's branch is the same on every `.next()` call, it's perfectly predicted and the `match` compiles to inlineable code. `dyn` would force an indirect call and prevent the inlining. It would also introduce a heap allocation per `iter()` call. It gave me references to `itertools` and `either::Either` which both use this pattern. So I could've *accidentally* gotten this one right.

I'll benchmark both and get an idea of runtime impact.

yHSJ: Ok, so I tested both. The `dyn` was between 5 and 20 times slower than the enum variant. Memory was the same and `dyn` had more allocations (obviously)

crates/amaru-kernel/src/data_structures/compact_collections/compact_map.rs
```diff
@@ -115,11 +114,16 @@ impl<K: Ord, V, const N: usize> CompactMap<K, V, N> {
...
-    fn promote(&mut self) {
+    /// Move all small entries into the tree, permanently, and return it. No-op when already
+    /// tree-backed.
+    fn promote(&mut self) -> &mut BTreeMap<K, V> {
         if let MapStorage::Small(entries) = &mut self.storage {
             self.storage = MapStorage::Tree(entries.take().into_iter().collect());
         }
+        match &mut self.storage {
+            MapStorage::Tree(entries) => entries,
+            MapStorage::Small(_) => unreachable!("promotion always ends in tree storage"),
```
rkuhn: If you do the conversion in this match then you won’t need a panic point (those are rather unfriendly to code optimisations and size).

crates/amaru-kernel/src/data_structures/compact_collections/compact_set.rs
```diff
@@ -60,43 +60,35 @@ impl<T: Ord, const N: usize> CompactSet<T, N> {
...
-                }
-                match &mut self.storage {
-                    SetStorage::Small(entries) => {
-                        entries.insert(index, value);
-                        true
+        if let SetStorage::Small(entries) = &mut self.storage {
+            match entries.binary_search_by(|candidate| candidate.cmp(&value)) {
+                Ok(_) => return false,
+                Err(index) => {
+                    if entries.len() < N {
```
rkuhn: This should be checked before entering this branch and then `return match ...` is more idiomatic and makes it immediately clear that this `if` branch always returns.

crates/amaru-kernel/src/data_structures/compact_collections/compact_set.rs
rkuhn: Yeah, if you microbenchmark this then the conclusion will always be “moar inlining”. This is not true for the bigger picture, though. All that inlining makes Amaru into the fatty >40MB it already is — I’m pretty sure that it could shrink by a factor ten without relevant reduction in runtime performance — consider also what this does to CPU I-caches. This is the uphill battle in Rust where any dynamic dispatch is demonised. I won’t continue fighting here because the ecosystem always makes projects slip down this slope one function signature at a time. But I might continue fighting for the source code lines reduction that results from the removal of `enum CompactSetIter`.

crates/amaru-kernel/benches/small_collections.rs
KtorZ, line 1: Not quite sure to understand what those benchmarks are measuring? Especially on sizes that are this small 🤔 ?

crates/amaru-kernel/src/data_structures/compact_collections/compact_map.rs
```diff
@@ -0,0 +1,457 @@
...
+/// promoting to a [`BTreeMap`].
+#[derive(Debug, Clone)]
+pub struct CompactMap<K, V, const N: usize> {
+    storage: MapStorage<K, V>,
+}
+
+/// Content-based equality: a promoted map equals a small one holding the same entries.
+impl<K: Ord, V: PartialEq, const N: usize> PartialEq for CompactMap<K, V, N> {
+    fn eq(&self, other: &Self) -> bool {
+        self.len() == other.len() && self.iter().zip(other.iter()).all(|(a, b)| a == b)
```
KtorZ, line 40: Interestingly; is there any test controlling that (for instance)`Small([(k1, v1)])` is equals to `Tree(k1 -> v1)` ?

crates/amaru-kernel/src/data_structures/compact_collections/compact_map.rs
```diff
@@ -0,0 +1,457 @@
...
+    /// tree-backed.
+    fn promote(&mut self) -> &mut BTreeMap<K, V> {
+        loop {
+            match &mut self.storage {
+                MapStorage::Tree(entries) => return entries,
+                MapStorage::Small(entries) => {
+                    self.storage = MapStorage::Tree(entries.take().into_iter().collect());
+                }
+            }
+        }
```
KtorZ, line 127: Strangely elegant.

crates/amaru-kernel/src/data_structures/compact_collections/small_sorted_buffer.rs
```diff
@@ -0,0 +1,77 @@
...
+    }
+}
+
+impl<T> Deref for SmallSortedBuffer<T> {
+    type Target = [T];
+
+    fn deref(&self) -> &Self::Target {
+        &self.entries
+    }
+}
```
KtorZ, line 77: This one deserves some unit and property tests as well.

crates/amaru-ledger/benches/state/volatile_db/common/scale.rs
```diff
@@ -20,24 +20,133 @@ use amaru_kernel::{MAINNET_DEFAULT_PROTOCOL_PARAMETERS, MAINNET_GLOBAL_PARAMETER
...
+    }
+}
+
+/// Relative frequency of each entity in the mixed workload,
+/// expressed as the mean number of items per collection over a 2160 block window.
+///
+/// The default values are sampled over blocks [13586560, 13588719].
+/// Entities absent from that window default to 1
+#[derive(Clone, Copy, Debug)]
+pub struct MixedWeights {
```
KtorZ, line 61: Please move, alongside its companion methods, into a separate module.

crates/amaru-ledger/src/state/volatile/fragment.rs
```diff
@@ -51,18 +48,27 @@ pub use tests::*;
...
 // ----------------------------------------------------------------------------------- VolatileFragment
 
+pub type AccountDiff = DiffBind<Credential, (PoolId, CertificatePointer), (DRep, CertificatePointer), Lovelace, 4, 2>;
+pub type DRepDiff = DiffBind<Credential, Box<Anchor>, Empty, DRepRegistration, 1, 1>;
+pub type CommitteeDiff = DiffBind<Credential, ConstitutionalCommitteeMemberStatus, Epoch, Empty, 1, 1>;
+pub type DRepDeregistrations = CompactMap<Credential, CertificatePointer, 1>;
+pub type Withdrawals = CompactSet<Credential, 4>;
+pub type Proposals = CompactMap<ProposalId, Arc<ProposalState>, 1>;
+pub type VoteDiff = DiffSet<BallotId, Ballot, 1, 1>;
+pub type UtxoDiff = DiffSet<TransactionInput, Arc<MemoizedTransactionOutput>, 64, 128>;
```
KtorZ, line 58: 64/128 seems like a lot 🤔? What's the rationale here?

KtorZ (commented):
Nice work; some remarks already while I try this out locally

crates/amaru-kernel/benches/small_collections.rs
yHSJ: I was measuring allocation counts, but I guess we don't really need that.

crates/amaru-kernel/src/data_structures/compact_collections/compact_map.rs
yHSJ: Not directly, no. I will add one directly comparing equalities, and I'll add one for `CompactSet` too

crates/amaru-kernel/src/data_structures/compact_collections/compact_map.rs
yHSJ: I was proud of this one, in a weird "I kind of hate it" way 😛

crates/amaru-ledger/src/state/volatile/fragment.rs
yHSJ: They came from the mainnet sample that I used to create the benchmarks. I was trying to find a number that meant that only the largest collections used a tree. Obviously these numbers were used for the benchmark runs, so it still sped things up, but I guess it doesn't make sense to have large Vec collections here. I can shrink them up.

KtorZ (approved):

PR #1290, opened 2026-08-27
run ledger on a dedicated thread

### LINKED ISSUE #1094: Move ledger to separate thread
All ledger operations are single-threaded and must be executed sequentially, while operations may be requested from separate entities (block validation, mempool, possible future cases like Leios integration). This matches the well-known Actor model, implemented using a dedicated thread that pulls requests from an mpsc channel and sends responses to oneshot channels that may be embedded in the work item messages as needed.

The integration with the current consensus & ledger architecture is straight-forward: instead of directly calling a ledger method, the BlockValidator creates the response oneshot channel, builds the request message, sends it to the ledger thread, and returns the oneshot receiver as the Future that will eventually contain the result.

Stake distribution computation can be moved to a different thread because it operates on immutable data. The result of that computation then needs to be injected into the consensus pure-stages in the same way as it is today with the `on_stake_dist_changed` callback.

bomanaps:
Closes #1094. The ledger now runs on its own dedicated thread that fully owns the ledger state, as described in EDR-012 callers send it request messages over a channel and get results back through a reply channel in each message, which removes the shared lock, the traits and effect wiring are unchanged, and stake distribution already ran on a background thread so nothing was needed there. Verified with the full workspace test suite plus a live preprod run, the node validated and adopted 24,830 blocks in 17 minutes through the new thread with no errors.

Screenshot 2026-08-27 at 16 53 00

## Summary by CodeRabbit

* **Changed**
  * Improved responsiveness during block and transaction validation.
  * Reduced contention during ledger and validation operations.
  * Improved fork switching and chain-tip handling.
  * Improved stake-pool relay candidate discovery.
  * Added clearer reporting when ledger validation services are unavailable.
  * Simplified block validator integration across node components.
  * Updated the release changelog with these improvements.


crates/amaru-consensus/src/block_validator.rs
```diff
@@ -26,55 +27,163 @@ use amaru_ledger::{
...
+        S: Store + Send + 'static,
+        HS: HistoricalStores + Send + Sync + 'static,
+    {
+        let (sender, mut receiver) = tokio::sync::mpsc::unbounded_channel();
+        thread::spawn(move || {
+            let mut ledger = LedgerThread { state, vm_eval_pool, chain_store };
+            while let Some(request) = receiver.blocking_recv() {
+                ledger.handle(request);
+            }
+        });
```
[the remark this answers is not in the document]
bomanaps: This design is intentional per EDR-012, which keeps the ledger outside pure_stage and prescribes exactly this channel-plus-oneshot shape, the effects boundary is unchanged, so simulation is unaffected.

crates/amaru-consensus/src/block_validator.rs
```diff
@@ -26,55 +27,172 @@ use amaru_ledger::{
...
-            .map_err(|error| TransactionValidationError::from(anyhow!(error)))
-    }
+/// Handle to the ledger state, which lives on a dedicated thread owning it exclusively.
+///
+/// Ledger operations are requested by sending a message to that thread and waiting for its
+/// response, so all of them execute sequentially regardless of how many entities (block
+/// validation, mempool, ...) hold a handle. The thread terminates once every handle is dropped.
+#[derive(Clone)]
+pub struct BlockValidator {
+    sender: UnboundedSender<LedgerRequest>,
```
rkuhn: Assuming a proof of the boundedness of the consensus processing and an audit that no other callers exist, it might be okay to punt on the bound of this channel. But even then I’d just use the proof to calculate the size of this channel instead of using `unbounded()`. As it stands, we should estimate a reasonable bound by counting call-sites, double that and fail hard when the queue is full.

crates/amaru-consensus/src/block_validator.rs
```diff
@@ -26,55 +27,172 @@ use amaru_ledger::{
...
     /// Set callback invoked when a new stake distribution is computed/available.
     /// The provided PoolSummaries should be used to update resources for header validation.
-    #[expect(clippy::unwrap_used)]
     pub fn set_on_stake_dist_updated(&self, callback: Arc<dyn Fn(PoolSummaries) + Send + Sync>) {
-        let mut state = self.state.lock().unwrap();
-        state.set_on_stake_dist_updated(callback);
+        let _ = self.sender.send(LedgerRequest::SetOnStakeDistUpdated(callback));
+    }
+
+    fn request<T>(&self, make: impl FnOnce(std::sync::mpsc::SyncSender<T>) -> LedgerRequest) -> Result<T, ThreadGone> {
```
rkuhn: This must be async, we cannot block Tokio threads waiting for a response on a sync channel.

crates/amaru-consensus/src/block_validator.rs
```diff
@@ -26,55 +27,172 @@ use amaru_ledger::{
...
+        response.recv().map_err(|_| ThreadGone)
+    }
+}
+
+#[derive(Debug)]
+struct ThreadGone;
+
+impl From<ThreadGone> for BlockValidationError {
+    fn from(_: ThreadGone) -> Self {
+        BlockValidationError::new(anyhow!("the ledger thread has terminated"))
```
rkuhn: This should be a separate error type so that the caller can properly distinguish it from other cases.

crates/amaru-consensus/src/block_validator.rs
```diff
@@ -26,55 +27,172 @@ use amaru_ledger::{
...
+    }
+}
+
+impl From<ThreadGone> for TransactionValidationError {
+    fn from(_: ThreadGone) -> Self {
+        TransactionValidationError::from(anyhow!("the ledger thread has terminated"))
```
rkuhn: same as above

crates/amaru-consensus/src/block_validator.rs
```diff
@@ -26,55 +27,172 @@ use amaru_ledger::{
...
+    }
+}
+
+impl CanValidateTxs for BlockValidator {
+    fn validate_tx(&self, tx: &Transaction) -> Result<(), TransactionValidationError> {
+        self.request(|reply| LedgerRequest::ValidateTx(Box::new(tx.clone()), reply))?
```
rkuhn: This interface will also need to become async, same reason as above.

crates/amaru-consensus/src/block_validator.rs
```diff
@@ -26,55 +27,172 @@ use amaru_ledger::{
...
+impl HasStakePools for BlockValidator {
+    fn registered_relay_socket_addrs(&self) -> Result<BTreeSet<SocketAddr>, BlockValidationError> {
+        self.request(LedgerRequest::RegisteredRelaySocketAddrs)?
+    }
+}
+
+struct LedgerThread<S: Store, HS: HistoricalStores> {
+    state: State<S, HS>,
+    vm_eval_pool: ArenaPool,
+    chain_store: Arc<dyn ChainStore>,
```
rkuhn: The chain store should be left out of the ledger thread: lookups should be performed before sending the request to the ledger thread.

rkuhn (commented):
Thanks for starting this, @bomanaps ! The part that is still missing is removing all concurrency infrastructure (Arc, Mutex, etc.) from the ledger code — this was the main driver behind the change, so that data structures can be optimised because concurrency no longer is a concern. We don’t have to do all of it in one PR, but it would be good to check whether this change actually allows us to perform such work.

bomanaps:
> Thanks for starting this, @bomanaps ! The part that is still missing is removing all concurrency infrastructure (Arc, Mutex, etc.) from the ledger code — this was the main driver behind the change, so that data structures can be optimised because concurrency no longer is a concern. We don’t have to do all of it in one PR, but it would be good to check whether this change actually allows us to perform such work.

Good point, I checked whether the change actually opens that door. It does `stable: Arc<Mutex<S>>` no longer has any concurrent user since every access now happens on the ledger thread, so we can turn it into a plain `S` right away, the remaining ones, `stake_distributions` and `snapshots`, are still shared with the rewards background thread, so they can only go once that handoff becomes message passing where the ledger thread sends the snapshot inputs over and receives the computed distribution back. I would be happy to take both of those in a follow-up PR.

bomanaps:
Addressed the remaining review comments in https://github.com/pragma-org/amaru/pull/1290/commits/dadf3c465620c86e8aaa303113a348b174c4db4e

rkuhn:
Yes, that sounds like a good plan. I have meanwhile merged a larger change splitting PeerCandidate out of Peer (which is now only a SocketAddr), which causes some conflicts. We currently can’t run full CI for external PRs anyway, so how about I resolve the conflicts and resubmit your branch as a new PR? (we’ll need to figure out this CI limitation soon … )

bomanaps:
> Yes, that sounds like a good plan. I have meanwhile merged a larger change splitting PeerCandidate out of Peer (which is now only a SocketAddr), which causes some conflicts. We currently can’t run full CI for external PRs anyway, so how about I resolve the conflicts and resubmit your branch as a new PR? (we’ll need to figure out this CI limitation soon … )

Funny timing, my push resolving exactly those conflicts landed about a minute after your comment, the branch now includes the PeerCandidate split and is conflict free, with the async interface adapted to registered_relay_candidates, and I reran the full workspace suite plus a short preprod sync through the merged code, all green. If you would still rather resubmit it org side so full CI can run, that works for me too, the commits are all signed off. Otherwise it's ready for review as is, and I will pick up the follow-up we discussed once it lands.

rkuhn:
Thanks, merged!

PR #1291, opened 2026-08-27
Increase Ledger Coverage

### LINKED ISSUE #885: Identify and close existing ledger coverage gaps
### Abstract

Close existing coverage gaps that exist between the rust-specific unit tests and the new fixtures.

### Why?

There are likely spots that are not covered by either the new fixtures nor the old rust-specific unit tests. We should resolve those by adding tests, in the form of the new fixtures.

### How?

1) Identify coverage gaps in the ledger using our codecov tooling
2) Create new fixture cases that execute uncovered logic

We have codecov tooling that reports code coverage and, importantly, percentage diffs in PRs. Coverage should be increasing, and we can verify specific coverage by looking at the reports.

### Discussion points

_No response_

This is a specific task spun out from the larger #797 issue

### Checklist

yHSJ:
Closes #885

After introducing the new JSON fixtures, there were some parts of the ledger left without coverage. This PR closes that gap, introducing fixtures to cover untested cases.

This brings the coverage percentage of `../rules/transaction` to 98.2%. Remaining uncovered lines in the `amaru-ledger` crate are mostly state/setup logic these fixtures cannot cover.

It's important to note that the JSON fixtures cannot test every aspect of the code. They only compare error state (pass or fail with specific case), they do *not* compare a final state to an initial state. So coverage here does not necessarily suggest complete and thorough testing.

I also noticed a panic when `is_valid=false` and a tx is negatively balanced, which has been fixed as well.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected negative certificate balance accounting and prevented panics when processing invalid transactions.
  - Improved validation for script witnesses, collateral inputs, reward-account networks, proposal return-account networks, and oversized Byron outputs.
  - Fixed handling of invalid transactions containing certificates, deposits, and stake-pool updates.

- **Tests**
  - Expanded coverage for governance, delegation, stake, pool, collateral, reference-input, and invalid-transaction scenarios.
  - Added validation cases for missing witnesses, invalid deposits, duplicate registrations, network mismatches, and oversized outputs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,10 +62,10 @@ Other guiding principles:
 - **amaru-ledger**: do not re-encode locally submitted transactions to obtain their size.
 - **amaru-protocols**: preserve bytes of transactions flowing through the mempool.
 - **amaru-tui**: tweak block dissemination metrics headers (fetch → select, sync → fetch)
-- **amaru-ledger**: Do not validate disjoint input sets in protocol version 11+. Similarly, do not allow non-disjoin input sets in PV3, regardless of protocol version.
 - **amaru-kernel**: decode reward accounts as a network tag plus stake credential, rejecting malformed reward accounts at deserialization.
 - **amaru-ledger**: Do not validate disjoint input sets in protocol version 11+. Similarly, do not allow non-disjoint input sets in PV3, regardless of protocol version.
 - **amaru-ledger**: do not allow the same script to exist in both the witness set and in a reference input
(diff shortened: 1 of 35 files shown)


rkuhn (commented):
looks clean, but I cannot judge the bulk of the change (i.e. the fixtures)

rkuhn:
(I did add a commit with tiny improvements, let me know if it is correct and desirable)

yHSJ:
Yes, thanks for the changes, they all look very reasonable @rkuhn.

PR #1293, opened 2026-08-28
chore: improve otel logs

jeluard:
Improve OpenTelemetry diagnostics when the configured collector does not receive every enabled signal.
Instead of repeatedly printing low-level SDK export errors, Amaru now emits batched state-transition events that identify the affected signals: metrics, traces, and/or logs.

### Behavior
- Repeated OpenTelemetry SDK export errors are suppressed.
- Failures observed during a five-second window are batched into one warning.
- Exporters remain active and continue retrying.
- A recovery info event is emitted when previously unavailable signals become available.
- If a recovered signal fails again, a new warning is emitted.

### Example logs

Collector does not receive logs and traces:

```rust
WARN observability.export_failed unavailable_signals=traces,logs
```

Those signals later recover:

```rust
INFO observability.export_recovered recovered_signals=traces,logs
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced repetitive OpenTelemetry export errors by grouping unavailable signals into a single warning.
  * Added batched recovery messages when previously unavailable signals become available.
  * Reports a fresh warning when an export failure occurs again after recovery.

* **Documentation**
  * Updated CLI help and observability trace documentation to cover metrics, traces, logs, export failures, and recoveries.
  * Added documented observability events for export failure and recovery details.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,10 @@ Other guiding principles:
 
 ## v10.11.20260903 _[unreleased; planned for 2026-09-03]_
 
+### Fixed
+
+- **amaru**: replace repeated low-level OpenTelemetry export errors with batched state-transition messages: one warning listing unavailable signals, one info listing recovered signals, and a fresh warning after a later failure.
+
 ## [v10.11.20260827](https://github.com/pragma-org/amaru/releases/tag/v10.11.20260827)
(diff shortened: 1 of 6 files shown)


jeluard:
@arwlf wdyt?

PR #1298, opened 2026-08-31
fix: incorrect cleanup procedure

jeluard:
Fixes sporadic bug https://github.com/pragma-org/amaru/actions/runs/33373621169/job/99430035168?pr=1287

Skip-Changelog

## Bug Fixes

* Ensured background rewards and stake-distribution processing completes cleanly before ledger state shuts down.
* Improved shutdown reliability by preventing incomplete background work from affecting ledger state cleanup.
* Avoided shutdown errors in scenarios where background processing finishes on the same execution thread.
* Reduced the risk of interrupted or inconsistent ledger state during application shutdown.

### DIFF
--- a/crates/amaru-ledger/src/state.rs
+++ b/crates/amaru-ledger/src/state.rs
@@ -128,6 +128,24 @@ where
     tip_update_emitter: TipUpdateEmitter,
 }
 
+impl<S: Store, HS: HistoricalStores> Drop for State<S, HS> {
+    fn drop(&mut self) {
+        // Dropping a JoinHandle detaches its worker, which may otherwise keep using snapshot stores
+        // after the ledger and even the process-wide RocksDB environment start shutting down.
+        if let Some(handle) = self.rewards_join_handle.take() {
+            join_background_thread(handle);
+        }
+    }
+}
+
+fn join_background_thread<T>(handle: JoinHandle<T>) {
+    // A synchronous callback can release the last State owner on this worker. The thread is
+    // already winding down in that case, and attempting to join it from itself would panic.
+    if handle.thread().id() != std::thread::current().id() {
+        let _ = handle.join();
+    }
+}
+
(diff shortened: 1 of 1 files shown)


jeluard:
@coderabbitai review

crates/amaru-ledger/src/state.rs
```diff
@@ -128,6 +128,24 @@ where
...
+        // after the ledger and even the process-wide RocksDB environment start shutting down.
+        if let Some(handle) = self.rewards_join_handle.take() {
+            join_background_thread(handle);
+        }
+    }
+}
+
+fn join_background_thread<T>(handle: JoinHandle<T>) {
+    // A synchronous callback can release the last State owner on this worker. The thread is
+    // already winding down in that case, and attempting to join it from itself would panic.
```
rkuhn, line 143: This comment is unclear: since the rewards calculation is spawned in a separate thread, how would a callback across threads trigger the drop? Or could it be the case that the rewards calculation is the last user and then drops the state itself? Would it not be better to use a weak reference in that case and control the lifecycle from the outside?

jeluard:
@coderabbitai full review

rkuhn (commented):
I think the fix is correct, but the comment may not be. And depending on what the correct comment will turn out to be, the fix may also be incomplete.

crates/amaru-ledger/src/state.rs
jeluard: So I did more digging with codex and indeed there might be a more holistic way to fix this. Would you want to take a look if you have some time? It's definitely more your alley :)
Although it's pretty low priority I would say.

rkuhn: Any hint at what that holistic way would be? I’m definitely interested in more background information.

jeluard:
Will reopen after deeper assessment.

PR #1299, opened 2026-08-31
feat: allow to selectively export OTLP signals

jeluard:
By default all OLTP signals are exported (metrics, logs and spans) while end users might be interested in only a subset of those. If some collector path is missing then errors will be printed amaru side (client here in this scenario).

Now `AMARU_OPEN_TELEMETRY_SIGNALS` can be defined to only push a subset of signals (e.g. `AMARU_OPEN_TELEMETRY_SIGNALS=metrics,traces`).

## Summary by CodeRabbit

* **New Features**
  * Added configurable OpenTelemetry export for metrics, traces, and logs.
  * Added `--open-telemetry-signals` and environment-variable support for selecting signals.
  * Signal selection accepts comma-separated values and defaults to all supported signals.

* **Bug Fixes**
  * Improved startup error reporting for invalid telemetry configuration.
  * Ensured all configured telemetry providers shut down reliably, even when one reports an error.

* **Documentation**
  * Updated monitoring guidance for telemetry configuration, signal selection, and validation.


crates/amaru-node/src/telemetry.rs
```diff
@@ -140,7 +143,11 @@ impl Telemetry {
         // also require a current runtime handle when constructed below.
         tokio::task::yield_now().await;
 
-        if with_otlp { Self::install_otlp(format, options) } else { Self::install_local(format) }
+        if with_otlp {
+            Self::install_otlp(format, options, OpenTelemetrySignals::from_env()?)
```
rkuhn: shouldn’t this also be available via cmdline option?

crates/amaru-node/src/telemetry.rs
```diff
@@ -152,47 +159,84 @@ impl Telemetry {
...
         let service_name = resource
             .get(&opentelemetry::Key::from_static_str(SERVICE_NAME))
             .map(|v| v.as_str().to_string())
             .unwrap_or_else(|| DEFAULT_SERVICE_NAME.to_string());
 
-        let traces_provider = SdkTracerProvider::builder()
-            .with_resource(resource.clone())
-            .with_batch_exporter(
-                opentelemetry_otlp::SpanExporter::builder().with_tonic().build().context("build OTLP span exporter")?,
+        let traces_provider = if signals.traces() {
```
rkuhn: since the then-branch only has `None` I’d find it easier to read as `let ... = signals.trace().then(|| { ... });`

crates/amaru-observability/src/open_telemetry.rs
```diff
@@ -0,0 +1,100 @@
...
+    const ALL: [Self; 3] = [Self::Metrics, Self::Traces, Self::Logs];
+}
+
+/// The OpenTelemetry signals enabled for OTLP export.
+///
+/// The environment representation is a comma-separated list containing one or
+/// more of `metrics`, `traces`, and `logs`. When
+/// [`AMARU_OPEN_TELEMETRY_SIGNALS`] is unset, all signals are enabled.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct OpenTelemetrySignals(BTreeSet<OpenTelemetrySignal>);
```
rkuhn: This isn’t used in any hot paths and not often, but it still hurts to see three bits of information represented by this immensely powerful data structure.

How about the following instead (as cmdline args with env fallback):

```rust
use clap::{Parser, ValueEnum};

#[derive(Clone, Copy, Debug, PartialEq, Eq, ValueEnum)]
enum OtelSignal {
    Traces,
    Metrics,
    Logs,
}

#[derive(Parser, Debug)]
struct Args {
    /// Enabled signals
    #[arg(long, value_enum, value_delimiter = ',')]
    otel_signals: Vec<OtelSignal>,
}
```

Then wrap in `struct OtelSignals(Vec<OtelSignal>)` to get nice names for `signals.0.contains(OtelSignal::Traces)`. If clap supported multi-arg parsing into any other type than `Vec` then there would be more compact solutions, but here I’d vote for minimal mental load.

crates/amaru/src/bin/amaru/main.rs
```diff
@@ -91,13 +91,14 @@ fn try_main() -> anyhow::Result<()> {
...
-        let handle = setup_observability(
+        let handle = try_setup_observability(
             with_open_telemetry,
             with_json_traces,
             local,
             color_enabled,
             &ListenAddressHint(listen_address.as_deref()),
-        );
+        )
+        .inspect_err(|error| eprintln!("amaru: failed to configure observability: {error}"))?;
```
rkuhn: This feels wrong: if the error is passed upwards, then it should be logged there as well, so why log it here? If the point is just to add context, then use `anyhow::Context` to create a cause chain.

crates/amaru-node/src/telemetry.rs
rkuhn: I see that the `observability` module does this already. Hmm, I must have missed this duplication while doing the amaru-node library work — this should probably be unified.

rkuhn (commented):
good point

crates/amaru-node/src/telemetry.rs
jeluard: oupsy

crates/amaru/src/bin/amaru/cli.rs
```diff
@@ -183,6 +184,17 @@ pub(crate) struct Cli {
     /// Export metrics, traces, and logs via OpenTelemetry (OTLP).
     #[clap(long, action, env = "AMARU_WITH_OPEN_TELEMETRY")]
     pub(crate) with_open_telemetry: bool,
+
+    /// Select which OpenTelemetry signals to export.
+    #[clap(
+        long,
+        global = true,
```
rkuhn, line 191: If this makes sense here, does it also make sense on the options above?

rkuhn (approved):
Nice! Just one nitpick remaining.

PR #1304, opened 2026-09-02
handshake: combine version data per the node-to-node spec

rkuhn:
Stacked on #1303.

Handshake now picks the greatest common version and combines that version's records as specified: network magics must match, `initiatorOnlyDiffusionMode` and `query` are OR, `peerSharing` is AND. A true `query` flag is not a session (`MsgQueryReply` with our table). The initiator checks that `MsgAcceptVersion` carries the agreed record.

Outbound still offers initiator-only diffusion. Duplex is not advertised until a later PR actually runs both halves.

## Test plan
- `cargo test -p amaru-protocols --lib handshake`
- Negotiation unit tests for magic, OR/AND, query, accept verification
- Blueprint handshake test-data vectors decode

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,7 @@ Other guiding principles:
 
 ### Changed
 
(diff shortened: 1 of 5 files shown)


rkuhn:
opened prematurely

PR #1306, opened 2026-09-02
mux: SDU assembly and send timers

rkuhn:
Stacked on #1304.

The mux now implements the blueprint SDU timer: unbounded wait for the leading edge of a header, then 10s (first Handshake) / 30s (after `SetSduTimeout`) for the rest of that SDU on receive and for writing the matching segment. Unknown protocol IDs still terminate the bearer.

Simulation cannot yet race an `UntilResolved` recv against the stage clock (that needs the later timeout-verification work). The stall test completes the post-leading-edge recv with `ReceiveError::sdu_timeout`.

## Test plan
- `cargo test -p amaru-protocols --lib mux`
- `cargo test -p amaru-protocols --lib handshake`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,7 @@ Other guiding principles:
 
 ### Changed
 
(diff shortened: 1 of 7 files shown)


rkuhn:
opened prematurely

PR #1307, opened 2026-09-02
connection: Established session and no manager redial

rkuhn:
Stacked on #1306.

Connection is one `Established` session (not Initiator XOR Responder) with `LocalUse::{None, Maintenance, Diffusion}`. Defaults remain outbound Diffusion / inbound None and still start only the TCP-direction half.

The manager reports `Disconnected { will_retry: false }` (or `ConnectFailed` before handshake) and does not reconnect. Peer selection already refills on that path. `SetLocalUse` is recorded; MsgDone group stop is the next PR.

## Test plan
- `cargo test -p amaru-protocols --lib`
- Integration reconnect-after-responder-restart now re-issues `AddPeer`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,7 @@ Other guiding principles:
 
 ### Changed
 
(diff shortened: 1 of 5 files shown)


rkuhn:
opened prematurely

PR #1308, opened 2026-09-02
feat: allow to cancel bootstrap process

jeluard:
Allow the bootstrap process to be cancelled using a `tokio_util::sync::CancellationToken`.
Especially useful for 3rd party apps doing a bootstrap, as bootstrap process can take some significant time.

## Summary by CodeRabbit

- **New Features**
  - Added the ability to cancel an in-progress node bootstrap operation.
  - Bootstrap now checks for cancellation at safe processing points and reports when cancellation occurs.
  - Cancellation stops lengthy state, UTxO, snapshot, and packaged-block imports more responsively.
  - Import progress is checkpointed incrementally, improving resumability and resource cleanup.

- **Documentation**
  - Added guidance for creating and passing a bootstrap cancellation handle.
  - Updated OpenTelemetry configuration documentation, including optional signal selection and enabling all signals by default.

### DIFF
--- a/crates/amaru-bootstrap/src/bootstrap/mod.rs
+++ b/crates/amaru-bootstrap/src/bootstrap/mod.rs
@@ -66,9 +66,6 @@ pub enum BootstrapError {
     #[error("bootstrap cancelled")]
     Cancelled,
 
-    #[error(transparent)]
-    Other(#[from] anyhow::Error),
-
     #[error("Can not create snapshots directory {0}: {1}")]
     CreateSnapshotsDir(PathBuf, io::Error),
 
@@ -103,10 +100,6 @@ pub enum BootstrapError {
 
 pub(crate) fn checkpoint(cancellation: &BootstrapCancellation) -> Result<(), BootstrapError> {
     if cancellation.is_cancelled() { Err(BootstrapError::Cancelled) } else { Ok(()) }
-}
-
-fn operation_error(cancellation: &BootstrapCancellation, error: impl FnOnce() -> BootstrapError) -> BootstrapError {
-    if cancellation.is_cancelled() { BootstrapError::Cancelled } else { error() }
 }
 
 pub const BOOTSTRAP_HEADERS_PER_POINT: usize = 2;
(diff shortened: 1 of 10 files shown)


crates/amaru-bootstrap/src/bootstrap/mod.rs
```diff
@@ -491,15 +563,20 @@ pub async fn bootstrap(
...
     // Nonces for both packaged tips are stored before the headers so
     // `import_packaged_blocks` can attach each header via `store_validated_header`.
-    store_chain_state(imported_second_snapshot.epoch, &chain_db, second_chain_state)?;
-    store_chain_state(imported_third_snapshot.epoch, &chain_db, third_chain_state)?;
+    checkpoint(&cancellation)?;
+    store_chain_state(imported_second_snapshot.epoch, &chain_db, second_chain_state)
+        .map_err(|error| operation_error(&cancellation, || error.into()))?;
+    checkpoint(&cancellation)?;
+    store_chain_state(imported_third_snapshot.epoch, &chain_db, third_chain_state)
+        .map_err(|error| operation_error(&cancellation, || error.into()))?;
```
rkuhn: This one I don’t understand: why swallow the error if cancellation occurred? It seems that cancellation is not passed into the `store_chain_state` logic, so I’d surface the error as is and leave cancellation to the `checkpoint` on the line below.

(applies to all `operation_error` use sites)

rkuhn (commented):
General question: since the DB will be incomplete after a cancellation anyway, why not just `exit(1)`? That would be one tiny change to the `amaru` binary.

jeluard:
@rkuhn that's intended to be used embedded in others apps.

crates/amaru-bootstrap/src/bootstrap/mod.rs
jeluard: Makes sense. Made some relevant changes and overall cleanup.

rkuhn:
> @rkuhn that's intended to be used embedded in others apps.

Then the issue remains that ctrl-c will lead to an unusable DB — it would arguably be better to remove the directory in that case because we don’t have a way of recovering a partially fill DB, and I don’t know whether the effort is worth it.

jeluard:
@rkuhn I think our strategy is to detect corrupted db at startup then detail the different options user can do to move forward (including dropping the db). So that would be ok.

rkuhn:
The “user” in this case isn’t a human reading logs, it is another code module calling amaru-node as a library. I’m not sure that the error we return from build_node is already good enough for what you say, and I also don’t think it is good customer service to leave a known-bad DB on the filesystem — it might not even be malformed, so startup would “work” but the node wouldn’t actually work.

jeluard:
Bad db can already happen for a number of reason during bootstrap (not atomic) or failure during node running. So it's a situation we already have to deal with and how we decided to deal with it is identifying incorrect db at startup (maybe not sufficient yet).

PR #1309, opened 2026-09-02
remove concurrency infrastructure from ledger state

bomanaps:
Closes the follow-up agreed in #1290, ref #1094 removes all remaining concurrency infrastructure from the ledger state, so stable and stake_distributions lose their Arc<Mutex> and the rewards background task now receives snapshot handles by value and returns the rotated stake distribution through its join handle instead of writing into shared state, verified with the full workspace suite plus a live preprod sync to tip of about 9,300 blocks including an epoch transition and rewards computation with no errors the read only era_history and global_parameters Arcs are left in place and I am happy to remove those too if desired.

## Summary by CodeRabbit

* **Changed**
  * Improved ledger state management for snapshots and stake distributions.
  * Updated rewards processing to support background computation and in-memory stake distribution updates.
  * Simplified ratification processing by using stake distribution data directly.
  * Enabled historical snapshots to be transferred safely for background processing.
  * Reduced synchronization requirements for ledger and consensus processing.
* **Documentation**
  * Updated changelog references without changing documented content.


rkuhn (commented):
Looks reasonable to me, but should be cross-checked by @yHSJ

CHANGELOG.md
```diff
@@ -51,6 +51,7 @@ Other guiding principles:
 ### Changed
 
 - **amaru-consensus**: `BlockValidator` now runs the ledger on a dedicated thread that owns the ledger state exclusively; operations are requested through a bounded channel and answered via per-request reply channels, replacing the shared lock around the state. `BlockValidator` is no longer generic over the store types. ([#1094][])
+- **amaru-ledger**: the ledger state no longer uses locks: the stable store and the in-memory stake distributions are owned directly by `State`. The rewards background task receives its snapshot handles by value and returns the rotated stake distribution through its join handle instead of writing into shared state; `HistoricalStores::for_epoch` snapshots are now `'static`, `RatificationContext` takes `&StakeDistribution` in place of the removed `StakeDistributionView`, and `StateError::FailedToAcquireStakeDistrLock` is gone. ([#1094][], [#1309][])
```
yHSJ: A general rule of thumb for the changelog: If this doesn't change anything that a *user* would see, it doesn't need to go in the changelog. That would be things like ledger rule changes, API changes, etc.

If a user wants to know more about changes, they can look at the commit history, which will clearly show we're modifying the internals (like this PR).

yHSJ (approved):
Yep, this looks good to me. I left one comment about the changelog, which we should cleanup.

bomanaps:
> Yep, this looks good to me. I left one comment about the changelog, which we should cleanup.

Addressed here https://github.com/pragma-org/amaru/pull/1309/commits/a677961588d95a483ac7a6e3cb1f0b2aaa35cbe5

rkuhn:
It doesn’t seem like a review is missing, just a merge.

crates/amaru-ledger/src/state.rs
```diff
@@ -99,12 +98,15 @@ where
     /// be updated but grouped here to avoid dealing with magic values everywhere.
     global_parameters: Arc<GlobalParameters>,
 
-    /// A shared collection of the latest slim stake distributions.
+    /// The latest slim stake distributions.
     ///
     /// These are used by the runtime for leader schedule verification and governance ratification.
     /// Full stake distributions remain reconstructible from on-disk snapshots when rewards need
     /// them, which avoids retaining large account maps in steady-state memory.
-    stake_distributions: Arc<Mutex<VecDeque<StakeDistribution>>>,
```
KtorZ, line 107: This one is likely _required_; the stake distributions are computed asynchronously and accessed concurrently within the ledger itself. 

It could be however weakened to a RWLock; since we know there's only one writer.

crates/amaru-ledger/src/state.rs
```diff
@@ -530,26 +528,44 @@ impl<S: Store, HS: HistoricalStores + Send + Sync + 'static> State<S, HS> {
...
         Ok(())
     }
 
+    /// Record a freshly rotated stake distribution, keeping only the two most recent ones.
+    /// Skipped when the distribution is already held, which can happen after a fork switch
+    /// re-crosses the epoch boundary.
+    fn install_stake_distribution(&mut self, distribution: StakeDistribution) {
+        if self.stake_distributions.front().is_none_or(|front| front.epoch < distribution.epoch) {
+            self.stake_distributions.push_front(distribution);
+            self.stake_distributions.truncate(2);
```
KtorZ, line 565: This is bad because it pushes a large object into memory only to truncate it later.

crates/amaru-ledger/src/state.rs
```diff
@@ -1179,128 +1188,102 @@ fn pool_summaries_for<'iter>(stake_distributions: impl Iterator<Item = &'iter St
...
+/// progresses. The thread receives everything it needs by value; results travel back through
+/// the [`JoinHandle`] consumed at the epoch transition.
+struct BackgroundTasks<Snap> {
+    /// Snapshot of the previous epoch; source of the rotated stake distribution as well as the
+    /// block issuers and pots for the rewards calculation.
+    previous_epoch: Snap,
+    /// Snapshot the rewards stake distribution is computed from, three epochs back.
+    rewards_snapshot: Snap,
+    /// Set when the in-memory stake distributions do not include the previous epoch yet.
+    rotation: Option<StakeDistributionRotation>,
```
KtorZ, line 1211: Not sur what's going on here 🤔? This is needlessly complicated and introduces room for mistakes (the caller is now responsible for providing the right snapshots). Please revert to use the historical store as it were.

crates/amaru-ledger/src/state.rs
KtorZ: On second thoughts, it _might be fine_ given that we have now introduced the `PoolSummaries`. So the only other consumer of these stake distribution is likely just the epoch boundary for the governance ratification. 

- This step is synchronous and only happens after the calculation is done. 
- We only starts a new computation after `k` blocks (or 3*k/f slots) in the epoch, well after the ratification is over and _stable_. 

That means we technically only need the latest stake distribution, which can be held by the `rewards_join_handle` altogether. 

Point being, with all the recent changes we did on that front, I don't think we need the `stake_distributions` field _at all_. We can simply augment (and rename) the `rewards_join_handle` to also hold the ongoing stake distribution. 

This is not trivial though because we must handle the restart and epoch boundaries correctly.

bomanaps: The lock isn't needed in the new design, `BackgroundTasks` takes everything it needs by value and returns the new distribution through the JoinHandle  it never touches `stake_distributions` directly and also the install happens synchronously in `epoch_transition` via &mut self after joining, and since State is exclusively owned by the ledger thread (from #1290), there's no concurrent access to this field at all.
The idea of folding stake_distributions into the join handle is interesting though I think it worth exploring as a follow-up, but pool_summaries() and the ratification lookup (next_epoch - 2) need it available outside the epoch transition, so it's not a straight swap.

crates/amaru-ledger/src/state.rs
bomanaps: Fair point on this I can replace with pop_back if len == 2 before push_front, happy to fix in a follow-up.

crates/amaru-ledger/src/state.rs
bomanaps: The trade-off here was removing the Sync bound on `HistoricalStores`, and passing `Arc<HS>` into the background task requires `HS: Sync`, which we explicitly removed from BlockValidator in this PR the current approach calls `for_epoch` on the ledger thread before spawning and passes the snapshots by value, which is exactly what `for_epoch` returning impl Snapshot + Send + 'static was designed for, am not sure we have any  clean alternative that avoids Sync without this pattern, if you have a better way on this am happy to follow

KtorZ: But, what's the problem with having a `Sync` bound on HistoricalStores? This is fine 😅... The goal isn't to erase all possible traces of concurrency, but remove the unnecessary ones.

crates/amaru-ledger/src/state.rs
KtorZ: > it never touches stake_distributions directly 

Point being, I believe it does not even need to touch it _at all_. The `JoinHandle` now indeed already returns the required stake distribution. So it's already half-way through what I am suggesting. 

> but pool_summaries()

This one is a bit out of the picture now. We used to have the ledger be the owner of the pool stake distribution data required by the consensus, with a shared mutex between both (the ledger being the sole writer and consensus the sole reader). Since a couple of weeks now, this has changed into a notification pattern where the consensus is made aware of a new stake distribution / pool summary when they're available. It then takes care of their lifecycle and their pruning. 

So when it comes to pool summaries, the ledger no longer needs to keep historical summaries around to cope with dynamic requests. It only needs to notify the consensus when new data is available. 

 > ratification lookup (next_epoch - 2)

This is a bit confusing but here, `next_epoch - 2` is simply `current_epoch - 1`; which is exactly the stake distribution being computed by the background task. Currently, this is being pushed into the stake distribution array, only to get immediately pulled out of it for ratification. 

---

This isn't a major issue w.r.t memory because the stake distribution is already slim and only a few kilobytes. But the design has become confusing and needlessly complex due to many of our recent changes.

crates/amaru-ledger/src/state.rs
rkuhn: StakeDistribution is 96 bytes large, so I don’t think it matters much.

PR #1310, opened 2026-09-02
Block Forging EDR

yHSJ:
This PR introduces the first draft of our block forging EDR. Refining this document will help is iron out the details of the design of block forging.

Once this is finalized (finalized doesn't mean it will never change, but rather that it is ready for implementation), we will create issues for the implementation work.

## Summary by CodeRabbit

* **Documentation**
  * Expanded the architecture decision record for block production with Ouroboros Praos and Leios context.
  * Clarified slot leadership, epoch scheduling, nonce handling, rollback recovery, and asynchronous leader-schedule computation.
  * Documented block construction and validation flow, including mempool-provided block bodies and later normal validation.
  * Added guidance for schedule telemetry, header build identification, NTP configuration, and TUI observability.
  * Clarified forging credentials and operational handling, including OCERT counter management.


engineering-decision-records/035-block-forging.md
```diff
@@ -0,0 +1,127 @@
+---
+type: architecture
+status: proposed
+---
+
+# Block Forging
+
+Currently, Amaru follows the chain but does not extend it. This document describes how a stake pool running Amaru would produce blocks.
+The goal is a valid block with limited wasted work. The Haskell node is our reference for what "valid" means, and this document links to the relevant code, but the shape of our implementation is our own.
```
jeluard, line 9: We should consider leios here too as it might have architectural impacts.

engineering-decision-records/035-block-forging.md
```diff
@@ -0,0 +1,127 @@
...
+
+### Keys
+
+There are four keys used in the forging process. The Haskell node reads them in [`readLeaderCredentials`](https://github.com/IntersectMBO/cardano-node/blob/a38eac60bceb1a64a4ffa29e2d49d802787ce171/cardano-node/src/Cardano/Node/Protocol/Shelley.hs#L161-L211) and bundles them as [`PraosCanBeLeader`](https://github.com/IntersectMBO/ouroboros-consensus/blob/0ce94ef215312633d44d049ea1a7e30eb4b3deb6/ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos/Common.hs#L272-L297).
+
+- The **cold key** is an Ed25519 key kept offline. Its hash is the pool id. It signs the pool registration certificate and the operational certificate. The node only needs the verification key.
```
rkuhn: “verification key” has not been introduced here, please give the proper definition

engineering-decision-records/035-block-forging.md
```diff
@@ -0,0 +1,127 @@
...
+- The **operational certificate** binds a KES verification key to the pool. It carries the KES verification key, a counter, a start period, and a cold-key signature over the three ([`OCert`](https://github.com/IntersectMBO/cardano-ledger/blob/bc7df956110912d3b1e3501dbb66e466f7220398/libs/cardano-protocol/src/Cardano/Protocol/TPraos/OCert.hs#L84-L91)). Every header carries it. The [OCERT rule](https://github.com/IntersectMBO/cardano-ledger/blob/bc7df956110912d3b1e3501dbb66e466f7220398/libs/cardano-protocol-tpraos/src/Cardano/Protocol/TPraos/Rules/OCert.hs#L71-L106) rejects a header whose slot is before the start period or more than 62 periods after it, and a counter lower than the last one seen for that pool.
```
rkuhn: “… and a counter lower than the last one seen for that pool _on the current chain_.” This does make a difference, Eric and I discussed this in depth when he implemented opcert tracking.

engineering-decision-records/035-block-forging.md
```diff
@@ -0,0 +1,127 @@
...
+
+1. Pick the parent: the tip, or the tip's parent if the tip already sits in this slot.
+2. Tick the protocol state to the slot and evolve the KES key if the period changed.
+3. Run the leader check, then the "can forge" check.
+4. Take the longest prefix of the mempool that fits the block ([`getSnapshotFor`](https://github.com/IntersectMBO/ouroboros-consensus/blob/0ce94ef215312633d44d049ea1a7e30eb4b3deb6/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/API.hs#L184-L200), [`snapshotTake`](https://github.com/IntersectMBO/ouroboros-consensus/blob/0ce94ef215312633d44d049ea1a7e30eb4b3deb6/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Mempool/API.hs#L432-L440)). The mempool keeps its transactions validated against the tip, so the body is not validated again here.
+5. Build and sign the header ([`forgeShelleyBlock`](https://github.com/IntersectMBO/ouroboros-consensus/blob/0ce94ef215312633d44d049ea1a7e30eb4b3deb6/ouroboros-consensus-cardano/src/shelley/Ouroboros/Consensus/Shelley/Ledger/Forge.hs#L42-L100)).
+6. Hand the block to the chain database and wait for it to be adopted or rejected.
```
rkuhn, line 52: The “… and wait for it to be adopted or rejected” needs a bit more detail: I assume the decision is made by the network, correct? And I further assume that there is no active waiting involved, the procedure has already finished and doesn’t care one way or the other, right?

engineering-decision-records/035-block-forging.md
```diff
@@ -0,0 +1,127 @@
...
+## Decision
+
+We add one stage, `forge_block`, wired into the graph only when forging credentials are configured. It receives the adopted tip from `adopt_chain`, like the mempool does, and it sends exactly one message: a new tip to `select_chain`. From there our block is treated like any block a peer sent us.
+
+```mermaid
+flowchart LR
+    subgraph existing
+        track_peers --> select_chain --> fetch_blocks --> validate_block --> adopt_chain
+        adopt_chain -->|NewTip| mempool
+        adopt_chain -->|NewTip| manager
```
rkuhn, line 71: `validate_block` also sends the validation result back to `select_chain`

engineering-decision-records/035-block-forging.md
```diff
@@ -0,0 +1,127 @@
...
+
+Steps 2, 3 and 5 are callbacks in the [`BlockForging`](https://github.com/IntersectMBO/ouroboros-consensus/blob/0ce94ef215312633d44d049ea1a7e30eb4b3deb6/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/Forging.hs#L82-L156) record, not in the loop.
+
+### Amaru today
+
+Consensus is a graph of pure stages ([EDR 011](./011-deterministic-simulation-testing.md)). Headers arrive in `track_peers`, `select_chain` ranks tips, `fetch_blocks` downloads bodies, `validate_block` asks the ledger to apply them, and `adopt_chain` updates the best chain and tells the mempool and the network manager about the new tip. Every effect with the outside world, including the ledger, the stores and the clock, goes through a resource so the graph can run under simulation. Nothing in the graph produces a block.
+
+## Decision
+
+We add one stage, `forge_block`, wired into the graph only when forging credentials are configured. It receives the adopted tip from `adopt_chain`, like the mempool does, and it sends exactly one message: a new tip to `select_chain`. From there our block is treated like any block a peer sent us.
```
rkuhn, line 62: perhaps explain that `block_source` and `peer_selection` are omitted since they don’t contribute to block forging

engineering-decision-records/035-block-forging.md
```diff
@@ -0,0 +1,127 @@
...
+    forge_block -.->|schedule_at| clock[(clock)]
+```
+
+### Once per epoch
+
+On the first `NewTip` whose header is inside the stability window, the stage:
+
+1. Computes the next epoch's nonce with `Nonces::next_active` from the tip's stored nonces.
+2. Asks the credentials resource for the slots we lead in that epoch, given the nonce and our stake share from `PoolSummaries`.
+3. Converts each led slot to a wall-clock instant with `EraHistory::slot_to_posix_time` and schedules a `LeadSlot(slot)` message for it with `Effects::schedule_at`.
```
rkuhn, line 87: logically yes, in practice we should only schedule the next occurrence

engineering-decision-records/035-block-forging.md
```diff
@@ -0,0 +1,127 @@
...
+
+On startup the same steps run for the current epoch from the tip's active nonce, scheduling only the slots still ahead. Nothing runs on the slots we do not lead.
```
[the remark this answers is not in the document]
rkuhn: Yes, right after node startup, compute the schedule for the current epoch as well as for the following epoch if applicable.

engineering-decision-records/035-block-forging.md
```diff
@@ -0,0 +1,127 @@
...
+
+### Once per led slot
+
+When `LeadSlot(slot)` fires:
+
+1. Check that the operational certificate covers the slot's KES period. If not, log a warning and stop.
+2. Pick the parent: the current tip, or the tip's parent if the tip's slot equals ours.
```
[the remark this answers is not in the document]
rkuhn: Yes, we should check this to be on the safe side, but if other parts of Amaru are working as designed, this cannot happen: headers are only validated once their designated slot time has started. No headers are stored before validation.

engineering-decision-records/035-block-forging.md
```diff
@@ -0,0 +1,127 @@
...
+3. Ask the mempool for a sequence of transactions that is valid on the parent's state as of our slot and fits in a block. Ask the ledger to apply it and keep the resulting state fragment, keyed by the body hash.
```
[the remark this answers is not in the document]
rkuhn: @yHSJ This is a good point: if there is already a block at the current slot then our mempool will very likely be empty. In that case we probably shouldn’t forge a block. Putting the transactions back would be rather specific logic that we might want to avoid for the MVP.

engineering-decision-records/035-block-forging.md
```diff
@@ -0,0 +1,127 @@
...
+4. Ask the credentials resource to forge the header: VRF proof for the slot, block body hash, KES signature for the period.
+5. Run the header through the same `validate_header` every peer header passes. This costs one VRF verify and one KES verify, and yields the evolved nonces we must store with the header anyway.
+6. Store the header and the block, and send the new tip to `select_chain`.
```
rkuhn, line 100: We should probably mark the block as valid in the chain store right away, in which case we’ll need to audit whether the select_chain + fetch_blocks + validate_block combo may be confused.

engineering-decision-records/035-block-forging.md
```diff
@@ -0,0 +1,127 @@
...
+
+### Rules
+
+- **Secrets never enter stage state.** Stage state is serialised into the trace buffer on every message. The VRF and KES keys live in a resource and answer two effects, `leader_schedule` and `forge_header`. The stage keeps only public facts: the led slots, the certificate's start period and evolution limit.
+- **Compute the schedule once.** Every input to the leader check is fixed before the epoch starts. There is no per-slot loop.
+- **Enter the pipeline at `select_chain`, not `adopt_chain`.** `adopt_chain` assumes the ledger has applied the block and that `validate_block` and `select_chain` have moved their tip. Skipping them leaves `validate_block` believing the old tip is current, so the next upstream sibling of our block would be applied as an extension and fail. Entering at `select_chain` keeps every stage's bookkeeping right, and the kept fragment removes the double validation that route would otherwise cost.
```
rkuhn, line 108: Well, it’s a trade-off, we could make it work either way: sending to `select_chain` may require some updates so `fetch_blocks` correctly informs `validate_block` even though the block is already stored, and `validate_block` needs some updates to correctly deal with an already valid block.

engineering-decision-records/035-block-forging.md
```diff
@@ -0,0 +1,127 @@
...
+- **Everything is simulatable.** Time comes from `schedule_at`, keys from a resource, the ledger and mempool from resources. A pure-stage test can drive an epoch boundary and a led slot with a mocked credentials resource and assert the exact effect trace.
```
rkuhn, line 109: yes, that was the goal

engineering-decision-records/035-block-forging.md
```diff
@@ -0,0 +1,127 @@
...
+
+### Ledger
+
+The ledger gains two entry points: build a body on a given parent at a given slot, and commit a kept fragment on roll forward. Building on the tip's parent means producing the state at tip minus one. The volatile store holds the last `k` blocks and `switch_to_fork` already rolls back, so the pieces exist, but the builder needs a parent argument rather than assuming the tip. Block limits (max body size, max execution units, max header size) come from the protocol parameters in the parent state after the epoch tick, not from the stage.
+
+### Operations
+
+Forging depends on the wall clock, so the NTP requirement from [EDR 014](./014-time-in-amaru.md) becomes a hard requirement for pools.Operators must also rotate the KES key and certificate before the 62-period limit, as with the Haskell node. A missed slot is logged with the reason: certificate not yet valid, expired, or the tip moved under us.
```
rkuhn, line 123: we should perhaps consider the NTP check at startup that Matthias suggested, like pinging one of the “reliable” time sources on the Internet

rkuhn (commented):
Yes, good first version, only small adjustments.

engineering-decision-records/035-block-forging.md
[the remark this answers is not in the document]
yHSJ: It's worth noting that under heavy usage, such as during the NFT craze, the mempools were more full than blocks would allow. That was in part due to budget limitations, but it may not be the case the the mempool is empty. But let's just assume worst case for now, that it would be.

As a block producer, we always want to produce a block in our slot. It could be our only chance to produce a block, and we want to collect fees (which only happens if we produce a block). However, it would be worse for the chain if they were to accept our empty block instead of one full of transactions (Which is not part of the process of a height battle, but we should still strive to have Amaru leave a positive impact on the network).

So I would say the first iteration of this can be one where we just build a block with no transactions, since that is what is best for the node. Then we could explore keeping "tip's transactions" in the mempool.

engineering-decision-records/035-block-forging.md
yHSJ: I don't know enough about "real leios" (meaning what is actually being implemented versus what was originally described in the paper) to really weigh in here. Might be worth having a conversation with current leios folks here.

engineering-decision-records/035-block-forging.md
yHSJ: Yes. Though, Haskell's block forging loop does monitor the chainDB and, if the block was rejected (i.e. not in the chain), it removes *all* transactions from the invalid block to prevent forging an invalid block again.

engineering-decision-records/035-block-forging.md
yHSJ: You mean even if we have v10 slots which we can produce a block, we should only schedule the first occurrence, and then schedule the *next* occurrence after the first slot is completed?

engineering-decision-records/035-block-forging.md
yHSJ: Yeah, I think that's a good idea, but probably outside the scope of this EDR.

engineering-decision-records/035-block-forging.md
rkuhn: yes, this is how pure-stage wants to work

engineering-decision-records/035-block-forging.md
```diff
@@ -0,0 +1,129 @@
...
+# Block Forging
+
+Currently, Amaru follows the chain but does not extend it. This document describes how a stake pool running Amaru would produce blocks.
+The goal is a valid block with limited wasted work. The Haskell node is our reference for what "valid" means, and this document links to the relevant code, but the shape of our implementation is our own.
+
+## Context
+
+### Slot leadership is decided before the epoch starts
+
+A pool leads a slot when its VRF output for that slot, hashed with the epoch nonce, falls below a threshold that depends only on its stake share and the active slot coefficient `f`. See the Praos paper ([eprint 2017/573](https://eprint.iacr.org/2017/573)) for the protocol, and [`checkLeaderValue`](https://github.com/IntersectMBO/cardano-ledger/blob/bc7df956110912d3b1e3501dbb66e466f7220398/libs/cardano-protocol/src/Cardano/Protocol/TPraos/BlockHeader.hs#L337-L371) for the check as implemented.The VRF input is `hash(slot ‖ epoch nonce)` ([`mkInputVRF`](https://github.com/IntersectMBO/cardano-ledger/blob/bc7df956110912d3b1e3501dbb66e466f7220398/libs/cardano-protocol/src/Cardano/Protocol/Praos/VRF.hs#L63-L78)).
```
KtorZ, line 15: Note that there's also a reference implementation of this in amaru 🤓; since we have to perform that very check for _every slot leader_ already as part of header validations:

https://github.com/pragma-org/amaru/blob/e77e1947c5e12fe9067f0f889be299181085d07e/crates/amaru-ouroboros/src/praos/header.rs#L283-L303

engineering-decision-records/035-block-forging.md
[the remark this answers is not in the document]
KtorZ: Not sure about the meaning of "keep the resulting state fragment"; are you suggesting that the block is applied through a special handler instead of simply coming out of the `validate_block` stage; and this in order to avoid re-validating our own block?

engineering-decision-records/035-block-forging.md
rkuhn: meeting notes:
- mempool will need to become more powerful
- independent processes for votes, announcements, EB handling to be added
- ledger needs to become better at handling tip changes, minimize effort to keep state up-to-date

engineering-decision-records/035-block-forging.md
```diff
@@ -0,0 +1,129 @@
...
+
+### Amaru today
+
+Consensus is a graph of pure stages ([EDR 011](./011-deterministic-simulation-testing.md)). Headers arrive in `track_peers`, `select_chain` ranks tips, `fetch_blocks` downloads bodies, `validate_block` asks the ledger to apply them, and `adopt_chain` updates the best chain and tells the mempool and the network manager about the new tip. Every effect with the outside world, including the ledger, the stores and the clock, goes through a resource so the graph can run under simulation. Nothing in the graph produces a block.
+
+## Decision
+
+We add one stage, `forge_block`, wired into the graph only when forging credentials are configured. It receives the adopted tip from `adopt_chain`, like the mempool does, and it sends exactly one message: a new tip to `select_chain`. From there our block is treated like any block a peer sent us.
+
+Note the graph below only contains the stages that are relevant to block forging.
```
rkuhn, line 64: Also add the relationship between adopt_chain and mempool to highlight that there shall be a delayed purge of transactions from the mempool driven by chain progress.

engineering-decision-records/035-block-forging.md
```diff
@@ -0,0 +1,129 @@
...
+
+### Once per epoch
+
+On the first `NewTip` whose header is inside the stability window, the stage:
+
+1. Computes the next epoch's nonce with `Nonces::next_active` from the tip's stored nonces.
+2. Asks the credentials resource for the slots we lead in that epoch, given the nonce and our stake share from `PoolSummaries`.
+3. Converts each led slot to a wall-clock instant with `EraHistory::slot_to_posix_time` and schedules a `LeadSlot(slot)` message for it with `Effects::schedule_at`.
+
+On startup the same steps run for the current epoch from the tip's active nonce, scheduling only the slots still ahead. Nothing runs on the slots we do not lead.
```
rkuhn, line 89: Decision: we start computation k blocks after slot `NextEpoch-4k/f` or when the epoch changes, whichever comes first. If this means that we miss a block opportunity very early in the next epoch, so be it.

We compute speculatively in the background ASAP so that we can show a tentative schedule, marked with the remark of how much of `k` has passed (a.k.a. confidence level, chain depth, … )

engineering-decision-records/035-block-forging.md
```diff
@@ -0,0 +1,129 @@
...
+
+### Once per led slot
+
+When `LeadSlot(slot)` fires:
+
```
rkuhn, line 94: We might want to schedule the forging a couple ms before the onset of the slot so that we have the block ready upon slot onset.

Important: once we start forging, we have picked a parent block. If the adopted chain (best tip) changes while forging, we don’t abort/restart/whatever, we just keep on forging.

engineering-decision-records/035-block-forging.md
```diff
@@ -0,0 +1,129 @@
...
+1. Check that the operational certificate covers the slot's KES period. If not, log a warning and stop.
+2. Pick the parent: the current tip, or the tip's parent if the tip's slot equals ours.
+3. Ask the mempool for a sequence of transactions that is valid on the parent's state as of our slot and fits in a block. Ask the ledger to apply it and keep the resulting state fragment, keyed by the body hash.
+4. Ask the credentials resource to forge the header: VRF proof for the slot, block body hash, KES signature for the period.
+5. Run the header through the same `validate_header` every peer header passes. This costs one VRF verify and one KES verify, and yields the evolved nonces we must store with the header anyway.
```
rkuhn, line 99: KES evolution happens as part of the signing procedure, persistence is not required by Praos rules.

Technically, we need only the `evolve_nonces` effect, not the full `validate_header()` function computation.

engineering-decision-records/035-block-forging.md
```diff
@@ -0,0 +1,129 @@
...
+6. Store the header and the block, and send the new tip to `select_chain`.
+
+`select_chain` ranks the tip. `fetch_blocks` sees the body is already stored. `validate_block` asks the ledger to roll forward, and the ledger recognises the block it just built and commits the kept fragment instead of running the rules again. `adopt_chain` then tells the mempool, the manager and `forge_block` about the new tip, and the manager serves the block to peers.
+
```
rkuhn, line 103: Need to add a section describing the TUI changes showing everything forging-related ⇒ @KtorZ 

Key management should be CLI commands as well as TUI functions.

Possibly highlighting transactions involving the pool address, etc.

engineering-decision-records/035-block-forging.md
```diff
@@ -0,0 +1,129 @@
...
+- **Enter the pipeline at `select_chain`, not `adopt_chain`.** `adopt_chain` assumes the ledger has applied the block and that `validate_block` and `select_chain` have moved their tip. Skipping them leaves `validate_block` believing the old tip is current, so the next upstream sibling of our block would be applied as an extension and fail. Entering at `select_chain` keeps every stage's bookkeeping right, and the kept fragment removes the double validation that route would otherwise cost.
+- **Everything is simulatable.** Time comes from `schedule_at`, keys from a resource, the ledger and mempool from resources. A pure-stage test can drive an epoch boundary and a led slot with a mocked credentials resource and assert the exact effect trace.
+
+## Consequences
+
+### Mempool
+
+Forging needs one thing from the mempool: given a parent state and a slot, a sequence of transactions that is valid in that order on that state and fits within the block limits. How the mempool produces it is its own concern and is not decided here. Today's `Mempool::take` does not offer this.
```
rkuhn, line 115: fix: purging mempool might be different operation than getting valid transactions

engineering-decision-records/035-block-forging.md
```diff
@@ -0,0 +1,129 @@
...
+
+### Ledger
+
+The ledger gains two entry points: build a body on a given parent at a given slot, and commit a kept fragment on roll forward. Building on the tip's parent means producing the state at tip minus one. The volatile store holds the last `k` blocks and `switch_to_fork` already rolls back, so the pieces exist, but the builder needs a parent argument rather than assuming the tip. Block limits (max body size, max execution units, max header size) come from the protocol parameters in the parent state after the epoch tick, not from the stage.
```
rkuhn, line 119: change: move the forging computations from ledger into the `forge_block` stage

engineering-decision-records/035-block-forging.md
```diff
@@ -0,0 +1,129 @@
...
+
+- **Secrets never enter stage state.** Stage state is serialised into the trace buffer on every message. The VRF and KES keys live in a resource and answer two effects, `leader_schedule` and `forge_header`. The stage keeps only public facts: the led slots, the certificate's start period and evolution limit.
+- **Compute the schedule once.** Every input to the leader check is fixed before the epoch starts. There is no per-slot loop.
+- **Enter the pipeline at `select_chain`, not `adopt_chain`.** `adopt_chain` assumes the ledger has applied the block and that `validate_block` and `select_chain` have moved their tip. Skipping them leaves `validate_block` believing the old tip is current, so the next upstream sibling of our block would be applied as an extension and fail. Entering at `select_chain` keeps every stage's bookkeeping right, and the kept fragment removes the double validation that route would otherwise cost.
+- **Everything is simulatable.** Time comes from `schedule_at`, keys from a resource, the ledger and mempool from resources. A pure-stage test can drive an epoch boundary and a led slot with a mocked credentials resource and assert the exact effect trace.
+
```
rkuhn, line 110: we should use the minor protocol version in the header to identify Amaru as the forger (should include git commit hash, we have 64bits of space)

engineering-decision-records/035-block-forging.md
rkuhn: We shall add an opt-out facility to check the local clock sync periodically, which may emit WARN messages. SPOs are expected to correctly configure NTP.

engineering-decision-records/035-block-forging.md
```diff
@@ -0,0 +1,139 @@
+---
+type: architecture
+status: proposed
+---
+
+# Block Forging
+
+Currently, Amaru follows the chain but does not extend it. This document describes how a stake pool running Amaru would produce blocks in Ouroboros Praos. The goal is a valid block with limited wasted work. The Haskell node is our reference for what "valid" means, and this document links to the relevant code, but the shape of our implementation is our own.
+
+Ouroboros Leios will introduce many changes for Amaru overall, but they are orthagonal to this document. For example, there will be mempool improvements, new independent processes for votes, announcements, and handling of endorsement blocks (EBs), and changes to the ledger. In theory, there will be no changes directly to the block forging architecture, though that may change as we learn more about Leios.
```
etorreborre, line 10: `s/orthagonal/orthogonal`

etorreborre: 2 consecutive `and` in the sentence.

engineering-decision-records/035-block-forging.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Context
+
+### Slot leadership is decided before the epoch starts
+
+A pool leads a slot when its VRF output for that slot, hashed with the epoch nonce, falls below a threshold that depends only on its stake share and the active slot coefficient `f`. See the Praos paper ([eprint 2017/573](https://eprint.iacr.org/2017/573)) for the protocol, and [`checkLeaderValue`](https://github.com/IntersectMBO/cardano-ledger/blob/bc7df956110912d3b1e3501dbb66e466f7220398/libs/cardano-protocol/src/Cardano/Protocol/TPraos/BlockHeader.hs#L337-L371) for the check as implemented.The VRF input is `hash(slot ‖ epoch nonce)` ([`mkInputVRF`](https://github.com/IntersectMBO/cardano-ledger/blob/bc7df956110912d3b1e3501dbb66e466f7220398/libs/cardano-protocol/src/Cardano/Protocol/Praos/VRF.hs#L63-L78)).
```
etorreborre, line 16: `s/implemented.The VRF/implemented. The VRF`

engineering-decision-records/035-block-forging.md
```diff
@@ -0,0 +1,139 @@
...
+
+Every input to that check is fixed before the epoch starts. The stake share comes from the snapshot taken two epochs earlier, which Amaru already exposes through `PoolSummaries`.The epoch nonce is fixed two days before the boundary (next section). So a pool could compute its whole leader schedule for an epoch in one pass. On mainnet that is 432,000 VRF evaluations.
```
etorreborre, line 18: `s/.The epoch nonce/. The epoch nonce`

engineering-decision-records/035-block-forging.md
```diff
@@ -0,0 +1,139 @@
...
+
+### Keys
+
+There are four keys used in the forging process. The Haskell node reads them in [`readLeaderCredentials`](https://github.com/IntersectMBO/cardano-node/blob/a38eac60bceb1a64a4ffa29e2d49d802787ce171/cardano-node/src/Cardano/Node/Protocol/Shelley.hs#L161-L211) and bundles them as [`PraosCanBeLeader`](https://github.com/IntersectMBO/ouroboros-consensus/blob/0ce94ef215312633d44d049ea1a7e30eb4b3deb6/ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos/Common.hs#L272-L297).
+
+- The **cold key** is an Ed25519 key kept offline. Its hash is the pool id. It signs the pool registration certificate and the operational certificate. The running node only needs the assosciated public key.
```
etorreborre, line 37: `s/assosciated/associated`

engineering-decision-records/035-block-forging.md
```diff
@@ -0,0 +1,139 @@
...
+- The **VRF key** lives on the block producer. It proves slot leadership. Cardano uses ECVRF-Ed25519-SHA512-Elligator2 from [draft-irtf-cfrg-vrf-03](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-vrf-03), bound in [`Cardano.Crypto.VRF.Praos`](https://github.com/IntersectMBO/cardano-base/blob/d92e2e3841eaad354c5e1ef77b458b347f471271/cardano-crypto-praos/src/Cardano/Crypto/VRF/Praos.hs#L524-L526). Amaru already has the verify side in `amaru-ouroboros/src/vrf`.
+- The **KES key** (key evolving signature) lives on the block producer and signs headers. It is a forward-secure scheme(defined in [eprint 2001/034](https://eprint.iacr.org/2001/034)), instantiated as [`Sum6KES`](https://github.com/IntersectMBO/cardano-base/blob/d92e2e3841eaad354c5e1ef77b458b347f471271/cardano-crypto-class/src/Cardano/Crypto/KES/Sum.hs#L104) over Ed25519: 64 periods, each `slotsPerKESPeriod` long. Evolving the key to the next period erases the previous period's secret, so a stolen key cannot sign old slots. Only the verification side exists in Amaru today.
```
etorreborre, line 39: `s/scheme(defined/scheme (defined`

engineering-decision-records/035-block-forging.md
```diff
@@ -0,0 +1,139 @@
...
+- The **operational certificate** binds a KES verification key to the pool. It carries the KES verification key, a counter, a start period, and a cold-key signature over the three ([`OCert`](https://github.com/IntersectMBO/cardano-ledger/blob/bc7df956110912d3b1e3501dbb66e466f7220398/libs/cardano-protocol/src/Cardano/Protocol/TPraos/OCert.hs#L84-L91)). Every header carries it. The [OCERT rule](https://github.com/IntersectMBO/cardano-ledger/blob/bc7df956110912d3b1e3501dbb66e466f7220398/libs/cardano-protocol-tpraos/src/Cardano/Protocol/TPraos/Rules/OCert.hs#L71-L106) rejects a header whose slot is before the start period or more than 62 periods after it, and a counter lower than the last one seen for that pool on the current chain.
```
etorreborre, line 40: I'm not sure how much we need to go into the details for validating the operational certificate but I would propose:

The operational certificate binds a KES verification key to the pool. It carries the KES verification key, a counter, a start period, and a cold-key signature over those three fields ([OCert](https://github.com/IntersectMBO/cardano-ledger/blob/bc7df956110912d3b1e3501dbb66e466f7220398/libs/cardano-protocol/src/Cardano/Protocol/TPraos/OCert.hs#L84-L91)). Every header carries it. 

The OCERT rule ([Praos](https://github.com/IntersectMBO/ouroboros-consensus/blob/0ce94ef215312633d44d049ea1a7e30eb4b3deb6/ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos.hs)) checks four things:
 1. The cold-key signature must verify against the header's issuer key. 
 2. The header's KES signature must verify under the hot key evolved by current period - start period. 
 3. The current KES period must lie within the certificate's window: at or after the start period and before start + maxKESEvolutions (62 on mainnet, leaving a margin of two under the 64 periods of Sum6KES), so a certificate covers 62 consecutive periods. 
 4. The counter must be equal to or one greater than the last counter seen for that pool on the current chain.

engineering-decision-records/035-block-forging.md
```diff
@@ -0,0 +1,139 @@
...
+
+A pool can therefore forge only when its certificate covers the current KES period. The Haskell node checks this per slot in [`praosCheckCanForge`](https://github.com/IntersectMBO/ouroboros-consensus/blob/0ce94ef215312633d44d049ea1a7e30eb4b3deb6/ouroboros-consensus-protocol/src/ouroboros-consensus-protocol/Ouroboros/Consensus/Protocol/Praos.hs#L697-L716).
+
+### How the Haskell node forges
+
+For reference, the per-slot loop in [`forkBlockForging`](https://github.com/IntersectMBO/ouroboros-consensus/blob/0ce94ef215312633d44d049ea1a7e30eb4b3deb6/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs#L573-L760) does the following at every slot:
+
+1. Pick the parent: the tip, or the tip's parent if the tip already sits in this slot.
```
etorreborre, line 48: We could add that this is the case if the node has already received (and validated) another node's block for the same slot.

engineering-decision-records/035-block-forging.md
```diff
@@ -0,0 +1,139 @@
...
+
+Steps 2, 3 and 5 are callbacks in the [`BlockForging`](https://github.com/IntersectMBO/ouroboros-consensus/blob/0ce94ef215312633d44d049ea1a7e30eb4b3deb6/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/Forging.hs#L82-L156) record, not in the loop.
+
+### Amaru today
+
+Consensus is a graph of pure stages ([EDR 011](./011-deterministic-simulation-testing.md)). Headers arrive in `track_peers`, `select_chain` ranks tips, `fetch_blocks` downloads bodies, `validate_block` asks the ledger to apply them, and `adopt_chain` updates the best chain and tells the mempool and the network manager about the new tip. Every effect with the outside world, including the ledger, the stores and the clock, goes through a resource so the graph can run under simulation. Nothing in the graph produces a block.
+
+## Decision
+
+We add one stage, `forge_block`, wired into the graph only when forging credentials are configured. It receives the adopted tip from `adopt_chain`, like the mempool does, and it sends exactly one message: a new tip to `select_chain`. From there our block is treated like any block a peer sent us.
```
etorreborre, line 63: I would complete the sentence by saying that it forges and stores a block and its header before sending a `NewTip` out to `select_chain`.

engineering-decision-records/035-block-forging.md
```diff
@@ -0,0 +1,139 @@
...
+
+1. Check that the operational certificate covers the slot's KES period. If not, log a warning and stop.
+2. Pick the parent: the current tip, or the tip's parent if the tip's slot equals ours.
+3. Ask the mempool for a sequence of transactions that is valid on the parent's state as of our slot and fits in a block. That sequence is the block body as-is; the stage does not validate it and does not consult the ledger.
+4. Ask the credentials resource to forge the header: VRF proof for the slot, block body hash, KES signature for the period.
+5. Run the header through the same `validate_header` every peer header passes. In theory, we only need the `evolve_nonces` effect, but the full `validate_header` function costs us almost nothing.
+6. Store the header and the block, and send the new tip to `select_chain`.
+
+`select_chain` ranks the tip. `fetch_blocks` sees the body is already stored. `validate_block` asks the ledger to roll forward, applying our block exactly as it would a peer's. This is the first and only time the ledger sees the body. `adopt_chain` then tells the mempool, the manager and `forge_block` about the new tip, and the manager serves the block to peers.
```
etorreborre, line 109: We could add a line explaining what happens if another block for the same slot has still been processed by `select_chain` in the meantime.

PR #1321, opened 2026-09-06
correctly implement bearer lifecycle

### LINKED ISSUE #736: Properly handle in/out/duplex peers
### Abstract

Currently, the manager just keeps a single `Connect(...)` state. We’ll need to distinguish between responder, initiator, and full duplex connections. And we need to handle peers who bind to their listen port for outbound connections, meaning we can get the same peer twice.

### Why?

Because otherwise we break the protocols.

### How?

By tracking the state of each peer’s state in the manager. This will then inform where we can send requests for chain sync and fetch blocks or transactions.

Add simulation tests connecting two Amaru nodes such as to exercise all cases:

- A → B with initiatorOnly
- A → B without initiatorOnly
- A → B ∧ B → A
- A → B ∧ B → A with the second one using initiatorOnly=false

Headers, blocks, and transactions need to flow in the respectively possible directions.

### Discussion points

_No response_

_No response_

### Checklist

rkuhn:
- model local use for maintenance or diffusion
- tear down based on error or selection decision

fixes #736

---

## Summary by CodeRabbit

* **Changed**
  * Connections now use a unified session model with local-use states: None, Maintenance, or Diffusion.
  * The connection manager no longer automatically redials after a session drops; peer selection manages available outbound connections.
* **Observability**
  * Connection local-use changes are now recorded with peer, connection, and usage details.
* **Documentation**
  * Added documentation for the new connection local-use trace event.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,6 +78,11 @@ Other guiding principles:
 ### Changed
 
 - **amaru-consensus**: `BlockValidator` now runs the ledger on a dedicated thread that owns the ledger state exclusively; operations are requested through a bounded channel and answered via per-request reply channels, replacing the shared lock around the state. `BlockValidator` is no longer generic over the store types. ([#1094][])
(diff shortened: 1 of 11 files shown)


rkuhn:
@coderabbitai review

PR #1330, opened 2026-09-08
Address CBOR Decoding FIXMEs

### LINKED ISSUE #1174: Fix the decoding FIXMEs

yHSJ:
This PR progresses #1174, but doesn't necessarily close it entirely. Some of the FIXMEs there are left for other PRs or future work because they are more significant overhauls, or they need to be considered more deeply.

Diff is huge because of test regeneration and manual fixture introductions.


crates/amaru-kernel/src/cardano/address/byron.rs
```diff
@@ -36,20 +36,21 @@ impl Deref for ByronAddress {
...
+    /// witness yields the same root on both implementations.
     pub fn root(witness: &BootstrapWitness) -> Hash<28> {
-        let mut xpub = [0u8; 64];
-        xpub[..32].copy_from_slice(&witness.public_key[..]);
-        xpub[32..].copy_from_slice(&witness.chain_code[..]);
+        // Serialised `SpendingData::VerificationKey`, with the byte string length pinned to 64
+        // regardless of the actual chain code length.
+        let mut xpub = vec![0x82, 0x00, 0x58, 0x40];
+        xpub.extend_from_slice(&witness.public_key[..]);
+        xpub.extend_from_slice(&witness.chain_code[..]);
```
rkuhn, line 51: This is only valid CBOR if the slice lengths sum up to 64. Does the Haskell node really construct invalid CBOR?!?

crates/amaru-kernel/src/cardano/address/byron.rs
```diff
@@ -130,32 +131,46 @@ impl<C> cbor::Encode<C> for ByronAddress {
...
-            // (via `decodeCrcProtected`, always at the Byron protocol version), which rejects
-            // indefinite-length byte strings, so we reject them too.
-            #[allow(clippy::disallowed_methods)]
-            let payload = d.bytes()?.to_vec();
-            let crc = d.u32()?;
+        // Conformance: the Haskell node reads the tag-24 payload with cborg's `decodeBytes`
+        // (via `decodeCrcProtected`, always at the Byron protocol version), which rejects
+        // indefinite-length byte strings, so we reject them too.
+        #[allow(clippy::disallowed_methods)]
+        let payload = d.bytes()?.to_vec();
```
rkuhn: why use `to_vec` here?

crates/amaru-kernel/src/cardano/address/byron.rs
```diff
@@ -130,32 +131,46 @@ impl<C> cbor::Encode<C> for ByronAddress {
...
-            // Conformance: the Haskell node reads the tag-24 payload with cborg's `decodeBytes`
-            // (via `decodeCrcProtected`, always at the Byron protocol version), which rejects
-            // indefinite-length byte strings, so we reject them too.
-            #[allow(clippy::disallowed_methods)]
-            let payload = d.bytes()?.to_vec();
-            let crc = d.u32()?;
+        // Conformance: the Haskell node reads the tag-24 payload with cborg's `decodeBytes`
+        // (via `decodeCrcProtected`, always at the Byron protocol version), which rejects
+        // indefinite-length byte strings, so we reject them too.
+        #[allow(clippy::disallowed_methods)]
```
rkuhn, line 145: This should be `#[expect(...)]` (or removed).

crates/amaru-kernel/src/cardano/chain_code.rs
```diff
@@ -0,0 +1,105 @@
...
+        e: &mut cbor::Encoder<W>,
+        ctx: &mut C,
+    ) -> Result<(), cbor::encode::Error<W::Error>> {
+        e.encode_with(&self.0, ctx)?.ok()
+    }
+}
+
+impl<'b, C: cbor::HasProtocolVersion> cbor::Decode<'b, C> for ChainCode {
+    fn decode(d: &mut cbor::Decoder<'b>, ctx: &mut C) -> Result<Self, cbor::decode::Error> {
+        let bytes = cbor::decode_bytes_with(d, ctx)?;
```
rkuhn, line 56: `decode_bytes_with` is the wrong name, sent me in the wrong direction: this is about conditionally accepting indefinite byte strings and should be named as such.

crates/amaru-ledger/src/rules/transaction/phase_one/outputs.rs
```diff
@@ -58,14 +58,16 @@ pub enum SupplementalDatumPolicy {
...
+#[expect(clippy::too_many_arguments)]
 pub fn execute<C>(
     context: &mut C,
     arena_pool: &ArenaPool,
     protocol_parameters: &ProtocolParameters,
     network: Network,
     outputs: Vec<MemoizedTransactionOutput>,
     supplemental_datum_policy: SupplementalDatumPolicy,
-    construct_utxo: impl Fn(&mut C, u64, &Value) -> Option<TransactionInput>,
+    first_index: usize,
```
rkuhn: Why is this `usize` when it must fit in `u16` below? It even leads to a panic.

Important: can this panic be caused by data we receive from the network? Then it MUST be removed.

rkuhn (commented):
Lots of good work! Still have a couple of points.

crates/amaru-ledger/src/rules/transaction/phase_one/outputs.rs
yHSJ: Because the source of this data is a `.len()` call, but that is backwards logic.

yHSJ: And no, the CDDL restricts the indexes to a `u16`, and the size limits mean we can never get more outputs in a tx than a u16.

crates/amaru-kernel/src/cardano/address/byron.rs
yHSJ: Force of habbit

crates/amaru-kernel/src/cardano/address/byron.rs
yHSJ: Yes, it does... I'm not sure I understand *why*, but I made sure I was consistent

https://github.com/IntersectMBO/cardano-ledger/blob/ae2c8912b204d11e102faeb617b6faa6a8c9f934/libs/cardano-ledger-core/src/Cardano/Ledger/Keys/Bootstrap.hs#L158-L188

crates/amaru-kernel/src/cardano/address/byron.rs
yHSJ: Why would it be `expect`? `.bytes()` is disallowed by our clippy rules, with the note:
```
note: does not handle indefinite-length byte strings; use amaru_minicbor_extra::decode_bytes instead
```

Here we specifically do NOT want to support indefinite-length byte strings, so I am explicitly choosing to use this disallowed method, with a conformance comment.

crates/amaru-kernel/src/cardano/chain_code.rs
yHSJ: I think I agree with you, but I didn't introduce that here. What do you think a better name is? `decode_byte_string`? `decode_with_indefinite`?

crates/amaru-kernel/src/cardano/address/byron.rs
rkuhn: This is disgusting, I’ll need a strong drink now. But reading the Haskell comment: if the key was previously deserialized, does that not ensure that either part is 32 bytes long?

crates/amaru-kernel/src/cardano/address/byron.rs
rkuhn: Right, but `allow` means that this annotation could also be doing nothing, while our stated approach is to enforce that we only retain annotations that do something. It makes reading code on github easier because the semantics are clear.

crates/amaru-kernel/src/cardano/chain_code.rs
rkuhn: boyscout rule: no matter who left it, the new PR will clean it up :-)

How about `decode_bytes_v12_indefinite`?

crates/amaru-ledger/src/rules/transaction/phase_one/outputs.rs
rkuhn: Do I understand correctly that you’re saying that more outputs will fail during deserialization, hence we can trust the index if its type is `u16`? I still would be very careful around panics possibly resulting from invalid network input — perhaps someone touches the deserialization code later and removes the guarantee, then we have a DoS at our hands. Better to remain defensive and convert to infallible types ASAP.

yHSJ: No, I mean the CDDL enforces the `transaction_index` *is* a u16: https://github.com/IntersectMBO/cardano-ledger/blob/master/eras/conway/impl/cddl/data/conway.cddl#L801

The other restriction it has (which can't be represented in CDDL) is here: https://github.com/IntersectMBO/cardano-ledger/blob/master/eras/conway/impl/cddl/data/conway.cddl#L3-L7.

We also benefit from the fact that the outputs list can never have more elements than a u16 can hold, due to protocol parameters.

crates/amaru-kernel/src/cardano/address/byron.rs
```diff
@@ -40,16 +40,17 @@ impl ByronAddress {
     }
 
     /// Re-compute an address (verification key) root from a transaction witness.
+    ///
+    /// The chain code is hashed as-is even when it is not 32 bytes long so that such a
+    /// witness yields the same root on both implementations.
```
etorreborre: "Both implementations" means Amaru vs Haskell, we could be explicit about that.

etorreborre (approved):
Nice set of fixes!

crates/amaru-kernel/src/cardano/address/byron.rs
yHSJ: Ah, I see what you meant. Yeah, that's fine.

crates/amaru-kernel/src/cardano/chain_code.rs
yHSJ: I don't love that either, but I guess it's an improvement

crates/amaru-kernel/src/cardano/address/byron.rs
yHSJ: It depends on how it was decoded, of course, the `vkey` decodes with a fixed size, so yes we know the vkey is 32 bytes long. The chain code does not (at least, pre pv12).

PR #1335, opened 2026-09-09
fix: fetch candidate CC members from volatile and stable proposals

yHSJ:
@arwlf reported an invalid block on mainnet with the following logs:

```
Sep 09 18:41:35 Fast amaru[1265004]: 2026-09-09T16:41:35.364783Z  WARN amaru::consensus: block.invalid failed_tip="[197031688, h'cbb394a0e1ad78f740f210a4e0c9a09d4e4e0f9d14617702635a77ef80cf1be5', 13900703]" parent="[197031676, h'955946a4d2a9fe087be319bbb2c21fd93f58c933bbbe1c0c4e591c16a3e55e9e', 13900702]" error="BlockValidationError: Invalid block: Transaction d0f8573a76cf4dd49ffd3f263a37d820ad9e3bd0094f08de346e33c5233cde51 at index 0 is invalid: transaction failed phase one validation: invalid certificates: invalid cc member hot credential delegation: unknown source entity: ScriptHash(Hash<28>(\"7c34e0240b84029e0932f5e8d81af42a63f55de6da31f16e19b1f5b4\"))" detail="failed to advance the ledger to a new tip"
Sep 09 18:41:35 Fast amaru[1265004]: 2026-09-09T16:41:35.364897Z  WARN amaru::consensus: block.invalid failed_tip="[197031701, h'5f6b23b5b3866a7f9583b11241dc6ab56560d03df4d14b26cf5bdd34a1b09cab', 13900704]" parent="[197031688, h'cbb394a0e1ad78f740f210a4e0c9a09d4e4e0f9d14617702635a77ef80cf1be5', 13900703]" error="the block descends from an invalid block" detail="refusing to validate the descendant of an invalid block"
Sep 09 18:41:35 Fast amaru[1265004]: 2026-09-09T16:41:35.364981Z  INFO amaru::consensus: chain.best_tip_invalidated removed=1
```

This failure was because a candidate CC member was delegating a hot credential before the proposal was enacted to add them to the CC. This *is* an allowed action, however in our block preparation, we were not fetching candidates correctly from proposals. This PR fixes that, and introduces some unit tests to assert the behavior is correct.

## Summary by CodeRabbit

- **Bug Fixes**
  - Committee members proposed by a pending committee update can now authorize hot-key actions before the update is finalized.
  - Members who are proposed to join the committee can also submit resignations during this pending period.
  - Improved handling ensures these credentials are recognized consistently while committee changes are awaiting approval.

### DIFF
--- a/crates/amaru-ledger/src/context/default/preparation.rs
+++ b/crates/amaru-ledger/src/context/default/preparation.rs
@@ -438,28 +438,34 @@ fn resolve_committee<'block, 'volatile>(
             }
         }
 
+        // NOTE: Scanning stable proposals when resolving committee
+        //
         // A credential neither layer resolves may still be a candidate of a pending `UpdateCommittee`
         // proposal, entitled to authorize a hot key or resign before holding any committee row. Only
         // the proposals vouch for it, discounting those the pending boundary pruned.
+        //
+        // Note that pending `UpdateCommittee` in the volatile are already taken into account by
+        // resolve_cc_members. So here, an unresolved credential can only come from a pending
+        // proposal in the stable store we had no information about.
         let candidates: BTreeSet<&Credential> = cold_credentials_in_certificates
             .into_iter()
             .filter(|cold_credential| !cc_members.contains_key(cold_credential))
             .collect();
 
         if !candidates.is_empty() {
-            let mut proposed = volatile.resolve_committee_candidates();
-
             for (id, row) in db.iter_proposals().map_err(ContextHydratationError::ResolveCommittee)? {
-                if let GovernanceAction::UpdateCommittee(_, _, added, _) = row.proposal.gov_action
-                    && !matches!(volatile.resolve_proposal(&id), Existence::Gone)
-                {
-                    proposed.extend(added.into_iter().map(|(cold_credential, _)| cold_credential));
+                if matches!(volatile.resolve_proposal(&id), Existence::Gone) {
+                    // Discard any proposal that is stable but has been pruned at the not-yet-stable
+                    // epoch boundary.
+                    continue;
                 }
-            }
 
-            for cold_credential in candidates {
-                if proposed.contains(cold_credential) {
-                    cc_members.entry(*cold_credential).or_default();
+                if let GovernanceAction::UpdateCommittee(_, _, added, _) = row.proposal.gov_action {
+                    for (cold_credential, _) in
+                        added.into_iter().filter(|(candidate, _)| candidates.contains(candidate))
+                    {
+                        cc_members.entry(cold_credential).or_default();
+                    }
                 }
             }
         }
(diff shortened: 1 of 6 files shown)


crates/amaru-ledger/src/context/default/preparation.rs
```diff
@@ -420,57 +422,44 @@ fn resolve_committee<'block, 'volatile>(
...
-                    if cold_credentials_in_certificates.contains(cold_credential) {
-                        gone_but_requested.insert(*cold_credential);
-                    }
+        // A credential neither layer resolves may still be a candidate of a pending `UpdateCommittee`
+        // proposal, entitled to authorize a hot key or resign before holding any committee row. Only
+        // the proposals vouch for it, discounting those the pending boundary pruned.
+        let candidates: BTreeSet<&Credential> = cold_credentials_in_certificates
+            .into_iter()
+            .filter(|cold_credential| !cc_members.contains_key(cold_credential))
+            .collect();
```
KtorZ, line 453: Unnecessary collect

crates/amaru-ledger/src/state/volatile/aggregate.rs
```diff
@@ -125,6 +125,21 @@ impl VolatileAggregate {
...
+            .iter()
+            .filter_map(|(id, state)| {
+                if let GovernanceAction::UpdateCommittee(_, _, added, _) = &state.proposal.gov_action {
+                    Some(added.iter().map(move |(candidate, _)| (id, candidate)))
+                } else {
+                    None
+                }
+            })
+            .flatten()
+    }
```
KtorZ, line 142: I don't like the fact we pull the committee members from yet-another-source instead of returning them through `resolve_cc_members`; but we do need the proposal id which is lacking from the previous abstraction.

crates/amaru-ledger/src/context/default/preparation.rs
```diff
@@ -546,28 +535,39 @@ mod tests {
...
             type Pool = ();
             type Account<'a> = ();
             type DRep<'a> = ();
-            type Proposal = ();
+            type Proposal = Existence<ProposalStateSlim>;
             type CCMembers<'a> = BTreeMap<&'a Credential, Existence<CommitteeMemberBind<'a>>>;
 
+            fn resolve_proposal(&self, _: &ProposalId) -> Self::Proposal {
+                if self.proposals_pruned_at_boundary { Existence::Gone } else { Existence::Unknown }
+            }
```
KtorZ: This is... a recipe a for mistake. The mock already have `proposals`, so why not lookup in those instead of returning a potentially inconsistent response?

crates/amaru-ledger/src/context/default/preparation.rs
```diff
@@ -546,28 +535,39 @@ mod tests {
...
             },
         };
 
+        #[derive(Default)]
         struct Mock {
             volatile_cc_members: Vec<(Credential, Existence<Bind<ConstitutionalCommitteeMemberStatus, Epoch, Empty>>)>,
             stable_cc_members: Vec<(Credential, Option<Epoch>, Option<ConstitutionalCommitteeMemberStatus>)>,
             proposals: Vec<Proposal>,
+            volatile_candidates: Vec<Credential>,
+            proposals_pruned_at_boundary: bool,
```
KtorZ: What really seems missing from the mock is the ability to represent volatile proposals, since the current `proposals` represents stable proposals.

crates/amaru-ledger/src/context/default/preparation.rs
```diff
@@ -420,57 +422,44 @@ fn resolve_committee<'block, 'volatile>(
...
+            .into_iter()
+            .filter(|cold_credential| !cc_members.contains_key(cold_credential))
+            .collect();
+
+        if !candidates.is_empty() {
+            let mut proposed = volatile.resolve_committee_candidates();
+
+            for (id, row) in db.iter_proposals().map_err(ContextHydratationError::ResolveCommittee)? {
+                if let GovernanceAction::UpdateCommittee(_, _, added, _) = row.proposal.gov_action
+                    && !matches!(volatile.resolve_proposal(&id), Existence::Gone)
```
KtorZ: 👍

KtorZ (changes requested):
Thanks for the quick turn-around here. I completely forgot to include the pending proposals in the volatile state when I reworked this.

And, this is really the only thing that's missing. I don't quite like the newly introduced `resolve_committee_candidates` because it creates an unsafe API surface; this one must be called in tandem with `resolve_cc_members`; and the latter now returns only a partial information. 

I have a concrete suggestion that merges `resolve_committee_candidates` with `resolve_cc_members` which I'll push shortly.

PR #1350, opened 2026-09-14
add session type verification of typestate protocol handlers

rkuhn:
This is the last preparatory step before moving all mini-protocol handlers to the new typestate API. It gives us full binary session type verification of the wire protocol in mermaid syntax, including timeouts (*) and `WantNext` and with a generic (i.e. independently auditable) implementation of pipelining.

```rust
mod responder {
    make_states!(pub Proto { Idle; Done });
    define_role_tag!(pub ToInitiator);

    on_receive!(Idle as ServerIdleIn {
        Pull => { Send<ToMux, WantNext> => Idle }
        RequestRange => {
            Call<ToInitiator, StartBatch>, Repeat<Call<ToInitiator, Block>>,
                Call<ToInitiator, BatchDone>, Send<ToMux, WantNext> => Idle
            | Call<ToInitiator, NoBlocks>, Send<ToMux, WantNext> => Idle
        }
        ClientDone => { Send<ToMux, WantNext> => Done }
    });
    on_receive!(Done as DoneIn {});
}

pub(crate) fn session_spec() -> SessionSpec {
    session_spec! {
        Message;
        [*] --> Idle
        Idle --> Busy: RequestRange
        Idle --> Done: ClientDone
        Busy --> Idle: NoBlocks
        Busy --> Streaming: StartBatch
        Streaming --> Streaming: Block
        Streaming --> Idle: BatchDone
        note left of Idle: Initiator
        note left of Busy: Responder timeout BLOCKFETCH_AGENCY_TIMEOUT
        note left of Streaming: Responder timeout BLOCKFETCH_AGENCY_TIMEOUT
    }
}

#[test]
fn protocol_conformance() {
    check_want_next(&initiator::Proto::type_graph(), &blockfetch_initiator()).unwrap();
    check_want_next(&responder::Proto::type_graph(), &blockfetch_responder()).unwrap();
    let spec = session_spec();
    let h_i = assert_projects(&initiator::Proto::type_graph(), &blockfetch_initiator(), &spec);
    let h_r = assert_projects(&responder::Proto::type_graph(), &blockfetch_responder(), &spec);
    h_r.dual().assert_bisimilar(&h_i);
}
```

Skip-Changelog

No operator-facing changelog entry: this checks handler code against the network spec and does not change node behaviour.

There’s also an EDR explaining the approach. Please let me know if that level of detail / abstraction is useful or I should do it differently.

(*) The `Duration` is not yet captured on the type-level.

---
<a href="https://app.blacksmith.sh/pragma-org/codesmith/amaru/pr/1350"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792006576&installation_model_id=436488&pr_number=1350&repository=pragma-org%2Famaru&return_to=https%3A%2F%2Fgithub.com%2Fpragma-org%2Famaru%2Fpull%2F1350&signature=fffe2858a03ed24cfb7f030078e8c2d0a2cceb1332cfb345517a7218af3a424d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith-bot</code> with what you need. Autofix is disabled.</sup>

## Summary by CodeRabbit

- **New Features**
  - Added session-type projection and conformance validation against protocol specifications.
  - Added support for session diagrams, message labels, state graphs, agency, timeout, and receive-coverage validation.
  - Added protocol introspection utilities for describing states, transitions, and message alphabets.
  - Added display formatting for BlockFetch messages.

- **Bug Fixes**
  - BlockFetch completion now passes through an explicit terminal state, preventing unintended message pipelining.
  - Standardized BlockFetch agency timeout handling.

- **Documentation**
  - Added an engineering decision record describing session-type and typestate projection.


crates/amaru-protocols/src/blockfetch/spec.rs
```diff
@@ -0,0 +1,96 @@
...
+use super::{
+    BLOCKFETCH_AGENCY_TIMEOUT, BatchDone, Block, ClientDone, Message, NoBlocks, RequestRange, StartBatch,
+    initiator::{self, Busy, Done, Idle, Streaming},
+    responder,
+};
+use crate::protocol::{Pull, ToMux, check_want_next};
+
+pub(crate) fn session_spec() -> SessionSpec {
+    session_spec! {
+        Message;
```
etorreborre, line 32: What's the use of `Message` here, since it is disconnect from the rest?

engineering-decision-records/035-session-types-typestate-projection.md
```diff
@@ -0,0 +1,104 @@
...
+status: proposed
+---
+
+# Checking typestate mini-protocol handlers against the network spec
+
+Does not supersede [EDR-021](./021-switching-to-own-mini-protocols.md).
+
+## Motivation
+
+[EDR-021](./021-switching-to-own-mini-protocols.md) promised that each mini-protocol’s network behaviour could be checked against the [Ouroboros network spec](../agent-inputs/network-spec.pdf) by looking at the handler’s state machine, not only by running simulations. That used to mean `ProtoSpec` walking a `ProtocolState` object.
```
etorreborre, line 12: It would be good to have links to the `ProtoSpec` and `ProtocolState` code with an example of instantiation (the fetch block protocol since that's the running example) to show where we're coming from.

engineering-decision-records/035-session-types-typestate-projection.md
```diff
@@ -0,0 +1,104 @@
...
+
+Typestate remainders (sequences of effects along the transition from one state to the next) replace that object. They are richer than the spec diagram: they also mention mux demand (`WantNext`), injected `Pull`, agency timers, and node-local traffic (collector, store, `Fetch` / `Close`). Named typestate constructors need not match spec states — BlockFetch’s responder handles Busy and Streaming as a sequence from `Idle`. So the remainder is not 1:1 with Table 3.7 of the network spec, for example.
```
etorreborre, line 14: I think that "typestate remainders" deserves a longer introduction:

 - The name.
 - What it means in connection to session types.
 - What it means when used in Rust code (what can the compiler check with them). 
 - An example.
 - The full grammar used in `amaru` (the full thing could be presented in a "reference" section because that might be too much for the "motivation" section. Also to explain even simple things like `on_receive!(Done as DoneIn {});`).
 - How do they compare to the kind of state diagrams that are used in the network spec.
 - Why they are useful (that's a motivation section after all :-)).

engineering-decision-records/035-session-types-typestate-projection.md
```diff
@@ -0,0 +1,104 @@
...
+
+We still want a test, on ordinary Rust values, that the *wire conversation* implied by those typestate remainders is exactly the spec machine. The Rust type system already enforces local linear use of a remainder; it does not compare that remainder to the PDF.
+
+## Decision
+
+A protocol’s spec is a mermaid-like state diagram in code. BlockFetch lives in [`crates/amaru-protocols/src/blockfetch/spec.rs`](../crates/amaru-protocols/src/blockfetch/spec.rs):
```
etorreborre, line 20: I would say that the decision is to specify protocol handlers with typestate remainders handlers.
Then we can say how we can check that we have the correct implementation by:

 1. By describing the network spec diagram as a mermaid like diagram in the code
 2. Extracting from the typestate remainders the implied state machine
 3. Comparing them structurally (address the question of labels there).

engineering-decision-records/035-session-types-typestate-projection.md
```diff
@@ -0,0 +1,104 @@
...
+```
+
+That is the network-spec picture: states, messages, *who is allowed to speak* (agency), and how long the silent party may wait. The identifiers are the real types (`Idle`, `RequestRange`), so go-to-definition works. Building the spec also checks that every `Message` variant is either an edge or listed `unused […]` (needed later for Handshake).
+
+A handler is a typestate remainder graph (`Proto::type_graph()`). To compare it to the spec we *project*: drop mux plumbing, timers, and local roles, and keep only sends and receives on the peer channel. A remainder sequence such as `StartBatch; Block*; BatchDone` becomes a path of labeled edges. Vertices need not keep spec names; only the message types and who speaks matter.
+
+The projected graph is a communicating finite-state machine (CFSM): two parties, FIFO-ish messages, *exactly one side has the floor* in each state (no mixed “I might send or receive”). That is the network spec’s §3.2 model, and it is the binary case of session types. Honda–Vasconcelos–Kubo is the usual syntax (`!` send, `?` receive, choice, loop). We do not type-check the handler in that syntax; we build the machine in tests and compare.
```
etorreborre, line 54: `FIFO-ish` deserves an explanation.

etorreborre: > Honda–Vasconcelos–Kubo is the usual syntax (`!` send, `?` receive, choice, loop)

I would leave out that part if we don't use it. I'm also not sure if mentioning session types here helps that much. Maybe, instead, a section at the end connecting what we are doing to the literature could be useful as "further reading" for people interested.

engineering-decision-records/035-session-types-typestate-projection.md
```diff
@@ -0,0 +1,104 @@
...
+
+Comparison is *structural*. Walk both machines from the start, matching edge labels and agency. Names are diagnostic only (e.g. `Idle` in error dumps). Initiator and responder must be *duals*: one’s send is the other’s receive; who-has-the-floor is a property of the *state*, not swapped. After `ClientDone` the spec is `Done`. Restarting a new session (which is automatic for responders) is handler code (`state = initial_state()`), not a cycle in the diagram.
+
+We require an *exact* match of labels, not “the handler may receive extra kinds.” In session-type theory, Gay–Hole *subtyping* treats a participant that still accepts more incoming messages as a subtype of a pickier one (it is a more liberal receiver). The network spec does the opposite: an unexpected message *aborts* the bearer. Extra receive arms would swallow a peer bug. Extra send arms would be our bug. So projection must not leave leftover wire labels.
```
etorreborre, line 58: Those precisions on how the comparison is done are indeed useful. Maybe a proper list or table of all the checks: what + why could be even better.

engineering-decision-records/035-session-types-typestate-projection.md
```diff
@@ -0,0 +1,104 @@
...
+
+Pipelining (CIP-0164) is *N* lock-step instances plus occupancy. The check is performed on one instance, multiplexing is layered on top in an almost transparent fashion. Mux `WantNext` / `Pull` well-formedness is an extra test in `amaru-protocols` (`check_want_next`); it is not part of the session library, because the mux does not live in `amaru-pure-stage`.
```
etorreborre, line 60: This deserves a sub-section IMO to expose the problem with modeling / implementation of pipelining:
 - "the" check -> of what precisely?
 - "almost" transparent -> what's not transparent then?
 - well-formedness -> why is that even a problem? How does that relate to `check_want_next` below?

engineering-decision-records/035-session-types-typestate-projection.md
```diff
@@ -0,0 +1,104 @@
...
+
+A protocol’s conformance test is then: build the mermaid spec, project both handlers, refine the spec, check duality, plus `check_want_next`. See `protocol_conformance` in `spec.rs`.
```
etorreborre, line 62: Why does the spec need to be refined?

etorreborre: This sentence could be replace with the list or table mentioned above + a link to the code for `protocol_conformance`.

engineering-decision-records/035-session-types-typestate-projection.md
```diff
@@ -0,0 +1,104 @@
...
+
+### What this does not prove
+
+Typestate already forbids skipping remainder effects. `convert_input` still aborts unexpected mail at runtime; mux overflow is unchanged. The test does not prove local decisions (`StartBatch` vs `NoBlocks`), store I/O, timeout *durations* (only that a timer is armed), pipelining cursors, or that `Done` is followed by `initial_state()`.
```
etorreborre, line 66: > Typestate already forbids skipping remainder effects

This can be reminded here but must have been explained when introducing Typestate remainders.

> `convert_input` still aborts unexpected mail at runtime

Not clear what this is about.

> mux overflow is unchanged

What is it exactly, why is that a problem, what can we do about it?

> The test does not prove local decisions

What is a "local" decision? "local" with contrast to what? What needs to be proven with those decisions? That we the responders sends "NoBlocks" there are indeed no blocks to be sent? How do we check that then?

> timeout *durations* (only that a timer is armed)

By this sentence the reader should know how a timeout is represented in a typestate remainder graph. Is it a regular message?

> pipelining cursors

What are they? That should be covered by the section on pipelining.

> or that `Done` is followed by `initial_state()`

How do we cover that?

engineering-decision-records/035-session-types-typestate-projection.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Consequences
+
+- Each typestate protocol gets a `spec.rs` that is the network diagram, readable by humans (to cross-check with the PDF) and by `cargo test`.
```
etorreborre, line 70: Points 1, 2, 4 seem like decisions to me, under some implementation / migration sections.

The part about the expressiveness of the typestate API deserves a more developed example. With `Repeat(A, B)` for example to show that this cannot be projected while being potentially desirable in a protocol. And then describe what should be done if that cases arises (possibly change the projection code?).

Otherwise I'd say that the consequences are:

- The wire subset of the typestate handler can be checked against the network spec.
- Any change to the spec has to be reflected in the code (we don't magically pull the latest `network.pdf` spec).
- Internal effects / decision need to be tested separately (provide links on how it's done)..

engineering-decision-records/035-session-types-typestate-projection.md
```diff
@@ -0,0 +1,104 @@
...
+
+### Session types and CFSMs
+
+- Honda, Vasconcelos, Kubo, *Language primitives and type discipline for structured communication-based programming*, ESOP 1998 — binary session types (`!` / `?` / choice / recursion)
+- Honda, Yoshida, Carbone, *Multiparty asynchronous session types*, POPL 2008 — projection from a global type onto one role (we project a *handler* onto the mux, which is a different hiding step)
+- Brand, Zafiropulo, *On communicating finite-state machines*, JACM 1983 — CFSMs
+- Deniélou, Yoshida, *Multiparty session types meet communicating automata*, ESOP 2012 — binary session types ≈ deterministic two-machine CFSMs without mixed states (the network spec)
+- Gouda, Manning, Yu, *On progress for two communicating finite state machines*, 1984 — half-duplex two-machine progress
+- Gay, Hole, *Subtyping for session types in the pi calculus*, Acta Informatica 2005 — extra inputs as a subtype; we reject that under abort-on-unexpected
+- Kozen, *A completeness theorem for Kleene algebras*, 1994 — `Repeat` as star; we only expand a single remaining wire effect
```
etorreborre, line 103: Not sure how the completeness theorem helps here.

engineering-decision-records/035-session-types-typestate-projection.md
```diff
@@ -0,0 +1,104 @@
...
+- [`amaru-pure-stage::session`](../crates/amaru-pure-stage/src/session.rs) — spec, projection, structural compare
+
+### Ouroboros / Cardano
+
+- [Network spec](../agent-inputs/network-spec.pdf) §3.2 (agency CFSMs), §3.8 BlockFetch
+- [CIP-0164](https://cips.cardano.org/cip/CIP-0164) — pipelining as *N* lock-step copies; not in this projection
+- `typed-protocols` (Haskell) — GADT encoding of the same exclusive-agency machines
+
+### Session types and CFSMs
+
```
etorreborre, line 96: This section could be more useful with a bit more explanations on each paper:

 - "It introduces the notion of..."
 - "It shows that we can model communication with..."
 - "It would support X if we wanted to check further properties..."

crates/amaru-pure-stage/src/session.rs
```diff
@@ -0,0 +1,1986 @@
...
+            && self.transitions == other.transitions
+    }
+}
+
+impl Eq for Cfsm {}
+
+impl Cfsm {
+    /// Panic unless the reachable fragments are isomorphic (edge labels + agency).
+    #[track_caller]
+    pub fn assert_refines(&self, spec: &Cfsm) {
```
etorreborre, line 144: This function duplicates the next one but also doesn't do what it says. I was expecting some test for behavior inclusion.

crates/amaru-pure-stage/src/session.rs
```diff
@@ -0,0 +1,1986 @@
...
+                "assert_refines mismatch: {reason}\nprojected:\n{}\n\nspec:\n{}",
+                self.fmt_table(),
+                spec.fmt_table()
+            );
+        }
+    }
+
+    /// Same as [`assert_refines`](Self::assert_refines): names are not compared.
+    #[track_caller]
+    pub fn assert_bisimilar(&self, other: &Cfsm) {
```
etorreborre, line 156: I think that "bisimilar" is a bit too strong of a name for a structural equality check. I guess we can have  `Cfsm`s that are bisimilar but not non-structurally equal.

etorreborre (commented):
@rkuhn here is a bunch of comments. I haven't yet fully read everything but I though it'd be good to send a batch already :-).

crates/amaru-protocols/src/blockfetch/spec.rs
rkuhn: Exactly all of the message types mentioned in the graph must have `Into<Message>`, which is checked by the code the macro expands to. Basically, this formalism expects the wire protocols to use one message enum as a design choice.

engineering-decision-records/035-session-types-typestate-projection.md
rkuhn: Hmm, perhaps a link to a suitable git tree? Otherwise it would be somewhat large.

engineering-decision-records/035-session-types-typestate-projection.md
rkuhn: Ah, right, I forgot that even though the typestate API is not introduced in this PR, it is still rather new and unfamiliar to the audience. Will add.

engineering-decision-records/035-session-types-typestate-projection.md
rkuhn: Right, that’s better.

engineering-decision-records/035-session-types-typestate-projection.md
rkuhn: The notation is used in `fmt_table()`, I’ll think about what to do with that. The FIFO is strictly speaking true, but pipelining makes it confusing. I’ll disentangle this.

engineering-decision-records/035-session-types-typestate-projection.md
rkuhn: Yeah, this is the one bit that doesn’t yet feel perfect: WantNext is the back-pressure mechanism, which in other words means that each receive from the network must be preceded by sending this message to the mux. So far so neat, because the correct placement of this effect can be checked automatically given the expected network behaviour. But pipelining destroys this because the later protocol instances in the pipeline must not send their WantNext until their turn arrives, otherwise back-pressure would be sabotaged. Hence, when switching control for receiving to the next instance, then that instance will need to be prodded into sending the WantNext using the Pull message. This is the compromise that allows the same handler logic to run with and without pipelining (without pipelining the Pull is synthesized in a shallow wrapper).

engineering-decision-records/035-session-types-typestate-projection.md
rkuhn: Technically, the implemented behaviour doesn’t need to match the spec precisely, it only needs to be a subtype a.k.a. refinement. In this setting there isn’t much freedom for change (timeouts might be different, for example), but this is still the established terminology.

engineering-decision-records/035-session-types-typestate-projection.md
rkuhn: Yeah, this paragraph slipped through from Grok, sorry, it needs to be rewritten.

engineering-decision-records/035-session-types-typestate-projection.md
rkuhn: Hmm, yes, could be a decision as well. When read that way, a consequence would be that we add a secondary type-checker that checks the type, meaning that “it compiles” is no longer enough for the typestate API. An approach without this consequence would have been to encode the necessary computations on the type-level, but unfortunately Rust doesn’t seem to want that as of now (at least I couldn’t make it work).

engineering-decision-records/035-session-types-typestate-projection.md
rkuhn: Good catch, I’ll need to ask Grok where it needed that. For the session type papers I had the background myself.

crates/amaru-pure-stage/src/session.rs
rkuhn: I tried to come up with an example because what you say feels right, but so far the “redundant” states always trivially collapse … I’ll read some more to find the appropriate name for the function.

PR #1378, opened 2026-09-17
Fix Stack Overflows

yHSJ:
On preprod, [`f90dce5765108da976abdbb9fc618f9a6ffd9fa4d93b2f288eed1808545424c9`](https://preprod.cexplorer.io/tx/f90dce5765108da976abdbb9fc618f9a6ffd9fa4d93b2f288eed1808545424c9) was submitted which contains a native script nested 5,383 levels deep. That caused a stack overflow while decoding. This PR flattens `NativeScript` into it's node in pre-order.

@scarmuega gets credit for diagnosing the issue in Pallas originally: https://github.com/txpipe/pallas/pull/802/changes

While testing these changes, I realized we also recursively walk the cbor structures in our multiplexer, which was *actually* causing the stack overflow.

Additionally, I flattened the `Metadatum` data type, which had the same issue.

There are remaining recursions in the plutus VM, but they should be addressed in their own PR.

EDIT: I also completed the `TODO` of merging the `MemoizedNativeScript` and `NativeScript` types. I decided to keep the name as `NativeScript`, instead of `MemoizedNativeScript` because the fact that it carries it's own bytes is an implementation detail. If we want to follow the pattern of *any* type that owns it's original bytes is `Memoized`, I can change that.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented crashes when processing deeply nested CBOR messages, native scripts, and transaction metadata.
  * Improved handling of deeply nested native scripts while preserving script evaluation and encoding behavior.
  * Improved CBOR frame parsing for incomplete, tagged, and deeply nested data.

* **Refactor**
  * Updated native script and metadata handling to support deep nesting more reliably.
  * Simplified native script storage and removed obsolete memoization behavior.


crates/amaru-kernel/src/cardano/memoized/native_script.rs
```diff
@@ -1,205 +0,0 @@
...
-// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-// See the License for the specific language governing permissions and
-// limitations under the License.
-
-use crate::{NativeScript, cbor, utils::string::blanket_try_from_hex_bytes};
-
-#[derive(Debug, Clone, PartialEq, Eq, serde::Deserialize, serde::Serialize)]
-#[serde(try_from = "&str")]
-pub struct MemoizedNativeScript {
-    original_bytes: Vec<u8>,
```
KtorZ: !!

KtorZ (commented):
So, this is getting a little out of hands IMO. Especially because, this needs to ALSO be applied to `Eq`, `PartialEq`, `Clone`, `Drop`, etc... and basically any derive trait method that is performing a traversal of the structure. 

This must applied to Metadatum, NativeScript and PlutusData; yet because the actual traversal depends on the type itself, it is hard to abstract (though some parts can be). Still, that's thousands and thousands of lines of code on rather critical operations where the probability of it going wrong is high. And, it is certainly not code that we're going to enjoy maintaining. 

After some research, it seems that a common solution in Rust is to actually use a growing stack, that can "spill over" the heap when the actual stack is exhausted. This what the rust compiler (a.k.a `rustc` uses) internally, as well as 1600+ other rust libraries in more than -- check notes -- 100M+ downloads: https://crates.io/crates/stacker

There's a little utility that's been build on top to provide stack-safe wrappers and derived traits through just a couple of annotations: https://crates.io/crates/stacksafe

This looks reasonable. One caveat: with this approach, the stack/heap can technically _grow arbitrarily_. BUT, we know that all types in the kernel are bounded in size (block size), so this can be computed. So my suggestion would be instead to:

- use a growing stack approach via stacksafe
- increase the default stack size _anyway_ just to give us more headspace in many situation. 2MiB is extremely small, we could bump to 8MiB or so for all Tokio threads.
- make sure that usages of recursive structures is tightly controlled in the kernel (functions that traverse the structure must be `stacksafe`).

jeluard:
Default stack size is platform dependent, would be nice to make this in a way that doesn't mess with platform with smaller defaults.

> increase the default stack size anyway just to give us more headspace in many situation. 2MiB is extremely small, we could bump to 8MiB or so for all Tokio threads.

We might want to keep those under scrutiny ? There might be lots of places where tokio is used by 3rd parties etc.

KtorZ:
I believe this only concerns _our_ threads created through the Tokio runtime. We can adjust the stack when spawning a thread; it's not a general setting.

KtorZ:
Note: superseded by #1380; however, micro-benchmarks have shown that the "flattened" approach and the introduction of internals arena with indirect references offered great speed up (~6x faster) over the current approach with nested structures. So definitely worth considering in the long-run.

yHSJ:
Closing this PR with the note that there should be future work that uses the flattened approach, as mentioned above.

PR #1384, opened 2026-09-19
fix: decode uplc with smaller stack

jeluard:
Playing on iOS, I got some issues with stack sizes and uplc decoding.
Building on top of recent related changes, use similar technics to solve that.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed stack overflow crashes when processing deeply nested UPLC programs and constants.
  - Improved reliability during FLAT encoding, FLAT decoding, and result reconstruction for highly nested content.
  - Deeply nested delay terms and complex data structures can now be processed successfully, including programs with substantial nesting depth.
  - Added regression coverage for decoding programs containing tens of thousands of nested terms.


KtorZ (approved):
approving following offline discussions regarding benchmarks; this changes makes decoding about 10% slower; which is _acceptable_ as a short term mitigation while we work on something more long-term.

PR #1385, opened 2026-09-20
add forge_block stage

### LINKED ISSUE #1361: forge_block Stage: Compute and Schedule the Leadership Schedule
### Abstract

Introduce the new `forge_block` stage and implement the [EDR's](https://github.com/pragma-org/amaru/pull/1310) "once per epoch" behavior.

### Why?

Block Forging

### How?

New module under `stages/` wired after `adopt_chain` only when we have the necessary resources for block forging. State should hold only led slots and certificate window. Start calculating leadership schedule via resource when `NewTip`'s slot is within `NextEpoch -4k/f` (`randomess_stability_window` exists for this). `schedule_at` for the next led epoch, and the remaining schedule. `cancel_schedule` when the nonce changes before `k` blocks deep.

At startup, compute current *and* next epoch and drop past slots, skipping when *all* slots are in the past.

Node behavior and graph unchanged without credentials; `pure_stage` tests for window entry, rollback inside vs past the window, startup mid-epoch, sync catch-up.

### Discussion points

_No response_

#1310 introduced the block forging EDR

#1355 is required for this, though the mocks are enough for the `pure_stage` tests.

### Checklist

### LINKED ISSUE #1362: forge_block Stage: forge and Inject a Block on LeadSlot
### Abstract

Implement the [EDR's](https://github.com/pragma-org/amaru/pull/1310) "once per led slot" behavior.

### Why?

Block forging

### How?

KES window check, parent must be the tip, or the tip-1 when the tip is in our slot. Create an empty block body, build the block/header via #1359, sign via #1355, call `evolve_nonces` `store_validated_header` and `store_block`, then send the tip to `select_chain`. In theory, this can fire a few ms *before* slot to "prepare the work". Ignore tip changes once the parent is picked.

Log missed slots with a reason.

Trace-based tests for happy path, expired certificate, not-yet-valid certificate, tip in our slot.

### Discussion points

_No response_

#1350 introduces the block forging EDR.

Depends on #1355, #1356, #1357, #1358, #1359, #1360.

### Checklist

rkuhn:
fixes #1361
fixes #1362

## Summary by CodeRabbit

- **New Features**
  - Added the consensus block-forging stage framework, including leader scheduling, freeze-window handling, stale-slot protection, and forging diagnostics.
  - Added scheduling, cancellation, detached effects, and clock/external effects to stage sessions.

- **Improvements**
  - Improved block-fetch message handling with borrowed inputs.
  - Added block-height difference calculations.
  - Improved freeze-window and leader-schedule tracking across chain changes.

- **Documentation**
  - Documented forging trace events and schemas for forged blocks, missed slots, schedules, and failures.


crates/amaru-consensus/src/stages/forge_block/protocol.rs
```diff
@@ -0,0 +1,352 @@
...
+            let store = Store::new(eff.clone());
+            handle_adopted_tip(state, tip, store, eff).await
+        }
+        Ok(IdleIn::LeadSlot(lead)) => handle_lead_slot(state, lead, eff).await,
+        Ok(IdleIn::LeaderSchedule(schedule)) => handle_leader_schedule(state, schedule, eff).await,
+        Err(_msg) => state,
+    }
+}
+
+async fn handle_adopted_tip(
```
yHSJ, line 137: Once we set `state.freeze` once, I believe this function effectively becomes a noop. This is a mental trace through this logic:

Step 1: first tip in epoch `N` that is after the stabilization window. `frozen = true` and `freeze = None`, so we set `to_schedule` and `state.freeze`, this is the expected behavior.

Step 2: First tip in epoch `N+1`. `frozen = false` but `freeze = Some(epoch N)`. the old freeze is still on the best chain, `candidate_rollback` returns `None`, so we just log (and compute the depth). Normal, I think.

Step 3: First tip in epoch `N+1` that is after the stabilization window. `freeze` is still `Some(epoch N)`, now `frozen = true`. `candidate_rollback` returns `None` again. So we just compute the depth again and log. Importantly, we never call `to_schedule(N+2)` or update the freeze state. so I think we don't compute a leadership schedule for any epoch after the first computation.

The simplest solution is that we need to set `self.freeze` to `None` again at some point, which never happens.

rkuhn: I moved this logic out into its own function, which makes it a bit easier and also fixed this issue and more, please have a look.

PR #1386, opened 2026-09-21
fix(tui): restore terminal before reporting panics

jeluard:
Make sure panic message is properly shown when tui is enabled.

## Summary by CodeRabbit

- **Bug Fixes**
  - Restores the terminal dashboard before displaying panic diagnostics, keeping crash information visible.
  - Improves terminal recovery after crashes or failed setup, including restoration of the primary screen, cursor, mouse capture, and input behavior.
  - Prevents terminal state from being restored multiple times during error handling.

- **Documentation**
  - Updated the unreleased changelog to document improved terminal recovery and crash-reporting behavior.


crates/amaru-tui/src/terminal_guard.rs
```diff
@@ -27,39 +40,95 @@ pub struct TerminalGuard {
...
+        TerminalState::Active => operation(),
+        TerminalState::Inactive | TerminalState::Shutdown => Err(terminal_shutdown_error()),
+    }
+}
+
+fn restore_if_active() {
+    let state = TERMINAL_STATE.lock();
+    if state.get() == TerminalState::Active {
+        restore_terminal();
+        if state.get() != TerminalState::Shutdown {
```
etorreborre: This condition doesn't seem to be necessary since we are in an `Active` state and `restore_terminal` does not change that.

crates/amaru-tui/src/lib.rs
```diff
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#![feature(reentrant_lock)]
```
etorreborre: Can we use `parking_lot::ReentrantMutex` instead of relying on a nightly feature?

etorreborre (approved):
Ok, as far as I understand what's going on :-).
It would be good to get rid of the `reentrant_lock` feature if possible.

crates/amaru-tui/src/terminal_guard.rs
jeluard: Embarassing :)

crates/amaru-tui/src/terminal_guard.rs
```diff
@@ -12,13 +12,26 @@
...
 use crossterm::{
-    event, execute,
+    cursor, event, execute,
     terminal::{EnterAlternateScreen, LeaveAlternateScreen, disable_raw_mode, enable_raw_mode},
 };
-use ratatui::{Terminal, backend::CrosstermBackend};
+use parking_lot::ReentrantMutex;
+use ratatui::{Frame, Terminal, backend::CrosstermBackend};
+
+static TERMINAL_STATE: ReentrantMutex<Cell<TerminalState>> = ReentrantMutex::new(Cell::new(TerminalState::Inactive));
```
rkuhn: The `Cell` doesn’t really serve a functional purpose here, right?

crates/amaru-tui/src/terminal_guard.rs
```diff
@@ -27,39 +40,93 @@ pub struct TerminalGuard {
...
+
+pub fn emergency_restore_terminal() {
+    let state = TERMINAL_STATE.lock();
+    if state.replace(TerminalState::Shutdown) == TerminalState::Active {
+        restore_terminal();
+    }
+}
+
+fn with_active<T>(operation: impl FnOnce() -> io::Result<T>) -> io::Result<T> {
+    let state = TERMINAL_STATE.lock();
```
rkuhn: How is it guaranteed that this lock will be released by some other thread if it was locked?

crates/amaru-tui/src/terminal_guard.rs
```diff
@@ -27,39 +40,93 @@ pub struct TerminalGuard {
...
+                restore_if_active();
+                Err(error)
+            }
+        }
     }
 
-    pub fn terminal(&mut self) -> &mut Terminal<CrosstermBackend<Stdout>> {
-        &mut self.terminal
+    pub fn draw(&mut self, render: impl FnOnce(&mut Frame<'_>)) -> io::Result<()> {
+        with_active(|| self.terminal.draw(render).map(|_| ()))
```
rkuhn: This runs arbitrary code under the lock, meaning that panic handler may hang forever.

rkuhn (commented):
My gut feeling is that that ReentrantLock will bite us later. Could we make this terminal restoration at panic time more aggressive by using an AtomicU8 instead and just ignoring any potentially unfinished `draw()` activity?

PR #1389, opened 2026-09-21
test: run decoding tests from a node-agnostic test suite

etorreborre:
Skip-Changelog


crates/amaru-kernel/tests/common/acknowledged.rs
```diff
@@ -0,0 +1,93 @@
...
+            Ok(contents) => Ok(Self::These(toml::from_str::<Declarations>(&contents)?.acknowledged)),
+            Err(e) if e.kind() == std::io::ErrorKind::NotFound => Ok(Self::None),
+            Err(e) => Err(e.into()),
+        }
+    }
+
+    /// Return true if the run is allowed to observe a failure of the given rule and class.
+    pub fn contains(&self, rule: &str, class: &str) -> bool {
+        match self {
+            Self::None => true,
```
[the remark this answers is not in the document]
etorreborre: That's the same thing. No file, None, no failure are acknowledged, as if all failures were in the acknowledged file.

crates/amaru-kernel/tests/common/test_results.rs
```diff
@@ -0,0 +1,164 @@
...
+    /// Mark some failures as acknowledged, and record any stale acknowledgements that are no longer observed in this run.
+    pub fn acknowledge_failures(&mut self, test_configuration: &TestConfiguration) {
+        // The distinct `(rule, class)` pairs the run produced, which is what acknowledgements are matched against.
+        let observed: BTreeSet<(String, String)> =
+            self.failures.iter().map(|(key, reason)| (key.rule().to_string(), error_class(reason))).collect();
+        let acknowledged = test_configuration.acknowledged_failures();
+
+        self.unacknowledged =
+            observed.iter().filter(|(rule, class)| !acknowledged.contains(rule, class)).cloned().collect();
+        self.stale = acknowledged.stale(&observed);
```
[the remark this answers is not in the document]
etorreborre: Makes sense, fixed.

scripts/fetch-cbor-dataset
```diff
@@ -0,0 +1,36 @@
...
+COMMIT="$(tr -d '[:space:]' < "$PIN_FILE")"
+URL="https://codeload.github.com/r2rationality/cardano-cbor-dataset/tar.gz/$COMMIT"
+
+# GNU tar needs --wildcards to glob member names. BSD tar (macOS) globs by default and rejects the flag.
+WILDCARDS=""
+if tar --version 2>/dev/null | grep -q 'GNU tar'; then
+  WILDCARDS="--wildcards"
+fi
+
+mkdir -p "$DEST_DIR"
```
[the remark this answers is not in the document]
etorreborre: Good idea

.github/workflows/ci-build-and-test.yml
```diff
@@ -11,8 +11,7 @@ on:
   push:
     branches: ["main"]
   pull_request:
-    types: [opened, synchronize, reopened, converted_to_draft, ready_for_review]
-    branches: ["*"]
+    branches: ["etorreborre/test/cbor-node-agnostic"]
```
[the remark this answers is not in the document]
etorreborre: Will remove later

crates/amaru-kernel/tests/common/test_results.rs
```diff
@@ -0,0 +1,166 @@
...
+        let mut totals = TestOutcome::default();
+        for outcome in self.per_rule.values() {
+            totals.merge(outcome.clone());
+        }
+        totals
+    }
+
+    /// Total number of samples executed, including both generated and zapped samples.
+    pub fn total(&self) -> usize {
+        self.per_rule.values().map(|t| t.generated_total + t.zap_must_be_rejected_expected).sum()
```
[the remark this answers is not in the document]
etorreborre: That's ok

crates/amaru-observability/macros/Cargo.toml
```diff
@@ -18,7 +18,6 @@ categories.workspace = true
 proc-macro = true
 
 [dependencies]
-inventory.workspace = true
```
jeluard: Why? I don't see that you removed it's usages?

.github/actions/fetch-cbor-dataset/action.yml
```diff
@@ -0,0 +1,22 @@
...
+  steps:
+    - name: Download the CBOR dataset
+      shell: bash
+      run: make fetch-cbor-dataset
+
+    - name: Verify the CBOR dataset
+      shell: bash
+      run: |
+        set -euo pipefail
+        dataset="crates/amaru-kernel/tests/data/cbor.dataset"
```
jeluard: At some point might make sense to make this an argument of the action

.github/workflows/ci-build-and-test.yml
```diff
@@ -127,6 +127,31 @@ jobs:
         env:
           AMARU_BENCH_BLOCK_SIZE: 4096
 
+  cbor-conformance:
```
jeluard: How fast is this? If too slow, maybe move to nightly

crates/amaru-kernel/tests/acknowledged-failures.toml
```diff
@@ -0,0 +1,67 @@
+# Failures the decoders are known to produce against the conformance corpus.
+#
+# Each entry lets the suite observe one defect without failing, so a partially conforming decoder
+# still publishes a report.
+#
+# Delete an entry as soon as its fix lands. An entry that no longer matches any failure fails the
```
jeluard: So that's like a TODO? If so, acknowledged sounds a bit misleading

crates/amaru-kernel/tests/common/normalization/0001-indefinite-array.in.cbor
```diff
@@ -0,0 +1 @@
+��
```
jeluard: Should those move to the dataset eventually?

.github/actions/fetch-cbor-dataset/action.yml
```diff
@@ -0,0 +1,22 @@
...
+        samples="$(find "$dataset" -name '*.cbor' 2>/dev/null | wc -l | tr -d ' ')"
+        if [[ "$samples" -eq 0 ]]; then
+          echo "no CBOR samples under $dataset; the download did not produce the corpus" >&2
```
jeluard: How much are we exposed to vulnerability happening in the remote repo?

crates/amaru-observability/macros/Cargo.toml
etorreborre: There was a warning saying it was unused. There are actually other warnings that we should remove too.

.github/workflows/ci-build-and-test.yml
etorreborre: That doesn't take long, like one minute as a separate workflow task.

crates/amaru-kernel/tests/acknowledged-failures.toml
etorreborre: This is what makes the build not fail for now and the conformance test report partial conformance. When I push the PR fixing the gaps, I'll remove that file so that any regression will fail the build

crates/amaru-kernel/tests/common/normalization/0001-indefinite-array.in.cbor
etorreborre: They were copied from the `cbor-dataset` repository because they are small and don't change often.

crates/amaru-observability/macros/Cargo.toml
jeluard: Still used [here](https://github.com/pragma-org/amaru/blob/main/crates/amaru-observability/src/registry.rs#L20), weird.
The only warning I have is related to some nightly feature not supported anymore.

.github/actions/fetch-cbor-dataset/action.yml
etorreborre: Good question. I think that, in order to attack:

1. That repo would need to have its history re-written to present adversary data for the pinned commit.
2. There would need to be a vulnerability in the `cbor` library that we use to allow that data to execute malicious code.

I'm not a security expert but I think that even if 1. is possible, I suppose that 2. cannot happen.

.github/actions/fetch-cbor-dataset/action.yml
etorreborre: Yes, possibly. This will also might have to evolve when we support another dataset like Dijkstra.

crates/amaru-observability/macros/Cargo.toml
etorreborre: Weird indeed. Anyway that shouldn't be part of this PR. I'm going to revert that change.

jeluard (approved):

Issue #1390, opened 2026-09-21
`amaru mithril sync` cannot continue from a store created by `amaru node bootstrap`


mpizenberg:
### What revision are you using?

`v10.11.20260918`, the code involved is the same on `main` (`20fa9361a`).

### What operating system are you using?

MacOS

### What architecture are you using?

aarch64/armv8

### What problem are you having? What should be the expected behaviour?

I'm trying to get the ledger state at the end of a past epoch, plus the blocks of the few epochs before it, using only the published snapshots. The route I tried is `amaru node bootstrap --epoch E`, then `amaru mithril sync --ingest-until-slot S` to replay the rest from Mithril's immutable files. The release notes and `make refresh` suggest the same combination.

The sync fails on a bootstrapped store, in two places:

1. The download step can't find the ledger tip in the immutable files. It reads the tip with a block height of 0, and the lookup compares the height along with the slot and hash. As far as I can tell this hits any store made by bootstrap.
2. With that patched locally, ingest stops because bootstrap leaves the chain store without a best chain. `amaru node run` accepts the same store: it realigns the chain store to the ledger tip at startup.

I expected `mithril sync` to continue from the ledger tip, like `node run` does.

Two more things I noticed on the same path:

- The download ignores `--ingest-until-slot` and always fetches up to the latest Mithril snapshot, which is a lot more than I need.
- From the LLM reading the code (I couldn't get far enough to run it): ingest builds its ledger config from `LedgerConfig::default()`, so the ledger gets preprod's era history and global parameters on every network. On preview that would silently shift the epoch boundaries and `k`.

### Would you take PRs for this?

I'm happy to work on fixes: match the resume point on slot and hash, realign the chain store before ingest the way `node run` does, pass the network's parameters to the ledger, and bound the download when an until-slot is given. Does that direction sound right to you, or is the strict tip check in ingest deliberate?

### If possible, provide reproductions steps.

In an empty directory, with the release binary:

```sh
amaru node bootstrap --network preview --epoch 1119 \
  --ledger-dir ledger.preview.db --chain-dir chain.preview.db
amaru mithril sync --network preview \
  --ledger-dir ledger.preview.db --chain-dir chain.preview.db \
  --ingest-until-slot 120800000
```

The sync downloads, rebuilds its cache, downloads again, and exits with the first error below. The second error comes from a local build where the lookup matches on slot and hash only.

### If any, provide logs / error outputs from the system when the problem occurs.

Release binary:

```
[...]

PR #1391, opened 2026-09-22
Fix mithril sync after bootstrap

mpizenberg:
This is a proposed fix for #1390

## Summary by CodeRabbit

- **Changed**
  - `amaru mithril sync --ingest-until-slot` now stops downloading immutable files beyond the requested slot.

- **Bug Fixes**
  - Mithril sync now resumes correctly from stores created by node bootstrap or previous syncs.
  - Sync now uses the selected network’s era history and protocol parameters, ensuring accurate epoch boundaries on preview and mainnet.
  - Ledger and chain tip recovery is more reliable when resuming interrupted or partially initialized stores.


jeluard:
See https://github.com/pragma-org/amaru/issues/1390#issuecomment-5773514911

